PR #324, opened 2018-11-29
Feature/d3 asim 800

faizan2590:
Please goto folder: /aggregated_results/KPI & /aggregated_results/trade-detail to find the desired result


src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -122,7 +127,18 @@ def _is_cell_tower_node(area):
 
 
 def _is_load_node(area):
-    return isinstance(area.strategy, LoadHoursStrategy)
+    return isinstance(area.strategy, (LoadHoursStrategy, CellTowerLoadHoursStrategy,
```
spyrostz: No need fo the CellTowerLoadHoursStrategy, isinstance is capable of detecting whether subclasses originate from a base class. Since CellTower inherits from LoadHours, it should detect it.

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -122,7 +127,18 @@ def _is_cell_tower_node(area):
...
+                                      ElectrolyzerStrategy))
+
+
+def _is_producer_node(area):
+    return isinstance(area.strategy, (PVStrategy, PVUserProfileStrategy,
```
spyrostz: Same as before, only the base classes should be sufficient here.

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -285,6 +315,19 @@ def _generate_intraarea_consumption_entries(accumulated_trades):
     return consumption_rows
 
 
+def generate_inter_area_trade_details(area, past_market_types):
+    accumulated_trades = _accumulate_grid_trades_all_devices(area, {}, past_market_types)
+    trade_details = list()
```
spyrostz: Is it necessary to be a list of tuples ? It would be much cleaner to my eyes if it was a dictionary

src/d3a/d3a_core/export.py
```diff
@@ -61,6 +73,24 @@ def __init__(self, root_area: Area, path: str, subdir: str):
             os.makedirs(self.plot_dir)
 
         self.export()
+        self.export_json_data(self.directory)
+
+    def export_json_data(self, directory: dir):
+        json_dir = os.path.join(directory, "aggregated_results")
+        mkdir_from_str(json_dir)
+        json_file = os.path.join(json_dir, "KPI")
```
spyrostz: Could you rename the json_file to kpi_file to be a bit more readable ?

src/d3a/d3a_core/export.py
```diff
@@ -44,9 +51,14 @@ def mkdir_from_str(directory: str, exist_ok=True, parents=True):
 
 class ExportAndPlot:
 
-    def __init__(self, root_area: Area, path: str, subdir: str):
+    def __init__(self, root_area: Area, path: str, subdir: str, endpoint_buffer):
         self.area = root_area
         self.export_data = FileExportEndpoints(root_area)
+        self.endpoint_buffer = endpoint_buffer
+        self.area_performance_index = dict()
```
spyrostz: I would abstract away the KPIs from this file to their own class. The ExportAndPlot class is not supposed to do any calculation, should only write the files.

src/d3a/d3a_core/export.py
```diff
@@ -44,9 +51,14 @@ def mkdir_from_str(directory: str, exist_ok=True, parents=True):
...
+        self.cep_device = list()
+        self.cep_energy = 0
+        self.accumlated_trade_energy = 0
```
spyrostz: 'accumulated_trade_energy'

src/d3a/d3a_core/export.py
```diff
@@ -103,9 +133,24 @@ def _export_area_with_children(self, area: Area, directory: dir):
             subdirectory.mkdir(exist_ok=True, parents=True)
             for child in area.children:
                 self._export_area_with_children(child, subdirectory)
+                if isinstance(child.strategy, (CommercialStrategy, FinitePowerPlant)):
```
spyrostz: This whole block should be moved to its own module that would calculate independently the KPI values

src/d3a/d3a_core/export.py
```diff
@@ -116,6 +161,44 @@ def _export_area_with_children(self, area: Area, directory: dir):
                                                     BalancingOffer, "offers",
                                                     area.past_balancing_markets)
 
+    def _export_house_pv_self_consumption(self, area):
```
spyrostz: Same as above, this function should be in its own module

src/d3a/d3a_core/export.py
```diff
@@ -61,6 +64,24 @@ def __init__(self, root_area: Area, path: str, subdir: str):
...
         self.export()
+        self.export_json_data(self.directory)
+
+    def export_json_data(self, directory: dir):
+        json_dir = os.path.join(directory, "aggregated_results")
+        mkdir_from_str(json_dir)
+        kpi_file = os.path.join(json_dir, "KPI")
+        with open(kpi_file, 'w') as outfile:
+            cep_share = self.kpi.cep_energy / self.kpi.total_energy
+            self.kpi.performance_index["global-non-renewable-energy-share"] = cep_share
```
spyrostz: Trivial, but why not moving this calculation inside the kpi as well? cep_energy, total_energy and performance_index are all members of the kpi class, therefore this calculation could be moved inside the class

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -125,9 +126,17 @@ def _is_load_node(area):
...
+
+
+def _is_prosumer_node(area):
+    return isinstance(area.strategy, StorageStrategy)
+
+
 def _accumulate_load_trades(load, grid, accumulated_trades, is_cell_tower):
     accumulated_trades[load.name] = {
-        "type": "cell_tower" if is_cell_tower else "load",
+        "type": load.name,
```
spyrostz: I would not do that, since it breaks the current UI implementation.

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -143,10 +152,27 @@ def _accumulate_load_trades(load, grid, accumulated_trades, is_cell_tower):
...
+                accumulated_trades[load.name]["produced"] += trade.offer.energy
+                accumulated_trades[load.name]["earned"] += trade.offer.price
+    return accumulated_trades
+
+
 def _accumulate_house_trades(house, grid, accumulated_trades, past_market_types):
     if house.name not in accumulated_trades:
         accumulated_trades[house.name] = {
-            "type": "house",
+            "type": house.name,
```
spyrostz: Same as above, this breaks the UI

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -211,3 +216,68 @@ def _specific_row(self, slot, market):
...
+        house_pv_device = list()
+        house_load_device = list()
+        trade_by_pv = 0
+        total_energy_bought = 0
+        traded_from_pv = 0
+        if not _is_house_node(area):
+            return
+        for child in area.children:
+            if isinstance(child.strategy,
+                          (PVStrategy, PVUserProfileStrategy, PVPredefinedStrategy)):
```
spyrostz: No need for checking for all these strategies, only PVStrategy is enough.

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -143,10 +152,27 @@ def _accumulate_load_trades(load, grid, accumulated_trades, is_cell_tower):
     return accumulated_trades
 
 
+def _accumulate_producer_trades(load, grid, accumulated_trades):
+    accumulated_trades[load.name] = {
+        "type": load.name,
```
spyrostz: Same goes for this, for uniformity I would suggest to keep the original purpose of type to map to either cell tower or house or unknown. You have the load name from the dict key anyways

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -211,3 +216,68 @@ def _specific_row(self, slot, market):
...
+
+
+class KPI:
+    def __init__(self):
+        self.performance_index = dict()
+        self.cep_device = list()
+        self.cep_energy = 0
+        self.total_energy = 0
+
+    def accumulated_trade_energy(self, area):
```
spyrostz: From what I understand, this function accumulates the sold energy from the commercial producer and the bought energy from the loads, but the naming does not reflect it ta all. Why not having 2 separate functions that handle these 2 separate tasks, with better names for each? There is no significant performance disadvantage either, since this function is not recursive (recursion is handled by its caller, not from this function).

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -211,3 +215,78 @@ def _specific_row(self, slot, market):
...
+            self_consumption_within_house = 0
+        if total_energy_bought != 0:
+            self_sufficiency = traded_from_pv / total_energy_bought
+        else:
+            self_sufficiency = 0
+
+        return {"self_consumption_within_house": self_consumption_within_house,
+                "self_sufficiency": self_sufficiency}
+
+    def area_performance_index(self, area):
```
spyrostz: Nitpicking, but this function would be better described if named 'update_kpis_from_area' (in order to be able to add more kpis in the future and not just the area_performance_index)

PR #331, opened 2018-12-06
Bug/d3 asim 843

faizan2590:


dirkbig:
I ran it through with the (stedin project) setup that was causing troubles previously, now I get: `

  File "/Users/dirkvandenbiggelaar/gsy/d3a/src/d3a/events/__init__.py", line 49, in event_listener
    self._event_mapping[event_type](**kwargs)
  File "/Users/dirkvandenbiggelaar/gsy/d3a/src/d3a/models/strategy/load_hours.py", line 210, in event_bid_traded
    assert self.energy_requirement_Wh[market.time_slot] >= -0.00001
AssertionError

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -117,15 +118,17 @@ def _match_offers_bids(self):
                     price_drop=True
                 )
             elif (bid.price/bid.energy) >= clearing_rate and \
-                    (clearing_energy - cumulative_traded_bids) <= bid.energy:
-                cumulative_traded_bids += (clearing_energy - cumulative_traded_bids)
+                    ((clearing_energy - cumulative_traded_bids) <= bid.energy and not 0):
```
spyrostz:
```suggestion
                    (0 < (clearing_energy - cumulative_traded_bids) <= bid.energy):
```

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -100,7 +100,8 @@ def _match_offers_bids(self):
         if clearing is None:
             return
         clearing_rate, clearing_energy = clearing
-        log.info(f"Market Clearing Rate: {clearing_rate} & Clearing Energy: {clearing_energy} ")
+        log.info(f"[AREA]: {self.owner.name} ||| [ENGINE]: {self.name} ||| MCR: {clearing_rate} "
```
spyrostz: Personal preference, but I would suggest to change these to lowercase, furthermore I was under the impression that some of these are already covered by the generic logger of IAAEngine (like the owner for instance), don't they?

PR #332, opened 2018-12-07
D3ASIM-802: Implementation of area connect and disconnect event, with…

spyrostz:
… a convenient DisconnectIntervalAreaEvent. Device events were added. Integration tests for this functionality.

### DIFF
--- a/integration_tests/steps/area_events.py
+++ b/integration_tests/steps/area_events.py
@@ -39,25 +39,20 @@ def no_trades_between(context, start_hour, end_hour):
                if start_hour <= market.time_slot.hour < end_hour)
 
 
-@then('load uses battery energy and battery charges from grid before {hour}:00')
+@then('load and battery use energy from grid before {hour}:00')
 def load_uses_battery_before(context, hour):
-    # TODO: Should be changed once https://gridsingularity.atlassian.net/browse/D3ASIM-852
-    # is fixed
     hour = int(hour)
     house1 = next(filter(lambda x: x.name == "House 1", context.simulation.area.children))
 
     for market in house1.past_markets:
         if market.time_slot.hour >= hour:
             continue
-        assert len(market.trades) == 1
-        assert all(t.seller == "IAA House 1" and t.buyer == "H1 General Load"
-                   for t in market.trades)
+        assert len(market.trades) == 2
+        assert all(t.seller == "IAA House 1" for t in market.trades)
 
 
-@then('load uses battery energy and battery charges from grid after {hour}:00')
+@then('load and battery use energy from grid after {hour}:00')
 def load_uses_battery_after(context, hour):
-    # TODO: Should be changed once https://gridsingularity.atlassian.net/browse/D3ASIM-852
-    # is fixed
     hour = int(hour)
     house1 = next(filter(lambda x: x.name == "House 1", context.simulation.area.children))
 
--- a/integration_tests/steps/area_events.py
+++ b/integration_tests/steps/area_events.py
@@ -49,8 +49,6 @@ def load_uses_battery_before(context, hour):
     for market in house1.past_markets:
         if market.time_slot.hour >= hour:
             continue
-        print(market.time_slot.hour)
-        print(market.trades)
         assert len(market.trades) == 1
         assert all(t.seller == "IAA House 1" and t.buyer == "H1 General Load"
                    for t in market.trades)
--- a/integration_tests/steps/area_events.py
+++ b/integration_tests/steps/area_events.py
@@ -1,0 +1,39 @@
+from behave import then
+
+
+def one_or_none_trade_before_or_after(context, hour, is_before, nr_expected_trades):
+    house1 = next(filter(lambda x: x.name == "House 1", context.simulation.area.children))
+    for market in house1.past_markets:
+        if is_before and market.time_slot.hour >= hour:
+            continue
+        elif not is_before and market.time_slot.hour < hour:
+            continue
+        assert len(market.trades) == nr_expected_trades
+
+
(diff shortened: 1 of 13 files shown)


src/d3a/setup/area_events/disconnect_interval_event.py
```diff
@@ -0,0 +1,50 @@
...
+from d3a.models.strategy.load_hours import LoadHoursStrategy
+
+
+def get_setup(config):
+    area = Area(
+        'Grid',
+        children=[
+            Area(
+                'House 1',
+                event_list=[DisconnectIntervalAreaEvent(6, 16)],
```
faizan2590: Doesn't event_list=[DisconnectAreaEvent(12), ConnectAreaEvent(16)] makes event_list=[DisconnectIntervalAreaEvent(6, 16)] it redundant

src/d3a/models/area/event_types.py
```diff
@@ -0,0 +1,43 @@
...
+
+    def tick(self, current_time):
+        self._active = not self.disconnect_start <= current_time.hour < self.disconnect_end
+
+
+class DisableIntervalAreaEvent(IntervalAreaEvent):
+    pass
+
+
+class DisconnectIntervalAreaEvent(IntervalAreaEvent):
```
faizan2590: Aren't DisconnectIntervalAreaEvent & DisableIntervalAreaEvent redundant and be achieved via IntervalAreaEvent

spyrostz: Not really, because these are different events, disableinterval is for blackouts, disconnectinterval is for islanding.

src/d3a/models/area/events.py
```diff
@@ -0,0 +1,87 @@
...
+class IndividualEvents:
+    def __init__(self, event_list, trigger_type):
+        self.event_list = event_list
+        self.trigger_type = trigger_type
+        self._active = True
+
+    def tick(self, current_time):
+        event_list = sorted(self.event_list, key=lambda e: e.event_time)
+        past_events = list(filter(lambda e: e.event_time <= current_time.hour, event_list))
+        self._active = len(past_events) == 0 or type(past_events[-1]) == self.trigger_type
```
hannesdiedrich: I have problems understanding the logic here. Shouldn't the Event be active if current_time in event_list.event_time ?

spyrostz: No, because the event_list.event_time contain distinct events, and not durations. An example for an event_list = [DisconnectEvent(3), ConnectEvent(5), DisconnectEvent(22)]. In this example the event is going to be active for hours 3-5 and 22-24, which cannot be matched that way. A typical algorithm for handling this case, is to sort the events in chronological order and get all the past events. The past event that is closer to the current time is defining whether the event is active or not. If it is not clear we can discuss this offline.

hannesdiedrich: Now I get it. Thanks!

src/d3a/setup/area_events/disconnect_interval_event.py
hannesdiedrich: I vote for keeping it. And how about adding DisableInterval as well?

hannesdiedrich: Sorry, overlooked that. Thanks!

hannesdiedrich: Sorry, I have a general comment concerning the two types (interval and individual):
IMO the code in area.events is quite convoluted (had a hard time understanding it). Maybe we can get rid of the individual events? IMO an interval is more straight forward fo implement for the user.
What do you think?

integration_tests/steps/area_events.py
```diff
@@ -0,0 +1,85 @@
...
+def load_uses_battery_before(context, hour):
+    # TODO: Should be changed once https://gridsingularity.atlassian.net/browse/D3ASIM-852
+    # is fixed
+    hour = int(hour)
+    house1 = next(filter(lambda x: x.name == "House 1", context.simulation.area.children))
+
+    for market in house1.past_markets:
+        if market.time_slot.hour >= hour:
+            continue
+        print(market.time_slot.hour)
```
hannesdiedrich: These prints can probably be omitted right?

src/d3a/setup/area_events/disconnect_interval_event.py
spyrostz: If it was just for the connect/disconnect and enable/disable I would definitely agree. But the story I am working on right now is about strategy events, in which case interval events do not make sense (the user will probably not be interested in reverting the parameters, just to see the result of the actual change), in which case we need the individual events.

hannesdiedrich: Ah ok, I see. Thanks!

src/d3a/models/area/event_types.py
```diff
@@ -0,0 +1,43 @@
+
```
hannesdiedrich: Please add license header here.

src/d3a/models/area/events.py
```diff
@@ -0,0 +1,87 @@
+from d3a.models.area.event_types import EnableAreaEvent, DisableAreaEvent, ConnectAreaEvent, \
```
hannesdiedrich: Please add license header here.

integration_tests/steps/area_events.py
spyrostz: Thanks for noticing, will do!

src/d3a/models/area/event_types.py
spyrostz: Adding the header, same goes for the following comment.

PR #341, opened 2018-12-17
Feature/d3 asim 854

hannesdiedrich:


hannesdiedrich:
Thanks for you comment @spyrostz . I would be in favour of your suggestion. @faizan2590 What do you think?

src/d3a/models/strategy/update_frequency.py
```diff
@@ -108,15 +109,17 @@ class OfferUpdateFrequencyMixin:
...
-                 energy_rate_decrease_per_update
+                 energy_rate_decrease_per_update,
                  ):
         self.initial_rate_option = InitialRateOptions(initial_rate_option)
         self.energy_rate_decrease_option = RateDecreaseOption(energy_rate_decrease_option)
         self.energy_rate_decrease_per_update = energy_rate_decrease_per_update
         self._decrease_price_timepoint_s = {}  # type: Dict[Time, float]
         self._decrease_price_every_nr_s = 0
         self.min_selling_rate = 0
+        self.initial_selling_rate = initial_selling_rate
```
spyrostz: I have 2 observations here. First is that there is no final_selling_rate. This implies that the final rate is going to be 0 ? What's the outcome of that, because it was not clear from the story description. 
Second, even if you configure the initial_selling_rate, the risk factor is used for interpolation. Is this actually correct behavior, or should these 2 be decoupled? I haven't checked the energy_rate_decrease_per_update, to see if it also uses the risk parameter (hope not).

hannesdiedrich: - 1st: We did not discuss this. IMO it can be an implicit 0, @sarahhambridge what do you think?
- 2nd: IMO they should not be decoupled the initial_selling_rate is only providing the starting point, and should not influence the price decrease. To reinsure you: const_energy_rate_decrease_per_update does not use the risk parameter

sarahhambridge: Hi, it seems to me that the final selling rate should be the min_selling_rate

faizan2590: @hannesdiedrich I agree, renaming it to final_selling_rate makes sense.

hannesdiedrich: Ok, will implement it.

src/d3a/models/const.py
```diff
@@ -28,7 +28,7 @@ class GeneralSettings:
         # Maximum number of area hierarchies. If None, this parameter will be
         # automatically calculated.
         MAX_OFFER_TRAVERSAL_LENGTH = None
-        # Number of times Market clearing rate has to be calculated per slot
+        # Number of times Market cwlearing rate has to be calculated per slot
```
spyrostz: Minor typo, 'clearing'

PR #343, opened 2018-12-20
Feature/d3 asim 805

spyrostz:
Added events for changing parameters on the load, pv and storage strategies.

### DIFF
--- a/src/d3a/models/strategy/load_hours.py
+++ b/src/d3a/models/strategy/load_hours.py
@@ -65,24 +65,8 @@ class LoadHoursStrategy(BaseStrategy, BidUpdateFrequencyMixin):
         self.energy_requirement_Wh = {}  # type: Dict[Time, float]
         self.hrs_per_day = {}  # type: Dict[int, int]
 
-        if hrs_of_day is None:
-            hrs_of_day = list(range(24))
-
-        # be a parameter on the constructor or if we want to deal in percentages
-        if hrs_per_day is None:
-            hrs_per_day = len(hrs_of_day)
-        if hrs_of_day is None:
-            hrs_of_day = list(range(24))
-
-        self.hrs_of_day = hrs_of_day
-        self._initial_hrs_per_day = hrs_per_day
+        self.assign_hours_of_per_day(hrs_of_day, hrs_per_day)
         self.balancing_energy_ratio = BalancingRatio(*balancing_energy_ratio)
-
-        if not all([0 <= h <= 23 for h in hrs_of_day]):
-            raise ValueError("Hrs_of_day list should contain integers between 0 and 23.")
-
-        if len(hrs_of_day) < hrs_per_day:
-            raise ValueError("Length of list 'hrs_of_day' must be greater equal 'hrs_per_day'")
 
     @property
     def active_markets(self):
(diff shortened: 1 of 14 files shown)


src/d3a/models/strategy/pv.py
```diff
@@ -62,15 +62,25 @@ def __init__(
...
                              "between 0 and 100, panel_count should be positive.")
-        if max_panel_output_W < 0:
+        if max_panel_output_W is not None and max_panel_output_W < 0:
             raise ValueError("Max panel output in Watts should always be positive.")
-        if initial_selling_rate < 0:
-            raise ValueError("Initial selling rate must be greater equal 0.")
+        if initial_selling_rate is not None and initial_selling_rate < 0:
+            raise ValueError("Min selling rate should be positive.")
+
+    def area_reconfigure_event(self, **kw):
```
hannesdiedrich: Very minor, but IMO kw is not self-explaining. What does it stand for? Keyword?

src/d3a/models/strategy/pv.py
```diff
@@ -99,14 +109,15 @@ def produced_energy_forecast_kWh(self):
...
         for slot_time in generate_market_slot_list(self.area):
-            difference_to_midnight_in_minutes = \
-                slot_time.diff(self.midnight).in_minutes() % (60 * 24)
-            self.energy_production_forecast_kWh[slot_time] = \
-                self.gaussian_energy_forecast_kWh(
-                    difference_to_midnight_in_minutes) * self.panel_count
-            self.state.available_energy_kWh[slot_time] = \
-                self.energy_production_forecast_kWh[slot_time]
-            assert self.energy_production_forecast_kWh[slot_time] >= 0.0
+            if slot_time >= self.area.now:
```
hannesdiedrich: Sorry, i don't get it: Why is this check needed?

src/d3a/models/strategy/pv.py
spyrostz: Stands for kwargs == 'keyword arguments'. Since it is confusing as is, I will rename it to kwargs

src/d3a/models/strategy/pv.py
spyrostz: This is needed now because this method is not only called at the start of the simulation, but can be called in-between via an event. In that case, if this check does not exist, self.state.available_energy_kWh will be modified even for past market slots, and this is not something that we want, since the event is supposed to modify the state of the strategy from this moment and on.

hannesdiedrich: Ah yes, thank you!

src/d3a/models/strategy/pv.py
```diff
@@ -34,7 +34,7 @@ class PVStrategy(BaseStrategy, OfferUpdateFrequencyMixin):
                 help="Change the risk parameter. Valid values are between 1 and 100.")
     ]
 
-    parameters = ('panel_count', 'risk')
+    parameters = ('panel_count', 'risk', 'max_panel_power_W', 'final_selling_rate')
```
hannesdiedrich: Can you please add initial_selling_rate here as well (i will add it in the d3a-web as well)

spyrostz: Will do, thanks!

src/d3a/setup/area_events/pv_event.py
```diff
@@ -0,0 +1,48 @@
...
+"""
+from d3a.models.appliance.switchable import SwitchableAppliance
+from d3a.models.area import Area
+from d3a.models.area.events import StrategyEvents
+from d3a.models.strategy.pv import PVStrategy
+from d3a.models.appliance.pv import PVAppliance
+from d3a.models.strategy.load_hours import LoadHoursStrategy
+
+
+def get_setup(config):
```
faizan2590: We used to have a short description of expected behaviour inside a setup file. Shouldn't we restart adding short description of expected behaviour?

spyrostz: I thought that the expected behavior is clear here, that's why I did not add it. The setup file consists of just one load and one PV, and on the PV there is an event that increases the number of panels from 1 to 10, and the max panel power from 160 to 320. This has the effect that the max power of the PV is multiplied by 20. Its purpose is to use it in the integration test to validate that the event actually triggers this increase in PV power.

src/d3a/models/strategy/load_hours.py
```diff
@@ -73,38 +85,45 @@ def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None, daily_budget=
...
-        self.balancing_energy_ratio = BalancingRatio(*balancing_energy_ratio)
 
         if not all([0 <= h <= 23 for h in hrs_of_day]):
             raise ValueError("Hrs_of_day list should contain integers between 0 and 23.")
 
         if len(hrs_of_day) < hrs_per_day:
             raise ValueError("Length of list 'hrs_of_day' must be greater equal 'hrs_per_day'")
 
-    @property
-    def active_markets(self):
```
faizan2590: It is deleted here and then re-added above. Any particular reason for change of position

PR #344, opened 2018-12-24
D3ASIM-867: Fixing infinite trade in Pay-As-Clear Market

faizan2590:


sarahhambridge:
@faizan2590 @dirkbig  See comment above ^^

PR #345, opened 2018-12-24
D3ASIM-804: Added implementation for config change events on d3a. Cur…

spyrostz:
…rently only cloud coverage and pv_user_profile settings are supported, more will be added on the future.

### DIFF
--- a/integration_tests/steps/area_events.py
+++ b/integration_tests/steps/area_events.py
@@ -1,4 +1,7 @@
 from behave import then
+from d3a.models.read_user_profile import read_profile_csv_to_dict, InputProfileTypes
+from d3a.d3a_core.util import d3a_path
+from pendulum import duration
 
 
 def one_or_none_trade_before_or_after(context, hour, is_before, nr_expected_trades):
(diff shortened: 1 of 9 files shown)


integration_tests/steps/area_events.py
```diff
@@ -153,3 +156,64 @@ def load_does_not_consume_between(context, start_time, end_time):
...
+    slot_length = context.simulation.area.config.slot_length
+    sunny = _read_solar_profile(d3a_path + '/resources/Solar_Curve_W_sunny.csv', slot_length)
+    house1 = next(filter(lambda x: x.name == "House 1", context.simulation.area.children))
+    house2 = next(filter(lambda x: x.name == "House 2", context.simulation.area.children))
+
+    for time_slot, market in house1._markets.past_markets.items():
+        if time_slot.hour > 12:
+            continue
+        h1_trades = [t for t in market.trades if t.seller == "H1 PV"]
+        h2_trades = [t for t in house2._markets.past_markets[time_slot].trades
```
hannesdiedrich: Why are the h2_trades determined different from the h1_trades?

src/d3a/models/config.py
```diff
@@ -86,6 +77,27 @@ def as_dict(self):
...
+        # TODO: Once the d3a uses a common API to the d3a-web, this should be removed
+        # since this limitation already exists on d3a-web
+        if 0 <= cloud_coverage <= 2:
+            self.cloud_coverage = cloud_coverage
+        else:
+            raise D3AException("Invalid cloud coverage value ({}).".format(cloud_coverage))
+
+    def read_pv_user_profile(self, pv_user_profile=None):
+        self.pv_user_profile = None \
+            if pv_user_profile is None \
```
hannesdiedrich: Why this check again? Isn't it performed in L84 already?

integration_tests/steps/area_events.py
spyrostz: This is necessary in order to get the trades that happen in the same timeslot in 2 different IAAs. The rationale behind this is to iterate over all markets of house 1, and get the trades that happened in house 2 at this market slot in order to see at this exact point in time what happened to all the IAAs.

hannesdiedrich: Ahh, now i get it. `market` is the house_1 market. 
Ok, Thanks!

src/d3a/models/config.py
hannesdiedrich: `Yes, but this method is also called from line 49 (Config constructor)` Sorry, did not see that.
And yes, I agree: If we add more user profiles, we need to refactor.
Thanks!

PR #346, opened 2019-01-04
Feature/d3 asim 869

hannesdiedrich:
This PR contains the following features:
- Implementation of a new IAA for the alternative price schemes, that buys energy surplus from PV in Houses
- Enabled PV and Storage strategy for alternative pricing
- refactored _max_selling_rate() in storage.py because it is a duplicate of OfferUpdateMixin. calculate_initial_sell_rate()
- added setup file for testing

This PR does not incorporate: 
- integration tests (will be added in D3ASIM-868)


src/d3a/d3a_core/export.py
```diff
@@ -71,15 +71,15 @@ def export_json_data(self, directory: dir):
         mkdir_from_str(json_dir)
         kpi_file = os.path.join(json_dir, "KPI")
         with open(kpi_file, 'w') as outfile:
-            json.dump(self.kpi.performance_index, outfile)
+            json.dump(self.kpi.performance_index, outfile, indent=2)
```
spyrostz: Not a problem at all to have 2-space indentation for JSON files, so either way if you prefer this format keep it, but out of curiosity was there a need/request/reason?

src/d3a/models/const.py
```diff
@@ -104,6 +104,13 @@ class IAASettings:
         # Option 2 stands for double sided pay as bid market
         # Option 3 stands for double sided pay as clear market
         MARKET_TYPE = 1
+        # PRICING_SCHEME
```
spyrostz: If there are more settings in the following subtasks (like for multiple processes), I would suggest to create yet another level of nesting, in order to distinguish it in a better way.

src/d3a/models/strategy/area_agents/one_sided_alternative_pricing_agent.py
```diff
@@ -0,0 +1,72 @@
...
+
+    @staticmethod
+    def _get_children_by_name(area, name):
+        for child in area.children:
+            if child.name == name:
+                return child
+        return None
+
+    def _buy_energy_alternative_pricing_schemes(self, area):
+        if _is_house_node(self.owner):
```
spyrostz: In order to avoid nesting, I would negate this if to:
if not _is_house_node: return

src/d3a/models/strategy/area_agents/one_sided_alternative_pricing_agent.py
```diff
@@ -0,0 +1,72 @@
...
+            try:
+                for offer in self.lower_market.sorted_offers:
+                    if offer.seller != self.name:
```
spyrostz: Same as above, to avoid levels of nesting the condition could be negated:
if offer.seller == self.name: continue

src/d3a/d3a_core/export.py
hannesdiedrich: This is just for better readability of the aggregated_result files.
Before, the results were just in one line.

src/d3a/models/const.py
hannesdiedrich: Ok! Good Idea!

src/d3a/models/strategy/area_agents/one_sided_alternative_pricing_agent.py
```diff
@@ -0,0 +1,72 @@
...
+                            offer.price = offer.energy * sell_rate
+                            self.accept_offer(offer.market, offer)
+
+            except MarketException:
+                self.log.exception("Alternative pricing scheme: "
+                                   "An Error occurred while buying an offer")
+
+    def event_tick(self, *, area):
+
+        # The following is an artificial number but has to be >= 2:
```
spyrostz: I would move the min_slot_age to the module scope and I would change the naming to upper-case, in order to denote that it is a const for this module and not a normal variable.

src/d3a/models/strategy/pv.py
```diff
@@ -93,10 +96,11 @@ def event_activate(self):
     def _incorporate_rate_restrictions(self, initial_sell_rate, current_time):
         energy_rate = max(initial_sell_rate, self.final_selling_rate)
         rounded_energy_rate = round(energy_rate, 2)
-        if rounded_energy_rate == 0.0:
-            # Initial selling offer
-            rounded_energy_rate =\
-                self.area.config.market_maker_rate[current_time]
+        # TODO: Find out the reason for this block, at the moment it does not make sense
```
spyrostz: My guess is that it is one of the last traces of the Brainbot implementation. There probably was a convention where if the initial energy rate configuration of the PV was 0, the market maker rate was used instead, that's why this still exists up to this day. I would even go ahead and remove this completely, but if you want to play it safe leave it as is.

src/d3a/models/strategy/storage.py
```diff
@@ -126,6 +126,22 @@ def area_reconfigure_event(self, risk=None, initial_rate_option=None,
             self.state.min_allowed_soc = min_allowed_soc
 
     def event_activate(self):
+        if ConstSettings.IAASettings.PRICING_SCHEME != 0:
```
spyrostz: Since this is a repetitive pattern to assign these on the event_activate method, I would create a base class method on the BaseStrategy that would be overridden by the subclasses with these conditions. That way there will be a common entry point for all strategies, and it might also be possible to use the BaseStrategy event_activate method for calling this entry point. That way the event_activate methods of each individual strategy will remain untouched. If it is not that clear let me know to discuss this offline.

src/d3a/models/strategy/storage.py
```diff
@@ -263,9 +279,13 @@ def buy_energy(self, market):
             if offer.seller == self.owner.name:
                 # Don't buy our own offer
                 continue
+            if "IAA" in offer.seller \
```
spyrostz: This is OK for now, but I would leave a TODO to find a better way to distinguish between inter area agents and devices. This is a pending issue in other places as well, which we have not dealt with yet.

src/d3a/d3a_core/export.py
spyrostz: I see, very nice! Thanks for explaining

src/d3a/models/strategy/pv.py
hannesdiedrich: Yes, I am always in favour of removing code. ;-) 
So will do.

src/d3a/models/strategy/storage.py
hannesdiedrich: I know what you mean, but here this pattern is used to set the break even point, which is storage-specific. 
Additionally, only in the storage this pattern is called in the event_activate. 
For setting the sell prices in the PV and Storage I already refactored this pattern into set_initial_selling_rate_alternative_pricing_scheme() . 

Or do you mean only the following two lines?:
```
        if ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0:
            self.assign_offermixin_arguments(3, 2, 0)
```

hannesdiedrich: For the latter, I could refactor it into the base class.

hannesdiedrich: Spyros and me did offline discussion and refactoring.

src/d3a/models/strategy/area_agents/one_sided_alternative_pricing_agent.py
spyrostz: Hmm, just in order to be visible, I would move it outside of the function (inside the class though, right under the class ... line should be ok)

hannesdiedrich: Unfortunately, it is not possible to put a "global" constant into the class. Either it is defined globally in the file or it is a self member. What do you prefer? I dont have a strong opinion on that.

spyrostz: You can do it, but you have to access the variable either via self. or via ClassName. On second thought though, I would suggest to move it globally in the file/module. 
It is not that important, I just would like to prevent someone changing this variable without knowing that it is actually a constant.

hannesdiedrich: Ahh, sorry, my bad. I have to access via self. OK. Applied! Thanks!

src/d3a/models/const.py
```diff
@@ -105,6 +105,14 @@ class IAASettings:
         # Option 3 stands for double sided pay as clear market
         MARKET_TYPE = 1
 
+        class AlternativePricing:
+            # Option 0: D3A_trading, 1=no scheme, 2=feed-in-tariff, 3=net-metering,
```
faizan2590: Probably just
```suggestion
            # Option 0: D3A_trading
```

PR #348, opened 2019-01-07
D3ASIM-884: Accepting bids in 2sidedPAC

faizan2590:


src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -115,7 +115,6 @@ def _match_offers_bids(self):
                     bid._replace(price=(bid.energy * clearing_rate), energy=bid.energy),
                     energy=bid.energy,
                     seller=self.owner.name,
-                    already_tracked=True,
```
spyrostz: From what I understand this change is not related to the main issue, why is it necessary? I am asking because in this case there are duplicate trades, which stem from the fact that the bid/offer match should result to one trade, but if both are tracked then there are going to be 2 trades tracked.

PR #350, opened 2019-01-08
D3ASIM-890: Added assertion on the storage state constructor with mor…

spyrostz:
…e detailed explanation if the user has misconfigured the initial capacity of the storage to exceed the min allowed soc.

Sample output on the terminal:
...
  File "/Users/spyrostzavikas/dev/d3a/src/d3a/models/state.py", line 97, in __init__
    f"Initial capacity ({initial_capacity_kWh} kWh) exceeds the " \
AssertionError: Initial capacity (0.0 kWh) exceeds the min allowed soc (10.0%).

### DIFF
--- a/src/d3a/models/state.py
+++ b/src/d3a/models/state.py
@@ -92,6 +92,11 @@ class StorageState:
 
         if min_allowed_soc is None:
             min_allowed_soc = StorageSettings.MIN_ALLOWED_SOC
+
+        assert limit_float_precision(initial_capacity_kWh / capacity) >= min_allowed_soc, \
+            f"Initial capacity ({initial_capacity_kWh} kWh) exceeds the " \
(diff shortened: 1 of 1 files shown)


src/d3a/models/state.py
```diff
@@ -92,6 +92,11 @@ def __init__(self,
 
         if min_allowed_soc is None:
             min_allowed_soc = StorageSettings.MIN_ALLOWED_SOC
+
+        assert limit_float_precision(initial_capacity_kWh / capacity) >= min_allowed_soc, \
+            f"Initial capacity ({initial_capacity_kWh} kWh) exceeds the " \
```
faizan2590:
```suggestion
            f"Initial capacity ({initial_capacity_kWh} kWh) less than the " \
```

PR #352, opened 2019-01-09
Bug/d3 asim 884 1

faizan2590:
The original bug was that Loads were shown as unmatched despite being accepted. The problem was that bids were accepted with flag already_tracked=True.
While solving it, another bug got apparent of having same buyers/sellers of some bids/offer. Since, bids/offers are not directly matched in 2PAC and the offer buyer is the IAA whereas bids' seller is also IAA.


integration_tests/steps/two_sided_pay_as_clear.py
```diff
@@ -32,6 +32,18 @@ def test_traded_energy_rate(context):
...
 
+@then('buyers and sellers are not same')
+def test_different_buyer_seller(context):
+    grid = context.simulation.area
+    for child in grid.children:
+        for a, b in child.dispatcher.interarea_agents.items():
+            for c in b:
+                for d in c.engines:
+                    for trade in d.markets.source.trades:
+                        print(f"Seller: {trade.offer.seller} & buyer: {trade.buyer}")
```
hannesdiedrich: Please remove print.

integration_tests/two_sided_pay_as_clear_market.feature
```diff
@@ -5,6 +5,7 @@ Feature: Two sided pay_as_clear market tests
      And d3a is installed
      When we run the simulation with setup file two_sided_pay_as_clear.default_2a and parameters [24, 15, 15, 0, 1]
      Then all trades are equal to market_clearing_rate
+     And buyers and sellers are not same
```
hannesdiedrich: So you introduced the `already_tracked` flag to disable that devices buy their own energy? Did I get it right? 
Was that the problem? Does not seam to coincide with the Jira ticket description.

integration_tests/steps/two_sided_pay_as_clear.py
```diff
@@ -32,6 +32,18 @@ def test_traded_energy_rate(context):
                                         for rate in d.clearing_rate])
 
 
+@then('buyers and sellers are not same')
+def test_different_buyer_seller(context):
+    grid = context.simulation.area
+    for child in grid.children:
```
spyrostz: Please remove the nested for loops as it makes it unreadable, you do not need to access the markets via the IAA class, you can achieve the same result via the area classes.

integration_tests/two_sided_pay_as_clear_market.feature
spyrostz: It was discussed already offline, but I would like to point out that we need to use Jira more an write comments if the story context has changed or if the root cause of the error is completely different from the description of the issue. That will make it easier for everyone to keep track of the issue, especially during reviews.

hannesdiedrich: And I also suggest to use the general comment field of the PR (when opening) for a short summary what has been done.
A lot of times I realised, that I forgot to add stuff when I was writing this summary.

integration_tests/steps/integration_tests.py
```diff
@@ -320,7 +320,8 @@ def test_offer_bid_files(context):
                  os.path.join(base_path, 'grid', 'house-2-balancing-offers.csv')]
 
     assert all(len(glob.glob(f)) == 1 for f in file_list)
-    assert all(len(open(glob.glob(f)[0]).readlines()) > 1 for f in file_list)
+    for f in file_list:
```
spyrostz: It will not be possible to see the filename like that. You need to add a string to the assert to display the exact filename that failed

integration_tests/steps/two_sided_pay_as_clear.py
```diff
@@ -32,6 +33,41 @@ def test_traded_energy_rate(context):
...
+    areas.append(grid)
+    house1 = next(filter(lambda x: x.name == "House 1", context.simulation.area.children))
+    areas.append(house1)
+    for area in areas:
+        for market in area.past_markets:
+            for trade in market.trades:
+                assert str(trade.offer.seller) != str(trade.buyer)
+
+
+@then('cummulative traded offer energy equal to cumulative bid energy')
```
spyrostz: Trivial typo, cummulative -> cumulative.

src/d3a/setup/two_sided_pay_as_clear/test_clearing_energy.py
```diff
@@ -0,0 +1,71 @@
...
+    ConstSettings.LoadSettings.INITIAL_BUYING_RATE = 35
+    ConstSettings.LoadSettings.FINAL_BUYING_RATE = 35
+    ConstSettings.StorageSettings.MIN_BUYING_RATE = 24.99
+    ConstSettings.StorageSettings.BREAK_EVEN_BUY = 25
+    ConstSettings.StorageSettings.BREAK_EVEN_SELL = 25.01
+
+    area = Area(
+        'Grid',
+        [
+            Area(
```
spyrostz: Well, this is ok for now, but the problem with this file is that it does not test offer/bid forwarding, because there is only one house. If there were 2 houses we would be able to see whether bids/offers are forwarded properly. But anyways, as I said before, it is ok for now

faizan2590: Well, the target of this setup file is-to specifically test if cumulative traded offered_energy equals to bid_energy within an area. Not sure if this is the right approach.

integration_tests/steps/integration_tests.py
hannesdiedrich: Wow! Great, did not know the functionality of assert (adding an error string) 
Thanks for that!

src/d3a/d3a_core/simulation.py
```diff
@@ -141,6 +141,7 @@ def _init(self, slowdown, seed, paused, pause_after):
         else:
             random_seed = random.randint(0, 1000000)
             random.seed(random_seed)
+            self.initial_params['seed'] = random_seed
```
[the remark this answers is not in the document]
spyrostz: Could you please assign it on its own variable though, and not use the initial_params for that ? Initial params is also used for defining the initial simulation state, therefore it could have an effect on the random test failures we have encountered. At least this is the only part where nondeterministic behavior could be introduced in the simulation.

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -167,5 +172,5 @@ def tick(self, *, area):
         self.mcp_update_point = \
             area.config.ticks_per_slot / \
             ConstSettings.GeneralSettings.MARKET_CLEARING_FREQUENCY_PER_SLOT
-        if current_tick_number % int(self.mcp_update_point) == 0:
+        if (current_tick_number+1) % int(self.mcp_update_point) == 0 and current_tick_number != 0:
```
hannesdiedrich: I understand this part: `and current_tick_number != 0`
But why the +1 for current_tick_number ?

faizan2590: current_tick_number = 0 is the first tick, last tick is 59 or 60th

PR #353, opened 2019-01-09
Feature/d3 asim 874

spyrostz:
Added ability on the alternative pricing agent to place an infinite energy offer with the current market maker rate at the beginning of every market cycle. This enables the devices to buy energy from the IAA in case the house energy production is not sufficient to cover the energy consumption.

### DIFF
--- a/src/d3a/models/strategy/area_agents/one_sided_alternative_pricing_agent.py
+++ b/src/d3a/models/strategy/area_agents/one_sided_alternative_pricing_agent.py
@@ -15,6 +15,8 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+import sys
+
 from d3a.models.strategy.area_agents.one_sided_agent import OneSidedAgent
 from d3a.models.strategy.area_agents.one_sided_engine import IAAEngine
 from d3a.d3a_core.exceptions import MarketException
@@ -51,11 +53,11 @@ class OneSidedAlternativePricingAgent(OneSidedAgent):
                         sell_rate = 0
                     elif ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME == 2:
                         sell_rate = \
-                            ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE * \
+                            area.config.market_maker_rate[self.lower_market.time_slot_str] * \
                             ConstSettings.IAASettings.AlternativePricing.FEED_IN_TARIFF_PERCENTAGE\
                             / 100
                     elif ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME == 3:
-                        sell_rate = ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE
+                        sell_rate = area.config.market_maker_rate[self.lower_market.time_slot_str]
                     else:
                         raise MarketException
                     offer.price = offer.energy * sell_rate
@@ -66,8 +68,19 @@ class OneSidedAlternativePricingAgent(OneSidedAgent):
                                "An Error occurred while buying an offer")
 
     def event_tick(self, *, area):
-        if area.current_tick_in_slot - self.MIN_SLOT_AGE >= 0 and \
+        if area.current_tick_in_slot >= self.MIN_SLOT_AGE and \
                 ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0:
             self._buy_energy_alternative_pricing_schemes(area)
 
-        super().event_tick(area=area)
+        # super().event_tick(area=area)
+
+    def event_market_cycle(self):
+        if ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0:
+            energy_per_slot = sys.maxsize
+            energy_rate = self.owner.config.market_maker_rate[self.lower_market.time_slot_str]
+
+            self.lower_market.offer(
+                energy_per_slot * energy_rate,
+                energy_per_slot,
+                ConstSettings.GeneralSettings.ALT_PRICING_MARKET_MAKER_NAME
+            )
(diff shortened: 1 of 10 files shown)


src/d3a/models/strategy/area_agents/one_sided_alternative_pricing_agent.py
```diff
@@ -66,8 +68,19 @@ def _buy_energy_alternative_pricing_schemes(self, area):
                                "An Error occurred while buying an offer")
 
     def event_tick(self, *, area):
-        if area.current_tick_in_slot - self.MIN_SLOT_AGE >= 0 and \
+        if area.current_tick_in_slot >= self.MIN_SLOT_AGE and \
```
hannesdiedrich: Ah yeah, thank you! This looks much better!

src/d3a/models/strategy/area_agents/one_sided_alternative_pricing_agent.py
```diff
@@ -66,8 +68,19 @@ def _buy_energy_alternative_pricing_schemes(self, area):
...
                 ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0:
             self._buy_energy_alternative_pricing_schemes(area)
 
-        super().event_tick(area=area)
+        # super().event_tick(area=area)
```
hannesdiedrich: Please remove this or add a comment with reason for keeping it. Thanks!

spyrostz: Ah, you are right! Removing it completely, no reason for it to stay.

src/d3a/models/strategy/area_agents/one_sided_alternative_pricing_agent.py
```diff
@@ -66,8 +68,19 @@ def _buy_energy_alternative_pricing_schemes(self, area):
...
+
+    def event_market_cycle(self):
+        if ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0:
+            energy_per_slot = sys.maxsize
```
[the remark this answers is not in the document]
spyrostz: Thanks for noticing that, I will remove it now.

src/d3a/models/strategy/commercial_producer.py
```diff
@@ -29,7 +28,7 @@ def __init__(self, energy_rate=None):
         if energy_rate is not None and energy_rate < 0:
             raise ValueError("Energy rate should be positive.")
         super().__init__()
-        self.energy_per_slot_kWh = Q_(int(sys.maxsize), ureg.kWh)
+        self.energy_per_slot_kWh = int(sys.maxsize)
```
faizan2590: Any particular reason to get rid of units?

spyrostz: Yes, because it was the last place in the source code that they were used. I remember we decided some time ago to remove these, because they are useful only if you enforce these throughout the source code (as you can see here, the units are casted away by using .m when using them, while this should not be the case, the units should be used as they were).

src/d3a/models/strategy/area_agents/one_sided_alternative_pricing_agent.py
```diff
@@ -66,8 +68,17 @@ def _buy_energy_alternative_pricing_schemes(self, area):
...
-        if area.current_tick_in_slot - self.MIN_SLOT_AGE >= 0 and \
+        if area.current_tick_in_slot >= self.MIN_SLOT_AGE and \
                 ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0:
             self._buy_energy_alternative_pricing_schemes(area)
 
-        super().event_tick(area=area)
+    def event_market_cycle(self):
+        if ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0:
+            energy_per_slot = int(sys.maxsize)
+            energy_rate = self.owner.config.market_maker_rate[self.lower_market.time_slot_str]
```
faizan2590: In case of no-scheme, the rate should be zero. So, shouldn't this be based on pricing scheme rather than equal to MMR?

spyrostz: This code is referring to the trade from the IAA to the house loads. Therefore, even in the no-scheme, the rate should not be zero for this, it should be the market maker rate, since the IAA does not sell energy for free to the house devices.

faizan2590: Then, I guess we should simply remove the no scheme option?

spyrostz: Not really, because the no-scheme option refers to the trade from the device to the IAA, it denotes that the energy surplus of the house is given away for free to the TSO.

src/d3a/models/strategy/storage.py
```diff
@@ -284,10 +284,10 @@ def buy_energy(self, market):
...
                 continue
 
-            # TODO: find a better was to distinguish between IAA and devices
-            if "IAA" in offer.seller \
-                    and ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0:
-                # dont buy from IAA if alternative pricing scheme is activated
+            alt_pricing_settings = ConstSettings.IAASettings.AlternativePricing
+            if offer.seller == alt_pricing_settings.ALT_PRICING_MARKET_MAKER_NAME \
+                    and alt_pricing_settings.PRICING_SCHEME != 0:
+                # don't buy from IAA if alternative pricing scheme is activated
```
faizan2590: if alt_pricing_settings.PRICING_SCHEME != 0:
Shouldn't this be enough?

spyrostz: It is a safety precaution, for the storage to never buy energy from the IAA, it is not necessary per se, but I feel safer to have it here.

faizan2590: Then I guess it should be with or. Right?
            if offer.seller == alt_pricing_settings.ALT_PRICING_MARKET_MAKER_NAME \
                    or alt_pricing_settings.PRICING_SCHEME != 0:

hannesdiedrich: No, the `offer.seller == alt_pricing_settings.ALT_PRICING_MARKET_MAKER_NAME` constraint should only be applied if an alternative pricing scheme is selected. Otherwise you disable buying from IAA for all cases.

PR #355, opened 2019-01-10
Feature/d3 asim 868

hannesdiedrich:
Whats has been done in the PR:
- Removed REST API and and its debris from code
- Refactored cli parameters as well as simulation run call
- Added multiprocessing for alternative pricing schemes and the export into subfolders

This branch MUST be merged at the same time with the merge of the corresponding d3a-web branch!


integration_tests/steps/integration_tests.py
```diff
@@ -203,14 +203,12 @@ def running_the_simulation(context):
     paused = False
     pause_after = duration()
     repl = False
-    export = False
+    no_export = True
```
spyrostz: I haven't looked at the d3a-web PR so this might already be handled, but we need to also check that the d3a on the production server does not export simulation results, because this could lead to filling up the disk space of the d3a simulation pod.

hannesdiedrich: Yes you are right, i tried to deal with that by added this line to the d3a-web.py:
`kwargs = {"no_export": True}`

src/d3a/d3a_core/cli.py
```diff
@@ -61,9 +64,9 @@ def main(log_level):
 @main.command()
 @click.option('-d', '--duration', type=IntervalType('D:H'), default="1d", show_default=True,
               help="Duration of simulation")
-@click.option('-t', '--tick-length', type=IntervalType('M:S'), default="1s", show_default=True,
+@click.option('-t', '--tick_length', type=IntervalType('M:S'), default="1s", show_default=True,
```
spyrostz: I would stick to use a dash for the CLI parameters, and I would also fix the CLI parameters that use underscore. It is somehow a standard in most command lines to never use _ and always use - for word separation.

hannesdiedrich: Ahh, i see, and it gets mapped from "-" to "_" automatically by click?

spyrostz: Yes, you can also check in the changes before the pause-after parameter, which is mapped to pause_after variable

src/d3a/d3a_core/simulation.py
```diff
@@ -72,24 +71,20 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig =
         )
 
         self.simulation_config = simulation_config
-        self.use_repl = use_repl
-        self.export_on_finish = export
+        self.use_repl = repl
+        self.export_on_finish = not no_export
         self.export_path = export_path
+        if export_subdir is None:
```
spyrostz: If the export_subdir is set to something other than None, there is no 'self.export_subdir = export_subdir' line, therefore its value would be lost. Probably I am not seeing something here though.

hannesdiedrich: No you are right! Adding it!

hannesdiedrich:
@spyrostz  @faizan2590 
I just discovered, that if you set the `ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME` in the setup file that you use for `compare_alt_pricing`, only THIS pricing scheme will be used in the different runs. This is because the setup file is read after setting the pricing scheme in run_simulation. 
I propose to move the setting of the PRICING_SCHEME into the Simulation class (after the reading of the setup_file) OK? 
Shall I implement it in this branch or in the integrationtest branch?

PR #358, opened 2019-01-17
Feature/d3 asim 846

spyrostz:
For this story I have created scripts that create minimal grid setups for PandaPower and PyPSA, in order to evaluate and to have a proof-of-concept for our cases. In order to run the scripts one has to install the corresponding python package first (I did not add either in the requirements yet because we need to decide which one to use first), therefore in order to test these one needs to 'pip install pypsa pandapower'. 
I am preparing a separate confluence page as we speak for documenting the results, but to our perspective the API of both tools is pretty similar, and it is quite easy to create clones of our setup files. The main difference that I noticed is that on PyPSA one needs to create a slack generator in order to simulate a 'main grid connection', because by default the buses operate according to the control mode of the connected devices/generators. This is not true in pandapower, where every bus bar tries to perform power flow and does not depend on its connected devices. Furthermore, I did not manage to tune PyPSA to perform nonlinear power flow, even though it should be possible according to documentation, therefore only the real part of the calculations take place (active power-frequency only, no reactive power and voltage calculations take place). Pandapower performs nonlinear power flow, therefore voltage results are available. 
Both tools have been configured to dump output files under d3a/tools/ directory, which display results of the simulation for all the components of the grid.

### DIFF
--- a/tools/pandapower_eval.py
+++ b/tools/pandapower_eval.py
@@ -1,3 +1,4 @@
+# In order to install PandaPower run 'pip install pandapower'
 import pandapower as pp
 from pandapower.plotting import to_html
 
--- a/tools/pandapower_eval.py
+++ b/tools/pandapower_eval.py
@@ -1,4 +1,4 @@
-# In order to install PandaPower run 'pip install pandapower'
+# In order to install PandaPower run 'pip install matplotlib pandapower'
 import pandapower as pp
 from pandapower.plotting import to_html
 
--- a/tools/pandapower_eval.py
+++ b/tools/pandapower_eval.py
@@ -1,0 +1,35 @@
+import pandapower as pp
+from pandapower.plotting import to_html
+
+
+# create empty net
+net = pp.create_empty_network()
+
+# create buses
+bus1 = pp.create_bus(net, vn_kv=20., name="Bus Grid 1")
+bus_trafo = pp.create_bus(net, vn_kv=0.4, name="Bus Trafo")
+bus_h1 = pp.create_bus(net, vn_kv=0.4, name="Bus House 1")
+bus_h2 = pp.create_bus(net, vn_kv=0.4, name="Bus House 2")
+
+# create bus elements
+pp.create_ext_grid(net, bus=bus1, vm_pu=1.0, name="Grid Connection")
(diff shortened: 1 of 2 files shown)


tools/pandapower_eval.py
```diff
@@ -0,0 +1,36 @@
+# In order to install PandaPower run 'pip install pandapower'
```
hannesdiedrich: You also need to install matplotlib with `pip install matplotlib`

spyrostz: Hmm, I had the impression that it was a dependency of pandapower. I will include it to the instructions here.

hannesdiedrich: Yeah, that would be smart of pandapower,  but it wasnt installed automatically in my case.

tools/pandapower_eval.py
```diff
@@ -0,0 +1,36 @@
...
+# create buses
+bus1 = pp.create_bus(net, vn_kv=20., name="Bus Grid 1")
+bus_trafo = pp.create_bus(net, vn_kv=0.4, name="Bus Trafo")
+bus_h1 = pp.create_bus(net, vn_kv=0.4, name="Bus House 1")
+bus_h2 = pp.create_bus(net, vn_kv=0.4, name="Bus House 2")
+
+# create bus elements
+pp.create_ext_grid(net, bus=bus1, vm_pu=1.0, name="Grid Connection")
+pp.create_load(net, bus=bus_h1, p_kw=0.1, q_kvar=0.005, name="H1 Load")
+pp.create_sgen(net, bus=bus_h1, p_kw=-0.1, name="H1 PV")
```
faizan2590: In case of PV, we have power profile rather than one power value. So do we have to update p_kw for every market_slot?

spyrostz: I think it was mentioned in the documentation too, none of these tools support dynamic analysis, only steady state analysis is supported for both pandapower and PyPSA. Therefore there is no straightforward way of providing 'power profiles' to the software and have the analysis ready (there is a way via 'snapshots', but it is the same as changing the values and manually performing the power flow analysis sequentially).

faizan2590: Sorry if I missed it, for PyPSA it is explicitly written that it supports steady-state analysis but for pandapower I couldn't find it there. Thanks for the clarification

PR #361, opened 2019-01-21
D3ASIM-903: Refactoring market_clearing_rate to state class

faizan2590:


src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -20,20 +20,23 @@
...
                  owner: "InterAreaAgent"):
-        super().__init__(name, market_1, market_2, min_offer_age, transfer_fee_pct, owner)
+        TwoSidedPayAsBidEngine.__init__(self, name, market_1, market_2, min_offer_age,
+                                        transfer_fee_pct, owner)
+        MarketClearingState.__init__(self)
         self.forwarded_bids = {}  # type: Dict[str, BidInfo]
         self.sorted_bids = []
         self.sorted_offers = []
-        self.clearing_rate = []  # type: List[int]
+        # self.clearing_rate = []  # type: List[int]
```
hannesdiedrich: Please remove.

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -87,7 +87,7 @@ def _delete_forwarded_bid_entries(self, bid):
         self.forwarded_bids.pop(bid_info.target_bid.id, None)
         self.forwarded_bids.pop(bid_info.source_bid.id, None)
 
-    def _match_offers_bids(self):
+    def _match_offers_bids(self, time=None):
```
hannesdiedrich: time is not used inside the function

src/d3a/models/state.py
```diff
@@ -234,3 +234,8 @@ def market_cycle(self, past_time_slot, time_slot):
         self.charge_history[time_slot] = 100.0 * self.used_storage / self.capacity
         self.charge_history_kWh[time_slot] = self.used_storage
         self.offered_history[time_slot] = self.offered_sell_kWh[time_slot]
+
+
+class MarketClearingState:
```
hannesdiedrich: Can you please explain, why this has to be a separate class? Cant and shouldnt `self.clearing_rate` be a member of TwoSidedPayAsBidEngine and TwoSidedPayAsClearEngine ?

spyrostz: I can elaborate a bit more on that, the problem here is that this is not actually a simple variable, but it actually is historical data. The root cause of the issue here was that clearing rate is not actually one value, but multiple ones. In order to denote that this value is for historical data, I find it good to have it in a separate class, and I think it would be good if we did the same thing for the other strategies as well. One step at a time though :)

hannesdiedrich: Ok, makes sense! Thanks for the explanation!

integration_tests/steps/two_sided_pay_as_clear.py
```diff
@@ -71,9 +66,24 @@ def test_cumulative_offer_bid_energy(context):
...
+                 os.path.join(base_path, 'grid-bids.csv'),
+                 os.path.join(base_path, 'grid-market-clearing-rate.csv'),
+                 os.path.join(base_path, 'grid', 'house-1-offers.csv'),
+                 os.path.join(base_path, 'grid', 'house-1-bids.csv'),
+                 os.path.join(base_path, 'grid', 'house-1-market-clearing-rate.csv'),
+                 os.path.join(base_path, 'grid', 'house-2-offers.csv'),
+                 os.path.join(base_path, 'grid', 'house-2-bids.csv'),
+                 os.path.join(base_path, 'grid', 'house-2-market-clearing-rate.csv')]
+
+    assert all(len(glob.glob(f)) == 1 for f in file_list)
```
hannesdiedrich: How about adding a check if the file is not empty, like @spyrostz did in his last PR:
`assert all(len(open(glob.glob(f)[0]).readlines()) > 0 for f in file_list)` ?

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -161,7 +161,7 @@ def _match_offers_bids(self):
             self._delete_forwarded_offer_entries(offer)
 
     def tick(self, *, area):
-        super().tick(area=area)
+        # super().tick(area=area)
```
hannesdiedrich: Please remove

hannesdiedrich: You dont need the whole function then, right?

integration_tests/steps/two_sided_pay_as_clear.py
faizan2590: I deliberately didn't added this check as it is not necessary to have market_clearing_rate

hannesdiedrich: Hmm, ok. So there the simulation returns an empty file in this case?

faizan2590: Yes

hannesdiedrich: Ok, Thanks!

spyrostz: The file should not be exported at all in case of pay-as-bid and one sided market. The reason for that is that even now there are quite a lot of files that are exported, and it would help reduce the clutter if the files were not exported at all in that case.

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
faizan2590: @spyrostz I removed it, as it was calling two_sided_pay_as_bid_engines event_tick which doesn't makes sense to me. Do you have any opinion on that?

spyrostz: @faizan2590 yes, some form of it should be here, because if you remove it completely the IAA will not forward the offers. The problem is that it is problematic as it is right now, since the way it is implemented, it forwards the bids twice, and it also performs market clearings for both two sided pay as clear AND pay as bid at the same time. Therefore this needs to change to not call super().tick() but the one sided engine constructor, bypassing the two sided pay as bid constructor. Even better would be to handle this via function overloading of course, which would remove the need of bypassing the constructor.

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -161,16 +161,22 @@ def _match_offers_bids(self):
...
 
     def tick(self, *, area):
-        super().tick(area=area)
 
         for bid_id, bid in self.markets.source.bids.items():
             if bid_id not in self.forwarded_bids and \
                     self.owner.usable_bid(bid) and \
                     self.owner.name != bid.seller:
                 self._forward_bid(bid)
+
```
spyrostz: This copy-pastes code, and apart from that it is still not succeeding, since there is a whole bunch of code that is related to offer forwarding that is not included here. Same applies to the 5 lines above this change, it is just for the two sided pay as bid market, and should not be repeated. Like in the previous comment, I would either refactor the code with function overloading to do exactly what each layer is supposed to do (offer forwarding in the one sided engine, bid forwarding in the pay as bid engine), OR to call the one sided engine constructor directly from here.

src/d3a/models/state.py
```diff
@@ -234,3 +234,12 @@ def market_cycle(self, past_time_slot, time_slot):
         self.charge_history[time_slot] = 100.0 * self.used_storage / self.capacity
         self.charge_history_kWh[time_slot] = self.used_storage
         self.offered_history[time_slot] = self.offered_sell_kWh[time_slot]
+
```
spyrostz: One last comment on the PR, the MarketClearingState class could be moved on the market_structures module, in order to be at the same place as Offer, Bid, Trade etc.

integration_tests/steps/two_sided_pay_as_clear.py
hannesdiedrich: What is the status on this issue? I dont see a check for empty file. So the market_clearing_rate is exported every time, yes?

faizan2590: For now yes, it would be imported everytime.

PR #362, opened 2019-01-22
Bug/d3 asim 906

spyrostz:


src/d3a/d3a_core/util.py
```diff
@@ -330,3 +334,20 @@ def recursive_retry(functor, retry_count, max_retries, *args, **kwargs):
...
+    # And the market type is not single sided market
+    # This is a wrong configuration and an exception is raised
+    if not ConstSettings.IAASettings.AlternativePricing.COMPARE_PRICING_SCHEMES and \
+       ConstSettings.IAASettings.MARKET_TYPE != 1 and \
+       ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0:
+        assert False, "Alternate pricing schemes are only usable with an one sided market."
+
+    # If an alternate price is selected on compare schemes
+    # There should be a single sided market
+    if ConstSettings.IAASettings.AlternativePricing.COMPARE_PRICING_SCHEMES and \
```
hannesdiedrich: This condition should never be True, as the user sets 
either
COMPARE_PRICING_SCHEMES 
or
PRICING_SCHEME
So I would change the `and` into an `or`. Do you agree?

hannesdiedrich: Additionally, how about adding an assert here for the validation of COMPARE_PRICING_SCHEMES and PRICING_SCHEME ? We first decided against it, but as you added this function, it would be beneficial for the user , IMO.

src/d3a/blockchain/utils.py
```diff
@@ -50,8 +50,8 @@ def unlock_account(chain, address):
     if ConstSettings.BlockchainSettings.START_LOCAL_CHAIN:
         return
     else:
+        account_pass = os.getenv("BLOCKCHAIN_ACCOUNT_PASSWORD", "testgsy")
```
hannesdiedrich: So I understand, that the user has to set the BLOCKCHAIN_ACCOUNT_PASSWORD in an environmental variable, right? If not done already, please add it to the Wiki. Thanks!

src/d3a/d3a_core/util.py
spyrostz: To answer both comments, the problem is not with the user in this case, but with the subprocesses. When COMPARE_PRICING_SCHEMES is set to true, simulation.py will change the PRICING_SCHEME according to the subprocess that is running. Therefore it is a valid scenario in our case, and we should not assert if we encounter it, because otherwise some processes would fail.

hannesdiedrich: OK, thanks, I understand

src/d3a/blockchain/utils.py
spyrostz: Will do, thanks!

PR #363, opened 2019-01-24
Feature/d3 asim 882

hannesdiedrich:
This PR contains several changes:
- introduced a GlobalConfig class
- changed and reformating of time-formating throughout the code including elimination of variations between pendulum and string objects  for time-stamps. 
  - Now every time-stamp in the code is a pendulum (DateTime) object. The market-member time_slot_str is only used for logging
  - The used time format is set to YYYY-MM-DDTHH:mm
- the SimulationConfig member duration was renamed to sim_duration due to name overlap with the pendulum object
- a script was added, that enables the user to create a multiday profile from a daily profile. 

The main changes were implemented in the read_user_profile.py


integration_tests/load_tests.feature
```diff
@@ -16,19 +16,26 @@ Feature: Load Tests
...
-    When we run the d3a simulation with strategy_tests.user_profile_load_csv [24, 15, 15]
-    Then the DefinedLoadStrategy follows the Load profile provided as csv
+    When we run the d3a simulation with strategy_tests.user_profile_load_csv [None, 24, 15, 15]
+    Then the DefinedLoadStrategy follows the single day Load profile provided as csv
+    And load only accepted offers lower than final_buying_rate
+
+  Scenario: DefinedLoadStrategy trades energy based on a multiday csv profile
+    Given we have a scenario named strategy_tests/user_profile_load_csv_multiday
+    And d3a is installed
+    When we run the d3a simulation with strategy_tests.user_profile_load_csv_multiday [2019-01-01, 24, 15, 15]
```
faizan2590: Since, this integrationtest targets multiday csv profile. It should probably be tested by running simulation for more then one day

src/d3a/models/const.py
```diff
@@ -137,3 +139,16 @@ class BalancingSettings:
         OFFER_SUPPLY_RATIO = 0.1
         # Adds flexible load support.
         FLEXIBLE_LOADS_SUPPORT = True
+
+
+class GlobalConfig:
```
faizan2590: Is there any particular advantage to not make it a subclass of ConstSettings

integration_tests/load_tests.feature
```diff
@@ -3,7 +3,7 @@ Feature: Load Tests
   Scenario: DefinedLoadStrategy follows the profile provided by the user as dict
     Given we have a scenario named strategy_tests/user_profile_load_dict
     And d3a is installed
-    When we run the d3a simulation with strategy_tests.user_profile_load_dict [24, 15, 15]
+    When we run the d3a simulation with strategy_tests.user_profile_load_dict [None, 24, 15, 15]
```
spyrostz: Well, you already made this change therefore no point in reverting it, but you could have created a separate behave step implementation that would automatically match the [d, h, m, s] format, and would redirect to the existing function. Anyways, since you already made the changes it's ok.

integration_tests/steps/integration_tests.py
```diff
@@ -584,7 +601,8 @@ def check_pv_profile_csv(context):
     house1 = list(filter(lambda x: x.name == "House 1", context.simulation.area.children))[0]
     pv = list(filter(lambda x: x.name == "H1 PV", house1.children))[0]
     input_profile = _readCSV(context._device_profile)
-    produced_energy = {f'{k.hour:02}:{k.minute:02}': v
+    # today_str = str(today(tz=TIME_ZONE).format(DATE_FORMAT))
```
spyrostz: Can be removed

integration_tests/load_tests.feature
hannesdiedrich: You are totally right! Will change that.

src/d3a/models/const.py
hannesdiedrich: It really is a matter of preference I think.
I see it as a singleton for the global simulation configuration, that is set for for each simulation run. The others are constants and default values for the strategies. Additionally, these parameters always where separate from the ConstSettings, even in the Settings json file.  This why we did not think about moving it into ConstSettings.

integration_tests/load_tests.feature
hannesdiedrich: OMG! You are right. Did not think about that. If you think this looks ugly and should rather be changed, I am happy to add another step. Please let me know. Thanks

hannesdiedrich: Actually, considering the change to @faizan2590 s comment, I will implement e separate step for multi-day simulations and get rid of the `None`s all over the place. :-)

src/d3a/d3a_core/simulation.py
```diff
@@ -88,21 +88,21 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig =
...
-            num_ticks_to_propagate = no_of_levels * 2
-            ConstSettings.GeneralSettings.MAX_OFFER_TRAVERSAL_LENGTH = int(num_ticks_to_propagate)
-            time_to_propagate_minutes = num_ticks_to_propagate * \
-                self.simulation_config.tick_length.seconds / 60.
-            log.error("Setup has {} levels, offers/bids need at least {} minutes "
-                      "({} ticks) to propagate.".format(no_of_levels, time_to_propagate_minutes,
-                                                        ConstSettings.GeneralSettings.
-                                                        MAX_OFFER_TRAVERSAL_LENGTH,))
+        no_of_levels = self._get_setup_levels(self.area) + 1
+        num_ticks_to_propagate = no_of_levels * 2
```
spyrostz: This is not in the context of this story, butI was thinking how could we remove this "2" magic value from here, and one way is to change the _get_setup_levels method to actually access the "min_offer_age" IAA parameter. That way we can calculate the exact amount of ticks required to propagate an offer without relying on having always min_offer_age = 1. Maybe we can open a small task for handling this

src/d3a/d3a_core/util.py
```diff
@@ -336,6 +359,12 @@ def recursive_retry(functor, retry_count, max_retries, *args, **kwargs):
         return recursive_retry(functor, retry_count+1, max_retries, *args, **kwargs)
 
 
+def change_global_config(**kwargs):
+    for arg, value in kwargs.items():
+        if hasattr(GlobalConfig, arg):
```
spyrostz: I would either raise an exception or assert in the else block here, because if I understand correctly only if the user tries to monkeypatch the GlobalConfig object with additional attributes.

src/d3a/d3a_core/simulation.py
hannesdiedrich: Yes, agreed. Bu I suggest to open a small task and deal with it separately. Will do so.

spyrostz: Cool, thanks!

src/d3a/d3a_core/util.py
hannesdiedrich: I think it is not possible for the user to monkeypatch the GlobalConfig object, but I will add the else block anyway to be on the save side and to be more explicit. Thanks!

src/d3a/resources/LOAD_DATA_1.csv
```diff
@@ -1,97 +1,97 @@
...
-21:30;10
-21:45;10
-22:00;10
-22:15;10
-22:30;10
-22:45;10
-23:00;10
-23:15;10
-23:30;10
+INTERVAL;POWER(W)
```
spyrostz: Hmm, what's the change here? It is not evident from the PR diff.

src/d3a/d3a_core/util.py
hannesdiedrich: Ahh no, this else block has to be empty, as I like to do this: `change_global_config(**configuration.__dict__)` 
but configuration.__dict__ might not only contain members of GlobalConfig.
Does this make sense?

spyrostz: You can filter out the class dict items like this : 
config_dict = {k: v for k, v in vars(GlobalConfig).items() if not k.startswith("__")}
but I am not sure if this is an overkill for us here.

hannesdiedrich: Thanks. I tried to implement it, but i came to the conclusion, that it is a bit of an overkill.
I you dont mind @spyrostz I will keep the else block empty. Or shall I use `pass` and add a comment why this block is empty?

PR #366, opened 2019-01-28
Feature/d3 asim 833.1

faizan2590:


src/d3a/d3a_core/export.py
```diff
@@ -386,6 +389,56 @@ def _plot_ess_soc_history(self, storage_list: list, subdir: str, root_name: str)
...
+        """
+        new_subdir = os.path.join(subdir, area.slug)
+        self._supply_demand_curve(new_subdir, area)
+        for child in area.children:
+            if child.children:
+                self.supply_demand_curve(child, new_subdir)
+
+    def _supply_demand_curve(self, subdir: str, area: Area):
+
+        for i in area.past_markets:
```
hannesdiedrich: I am not a big fan of using single-character variable names, because it decreases the readability.
How about renaming all single-character variables in this function, e.g. "i" --> "past_market" ?

src/d3a/d3a_core/export.py
```diff
@@ -517,6 +607,26 @@ def plot_bar_graph(cls, barmode: str, title: str, xtitle: str, ytitle: str, data
         fig = go.Figure(data=data, layout=layout)
         py.offline.plot(fig, filename=iname, auto_open=False)
 
+    @classmethod
+    def plot_line_graph(cls, title: str, xtitle: str, ytitle: str, data, iname: str, xmax: int):
+        layout = go.Layout(
```
hannesdiedrich: Isnt that the point where you can increase the line thickness? Or did I miss it somewhere else?

src/d3a/models/state.py
```diff
@@ -234,12 +234,3 @@ def market_cycle(self, past_time_slot, time_slot):
         self.charge_history[time_slot] = 100.0 * self.used_storage / self.capacity
         self.charge_history_kWh[time_slot] = self.used_storage
         self.offered_history[time_slot] = self.offered_sell_kWh[time_slot]
-
```
hannesdiedrich: Can you tell me the reason for reverting this again?
I can remember that I was asking for the reason for this separate class in one of you previous PRs and @spyrostz answered:
https://github.com/gridsingularity/d3a/pull/361#discussion_r249437805

Did something change in the meantime?

src/d3a/d3a_core/export.py
```diff
@@ -450,6 +503,43 @@ def __init__(self, dataset: dict, key: str):
         self.key = key
         self.dataset = dataset
         self.umHours = dict()
+        self.rate = list()
+        self.energy = list()
+
+    def sd_curve(self, supply=True):
```
hannesdiedrich: How about renaming it to supply_demand_curve?

src/d3a/d3a_core/export.py
```diff
@@ -450,6 +503,43 @@ def __init__(self, dataset: dict, key: str):
...
+        try:
+            if supply:
+                sort_values = SortedDict(self.dataset)
```
hannesdiedrich: you can pull this lien out of the if block as it is duplicated.

src/d3a/d3a_core/export.py
```diff
@@ -450,6 +503,43 @@ def __init__(self, dataset: dict, key: str):
...
+            cond_rate = list()
+            cond_energy = list()
+
+            for i in range(len(self.energy)):
+
+                if i == 0:
+                    cond_rate.append(self.rate[0])
+                    cond_energy.append(0)
+                    cond_rate.append(self.rate[0])
+                    cond_energy.append(self.energy[i])
```
hannesdiedrich: As i is 0 here, this is a duplicate of line 527, right?

src/d3a/d3a_core/export.py
```diff
@@ -450,6 +503,43 @@ def __init__(self, dataset: dict, key: str):
...
+                else:
+                    if self.energy[i-1] == self.energy[i] and supply:
+                        continue
+                    cond_rate.append(self.rate[i])
```
hannesdiedrich: Sorry, can you please explain, what happens here? Again, there is duplication. Thanks!

src/d3a/d3a_core/export.py
```diff
@@ -386,6 +389,56 @@ def _plot_ess_soc_history(self, storage_list: list, subdir: str, root_name: str)
...
+                    xmax = max(xmax, o[1]) * 3
+
+            if len(data) > 0:
+                plot_dir = os.path.join(self.plot_dir, subdir, 'mcp')
+                mkdir_from_str(plot_dir)
+                output_file = os.path.join(plot_dir, f'supply_demand_{i.time_slot_str}.html')
+                BarGraph.plot_line_graph('supply_demand_curve', 'Energy (kWh)', 'Rate (ct./kWh)',
+                                         data, output_file, xmax)
+
+    def _line_plot(self, data_set, key, supply):
```
spyrostz: This can be a staticmethod

src/d3a/models/market/two_sided_pay_as_clear.py
```diff
@@ -27,7 +28,13 @@ class TwoSidedPayAsClear(TwoSidedPayAsBid):
...
                  notification_listener=None, readonly=False):
         super().__init__(time_slot, area, notification_listener, readonly)
-        self.state = MarketClearingState()
+        self.cumulative_offers = dict()  # type Dict[Datetime, dict()]
+        self.cumulative_bids = dict()  # type Dict[Datetime, dict()]
+        self.clearing = defaultdict(int)  # type: Dict[DateTime, tuple()]
+
+    @classmethod
+    def _csv_fields(cls):
+        return 'time', 'rate [ct./kWh]'
```
spyrostz: I think I have left a comment in another PR too, clearing rate should be a separate class, and not part of the two-sided pay as clear market.

src/d3a/d3a_core/export.py
```diff
@@ -386,6 +389,56 @@ def _plot_ess_soc_history(self, storage_list: list, subdir: str, root_name: str)
...
+    def _supply_demand_curve(self, subdir: str, area: Area):
+
+        for i in area.past_markets:
+            data = list()
+            xmax = 0
+            for j, k in i.cumulative_offers.items():
+                data.append(self._line_plot(k, j, True))
+            for l, m in i.cumulative_bids.items():
+                data.append(self._line_plot(m, l, False))
+            for n, o in i.clearing.items():
```
spyrostz: As stated in other comments, at least the 'o' parameter should be expanded to different names (it is a tuple, right?), because it is even more difficult to understand what this variable does if you have to keep in mind what is [0] and [1] in it.

src/d3a/d3a_core/export.py
```diff
@@ -517,6 +607,26 @@ def plot_bar_graph(cls, barmode: str, title: str, xtitle: str, ytitle: str, data
         fig = go.Figure(data=data, layout=layout)
         py.offline.plot(fig, filename=iname, auto_open=False)
 
+    @classmethod
+    def plot_line_graph(cls, title: str, xtitle: str, ytitle: str, data, iname: str, xmax: int):
```
spyrostz: It is quite bad that there is a function called 'plot_line_graph' inside a class called 'BarGraph'. I would convert the plot_line_graph to a free function, since there is not really any dependency on the class members (it is a classmethod anyways, and cls is not used).

src/d3a/d3a_core/export.py
faizan2590: Its not duplicate, Im trying to add two points in x-axis with same energy_rate. If it is not clear, I can draw and explain offline.

src/d3a/d3a_core/export.py
faizan2590: Same as above. Lets discuss offline.

src/d3a/models/state.py
faizan2590: https://github.com/gridsingularity/d3a/pull/361/files#r249868813

src/d3a/d3a_core/export.py
faizan2590: https://github.com/gridsingularity/d3a/pull/366/files#diff-9e95bed0f62db1b1f967c65f0823d49dR418

src/d3a/models/state.py
hannesdiedrich: Yes, but the mentioned comment states only that the MarketClearingState class should be moved to a different file, not deleted totally. Right?

faizan2590: Yes, I have moved it to market_structures in my last commit https://github.com/gridsingularity/d3a/pull/366/files#diff-286711868906f1b84a065750ab4e674fR164

hannesdiedrich: Ok! Thanks! No i see it. An explanatory sentence is always better than a link ;-)

src/d3a/d3a_core/export.py
```diff
@@ -384,7 +387,52 @@ def _plot_ess_soc_history(self, storage_list: list, subdir: str, root_name: str)
...
+                    data.append(data_obj)
+                    data_obj = go.Scatter(x=[clearing_point[1], clearing_point[1]],
+                                          y=[0, clearing_point[0]],
+                                          mode='lines+markers',
+                                          line=dict(width=5),
+                                          name=str(time_slot) + str('Clearing-Energy'))
+                    data.append(data_obj)
+                    xmax = max(xmax, clearing_point[1]) * 3
+
+            if len(data) > 0:
```
hannesdiedrich: You could move this check up to line 413 to skip the plotting of the clearing point if there is no supply and demand curve.

src/d3a/d3a_core/export.py
hannesdiedrich: As this is just a function for the layout, please rename.

src/d3a/d3a_core/export.py
```diff
@@ -384,7 +387,52 @@ def _plot_ess_soc_history(self, storage_list: list, subdir: str, root_name: str)
...
+                                          line=dict(width=5),
+                                          name=str(time_slot) + str('Clearing-Rate'))
+                    data.append(data_obj)
+                    data_obj = go.Scatter(x=[clearing_point[1], clearing_point[1]],
+                                          y=[0, clearing_point[0]],
+                                          mode='lines+markers',
+                                          line=dict(width=5),
+                                          name=str(time_slot) + str('Clearing-Energy'))
+                    data.append(data_obj)
+                    xmax = max(xmax, clearing_point[1]) * 3
```
hannesdiedrich: Why multiplying by 3?

src/d3a/d3a_core/export.py
```diff
@@ -384,7 +387,52 @@ def _plot_ess_soc_history(self, storage_list: list, subdir: str, root_name: str)
...
+                data.append(PlotlyGraph._line_plot(demand_curve, time_slot, False))
+            for time_slot, clearing_point in past_market.state.clearing.items():
+                # clearing_point[0] --> Clearing-Rate
+                # clearing_point[1] --> Clearing-Energy
+                if len(clearing_point) != 0:
+                    data_obj = go.Scatter(x=[0, clearing_point[1]],
+                                          y=[clearing_point[0], clearing_point[0]],
+                                          mode='lines+markers',
+                                          line=dict(width=5),
+                                          name=str(time_slot) + str('Clearing-Rate'))
```
hannesdiedrich:
```suggestion
                                          name=time_slot.format(DATE_TIME_FORMAT) + str('Clearing-Rate'))
```

src/d3a/d3a_core/export.py
```diff
@@ -384,7 +387,52 @@ def _plot_ess_soc_history(self, storage_list: list, subdir: str, root_name: str)
         plot_dir = os.path.join(self.plot_dir, subdir)
         mkdir_from_str(plot_dir)
         output_file = os.path.join(plot_dir, 'ess_soc_history_{}.html'.format(root_name))
-        BarGraph.plot_bar_graph(barmode, title, xtitle, ytitle, data, output_file)
+        PlotlyGraph.plot_bar_graph(barmode, title, xtitle, ytitle, data, output_file)
+
+    def supply_demand_curve(self, area, subdir):
```
hannesdiedrich: Ok, now there are two function called `supply_demand_curve` in this file. How about renaming this to `plot_supply_demand_curve`, similar to the other function in this class

src/d3a/d3a_core/export.py
faizan2590: It could be any value. Im not actually sure what should be the right value. My intuition was to have clearing point at 1/3rd of the graph

src/d3a/d3a_core/export.py
faizan2590: I followed the old convention here.
https://github.com/gridsingularity/d3a/blob/master/src/d3a/d3a_core/export.py#L273
https://github.com/gridsingularity/d3a/blob/master/src/d3a/d3a_core/export.py#L363
https://github.com/gridsingularity/d3a/blob/master/src/d3a/d3a_core/export.py#L403

faizan2590:
> Some comments and questions and a general comment:
> Why is this branch called D3ASIM833.1 ? Will there be a D3ASIM833.2 ?

I had one more implementation, don't want to delete it. So, I just created another branch with a version. You shouldn't expect anything like D3ASIM833.

src/d3a/setup/two_sided_market/default_2a.py
```diff
@@ -28,7 +28,7 @@
 
 def get_setup(config):
 
-    ConstSettings.IAASettings.MARKET_TYPE = 3
+    ConstSettings.IAASettings.MARKET_TYPE = 2
```
hannesdiedrich: What? How did this happen in the first place? Weird! Thanks for fixing!

src/d3a/d3a_core/export.py
hannesdiedrich: Sorry, I dont get it. Can we please discuss this offline?

src/d3a/d3a_core/export.py
hannesdiedrich: Hmm, ok. But why not showing all of the bids and offers? That was the whole point of the graph, right?

src/d3a/setup/two_sided_market/default_2a.py
faizan2590: Don't know, perhaps somehow I overlooked it.

hannesdiedrich:
> > Some comments and questions and a general comment:
> > Why is this branch called D3ASIM833.1 ? Will there be a D3ASIM833.2 ?
> 
> I had one more implementation, don't want to delete it. So, I just created another branch with a version. You shouldn't expect anything like D3ASIM833.2.

Ok, for tracking of the stories and branches this procedure is quite bad, IMO. I would suggest for the future:
Commit the first version to the branch with the ID provided by the ticket. So you're implementation is not deleted and the ID is still trackable. Does this make sense?

src/d3a/d3a_core/export.py
faizan2590: CEP, places the huge offer that makes the clearing_point unreadable. Its kind of a work around.

PR #367, opened 2019-01-30
D3ASIM-853: Changing KPI formula

faizan2590:
This is a first draft. Please review the parameters of self-sufficiency & self-consumption. For now it is only calculated for house_node. Once the calculation parameters are finalized, I will do it recursively for each hierarchy


src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -246,48 +247,61 @@ def _accumulate_energy(self, area, child, is_cep: bool):
...
-            elif _is_load_node(child) or _is_prosumer_node(child):
-                house_load_device.append(child.name)
+            if _is_load_node(child):
+                area_load_device.append(child.name)
+            elif _is_prosumer_node(child):
+                area_ess_device.append(child.name)
+            elif _is_producer_node(child):
+                area_producer_device.append(child.name)
+            else:
+                area_load_device.append(make_iaa_name(child))
```
hannesdiedrich: This block is for the storage?

hannesdiedrich: Sorry, no. Its further up.
But for what case is this block?

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -246,48 +247,61 @@ def _accumulate_energy(self, area, child, is_cep: bool):
...
 
-    def _export_house_pv_self_consumption(self, area):
-        house_pv_device = list()
-        house_load_device = list()
-        trade_by_pv = 0
-        total_energy_bought = 0
-        traded_from_pv = 0
-        if not _is_house_node(area):
-            return
+    def _export_area_sufficiency(self, area):
```
hannesdiedrich: How about to rename it to `_export_area_sufficiency_consumption` because it returns both ?

src/d3a/setup/default_5.py
```diff
@@ -29,105 +29,118 @@ def get_setup(config):
     area = Area(
         'Grid',
         [
-            Area(
```
spyrostz: I would strongly suggest to create a new setup file and do not use default_5 file just for this. The reason is that this file is used heavily for testing (it is the 5-house microgrid on the UI) and I would leave it as is in order for the people that test it to not be confused by the addition of the Street hierarchy.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
spyrostz: Maybe this handles the Area class?

faizan2590: Yes, this block is to handle the Area class

src/d3a/d3a_core/sim_results/file_export_endpoints.py
faizan2590: Or how about _export_area_performance_index

src/d3a/d3a_core/sim_results/file_export_endpoints.py
faizan2590: @mfatma, @sarahhambridge, @spyrostzand @hannesdiedrich. Can you guys give me little clarity to either consider area as load, producer or prosumer device?

hannesdiedrich: An area should be a prosumer, right?

PR #368, opened 2019-02-06
D3ASIM-932-1: Fixing energy balance

faizan2590:


src/d3a/models/const.py
```diff
@@ -35,7 +35,7 @@ class GeneralSettings:
         # Flag to enable supply/demand backend plots
         SUPPLY_DEMAND_PLOTS = True
         ENERGY_RATE_DECREASE_PER_UPDATE = 1  # rate decrease in cents_per_update
-        SETUP_FILE_PATH = None  # Default path of the available setup files
```
hannesdiedrich: Please readd the comment.

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -68,13 +68,14 @@ def energy_bills(area, past_market_types, from_slot=None, to_slot=None):
         if (from_slot is None or slot >= from_slot) and (to_slot is None or slot < to_slot):
             for trade in market.trades:
                 buyer = area_name_from_area_or_iaa_name(trade.buyer)
-                seller = area_name_from_area_or_iaa_name(trade.offer.seller)
+                seller = area_name_from_area_or_iaa_name(trade.seller)
```
hannesdiedrich: super! Just saw this in the discussion with dirk!

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -107,7 +110,7 @@ def _match_offers_bids(self):
             return
         clearing_rate, clearing_energy = clearing
         if clearing_energy > 0:
-            self.owner.log.info(f"Market Clearing Rate: {clearing_rate} "
+            self.owner.log.warn(f"Market Clearing Rate: {clearing_rate} "
```
hannesdiedrich: I guess you can revert this again, right?

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -60,7 +60,7 @@ def _discrete_point_curve(self, obj):
             if len(obj) <= 1 or i == (len(obj) - 1):
                 break
             rate = math.floor(obj[i+1].price / obj[i+1].energy)
-            cumulative[rate] += obj[i].energy
+            cumulative[rate] += obj[i+1].energy
```
hannesdiedrich: Why working with `i+1` in this loop. I suggest the following change for the loop:
```
        for i in range(len(obj)):
            rate = math.floor(obj[i].price / obj[i].energy)
            cumulative[rate] += obj[i].energy
```

hannesdiedrich: I have an even better suggestion for the whole function:
```
    def _discrete_point_curve(self, obj_list):
        cumulative = defaultdict(int)
        rate = math.floor(obj_list[0].price / obj_list[0].energy)
        cumulative[rate] = obj_list[0].energy
        for obj in obj_list:
            rate = math.floor(obj.price / obj.energy)
            cumulative[rate] += obj.energy
        return cumulative
```

hannesdiedrich: Sorry,
You can even omit two lines:
```
    def _discrete_point_curve(self, obj_list):
        cumulative = defaultdict(int)
        for obj in obj_list:
            rate = math.floor(obj.price / obj.energy)
            cumulative[rate] += obj.energy
        return cumulative
```

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
faizan2590: IMO looks better with warn, to see it with trade details

hannesdiedrich: OK

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
hannesdiedrich: Here you did not add my proposed change of renaming obj to obj_list.
I suggested it because it is a list of objects (list of bids or list of offers)
The positiv side effect is, that you get rid of the counting variable i.
Any reason against this?

faizan2590: Sorry, my bad

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
spyrostz: I think I added this comment on another PR yesterday, but I do not think it is that important, since you can see the market clearing rate directly from the trades. I don't have a strong opinion though, since it probably will not clatter the logs that much.
In addition to that, if you decide to stick to having the log, I would suggest to use the log.warning method instead of the deprecated log.warn.

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -72,6 +68,14 @@ def _smooth_discrete_point_curve(self, obj, limit, asc_order=True):
                 obj[i] = obj.get(i, 0) + obj.get(i+1, 0)
         return obj
 
+    def _get_clearing_point(self, max_rate):
+        for i in range(1, max_rate+1):
+            if self.markets.source.state.cumulative_offers[self.owner.owner.now][i] >= \
+                    self.markets.source.state.cumulative_bids[self.owner.owner.now][i]:
+                return i, self.markets.source.state.cumulative_bids[self.owner.owner.now][i]
+            else:
```
spyrostz: The else: continue block should not be needed, can be removed.

integration_tests/integration_tests.feature
```diff
@@ -35,6 +35,18 @@ Feature: Run integration tests
      |        default_5       |    24    |      60     |      60     |
      |        default_csv     |    24    |      60     |      60     |
 
+  Scenario Outline: Test Balanced Energy Bills
+     Given we have a scenario named <scenario>
+     And d3a is installed
+     When we run the d3a simulation with <scenario> [24, 60, 60]
+     Then we test energy bills
```
spyrostz: In order for the step to denote exactly what it does, I would change it to "Then the energy bills report the correct accumulated energy values".

integration_tests/steps/integration_tests.py
```diff
@@ -569,6 +569,31 @@ def test_output(context, scenario, sim_duration, slot_length, tick_length):
...
 
+@then('we test energy bills')
+def test_energy_bills(context):
+    bills = context.simulation.endpoint_buffer.bills
+
+    cell_tower = bills["Cell Tower"]["sold"] - bills["Cell Tower"]["bought"]
+
+    house1 = bills["House 1"]["sold"] - bills["House 1"]["bought"]
+    area_net_bill = 0
+    for k, v in bills["House 1"]["children"].items():
```
spyrostz: This for loop can be replaced with 
`area_net_bill = sum([v["sold"] - v["bought] for k, v in bills["House 1"]["children"].items()])`

integration_tests/steps/integration_tests.py
```diff
@@ -569,6 +569,31 @@ def test_output(context, scenario, sim_duration, slot_length, tick_length):
         assert all([permanent_load.strategy.energy == ei for ei in energy_profile])
 
 
+@then('we test energy bills')
+def test_energy_bills(context):
```
spyrostz: I know it is not on the scope of this story, but it would be great to add the corresponding test for the prices, to validate whether the "spent" and "earned" values are populated correctly.

integration_tests/steps/integration_tests.py
```diff
@@ -569,6 +569,49 @@ def test_output(context, scenario, sim_duration, slot_length, tick_length):
...
+def test_accumulated_energy_price(context):
+    bills = context.simulation.endpoint_buffer.bills
+
+    cell_tower = bills["Cell Tower"]["earned"] - bills["Cell Tower"]["spent"]
+
+    house1 = bills["House 1"]["earned"] - bills["House 1"]["spent"]
+    area_net_traded_energy_price = \
+        sum([v["earned"] - v["spent"] for k, v in bills["House 1"]["children"].items()])
+
+    assert isclose(round(area_net_traded_energy_price, 10), house1)
```
hannesdiedrich: I suggest either to use round or isclose() not both.

integration_tests/steps/integration_tests.py
```diff
@@ -569,6 +569,49 @@ def test_output(context, scenario, sim_duration, slot_length, tick_length):
...
 
+@then('the energy bills report the correct accumulated traded energy price')
+def test_accumulated_energy_price(context):
+    bills = context.simulation.endpoint_buffer.bills
+
+    cell_tower = bills["Cell Tower"]["earned"] - bills["Cell Tower"]["spent"]
+
+    house1 = bills["House 1"]["earned"] - bills["House 1"]["spent"]
+    area_net_traded_energy_price = \
+        sum([v["earned"] - v["spent"] for k, v in bills["House 1"]["children"].items()])
```
hannesdiedrich: You could use `.values()` instead and get rid of unused `k`.

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -100,6 +101,10 @@ def accept_bid(self, bid: Bid, energy: float = None,
                 final_price = energy * trade_rate
                 bid = Bid(bid.id, final_price, energy,
                           buyer, seller, self)
+            else:
+                if trade_rate is not None:
+                    bid = bid._replace(price=trade_rate * market_bid.energy)
+                    # print(f"BidID: {bid.id}")
```
spyrostz: Reminder, remove this at the end of the PR.

src/d3a/models/strategy/storage.py
```diff
@@ -77,7 +77,7 @@ def __init__(self, risk: int=GeneralSettings.DEFAULT_RISK,
             InputProfileTypes.IDENTITY,
             StorageSettings.MIN_BUYING_RATE
         )
-        self.max_buying_rate_profile = {k: v[1] for k, v in break_even.items()}
+        self.max_buying_rate_profile = {k: v[0] for k, v in break_even.items()}
```
spyrostz: It is ok for now, but in D3ASIM-968 I have already added a more future-proof fix for that which does not use a tuple, but a namedtuple instead. That way it will be easy for the reader to understand which one is the buy and the sell break even point.

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -78,7 +78,8 @@ def accept_bid(self, bid: Bid, energy: float = None,
         energy = market_bid.energy if energy is None else energy
         if trade_rate is None:
             trade_rate = market_bid.price / market_bid.energy
-        assert trade_rate <= (market_bid.price / market_bid.energy)
+        assert trade_rate <= (market_bid.price / market_bid.energy) + 0.0001, \
```
spyrostz: No need for this to be fixed in this PR (I will probably incorporate it on the 2-sided market story), but floating-point threshold values are scattered throughout our code (a quick search for 0.000 will immediately reveal quite a lot of these occurrences). I will take the time to refactor these to a common threshold variable for the whole project.

faizan2590: Integration tests are failing after removing floating threshold. So, should I use rounding, keep it the way it is or any alternative suggestion?

spyrostz: Leave it as it is. I am not suggesting to remove it, I will add a corresponding refactoring in the two-sided market PR.

src/d3a/d3a_core/simulation.py
```diff
@@ -265,7 +265,10 @@ def run(self, resume=False) -> (Period, duration):
                                 self.endpoint_buffer
                             )
 
-                    self.deactivate_areas(self.area)
+                    """
+                    TODO: D3ASIM-1001
+                    """
+                    # self.deactivate_areas(self.area)
```
hannesdiedrich: So you decided to remove this? I thought we found the reason somewhere else?

faizan2590: There were two different things that were causing energy imbalance after recent merges in master. It is commented and to be solved later in this story (https://gridsingularity.atlassian.net/browse/D3ASIM-1001). The other was due to the merge of branch 965 (https://github.com/gridsingularity/d3a/pull/377).  If you look into this link (https://github.com/gridsingularity/d3a/pull/368/files#diff-19189e4044c6277fcfe2d4bd1b9b99d7R104 & https://github.com/gridsingularity/d3a/pull/368/files#diff-ce55928b3f59db35206112ab5edebc9dR153), trade_rate was incorporated for the partial trade but was overlooked for the whole energy trade.

PR #372, opened 2019-02-06
D3ASIM-904:Adding total_energy & total_cost in bills.json file

faizan2590:


src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -72,9 +73,15 @@ def energy_bills(area, past_market_types, from_slot=None, to_slot=None):
                 if buyer in result:
                     result[buyer]['bought'] += trade.offer.energy
                     result[buyer]['spent'] += trade.offer.price
+                    result[buyer]['total_energy'] += trade.offer.energy
```
hannesdiedrich: Sorry, but `result[buyer]['total_energy']` will have the exact same value as `result[buyer]['bought']`.
The same goes for the other pairs. 
From the ticket i read: `can you just add the Total Energy (bought - sold) and the Total Cost (spent - earned) for each of the Areas/Devices (House 1, Load, PV, Energy Storage).` 
Or did i misunderstood?

faizan2590: Total Energy = bought - sold

So Im adding it here and then subtracting it in line https://github.com/gridsingularity/d3a/pull/372/files#diff-672c71151977db649f4d45c99e348f73R82.

In the end, i would be able to get Total Energy that would be the difference of bought and sold.

hannesdiedrich: Sorry, confused the signs.

src/d3a/models/market/one_sided.py
```diff
@@ -117,7 +117,7 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *, energy: in
                     if trade_rate is None:
                         trade_rate = offer.price / offer.energy
 
-                    assert trade_rate >= math.floor(offer.price / offer.energy)
+                    assert trade_rate + 0.001 >= math.floor(offer.price / offer.energy)
```
hannesdiedrich: @spyrostz another point were you can input your FLOATING_POINT_TOLERANCE

spyrostz: Indeed, once merged I will replace the the hardcoded value. Thanks!

PR #375, opened 2019-02-12
D3ASIM-974: Added global start_date to read_user_profile._readCSV.

hannesdiedrich:


src/d3a/models/read_user_profile.py
```diff
@@ -36,7 +36,27 @@ class InputProfileTypes(Enum):
...
+    """
+    Converts time_str into a pendulum (DateTime) object that either takes the global start date or
+    the provided one, dependant on the time_format
+    :return: DateTime
+    """
+    time = from_format(time_str, time_format, tz=TIME_ZONE)
+    if time_format == DATE_TIME_FORMAT:
+        return time
+    elif time_format == TIME_FORMAT:
+        return GlobalConfig.start_date.add(hours=time.hour, minutes=time.minute)
```
spyrostz: Does time not include seconds in that case ? I don't mind about this exact bug, but if we decide to use profiles with the format hh:mm:ss this will probably not take into account seconds.

hannesdiedrich: We decided earlier that we will omit seconds.
I am still in favour of that. But if we wont to add them, it should be done in a different story.
Let's discuss this in the next planing meeting!

spyrostz: My memory does not serve me well lately. Still, if the 'time' object is indeed a Datetime, then I see no problem adding the seconds=time.seconds as an additional function argument here, and be future-proof. Since our profiles do not use seconds, it will automatically populate this field with 0.

hannesdiedrich: You are right! Will add this. 
In the future, DATE_TIME_FROMAT has also be changed for that, end everything that comes with that.

PR #376, opened 2019-02-13
D3ASIM-969: Enabling offers/bids to follow original rate while reposting

faizan2590:
This is an initial implementation. If it looks good. Will fix the tests


src/d3a/models/const.py
```diff
@@ -36,6 +36,7 @@ class GeneralSettings:
         SUPPLY_DEMAND_PLOTS = True
         ENERGY_RATE_DECREASE_PER_UPDATE = 1  # rate decrease in cents_per_update
         SETUP_FILE_PATH = None  # Default path of the available setup files
+        # SETUP_FILE_PATH = "/Users/muhammad.faizan/d3a-private/src/LES"
```
spyrostz: Please delete that

src/d3a/models/market/one_sided.py
```diff
@@ -113,9 +113,12 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *, energy: in
                         if self.area is None or self.area.bc is None \
                         else offer.real_id
 
+                    if trade_rate is None:
```
spyrostz: It would also be nice to add an assert, which would validate whether the trade_rate respects the offer rate, since trade_rate should never fall below the initial offer rate.

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -68,14 +68,16 @@ def delete_bid(self, bid_or_id: Union[str, Bid]):
...
-                   price_drop: bool = True):
+                   price_drop: bool = True, trade_rate: float = None):
         market_bid = self.bids.pop(bid.id, None)
         if market_bid is None:
             raise BidNotFound("During accept bid: " + str(bid))
 
         seller = market_bid.seller if seller is None else seller
         buyer = market_bid.buyer if buyer is None else buyer
         energy = market_bid.energy if energy is None else energy
+        if trade_rate is None:
```
spyrostz: Similar to the one-sided market case, an assert needs to be added to validate that the trade rate is never higher than the bid rate.

PR #377, opened 2019-02-13
Bug/d3 asim 965

faizan2590:


integration_tests/commercial_power_plant_tests.feature
```diff
@@ -18,6 +18,6 @@ Feature: Commercial Power Plant tests
      Given we have a scenario named strategy_tests/commercial_producer_market_maker_rate
      And d3a is installed
      And we have a profile of market_maker_rate for strategy_tests.commercial_producer_market_maker_rate
-     When we run the d3a simulation with config parameters [1, 1] and strategy_tests.commercial_producer_market_maker_rate
-     Then we test that config parameters are correctly parsed for strategy_tests.commercial_producer_market_maker_rate [1, 1]
+     When we run the d3a simulation with config parameters [1, 0] and strategy_tests.commercial_producer_market_maker_rate
```
hannesdiedrich: Can you tell me the reason for this change or rather the reason for changing the iaa_fee to 0 in several integration tests?

src/d3a/models/strategy/area_agents/one_sided_engine.py
```diff
@@ -115,18 +115,19 @@ def event_trade(self, *, trade):
...
             try:
+                trade_offer_rate = (trade.offer.price / trade.offer.energy)
                 if trade.price_drop:
                     # Use the rate of the trade offer for accepting the source offer too
                     # Drop the rate of the trade offer according to IAA fee
-                    trade_offer_rate = trade.offer.price / trade.offer.energy
-                    offer_info.source_offer.price = \
-                        trade_offer_rate * offer_info.source_offer.energy \
-                        / (1 + (self.transfer_fee_pct / 100))
+                    # trade_offer_rate = trade.offer.price / trade.offer.energy
```
hannesdiedrich: please delete.

src/d3a/models/strategy/area_agents/one_sided_engine.py
```diff
@@ -115,18 +115,19 @@ def event_trade(self, *, trade):
...
+                    trade_offer_rate = trade_offer_rate * \
+                                       (1 + (self.transfer_fee_pct / 100))
                 trade_source = self.owner.accept_offer(
                     self.markets.source,
                     offer_info.source_offer,
                     energy=trade.offer.energy,
-                    buyer=self.owner.name
+                    buyer=self.owner.name,
+                    trade_rate=trade_offer_rate
```
hannesdiedrich: Set this to None for the single sided market.

src/d3a/models/market/one_sided.py
```diff
@@ -113,9 +114,14 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *, energy: in
                         if self.area is None or self.area.bc is None \
                         else offer.real_id
 
+                    if trade_rate is None:
+                        trade_rate = offer.price / offer.energy
+                    else:
+                        assert trade_rate >= math.floor(offer.price / offer.energy)
```
hannesdiedrich: Move this outside the else block.

integration_tests/commercial_power_plant_tests.feature
faizan2590: As discussed offline, reverting it

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -107,7 +107,7 @@ def _match_offers_bids(self):
             return
         clearing_rate, clearing_energy = clearing
         if clearing_energy > 0:
-            self.owner.log.info(f"Market Clearing Rate: {clearing_rate} "
+            self.owner.log.warn(f"Market Clearing Rate: {clearing_rate} "
```
spyrostz: I do not know how important it is in order to add it on the warning log (to my eyes not that important, since you can see the clearing rate from the trade rate, they should both be the same). Info looks more appropriate to me. In addition to that, if you decide to stick to warning log level, you should use the warning log function instead of the deprecated warn function.

faizan2590: If you both vouch to have it in info logs, then i will change it. But clearing are not too frequent to clutter the warn_logs and it might be easier while matching trade_rate & trade_energy with clearing_rate & clearing_energy.

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -60,7 +60,7 @@ def _discrete_point_curve(self, obj):
             if len(obj) <= 1 or i == (len(obj) - 1):
                 break
             rate = math.floor(obj[i+1].price / obj[i+1].energy)
-            cumulative[rate] += obj[i].energy
+            cumulative[rate] += obj[i+1].energy
```
spyrostz: Wait, I am confused here. Wasn't this loop refactored before? I think after a comment of @hannesdiedrich you replaced the index-based loop to iterating over the items without using indices. If I am imagining this, then this needs to be refactored, it looks much better and it is much safer if the indices are not used.

faizan2590: Yes, you are right. It is in branch 932.1 (https://github.com/gridsingularity/d3a/pull/368/files#diff-9727365ead9d7d435176c7f1c9a53c19R55)

PR #378, opened 2019-02-13
Feature/d3 asim 962

hannesdiedrich:
This PR contains:
- calculating device statistics and storing them in a json file in the aggregates_results folder
- prpopsal for bugfix of the missing last market rotation
- one fix of a integrationtest `D3ASIM-662, IAA should always track offers back to its source`


src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -62,7 +62,7 @@ def gather_area_loads_and_trade_prices(area, load_price_lists):
     return load_price_lists
 
 
-def gather_prices_pv_stor_energ(area, price_energ_lists):
+def gather_prices_pv_storage_energy(area, price_energ_lists):
```
spyrostz: Very good, I did not like the truncated name as well.

src/d3a/models/strategy/load_hours.py
```diff
@@ -112,6 +113,8 @@ def event_activate(self):
                             for day in range(self.area.config.sim_duration.days + 1)}
         self._simulation_start_timestamp = self.area.now
         self.assign_energy_requirement(self.avg_power_W)
+        self.energy_requirement_history_kWh = dict((k, v / 1000.)
```
spyrostz: Is this different from self.state.desired_energy_Wh (member of LoadState class) ? If not, then it is better to reuse the member from the state class, in order to not add another class member.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -70,10 +75,17 @@ def generate_json_report(self):
             "cumulative_grid_trades": self.cumulative_grid_trades,
             "bills": self.bills,
             "tree_summary": self.tree_summary,
-            "status": self.status
+            "status": self.status,
+            "device_stats": self.device_statistics_time_str_dict
         }
 
     def update_stats(self, area, simulation_status):
+        # if not (not hasattr(area.parent, "past_markets") or len(area.parent.past_markets) == 0):
```
spyrostz: Trivial, as a reminder please remove these before you merge.

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -0,0 +1,153 @@
...
+            else:
+                self.gather_device_statistics(child, subdict[child.name])
+
+    def _gather_device_statistics(self, area: Area, subdict: Dict):
+        if not hasattr(area.parent, "past_markets") or len(area.parent.past_markets) == 0:
+            return None
+
+        if isinstance(area.strategy, PVStrategy):
+
+            self._device_price_stats(area, subdict)
```
spyrostz: _device_price_stats and _device_energy_stats can be moved out of the if-elif blocks to remove duplication. You can check if area.strategy is not None, then you can safely deduce that this area is a device, therefore you can call safely these methods.

src/d3a/models/strategy/load_hours.py
hannesdiedrich: Yes, you are right, can not remember the reason for adding this here. Will try to remove it.

src/d3a/d3a_core/export.py
```diff
@@ -93,6 +93,9 @@ def export_json_data(self, directory: dir):
         trade_file = os.path.join(json_dir, "trade-detail")
         with open(trade_file, 'w') as outfile:
             json.dump(self.endpoint_buffer.trade_details, outfile, indent=2)
+        device_file = os.path.join(json_dir, 'device_statistics')
```
faizan2590: I guess these three line could be removed as https://github.com/gridsingularity/d3a/pull/378/files#diff-9e95bed0f62db1b1f967c65f0823d49dR100 it is taking care of dumping device statistics as json

hannesdiedrich: You are right! Thanks!

PR #379, opened 2019-02-14
Bug/d3 asim 969

spyrostz:


src/d3a/models/area/event_dispatcher.py
```diff
@@ -113,6 +113,8 @@ def event_listener(self, event_type: Union[MarketEvent, AreaEvent], **kwargs):
                 self.area.strategy.event_listener(event_type, **kwargs)
             if self.area.appliance:
                 self.area.appliance.event_listener(event_type, **kwargs)
+        elif not self.area.events.is_enabled and event_type == AreaEvent.MARKET_CYCLE:
+            self.area.strategy.event_on_disabled_area()
```
spyrostz: Add a None check for the self.area.strategy here.

PR #381, opened 2019-02-19
D3ASIM-938: Added start_date to d3a_jobs.

hannesdiedrich:


src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -46,20 +47,20 @@ def start(scenario, settings):
         update_advanced_settings(ast.literal_eval(advanced_settings))
 
     config = SimulationConfig(
-        sim_duration=pendulum.duration(
-            days=1 if 'duration' not in settings else settings['duration'].days
-        ),
-        slot_length=pendulum.duration(
+        sim_duration=duration(days=1 if 'duration' not in settings else settings['duration'].days),
```
spyrostz: It is not in the context of this story, but I was thinking whether it would be worth it to replace the default values (days=1, minutes=15, seconds=15, start_date=today()) with the GeneralSettings values, in order to not have the default values in 2 places. If you agree and it is not too much, it would be great to refactor it.

hannesdiedrich: You are right, will ad this!

PR #382, opened 2019-02-19
Feature/d3 asim 968

spyrostz:
It is easier to look at the changes of this PR commit-by-commit. 
The first 2 commits are dealing with a bug in the breakeven point calculation, where the incorrect tuple index was used (the sell BE index was used for defining the max buying rate). Throughout the process I refactored the BE to always use a namedtuple, in order to be easily recognisable which one is the sell and which one is the buy BE point. 
The third commit is a removal of some unneeded dependencies. After removing these the need for pyyaml is no longer there, therefore the annoying dependency warning on the main github page will disappear. 
The fourth commit is actually what deals with this issue. There is a new event called BID_CHANGED which is responsible for directly updating the bids, instead of deleting them and rely on the bid forwarding.
The fifth commit is a refactoring of the bid management for the strategies, which is moved on its own class.
The sixth commit is a refactoring of the floating point tolerance, because I noticed there were several points in our code where there was a comparison with 0.00001 or similar values. So I added a const value in order to be used globally.


src/d3a/models/strategy/storage.py
```diff
@@ -152,9 +146,29 @@ def _set_be_alternative_pricing(self):
             else:
                 raise MarketException
 
+    @staticmethod
+    def _validate_break_even_points(break_even):
+        if type(break_even) not in [tuple, list, dict]:
+            raise ValueError("Break even points have to be a tuple, list or dict.")
+
+        if type(break_even) is list:
```
hannesdiedrich: Can you please explain the reason for the whole if/elif block here?
AFAIK `read_arbitrary_profile` can only handle dicts, file-name stringe or json strings.
Did you try it with tupple?
Additionally, in the general PR comment you stated `Throughout the process I refactored the BE to always use a namedtuple`. However, in this `if type(break_even) is list:` block, it is not converted into a BreakEven instance.

requirements/dev.txt
```diff
@@ -6,97 +6,37 @@
...
 pycparser==2.19           # via cffi
-pycryptodome==3.7.3
 pygments==2.3.1           # via ptpython, readme-renderer
 pynacl==1.3.0             # via paramiko
-python-dateutil==2.7.5
-python-rex==0.4
-pytz==2018.9
-pytzdata==2018.7
-pyyaml==3.13
+pyyaml==3.13              # via aspy.yaml, pre-commit
```
hannesdiedrich: Sorry, but pyyaml==3.13 is still there which will lead to the github warning.

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -145,24 +146,36 @@ def event_bid_traded(self, *, bid_trade):
...
+    def after_successful_trade_event(self, source_trade, bid_info):
+        if source_trade.residual:
+            target_residual_bid = self.bid_trade_residual.pop(bid_info.target_bid.id)
+            assert target_residual_bid.id != source_trade.residual.id, \
+                f"Residual bid from the trade ({source_trade.residual}) is not the same as the" \
+                f" residual bid from event_bid_changed ({target_residual_bid})."
+            assert source_trade.residual.id not in self.forwarded_bids, \
+                f"Residual bid has not been forwarded even though it was transmitted via " \
+                f"event_bid_trade."
+            res_offer_info = BidInfo(source_trade.residual, target_residual_bid)
```
hannesdiedrich: I suggest to rename `res_offer_info` into `res_bid_info`.

src/d3a/models/strategy/storage.py
spyrostz: Point #1: Apparently the read_arbitrary_profile can also handle tuples. See https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/read_user_profile.py#L239. Break even always used to be a tuple, and it always converted the values correctly as far as I have seen. 
Point #2: If you pass a list in this function, it will be converted to a tuple (it is the next line in the diff) and will be treated as a tuple from then on, therefore it will correctly be converted to a namedtuple instance.

requirements/dev.txt
spyrostz: Shit, I thought that since it was removed from base.txt pyyaml was no longer there. I will fix the version to the one GitHub suggests. Thanks

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
spyrostz: Of course, you are correct. Will do

tests/test_strategy_load_hours_fb.py
```diff
@@ -310,8 +310,8 @@ def test_device_operating_hours_deduction_with_partial_trade(load_hours_strategy
         round(((0.1/0.155) * 0.25), 2)
 
 
-@pytest.mark.parametrize("partial", [False, True])
-def test_event_bid_traded_does_not_remove_bid_for_partial_trade(load_hours_strategy_test5,
+@pytest.mark.parametrize("partial", [None, Bid('test_id', 123, 321, 'A', 'B')])
```
hannesdiedrich: Super! Thanks for adding again!

src/d3a/models/strategy/load_hours.py
```diff
@@ -241,10 +222,6 @@ def event_bid_traded(self, *, market_id, bid_trade):
                 f"Energy requirement for load {self.owner.name} fell below zero " \
                 f"({self.energy_requirement_Wh[market.time_slot]})."
 
-        self.remove_bid_from_pending(bid_trade.offer.id, market)
```
hannesdiedrich: Sorry, this line was just added in the last commit and it is not in `super().event_bid_traded()`
Can you please explain?

src/d3a/models/strategy/storage.py
hannesdiedrich: Thanks for explaining! Makes sense! I was a little confused here because i though the user can upload a tuple of tuple (a profile of Breakeven tuples).
And I have to read more carefully in the future and dont confuse a elif with an if :-)

hannesdiedrich:
BTW: Thanks a lot for the general comment that guided us through the commits! Very helpful!!!!

src/d3a/models/strategy/load_hours.py
spyrostz: Ah yes, this is a bit misleading here. The bid deletion is handled by 'add_bid_to_bought' method of the BidEnabledStrategy class (the super() of the LoadHoursStrategy). This method has an optional parameter called 'remove_bid' which is true by default, so by default it deletes the bid from the pending bids (self._bids) and moves it to the bought bids (self._traded_bids). Therefore there is no need for having the bid removal twice here, that's why I deleted this line.

tests/test_strategy_storage.py
```diff
@@ -741,14 +741,17 @@ def test_storage_event_trade(storage_strategy_test11, market_test13):
...
     storage_strategy_test13.event_activate()
     storage_strategy_test13.event_market_cycle()
     storage_strategy_test13.event_balancing_market_cycle()
     assert len(area_test13.test_balancing_market.created_balancing_offers) == 0
+    ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET = False
 
 
 def test_balancing_offers_are_created_if_device_in_registry(
         storage_strategy_test13, area_test13):
+    ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET = True
```
faizan2590: Im wondering how it was able to place balancing offer before that

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -145,24 +146,36 @@ def event_bid_traded(self, *, bid_trade):
...
         # Bid was traded in the source market by someone else
         elif bid_trade.offer.id == bid_info.source_bid.id:
-            if self.owner.name == bid_trade.seller:
-                return
             self.delete_forwarded_bids(bid_info)
         else:
             raise Exception(f"Invalid bid state for IAA {self.owner.name}: "
                             f"traded bid {bid_trade} was not in offered bids tuple {bid_info}")
 
+    def after_successful_trade_event(self, source_trade, bid_info):
```
faizan2590: Would it be a good idea to move this method to IAAEngine class, and using it to handle both offers and bids?

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -179,3 +192,27 @@ def event_bid_deleted(self, *, bid):
...
+
+    def event_bid_changed(self, *, market_id, existing_bid, new_bid):
+        market = self.owner._get_market_from_market_id(market_id)
+        if market is None:
+            return
+
+        if market == self.markets.target:
+            # one of our forwarded bids was split, so save the residual bid for handling
+            # the upcoming trade event
+            assert existing_bid.id not in self.bid_trade_residual, \
```
faizan2590: Sorry, I didn't get it. Why should "Bid should only change once before each trade"

spyrostz: Event bid changed is emitted only once for every bid trade (and only if the bid trade is partial). If we receive twice an event that the same bid has changed on the target market, without receiving a bid trade event meanwhile, we know that something is wrong and we have received a duplicate event.

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
spyrostz: The logic is different in both cases. In the case of the offer the information about the residual offer has to be parsed before the accept_offer, because IIRC it needs to be associated with that. For the bids we do not need that, therefore the code here will actually be a bit different.

tests/test_strategy_storage.py
spyrostz: Me too, I was also surprised that we only saw that now.

PR #383, opened 2019-02-20
D3ASIM-1001: Moving last markets of all areas to past_market to maint…

faizan2590:
…ain energy_balance

### DIFF
--- a/integration_tests/integration_tests.feature
+++ b/integration_tests/integration_tests.feature
@@ -79,9 +79,9 @@ Feature: Run integration tests
      Then we test the export functionality of default_2a
 
   Scenario: Simulation returns same results for different market counts
-    Given we have a scenario named default_2b
+    Given we have a scenario named default_2a
     And d3a is installed
-    When we run the simulation with setup file default_2b with two different market_counts
+    When we run the simulation with setup file default_2a with two different market_counts
     Then the results are the same for each simulation run
 
   Scenario: Test aggregated results are exported
--- a/integration_tests/integration_tests.feature
+++ b/integration_tests/integration_tests.feature
@@ -79,9 +79,9 @@ Feature: Run integration tests
      Then we test the export functionality of default_2a
 
   Scenario: Simulation returns same results for different market counts
-    Given we have a scenario named default_2a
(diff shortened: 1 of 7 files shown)


integration_tests/integration_tests.feature
```diff
@@ -79,9 +79,9 @@ Feature: Run integration tests
      Then we test the export functionality of default_2a
 
   Scenario: Simulation returns same results for different market counts
-    Given we have a scenario named default_2a
+    Given we have a scenario named default_2b
```
hannesdiedrich: Sorry, what does this change solve?

src/d3a/models/area/__init__.py
```diff
@@ -241,7 +241,9 @@ def now(self) -> DateTime:
 
     @property
     def all_markets(self):
-        return list(self._markets.markets.values())
+        return [m for m in self._markets.markets.values()
```
hannesdiedrich: Just to clarify: Here, you omit the additional markets that are created due to market_count > 1?

integration_tests/integration_tests.feature
spyrostz: We tried different profiles when testing this, it should be possible by now to revert to the old profile. I agree to the comment btw.

src/d3a/models/strategy/load_hours.py
```diff
@@ -72,7 +72,9 @@ def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None, daily_budget=
     def active_markets(self):
         markets = []
         for market in self.area.all_markets:
-            if self._allowed_operating_hours(market.time_slot):
+            if self._allowed_operating_hours(market.time_slot) and \
+                    self.area.config.start_date <= market.time_slot <= \
```
spyrostz: Since it is used in 2 places, I wonder if it makes sense to refactor this to a simple util function named 'is_market_in_simulation_duration' or similar.

src/d3a/models/area/__init__.py
faizan2590: Yes

PR #385, opened 2019-02-21
Feature/d3 asim 960

hannesdiedrich:
This PR adds the export of device statistic plots.
The export is set to True by default. 
You can deactivate if by setting `ConstSettings.GeneralSettings.EXPORT_DEVICE_PLOTS=False`
One or all of [@mfatma, @dirkbig, @sarahhambridge ]: Please let me know how you like the plots. Is there something missing? Are there artefacts?
Thanks!
@spyrostz  and @faizan2590 I will let you know when you can review the code (once the others approved this)


src/d3a/d3a_core/export.py
```diff
@@ -113,10 +125,12 @@ def export(self):
         self.plot_all_unmatched_loads()
         self.plot_avg_trade_price(self.area, self.plot_dir)
         self.plot_ess_soc_history(self.area, self.plot_dir)
-        self.move_root_plot_folder()
+        if ConstSettings.GeneralSettings.EXPORT_DEVICE_PLOTS:
+            self.plot_device_stats(self.area, [])
         if ConstSettings.IAASettings.MARKET_TYPE == 3 and \
                 ConstSettings.GeneralSettings.SUPPLY_DEMAND_PLOTS:
             self.plot_supply_demand_curve(self.area, self.plot_dir)
+        self.move_root_plot_folder()
```
hannesdiedrich: @faizan2590 FYI: This change (moving `self.move_root_plot_folder()` further down fixes an issue with the supply-demand-plots. If you like, i can explain offline.

src/d3a/d3a_core/export.py
```diff
@@ -241,6 +255,51 @@ def _export_area_stats_csv_file(self, area: Area, directory: dir, balancing: boo
...
+            else:
+                address_list = new_node_address_list + [child.name]
+                self._plot_device_stats(address_list)
+
+    def _plot_device_stats(self, address_list: list):
+        """
+        Plots device graphs
+        """
+        # Dont use the root area name for address list:
+        device_address_list = address_list[1::]
```
hannesdiedrich: From offline discussion with @faizan2590 , this should not be handled like this. There should always be a "brig" member in `self.endpoint_buffer.device_statistics.device_stats_dict` even though there is not device directly connected to it. Will change this in `device_statistics.py`.

hannesdiedrich:
From offline discussion: €/kWh instead of EUR/kWh

src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -0,0 +1,584 @@
...
+import pendulum
+from sortedcontainers import SortedDict
+
+from d3a.constants import TIME_ZONE
+from d3a.models.strategy.storage import StorageStrategy
+from d3a.models.strategy.load_hours import LoadHoursStrategy
+from d3a.models.strategy.pv import PVStrategy
+from d3a import limit_float_precision
+
+_log = logging.getLogger(__name__)
```
faizan2590: Can you tell me where is it used?

src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -0,0 +1,584 @@
...
+    2: "feed_in_tariff_pricing",
+    3: "net_metering_pricing"
+}
+
+EXPORT_DEVICE_VARIABLES = ["trade_energy_kWh", "pv_production_kWh", "trade_price_eur",
+                           "soc_history_%", "load_profile_kWh"]
+
+green = 'rgba(20,150,20, alpha)'
+purple = 'rgb(156, 110, 177, alpha)'
+blue = 'rgba(0,0,200,alpha)'
```
faizan2590: Can you tell me the reason to have rgba for 'green' and 'blue, whereas rgb for purple?

src/d3a/d3a_core/sim_results/plotly_graph.py
hannesdiedrich: It is not used. Thanks! Will remove it.

src/d3a/d3a_core/sim_results/plotly_graph.py
hannesdiedrich: This is a typo. Will fix it. Thanks

src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -0,0 +1,584 @@
...
+        }
+        for key, value in self.trade_history.items():
+            fig["data"][0]["values"].append(value)
+            fig["data"][0]["labels"].append(key)
+
+        py.offline.plot(fig, filename=filename, auto_open=False)
+
+    @classmethod
+    def _plot_line_time_series(cls, device_dict, var_name):
+        color = _get_color(var_name, 1)
```
spyrostz: Trivial, but I would replace this 1 with a const variable named 'OPAQUE_ALPHA', and 0.4 to 'TRANSPARENT_ALPHA'

src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -0,0 +1,584 @@
...
+    def __init__(self, dataset: dict, key: str):
+        self.key = key
+        self.dataset = dataset
+        self.umHours = dict()
+        self.rate = list()
+        self.energy = list()
+        self.trade_history = dict()
+
+    @staticmethod
+    def _line_plot(curve_point, time, supply):
```
spyrostz: Since this method is not called only from this class anymore, I would make it public and rename it to 'render_line_plot' or similar.

src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -0,0 +1,584 @@
...
+            }
+        }
+        for key, value in self.trade_history.items():
+            fig["data"][0]["values"].append(value)
+            fig["data"][0]["labels"].append(key)
+
+        py.offline.plot(fig, filename=filename, auto_open=False)
+
+    @classmethod
+    def _plot_line_time_series(cls, device_dict, var_name):
```
spyrostz: Same for this, I think it is better this method to be public since it is called from outside of this class

src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -0,0 +1,584 @@
...
+        self.key = key
+        self.dataset = dataset
+        self.umHours = dict()
+        self.rate = list()
+        self.energy = list()
+        self.trade_history = dict()
+
+    @staticmethod
+    def _line_plot(curve_point, time, supply):
+        graph_obj = PlotlyGraph(curve_point, time)
```
spyrostz: This is actually weird, since the object of the class is instantiated inside the same class. To my eyes this looks very error-prone, and I would suggest to move this method out of the class, in the same module. It is a static method anyways therefore it does not need any member of the class.

src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -0,0 +1,584 @@
...
+        data_obj = go.Scatter(x=list(graph_obj.energy),
+                              y=list(graph_obj.rate),
+                              mode='lines',
+                              name=name)
+        return data_obj
+
+    def supply_demand_curve(self, supply=True):
+        sort_values = SortedDict(self.dataset)
+        if supply:
+            self.rate = list(sort_values.keys())
```
spyrostz: rate does not need to be a member of the class, a simple variable could suffice, same for self.energy.

src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -0,0 +1,584 @@
...
+        """
+        data_max = max([abs(x) for x in list(device_dict[var_name].values()) if x is not None])
+        data_max_margin = data_max + data_max * 0.1
+        if start_at_zero:
+            return [0, abs(data_max_margin)]
+        else:
+            return [-data_max_margin, data_max_margin]
+
+    @classmethod
+    def _plot_device_profile(cls, device_dict, device_name, output_file, device_strategy):
```
spyrostz: It is better this classmethod to be converted to public.

src/d3a/d3a_core/sim_results/plotly_graph.py
hannesdiedrich: You are right, will fix this

PR #389, opened 2019-02-25
D3ASIM-1012: Refactored integration tests to increase their slot and …

spyrostz:
…tick length, in order to accelerate runtime of the integration tests.


integration_tests/steps/integration_tests.py
```diff
@@ -193,10 +191,10 @@ def load_profile_scenario(context):
         }
       ]
     }
-    context._settings = SimulationConfig(tick_length=duration(seconds=15),
-                                         slot_length=duration(minutes=15),
+    context._settings = SimulationConfig(tick_length=duration(seconds=60),
+                                         slot_length=duration(minutes=60),
                                          sim_duration=duration(hours=24),
-                                         market_count=4,
+                                         market_count=1,
```
hannesdiedrich: Can you explain the reason for this change pleas? Thanks!

hannesdiedrich: Is this increasing the speed as well?

spyrostz: It increases the speed yes, but I think this is not the reason why I changed it. Let me try it out and I will come back at you with the reason.

spyrostz: Ok, I checked it out, and the reason I removed that was actually that it was checking the predefined load profile not only during the duration of the simulation, but on the first timeslots of the next day as well. I took the easy way out and just hardcoded the market count to 1, in order for the future markets to not span over the duration of the test.

hannesdiedrich: Sorry for being overly "chasing" here, but can you point me to the place were you check the first timeslots of the next day? I can not find it. Thanks!

spyrostz: Don't be sorry, it would be good to figure it out. From a brief search, the reason is [this line](https://github.com/gridsingularity/d3a/pull/389/files/bd204659ed7f55a45bcd0ca95c6143ad161935b6#diff-0e919c3f7c79c6df9954c6557f70efa5R621). Looks like the desired_energy_Wh is not populated according to the market objects, but it uses its own calculation (have a look at the 'generate_market_slots_list' of the util module, which does not clamp the values from the start till the end of the simulation, but generates values even after the end of the simulation). Personally I do not mind since it is only used for forecasts up to now, but if you want to have the same behavior as the markets, I can add checks for the start and the end of the simulation in this method.

hannesdiedrich: @spyrostz Can you please correct the link? It just points to this PR. I kind of still not understand what you mean. But maybe the link will make it clearer. If not, lets discuss offline.

PR #390, opened 2019-03-04
Feature/d3 asim 935

hannesdiedrich:


src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -52,10 +52,10 @@ def _sorting(self, obj, reverse_order=False):
                 obj.values(),
                 key=lambda b: b.price / b.energy))
 
-    def _discrete_point_curve(self, obj_list):
+    def _discrete_point_curve(self, obj_list, functor):
```
spyrostz: It was my bad naming this to something as generic as functor, I would rename this to something like round_functor

hannesdiedrich: Yes, will change that, thanks!

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -81,8 +81,9 @@ def _perform_pay_as_clear_matching(self):
         if len(self.sorted_bids) == 0 or len(self.sorted_offers) == 0:
             return
 
-        cumulative_bids = self._discrete_point_curve(self.sorted_bids)
-        cumulative_offers = self._discrete_point_curve(self.sorted_offers)
+        # TODO: math.floor and math.ceil distinguishing will be resolved in the frame of D3ASIM-907
+        cumulative_bids = self._discrete_point_curve(self.sorted_bids, math.floor)
+        cumulative_offers = self._discrete_point_curve(self.sorted_offers, math.ceil)
```
faizan2590: Can you help me understand the purpose of using different rounding method between bids and offers.

hannesdiedrich: Yes, spyros and me discovered this bug yesterday. ATM, clearing is only done in integer steps and here only math.floor was used. If you have an offer rate of 10.1, the clearing rate would be 10, which is lower than the offer rate and would trigger this assert:
https://github.com/gridsingularity/d3a/blob/7e7f2000f42947ccfa7344fad1e7ad6460d6023b/src/d3a/models/market/one_sided.py#L122
For the bids it is the other end which is why we had to distinguish between both cases. This is only a quick fix and should be dealt with in D3ASIM-907, as stated in the comment above.

PR #391, opened 2019-03-05
Feature/d3 asim 1014

spyrostz:


src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -72,11 +72,16 @@ def start(scenario, settings):
...
-                   redis_job_id=job.id,
-                   kwargs=kwargs)
+    try:
+        run_simulation(setup_module_name=scenario_name,
+                       simulation_config=config,
+                       slowdown=settings.get('slowdown', 0),
+                       redis_job_id=job.id,
+                       kwargs=kwargs)
+    except Exception:
+        import traceback
```
hannesdiedrich: Is there a reason for this local import? (also next line)

src/d3a/d3a_core/launcher.py
```diff
@@ -32,22 +32,23 @@
...
     def run(self):
-        self._start_worker()
-        n_jobs = 0
+        self.job_array.append(self._start_worker())
         while True:
             sleep(1)
-            if n_jobs <= self.n_jobs and self.is_crowded():
-                n_jobs += 1
+            if len(self.job_array) <= self.max_jobs and self.is_crowded():
                 self._start_worker()
```
hannesdiedrich: Shouldn't this also be `self.job_array.append(self._start_worker())` ?

src/d3a/d3a_core/d3a_jobs.py
spyrostz: I decided to add the imports here because they are really only needed in case the simulation crashes. By lazily importing these here we ensure that these modules are going to be imported only when the simulation crashes, and only in production mode. Therefore a CLI user will never import these modules.

src/d3a/d3a_core/launcher.py
spyrostz: You are right, thanks a lot! Fixing this

src/d3a/d3a_core/d3a_jobs.py
hannesdiedrich: Makes sense! Thanks!

src/d3a/d3a_core/redis_communication.py
```diff
@@ -39,7 +42,7 @@ def __init__(self, simulation, simulation_id):
                                    self._simulation_id + "/pause": self._pause_callback,
                                    self._simulation_id + "/resume": self._resume_callback,
                                    self._simulation_id + "/slowdown": self._slowdown_callback}
-        self.result_channel = "d3a-results"
+        self.result_channel = RESULTS_CHANNEL
```
faizan2590: Sorry, I didn't understand the advantage of this?

spyrostz: I wanted to try to isolate the output channels of d3a, in order for maybe in the future to move them to ConstSettings or something similar. Furthermore, it is in general good practice to never have hardcoded values anywhere in the code, even though we do that it is generally good to avoid that.

PR #397, opened 2019-03-12
D3ASIM-1025: Adapted cumulative grid trades for Redis to support the …

spyrostz:
…updated endpoint structure.

### DIFF
--- a/src/d3a/d3a_core/sim_results/area_statistics.py
+++ b/src/d3a/d3a_core/sim_results/area_statistics.py
@@ -23,6 +23,7 @@ from d3a.models.strategy.pv import PVStrategy
 from d3a.models.strategy.commercial_producer import CommercialStrategy
 from d3a.models.strategy.load_hours import CellTowerLoadHoursStrategy, LoadHoursStrategy
 from d3a.d3a_core.util import area_name_from_area_or_iaa_name, make_iaa_name
+from d3a.constants import FLOATING_POINT_TOLERANCE
 
 loads_avg_prices = namedtuple('loads_avg_prices', ['load', 'price'])
 prices_pv_stor_energy = namedtuple('prices_pv_stor_energy', ['price', 'pv_energ', 'stor_energ'])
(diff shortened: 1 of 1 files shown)


src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -152,19 +154,20 @@ def _accumulate_load_trades(load, grid, accumulated_trades, is_cell_tower):
     return accumulated_trades
 
 
-def _accumulate_producer_trades(load, grid, accumulated_trades):
-    accumulated_trades[load.name] = {
-        "id": load.area_id,
+def _accumulate_producer_trades(producer, grid, accumulated_trades):
+    assert producer.name not in accumulated_trades
```
hannesdiedrich: What happens if this assert is triggered? I dont see some catch, or did I miss it?

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -327,19 +330,53 @@ def generate_inter_area_trade_details(area, past_market_types):
...
+             "priceLabel":
+                 f"{child.name} Earned {str(round(abs(area_data['earned']), 3))} cents"}
+        )
+    if child.name in area_data["consumedFrom"]:
+        energy = area_data["consumedFrom"][child.name]
+        money = area_data["spentTo"][child.name]
+        results["bars"].append({
+            "energy": energy,
+            "targetArea": child.name,
+            "energyLabel": f"{child.name} Consumed {str(round(energy, 3))} kWh from {child.name}",
```
hannesdiedrich: Just very minor but is bugging me since this plot was added to the fronend: Why capital "Consumed" (and also Produced, Earned, Spent) ? I read the string as a sentence and consequently the first letter of this verb has to be lowercase, right?

spyrostz: Because Javascript uses mostly CamelCase format on its code, and since this code is mostly for consumption for the frontend, it makes sense to have these values in this format, instead of snake case that Python uses. At least this was the reason why we originally used camel case for these.

src/d3a/d3a_core/sim_results/area_statistics.py
spyrostz: No catch, the simulation terminates. It is deliberate, because it is a logical mistake in the code if get into this method twice for the same producer.

src/d3a/d3a_core/sim_results/area_statistics.py
hannesdiedrich: Sorry, i want clear. I did not mean the key string. I mean the value string.

spyrostz: Ahhhhh now I see it ! You are correct, I am fixing this. Thanks!

PR #398, opened 2019-03-14
Feature/d3 asim 1024

hannesdiedrich:
Once approved, i will also delete the test setup file jira/d3asim_1024.py.


hannesdiedrich:
Please disregard the commit messages. They dont make sense, i know. Will squash them once approved.

integration_tests/steps/jira_issues_tests.py
```diff
@@ -122,8 +123,8 @@ def check_matching_trades(context):
 
 @then('there should be no unmatched loads')
 def no_unmatched_loads(context):
-    unmatched = export_unmatched_loads(context.simulation.area)
-    assert unmatched["unmatched_load_count"] == 0
+    unmatched, unmatched_redis = ExportUnmatchedLoads(context.simulation.area).__call__()
```
spyrostz: Actually, there is no need for the __call__(), you can just (). This is the reason I suggested the __call__ magic method, to avoid having to even access a new method.

hannesdiedrich: OMG! You are right! Will fix this!

src/d3a/d3a_core/util.py
```diff
@@ -402,3 +402,13 @@ def convert_datetime_to_str_keys(indict, outdict):
                 convert_datetime_to_str_keys(indict[key], outdict[key])
 
     return outdict
+
+
+def get_number_of_unmatched_loads(indict):
```
spyrostz: This function is too specific to be placed in a generic module like utils. Can it not be located in the unmatched loads modules ?

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
```diff
@@ -15,127 +15,160 @@
...
+            self.name_uuid_map[child.name] = child.uuid
+            if child.children:
+                indict[area.name] = self.find_unmatched_loads(child, indict[area.name])
+            else:
+                if isinstance(child.strategy, LoadHoursStrategy):
+                    indict[area.name][child.name] = self._accumulate_unmatched_loads(child)
+        return indict
+
+    @classmethod
+    def _accumulate_unmatched_loads(cls, area):
```
spyrostz: Trivial, but this method does not actually accumulate, but just calculates the unmatched load for a leaf device. I would prefer if it was named similar to '_calculate_unmatched_loads_leaf_area' or similar

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
```diff
@@ -15,127 +15,160 @@
...
+        return outdict
+
+    @classmethod
+    def _get_hover_info(cls, indict, hour_time):
+        """
+        returns dict of UL for each subarea for the hover the UL graph
+        """
+        hover_dict = {}
+        ul_count = 0
+        for child_name, child_ul_list in indict.items():
```
spyrostz: I was wondering whether it is possible to use groupby to perform this grouping [like here](https://stackoverflow.com/a/25137532). No need to try it out, I was just wondering if it would be easier that way.

integration_tests/steps/integration_tests.py
```diff
@@ -554,8 +553,9 @@ def run_sim(context, scenario, total_duration, slot_length, tick_length, iaa_fee
 def test_output(context, scenario, sim_duration, slot_length, tick_length):
 
     if scenario in ["default_2a", "default_2b", "default_3"]:
-        unmatched = export_unmatched_loads(context.simulation.area)
-        assert unmatched["unmatched_load_count"] == 0
+        unmatched_loads, unmatched_loads_redis = \
+            ExportUnmatchedLoads(context.simulation.area).__call__()
```
spyrostz: Like in the other comment, please remove that too, thanks!

src/d3a/d3a_core/util.py
hannesdiedrich: Yes, you are right. will move it there.

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
hannesdiedrich: Sorry, I dont see how this could be helpful here. Maybe you could explain offline.

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
```diff
@@ -15,127 +15,171 @@
...
+        postprocessing: changing area names to uuids
+        """
+        new = {}
+        for k, v in indict.items():
+            new[self.name_uuid_map[k]] = v
+        return new
+
+    def _accumulate_all_uls_in_branch(self, subdict, unmatched_list) -> list:
+        """
+        aggregate all UL times in one branch into one list
```
faizan2590: I guess its okay concatenate the unmatched_loads to uls in method naming. But in comments it would be better to use full name for better understanding

PR #400, opened 2019-03-18
Feature/d3 asim 1028

faizan2590:


src/d3a/models/market/balancing.py
```diff
@@ -42,18 +42,22 @@ def __init__(self, time_slot=None, area=None, notification_listener=None, readon
 
         super().__init__(time_slot, area, notification_listener, readonly)
 
-    def offer(self, price: float, energy: float, seller: str):
+    def offer(self, price: float, energy: float, seller: str, iaaFee: bool = False):
```
spyrostz: Please use snake_case for variables and argument names. I wonder why flake did not complain about that though.

integration_tests/steps/balancing_market.py
```diff
@@ -119,20 +119,46 @@ def follow_device_registry_energy_rate(context, device_name):
...
+                           < FLOATING_POINT_TOLERANCE
+
+    for ii, device in enumerate([house1, house2, context.simulation.area]):
+        for market in device.past_balancing_markets:
+            for trade in market.trades:
+                if trade.offer.energy > 0:
+                    assert (trade.offer.price / trade.offer.energy) - positive_energy_rate \
+                           < FLOATING_POINT_TOLERANCE
+
+    # assert all((trade.offer.price / trade.offer.energy) - abs(negative_energy_rate)
```
spyrostz: I understand that the reason for that is that the balancing market fails with the new market fee calculation (or is it not the reason and I did not understand it correctly?), but the balancing market needs to also be adapted to ignore the market fee calculations.

src/d3a/models/area/__init__.py
```diff
@@ -109,12 +111,15 @@ def activate(self, bc=None):
...
                 self.budget_keeper.activate()
+        if self.transfer_fee_pct is None:
+            self.transfer_fee_pct = self.config.iaa_fee
 
         # Cycle markets without triggering it's own event chain.
         self._cycle_markets(_trigger_event=False)
 
         if not self.strategy and self.parent is not None:
             self.log.warning("No strategy. Using inter area agent.")
+            self.log.warning(f"PERCENTAGE GRID FEE {self.transfer_fee_pct}%")
```
spyrostz: Make sure to remove it after the PR, since it spams the logs

src/d3a/models/market/balancing.py
```diff
@@ -84,14 +87,23 @@ def accept_offer(self, offer_or_id: Union[str, BalancingOffer], buyer: str, *,
...
                 # Partial trade
                 if energy == 0:
                     raise InvalidBalancingTradeException("Energy can not be zero.")
+
                 elif abs(energy) < abs(offer.energy):
                     original_offer = offer
                     accepted_offer = Offer(
                         offer.id,
-                        abs((offer.price / offer.energy) * energy),
+                        abs((source_rate) * energy),
```
spyrostz: Extra brackets not needed anymore

src/d3a/models/area/__init__.py
faizan2590: Its only triggered at the start of simulation. Not sure if at least once in log could clutter it. Might not be super important but in future when there is separate grid fee for different area. It might come handy to analyse.

src/d3a/models/market/one_sided.py
```diff
@@ -106,25 +110,34 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *, energy: in
             if time is None:
                 time = self._now
             if energy is not None:
-                # Partial trade
                 if energy == 0:
                     raise InvalidTrade("Energy can not be zero.")
+                # partial energy is requested
                 elif energy < offer.energy:
                     original_offer = offer
+                    # Don't know why is it here
```
spyrostz: It is used in order to select the correct hash value for the case of smart contract. If we are not running on blockchain, the offer.id is selected. If we are running on blockchain, the offer.real_id is selected, which is the hash id of the offer on the smart contract. If this is clear you can remove the comment.

integration_tests/steps/balancing_market.py
```diff
@@ -119,16 +119,14 @@ def follow_device_registry_energy_rate(context, device_name):
     positive_energy_rate = DeviceRegistry.REGISTRY[device_name][1]
     house1 = next(filter(lambda x: x.name == "House 1", context.simulation.area.children))
     house2 = next(filter(lambda x: x.name == "House 2", context.simulation.area.children))
-    iaa_fee = context.simulation.simulation_config.iaa_fee
-    add_val = [0, 2*iaa_fee/100 + iaa_fee/100**2, iaa_fee/100]
```
spyrostz: Before moving on, I will take some time to dig a bit deeper on exactly how/why this was implemented. I will come back on the review with a response whether it really makes sense to remove this, or not.

src/d3a/models/area/__init__.py
spyrostz: The problem I have with this is that it simply displays a setting. If one wants to see the setting value, it is available throughout the usual config locations, that's why I do not find it useful to be on the log. Usually on the log, even in warning level, we display the output of some calculations and not config values.

integration_tests/steps/integration_tests.py
```diff
@@ -705,7 +705,8 @@ def test_finite_plant_energy_rate(context, plant_name):
             assert trade.buyer is not finite.name
             if trade.seller == finite.name:
                 trades_sold.append(trade)
-        assert all([isclose(trade.offer.price / trade.offer.energy, finite.strategy.energy_rate)
+        assert all([isclose(trade.offer.price / trade.offer.energy,
+                            finite.strategy.energy_rate, rel_tol=1e-02)
```
spyrostz: I wonder why the tests started to fail now and succeeded before these changes. I will take the time to debug it.

src/d3a/models/market/balancing.py
```diff
@@ -42,18 +42,22 @@ def __init__(self, time_slot=None, area=None, notification_listener=None, readon
...
+                        iaaFee: bool = False) -> BalancingOffer:
         if seller not in DeviceRegistry.REGISTRY.keys() and not from_agent:
             raise DeviceNotInRegistryError(f"Device {seller} "
                                            f"not in registry ({DeviceRegistry.REGISTRY}).")
         if self.readonly:
             raise MarketReadOnlyException()
         if energy == 0:
             raise InvalidOffer()
+        if iaaFee:
+            price = price * (1 + self.transfer_fee_pct / 100)
```
spyrostz: Why not holding a self.transfer_fee_fraction and eliminate these hard-to-maintain divisions/multiplications by 100 ?

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -105,8 +104,8 @@ def _match_offers_bids(self):
             return
         clearing_rate, clearing_energy = clearing
         if clearing_energy > 0:
-            self.owner.log.info(f"Market Clearing Rate: {clearing_rate} "
-                                f"||| Clearing Energy: {clearing_energy} ")
+            self.owner.log.warning(f"Market Clearing Rate: {clearing_rate} "
```
spyrostz: I would suggest to revert it, since it does not have anything to do with this story. There was also quite a long discussion about that in a past review, and I am under the impression we decided to keep it to info level.

tests/test_strategy_pvpredefined.py
```diff
@@ -19,7 +19,7 @@
 import pendulum
 import uuid
 import pathlib
-from pendulum import DateTime, duration, today
+from pendulum import DateTime, duration, today  # NOQA
```
spyrostz: Why is the NOQA useful here?

src/d3a/setup/two_sided_market/grid_fee_test.py
```diff
@@ -0,0 +1,69 @@
+"""
```
spyrostz: I did not see any integration tests that use this file, therefore this file should be either deleted or moved to d3a-private.

integration_tests/steps/balancing_market.py
spyrostz: It was what I was afraid of. Apparently the changes break the operation of the balancing market if I understand correctly. I tried to run:
`d3a -l WARN run --setup balancing_market.default_2a  -t 15s -s 15m `
which is one of the setup files that the integration tests use. There were multiple cases where the following log was observed:
`
15:40:57.500 WARNING  ( 148) d3a.models.market.balancing   : [BALANCING_TRADE][2019-03-20T16:00] {96eb4e} [BA House 2 -> House 2 Reserve] -0.005000000000000001 kWh (partial) @ 0.16831650000000006 -33.66330000000001 d4cec2a7-20ae-4fe6-a505-e46bbe3634bd
15:40:57.500 WARNING  ( 148) d3a.models.market.balancing   : [BALANCING_TRADE][2019-03-20T16:00] {584ea4} [BA House 1 -> BA House 2] -0.005000000000000001 kWh (partial) @ 0.16500000000000004 -33.0 ab74a59f-3da9-4462-8f41-d80213cd0ffa
15:40:57.501 WARNING  ( 148) d3a.models.market.balancing   : [BALANCING_TRADE][2019-03-20T16:00] {97c3e3} [H1 General Load -> BA House 1] -0.005000000000000001 kWh (partial) @ 0.16336633663366337 -32.67326732673267 ed3b34dd-e5d9-4010-b959-93935f259449
15:40:57.501 WARNING  ( 148) d3a.models.market.balancing   : [BALANCING_TRADE][2019-03-20T16:00] {ea90d3} [BA House 2 -> House 2 Reserve] 0.005000000000000001 kWh (partial) @ 0.17851750000000002 35.7035 117b5d4d-6e46-4371-962e-4396dfb11715
15:40:57.502 WARNING  ( 148) d3a.models.market.balancing   : [BALANCING_TRADE][2019-03-20T16:00] {a0c8fd} [BA House 1 -> BA House 2] 0.005000000000000001 kWh (partial) @ 0.17500000000000004 35.0 6dbb198c-e185-4f76-9501-fa5d2cec5f5f
15:40:57.502 WARNING  ( 148) d3a.models.market.balancing   : [BALANCING_TRADE][2019-03-20T16:00] {3668f7} [H1 Storage2 -> BA House 1] 0.005000000000000001 kWh (partial) @ 0.17326732673267328 34.65346534653465 a9b16405-ceb7-4f7f-a27c-a6d5ce77630a
`
The device registry is the following:
`
device_registry_dict = {
    "H1 General Load": (33, 35),
    "H2 General Load": (33, 35),
    "H1 Storage1": (33, 35),
    "H1 Storage2": (33, 35),
}
`
The evident problem is that the devices do not actually get paid according to the device registry, but less than that.  More specifically, Storage2 uses an energy rate of 34.65346534653465, while 35 was configured by the device registry. Note that this problem does not seem to affect the spot market, only the balancing market is affected by that. To summarise, I strongly suggest to revert the integration tests to their original form, since they obviously revealed the issue correctly, and to try to detect the reason why the balancing market in some cases does not use the device registry rate for the devices.

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
faizan2590: Sorry, left it there mistakenly. Had to visualize the clearing_rate in logs.

tests/test_strategy_pvpredefined.py
faizan2590: Removing, since one of the test was commented for a while. Had to do this then to avoid flake8

integration_tests/steps/load_tests.py
```diff
@@ -95,19 +95,16 @@ def check_min_user_rate_profile_dict(context):
...
     load2 = next(filter(lambda x: x.name == "H1 General Load 2", house.children))
 
-    num_of_hierarchies = 2
     for market in house.past_markets:
         assert len(market.trades) > 0
         for trade in market.trades:
             trade_rate = trade.offer.price / trade.offer.energy
             if trade.buyer == load1.name:
                 min_rate = load1.strategy.initial_buying_rate[market.time_slot]
-                assert trade_rate * (1 + context.simulation.simulation_config.iaa_fee /
```
spyrostz: Can you explain why was that removed ?

src/d3a/models/market/one_sided.py
```diff
@@ -117,14 +123,22 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *, energy: in
...
                         trade_rate = offer.price / offer.energy
+
                     # TODO: Remove the math.floor from the lower threshold once the pay-as-clear
                     # market is refactored (D3ASIM-907)
                     assert trade_rate + FLOATING_POINT_TOLERANCE >= \
                         math.floor(offer.price / offer.energy)
 
+                    # reducing trade_rate to be charged in terms of grid_fee
+                    if iaa_fee:
+                        source_rate = trade_rate / (1 + self.transfer_fee_ratio)
```
hannesdiedrich: Attention!   a / (1+b) != a - a*b 
I guess you were looking for `source_rate = trade_rate * (1 - self.transfer_fee_ratio)`, right?

src/d3a/models/strategy/area_agents/one_sided_engine.py
```diff
@@ -236,10 +233,11 @@ class BalancingEngine(IAAEngine):
 
     def _forward_offer(self, offer, offer_id):
         forwarded_balancing_offer = self.markets.target.balancing_offer(
-            offer.price + (offer.price * (self.transfer_fee_pct / 100)),
+            offer.price,
             offer.energy,
             self.owner.name,
-            True
+            True,
```
hannesdiedrich: I know it wasn't there before, but let's be more explicit here and change it to `from_agent=True` for the readability of the code.  Thanks!

src/d3a/models/area/event_dispatcher.py
```diff
@@ -151,9 +151,6 @@ def create_area_agents(self, is_spot_market, market):
                 owner=self.area,
                 higher_market=self.area.parent._markets.markets[market.time_slot],
                 lower_market=market,
-                transfer_fee_pct=0
-                if ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0
```
hannesdiedrich: I dont see this logic in the area (or market). Can you please double check with the science team whether it is still wanted, that the transfer_fee_pct is 0 for alternative Pricing schemes? Thanks!

integration_tests/steps/load_tests.py
hannesdiedrich: This is probably due to the fact that the iaa_fee is 0 if not set, right @faizan2590 ? For this Integration test setup it is not set, IIRC.

integration_tests/steps/balancing_market.py
hannesdiedrich: Sorry, I dont get it, Faizan only added spaces here. This implementation was done by me in an earlier ticket.
Before that the test was done only in integer space which IMO is not sufficient. Lets discuss offline.

integration_tests/steps/load_tests.py
hannesdiedrich: EDIT: Sorry, no, I misread. The iaa_fee is set to the global on if not set explicitly. So, yes, please explain. Thanks!

src/d3a/models/market/balancing.py
```diff
@@ -84,14 +90,22 @@ def accept_offer(self, offer_or_id: Union[str, BalancingOffer], buyer: str, *,
             if time is None:
                 time = self._now
             if energy is not None:
+
+                # reducing trade_rate to be charged in terms of grid_fee
+                if iaa_fee:
+                    source_rate = trade_rate / (1 + self.transfer_fee_ratio)
```
hannesdiedrich: Attention! a / (1+b) != a - a*b
I guess you were looking for source_rate = trade_rate * (1 - self.transfer_fee_ratio), right?

src/d3a/models/area/event_dispatcher.py
faizan2590: Thanks for pointing it, moving this logic to Area class.

src/d3a/models/market/balancing.py
faizan2590: No, its the other way around.
`target_rate/trade_rate = source_rate * (1 + self.transfer_fee_ratio)`.
I have rearranged it to get the source_rate.

src/d3a/models/market/one_sided.py
faizan2590: https://github.com/gridsingularity/d3a/pull/400/files#r267679532

src/d3a/models/market/balancing.py
hannesdiedrich: Ahh, I see. Sorry. Wanted to prevent you from a mistake that I was doing once. Thanks for explaining.

integration_tests/steps/integration_tests.py
hannesdiedrich: BTW, what happend here @spyrostz @faizan2590 ?

spyrostz: I never actually found out why this is necessary now. I strongly suggest to either try to remove the relative tolerance completely, since these balancing market fixes have been applied, and if it fails to try to change the tolerance to a much smaller value. 10^-2 is more or less the same as disabling the test.

PR #401, opened 2019-03-19
Feature/d3 asim 1007

spyrostz:


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -138,5 +141,52 @@ def calculate_prices(key, functor):
...
+    def _generate_external_and_total_bills(self, area, results, flattened):
+
+        all_child_results = [v for i in results[area.uuid] for _, v in i.items()]
+        results[area.uuid].append({"Accumulated Trades": {
+            'bought': sum(v['bought'] for v in all_child_results),
+            'sold': sum(v['sold'] for v in all_child_results),
+            'spent': sum(v['spent'] for v in all_child_results),
+            'earned': sum(v['earned'] for v in all_child_results),
+            'total_energy': sum(v['total_energy'] for v in all_child_results),
+            'total_cost': sum(v['total_cost'] for v in all_child_results),
```
hannesdiedrich: Not that i find it very important, but i dont see a member "type" here. Is this on purpose?

hannesdiedrich: Forget it, for the accumulated_bills for the area a type is not needed, right?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -77,7 +78,7 @@ def generate_json_report(self):
             "cumulative_loads": self.cumulative_loads,
             "price_energy_day": self.price_energy_day,
             "cumulative_grid_trades": self.cumulative_grid_trades,
-            "bills": self.bills,
+            "bills": self.bills_redis,
```
hannesdiedrich: Shouldn't this change be reverted to export self.bills locally?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -138,5 +141,52 @@ def calculate_prices(key, functor):
                 self._update_tree_summary(child)
 
     def _update_bills(self, area, past_market_types):
-        result = energy_bills(area, past_market_types)
-        return OrderedDict(sorted(result.items()))
+        result, external_trades = energy_bills(area, past_market_types, {})
+        return OrderedDict(sorted(result.items())), external_trades
+
+    def _calculate_redis_bills(self, area, energy_bills, external_trades):
+        from copy import deepcopy
```
hannesdiedrich: Any reason for this lazy import?

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -51,39 +51,56 @@ def total_avg_trade_price(markets):
...
+
+
+def _store_sold_trade(result_dict, trade_offer):
+    result_dict['sold'] += trade_offer.energy
+    result_dict['earned'] += trade_offer.price
+    result_dict['total_energy'] -= trade_offer.energy
+    result_dict['total_cost'] -= trade_offer.price
+
+
+def energy_bills(area, past_market_types, external_trades, from_slot=None, to_slot=None):
```
hannesdiedrich: Not part of this story, bit as far as I see, the from_slot to_slot functionality is not used at the moment. Is this already implemented for the future?

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -51,39 +51,56 @@ def total_avg_trade_price(markets):
...
-        return None
+        return None, None
     result = {child.name: dict(bought=0.0, sold=0.0,
                                spent=0.0, earned=0.0,
                                total_energy=0, total_cost=0,
                                type=get_area_type_string(child))
               for child in area.children}
+    external_trades[area.name] = dict(bought=0.0, sold=0.0,
+                                      spent=0.0, earned=0.0,
+                                      total_energy=0, total_cost=0)
```
hannesdiedrich: When I run the simulation, the "External Trade" member of the bills for each non-device area has a "type" member. Where is this set? I dont see it in the code. Can you help me please?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
spyrostz: Just to give some feedback on that, the type member is actually only needed for the former frontend, in order to know when to render a house, and when to render a cell tower image. Now that this is no longer the case, I think it is better to get rid of it globally. If you agree I will completely remove it.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
spyrostz: Oops, correct. I will revert this, thanks

src/d3a/d3a_core/sim_results/endpoint_buffer.py
spyrostz: You are right, this has to be global, moving this

src/d3a/d3a_core/sim_results/stats.py
spyrostz: It used to be useful, in order to calculate the energy bills for a time range, through the REST API of d3a. Now that the REST API has been removed, I think we can remove this, let me know if you agree and I will get rid of it. Thanks

src/d3a/d3a_core/sim_results/stats.py
spyrostz: The reason for that is this line https://github.com/gridsingularity/d3a/blob/feature/D3ASIM-1007/src/d3a/d3a_core/sim_results/endpoint_buffer.py#L190. External trades are actually generated from the flattened area results, since we only need to care about the parent area results. Thanks for pointing that out though, because I noticed that the external_trades are no longer used in the code. I will consequently remove these.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: Yes, please remove. I can not think of a usage later on. Thanks

src/d3a/d3a_core/sim_results/stats.py
hannesdiedrich: I agree, thanks!

src/d3a/d3a_core/sim_results/endpoint_buffer.py
spyrostz: Thanks, already deleted this.

src/d3a/d3a_core/sim_results/stats.py
spyrostz: Thanks, removed this

PR #406, opened 2019-03-29
Bug/d3 asim 1035

hannesdiedrich:


src/d3a/d3a_core/simulation.py
```diff
@@ -93,6 +93,7 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig =
         self._load_setup_module()
         self._init(**self.initial_params)
         self._init_events()
+        validate_const_settings_for_simulation()
```
spyrostz: I assume you want to validate the simulation const settings after loading the setup.py file ? This is nice.

hannesdiedrich: Yes, I should have added a comment in the PR.

dirkbig:
to be 100% sure: 
- if I want to run a simulation using a certain setup-up file, I can compare the results between a 2-sided d3a market scheme and an alternative scheme? By setting ConstSettings.IAASettings.MARKET_TYPE = 2 and adding the flag "--compare-alt-pricing";
- And it is not the case that when I try such a thing, the simulation changes its d3a market setting to one-sided market (...MARKET_TYPE = 1) automatically, thus then effectively comparing a 1-sided d3a market scheme with an an alternative scheme?

hannesdiedrich:
@dirkbig The first case is true now.

PR #408, opened 2019-04-01
Feature/d3 asim 1072

spyrostz:


src/d3a/d3a_core/util.py
```diff
@@ -388,14 +389,18 @@ def validate_const_settings_for_simulation():
...
     """
 
     for key, value in indict.items():
         if isinstance(key, DateTime):
-            outdict[key.format(DATE_TIME_FORMAT)] = indict[key]
+            if not ui_format:
+                outdict[key.format(DATE_TIME_FORMAT)] = indict[key]
+            else:
+                formatted_date = f"{calendar.month_name[key.month]} {key:%d %Y, %H:%M h}"
+                outdict[formatted_date] = indict[key]
```
hannesdiedrich: How about using pendulum instead:
`key.format("MMMM DD YYYY, HH:MM h")` ?

spyrostz: You are right, I totally forgot about pendulum. Fixing this, thanks!

PR #409, opened 2019-04-02
Feature/d3 asim 1043

hannesdiedrich:


src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -191,6 +195,16 @@ def _accumulate_house_trades(house, grid, accumulated_trades, past_market_types)
...
                 accumulated_trades[house.name]["spentTo"][seller_id] += trade.offer.price
+
+    for market in getattr(house, past_market_types):
+        for trade in market.trades:
+            if area_name_from_area_or_iaa_name(trade.seller) == house.name:
+                accumulated_trades[house.name]["consumedFromExternal"] += trade.offer.energy
+                accumulated_trades[house.name]["spentToExternal"] += trade.offer.price
+            elif area_name_from_area_or_iaa_name(trade.buyer) == house.name:
+                accumulated_trades[house.name]["producedForExternal"] -= trade.offer.energy
+                accumulated_trades[house.name]["earnedFromExternal"] -= trade.offer.price
```
spyrostz: This should be += and not -=, or did I not understand it correctly? The reason I am referring it is because it does not match the L190 of the same file.

hannesdiedrich: You are right! 
And additionally: I could change `-=` to `+=` one line above, too as these values are only needed once. Thus I can omit the `abs(` in L383 and L384 as well. 
Are you OK with that @spyrostz ?

spyrostz: I think for the energy you need the signs, because the graph is going to plot one bar on the negative axis for produced energy, and on the positive axis for consumed energy. Therefore the signs have to remain for the energy and they cannot be avoided. It is just for the prices that need to be removed.

hannesdiedrich: You are right! Thanks! *:face_palm:*

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -191,6 +195,16 @@ def _accumulate_house_trades(house, grid, accumulated_trades, past_market_types)
...
                 accumulated_trades[house.name]["spentTo"][seller_id] += trade.offer.price
+
+    for market in getattr(house, past_market_types):
+        for trade in market.trades:
+            if area_name_from_area_or_iaa_name(trade.seller) == house.name:
+                accumulated_trades[house.name]["consumedFromExternal"] += trade.offer.energy
+                accumulated_trades[house.name]["spentToExternal"] += trade.offer.price
+            elif area_name_from_area_or_iaa_name(trade.buyer) == house.name:
+                accumulated_trades[house.name]["producedForExternal"] -= trade.offer.energy
+                accumulated_trades[house.name]["earnedFromExternal"] += trade.offer.price
```
faizan2590: If we are following a sign convention. Then it has to be for both energy & price. For now, in this story it is irrelevant as you only need it for cumulative traded energy. But in case there is a need to plot cumulative energy bills, producer would be getting the money(to be shown in positive axis) whereas consumer would be paying the money(to be shown in negative axis).
So, I guess even now for uniformity to follow sign convention for both energy as well as for bills.

hannesdiedrich: You have a point, but @spyrostz was in favour of the += (see comment above).
I dont have a strong opinion on that. You are right, at the moment it is not used. So just let me know which one to implement. Thanks

spyrostz: @faizan2590 the reason why the prices need to be absolute values, is that the purpose of this function is not to provide general-purpose statistics, but to provide an output for the frontend graph. I would agree with you if we wanted to store these accumulated statistics for further processing later, but since the purpose of this is only for the graph, I would suggest to have the aggregated statistics in the way that the graph expects them. 
The prices in this graph are only reported in the tooltip window (and I am not even sure if they are going to decide to display these) as absolute values. Therefore I would say to stick on unsigned values here.

faizan2590: Ah sorry, didn't see @spyrostz  comment in your diff. Well then ignore it ATM. We will see later if there is some requirements later to plot "cumulative energy bills graph"

hannesdiedrich: Thank you @spyrostz this makes sense. Sticking with absolute values

PR #410, opened 2019-04-03
D3ASIM-1030: Reverted sign on the device producer entries.

spyrostz:


src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -155,7 +155,7 @@ def _accumulate_producer_trades(producer, grid, accumulated_trades):
     for market in grid.past_markets:
         for trade in market.trades:
             if trade.offer.seller == producer.name:
-                accumulated_trades[producer.name]["produced"] += trade.offer.energy
+                accumulated_trades[producer.name]["produced"] -= trade.offer.energy
```
faizan2590: We are explicitly using "produced" as a key, so why do we need to have signed value. Shouldn't the signed be used when it might either be positive or negative

spyrostz: TLDR: graphs. In a bit more detail, in the cumulative grid trades graph the produced energy is always represented as a negative value. Therefore when Airnauts tried to plot devices (in which case energy is always represented as positive) all the bars were on the positive part of the y-axis. This could obviously be distinguished by whether it is regarding produced or consumed energy, but it is easier to stick to sign convention.

PR #413, opened 2019-04-08
D3ASIM-1034: Setting max_job as env variable for staging server

faizan2590:


src/d3a/d3a_core/launcher.py
```diff
@@ -36,7 +37,7 @@ def __init__(self,
                  max_jobs=2,
                  max_delay_seconds=2):
         self.queue = queue or Queue('d3a', connection=StrictRedis.from_url(REDIS_URL))
-        self.max_jobs = max_jobs
+        self.max_jobs = max_jobs or MAX_JOBS
```
hannesdiedrich: AFAIK, this will return a boolean value and lead to an error. Please use an if statement.

PR #415, opened 2019-04-08
Post processing of iaa_fee in SimulationConfig

faizan2590:


src/d3a/models/config.py
```diff
@@ -110,3 +108,6 @@ def read_market_maker_rate(self, market_maker_rate):
         self.market_maker_rate = read_arbitrary_profile(InputProfileTypes.IDENTITY,
                                                         market_maker_rate_parsed)
         self.market_maker_rate = {k: float(v) for k, v in self.market_maker_rate.items()}
+
+    def read_iaa_fee(self, iaa_fee):
```
hannesdiedrich: I thought we agreed on a more general function like `read_config_params` or so that incorporates all parameters of the SimulationConfig class. This way the user does not have to call multiple functions to set the parameters but only one. Or did i get i wrong?

hannesdiedrich:
In general I propose to add integration test for this. Thanks!

hannesdiedrich:
Sorry, clicked the wrong button :-)

src/d3a/models/config.py
```diff
@@ -29,7 +29,8 @@
 
 class SimulationConfig:
     def __init__(self, sim_duration: duration, slot_length: duration, tick_length: duration,
-                 market_count: int, cloud_coverage: int, iaa_fee: int,
+                 market_count: int, cloud_coverage: int,
+                 iaa_fee: int=ConstSettings.IAASettings.FEE_PERCENTAGE,
```
hannesdiedrich: Sorry, as i am working on the same part of the code I am wondering, whether we want to keep the `iaa_fee` part of the SimulationConfig class. I am asking this because now as he iaa_fee is can also be set individually for each area, and is not a cli parameter any more, we could delete it here. 
The same applies IMO to the `cloud_coverage` and `pv_user_profile` . Is there a plan to move this to the PV class?

spyrostz: Hi @hannesdiedrich we should not move these 2 parameters from the config IMO, since these are useful for assigning global weather conditions for the whole grid. Otherwise one would need to upload independent profiles for each PV, instead of just uploading one profile.

hannesdiedrich: Ok, thanks! Considering small grids, it makes sense for now.

spyrostz: Regarding the iaa_fee though, to expand this discussion a bit more, I wonder if it makes sense to keep it as part of this class, since the user can always access it via ConstSettings.IAASettings.FEE_PERCENTAGE instead of using config.iaa_fee. Config parameters should be IMO only for configuration via d3a-web, otherwise the ConstSettings should be enough. This is a matter of preference though and I do not have a strong opinion on that.

hannesdiedrich: Hmm, tricky. It is indeed an important parameter for the simulation, it influences the whole behaviour of the trading. I think the user wants to set it on nearly every different setup. That is why I tend to be in favour of keeping it in the SimulationConfig class. 
It is true, that the user will provide this value for each area though.

src/d3a/models/config.py
faizan2590: How about updating it via already existed method(https://github.com/gridsingularity/d3a/pull/415/files#diff-ab2b09a6e41de97909debee94cf77eb1R83)?

hannesdiedrich: This could work (although this gets called every tick and might be a bit of an overkill)

src/d3a/d3a_core/cli.py
```diff
@@ -71,9 +71,6 @@ def main(log_level):
 @click.option('-c', '--cloud-coverage', type=int,
               default=ConstSettings.PVSettings.DEFAULT_POWER_PROFILE, show_default=True,
               help="Cloud coverage, 0 for sunny, 1 for partial coverage, 2 for clouds.")
-@click.option('-f', '--iaa-fee', type=int,
```
hannesdiedrich: Just to make sure: We decided in the offline discussion, that we remove this cli parameter but keep it in the SimulationConfig class for the d3a-web to be set, right? (need to know for my current story)

src/d3a/models/config.py
faizan2590: Okay, since the rest of parameters are explicitly updated via their own method. I suggest its better to have a separate method for iaa_fee too. But I have no strong opinion. Let me know, what looks good to you guys.

spyrostz: I agree with @faizan2590 , the event_update method is already used for area events (hence the 'event' part of the method), therefore it looks good to my eyes to not have a separate function to update these parameters via the setup file, but to reuse the same function. It is easy to adapt it and add the iaa_fee and market maker rate to it, even if the area events do not use these yet. 
I would vote against having a separate method for setting the IAA fee, because I find it more convenient to have one method that can set these parameters at once instead of having to call 2 methods.

src/d3a/models/config.py
spyrostz: To summarise the offline discussion with Sarah, there are plans to add this to the UI in the future, therefore I would vote to keep the iaa_fee parameter in the SimulationConfig class.

src/d3a/models/config.py
hannesdiedrich: Ok, makes sense, voting also for adding it to `event_update` (and also rename this function)

faizan2590: Sorry, I didn't get the reason to rename the method name

hannesdiedrich: Just for the user. I guess it is confusing to type `conf.event_update()`  but really want to set some configuration parameter. Does it make sense?

integration_tests/steps/integration_tests.py
```diff
@@ -781,3 +781,11 @@ def test_sim_market_count(context):
         market_4 = grid_4.get_past_market(market_1.time_slot)
         for area in market_1.traded_energy.keys():
             assert isclose(market_1.traded_energy[area], market_4.traded_energy[area])
+
+
+@then("we test the config parameters")
+def test_config_parameters(context):
+    grid = context.simulation.area
+    assert grid.config.iaa_fee == 5
+    assert all([isclose(rate, 35)
```
hannesdiedrich: why `isclose()` ?

src/d3a/setup/config_parameter_test.py
```diff
@@ -0,0 +1,69 @@
...
+from d3a.models.area import Area
+# from d3a.models.strategy.commercial_producer import CommercialStrategy
+# from d3a.models.strategy.finite_power_plant import FinitePowerPlant
+from d3a.models.strategy.storage import StorageStrategy
+from d3a.models.strategy.load_hours import CellTowerLoadHoursStrategy, LoadHoursStrategy
+from d3a.models.appliance.pv import PVAppliance
+from d3a.models.strategy.pv import PVStrategy
+
+
+def get_setup(config):
```
hannesdiedrich: I suggest to use a very simple setup file to save time. We dont need any trades here, do we?
Isnt there another setup file that can be reused?

faizan2590: I would vouch to make this setup file as simple as possible. I don't see any other setup where both parameters are modified together.

src/d3a/models/config.py
hannesdiedrich: So, any update here? I see that you did not rename it.

faizan2590: Sorry, missed it!

PR #416, opened 2019-04-10
Feature/d3 asim 901

hannesdiedrich:


tests/test_inter_area_agent.py
```diff
@@ -118,14 +120,16 @@ def delete_offer(self, *args):
     def delete_bid(self, *args):
         pass
 
-    def offer(self, price, energy, seller, agent=False, iaa_fee: bool = False):
+    def offer(self, price, energy, seller, iaa_fee: bool = False):
         self.offer_count += 1
+        if iaa_fee:
+            price = price * (1 + self.transfer_fee_ratio) + self.transfer_fee_const * energy
```
spyrostz: This is here because the constant fee is in cents/kWh, right ?

hannesdiedrich: Exacty

PR #419, opened 2019-04-10
D3ASIM-1087: Adding elements to the energy trade profile only in case…

spyrostz:
… of nonzero values.

### DIFF
--- a/src/d3a/d3a_core/sim_results/file_export_endpoints.py
+++ b/src/d3a/d3a_core/sim_results/file_export_endpoints.py
@@ -26,6 +26,7 @@ from d3a.models.strategy.pv import PVStrategy
 from d3a.models.strategy.commercial_producer import CommercialStrategy
 from d3a.models.strategy.finite_power_plant import FinitePowerPlant
 from d3a.d3a_core.util import convert_datetime_to_str_keys
+from d3a.constants import FLOATING_POINT_TOLERANCE
 
 
 class FileExportEndpoints:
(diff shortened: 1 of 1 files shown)


src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -92,17 +92,21 @@ def _calculate_devices_sold_bought_energy(self, area, past_markets):
                 if trade_seller.startswith("IAA "):
                     trade_seller = trade_seller[4:]
                 if trade_seller not in out_dict["sold_energy"]:
-                    out_dict["sold_energy"][trade_seller] = dict(
-                        (m.time_slot, 0) for m in area.past_markets)
-                out_dict["sold_energy"][trade_seller][market.time_slot] += trade.offer.energy
+                    out_dict["sold_energy"][trade_seller] = {}
+                if market.time_slot not in out_dict["sold_energy"][trade_seller]:
+                    out_dict["sold_energy"][trade_seller][market.time_slot] = trade.offer.energy
```
hannesdiedrich: Sorry, AFAIU there will still be an entry for each trade_seller, although the energy is 0. Or am i wrong?

spyrostz: The problem in this was not that the trade.offer.energy was 0 (in corner cases it can be 0, but it happens only for trades with small amounts of energy, which are less than 0.0001 kWh). The original problem was that in the initial implementation an entry with 0 energy was created for every seller AND for every market slot. Therefore there were market slots where this seller was not trading energy, but still there was an entry in this dictionary containing 0.

hannesdiedrich: Ahh, now i see it, sorry. So the trick is to delete LIne 95 and 96. Thanks!

spyrostz: Since you mentioned this though, I will also add another check that does not add elements if the traded energy is below our floating point threshold, in order to not fall to the corner case I described above. Adding it.

PR #421, opened 2019-04-11
Feature/d3 asim 901

hannesdiedrich:


src/d3a/setup/config_parameter_test.py
```diff
@@ -23,7 +23,7 @@
 
 
 def get_setup(config):
-    config.update_config_parameters(iaa_fee=5, market_maker_rate=35)
+    config.update_config_parameters(transfer_fee_pct=5, market_maker_rate=35)
```
faizan2590: Would be nice transfer_fee_const here too as well verify it in this integrationtest https://github.com/gridsingularity/d3a/blob/master/integration_tests/integration_tests.feature#L13

hannesdiedrich: True! Adding it.

PR #424, opened 2019-04-16
D3ASIM-907: Implemented line sweep algorithm for the pay as clear rat…

spyrostz:
…e calculation.

### DIFF
--- a/tests/test_inter_area_agent.py
+++ b/tests/test_inter_area_agent.py
@@ -530,11 +530,14 @@ def iaa_double_sided_pay_as_clear():
 
 @pytest.mark.parametrize("offer, bid, MCP", [
     ([1, 2, 3, 4, 5, 6, 7], [1, 2, 3, 4, 5, 6, 7], 4),
+    ([1, 2, 3, 4, 5, 6, 7], [7, 6, 5, 4, 3, 2, 1], 4),
     ([8, 9, 10, 11, 12, 13, 14], [8, 9, 10, 11, 12, 13, 14], 11),
     ([2, 3, 3, 5, 6, 7, 8], [1, 2, 3, 4, 5, 6, 7], 5),
 ])
+@pytest.mark.parametrize("algorithm", [1, 2])
 def test_iaa_double_sided_performs_pay_as_clear_matching(iaa_double_sided_pay_as_clear,
-                                                         offer, bid, MCP):
+                                                         offer, bid, MCP, algorithm):
+    ConstSettings.IAASettings.PAY_AS_CLEAR_AGGREGATION_ALGORITHM = algorithm
     low_high_engine = \
         next(filter(lambda e: e.name == "Low -> High", iaa_double_sided_pay_as_clear.engines))
     iaa_double_sided_pay_as_clear.lower_market.sorted_offers = \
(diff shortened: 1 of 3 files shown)


tests/test_inter_area_agent.py
```diff
@@ -557,6 +560,25 @@ def test_iaa_double_sided_performs_pay_as_clear_matching(iaa_double_sided_pay_as
...
     matched = low_high_engine._perform_pay_as_clear_matching()[0]
     assert matched == MCP
+    ConstSettings.IAASettings.PAY_AS_CLEAR_AGGREGATION_ALGORITHM = 1
+
+
+def test_iaa_double_sided_pay_as_clear_works_with_floats(iaa_double_sided_pay_as_clear):
+    ConstSettings.IAASettings.PAY_AS_CLEAR_AGGREGATION_ALGORITHM = 1
+    low_high_engine = \
+        next(filter(lambda e: e.name == "Low -> High", iaa_double_sided_pay_as_clear.engines))
+    iaa_double_sided_pay_as_clear.lower_market.sorted_offers = \
```
faizan2590:
```suggestion
    iaa_double_sided_pay_as_clear.lower_market.sorted_offers = \
        [Offer('id1', 1, 1, 'other'),
         Offer('id2', 2, 2, 'other'),
         Offer('id3', 3, 1, 'other')]

    iaa_double_sided_pay_as_clear.lower_market._bids = \
        [Bid('bid_id1', 3.5, 1, 'B', 'S'),
         Bid('bid_id2', 2.5, 1, 'B', 'S'),
         Bid('bid_id3', 1, 2, 'B', 'S')]

    matched = low_high_engine._perform_pay_as_clear_matching()[0]
    assert matched == 2
```
Please have a look into the suggestion, I tried to test with alternate offer/bid price. ClearingRate should be 2, but the test is failing

spyrostz: Clearing rate should be 2.5, have a look at the _clearing_point_from_supply_demand_curve method, this uses the bid rate as the clearing rate, not the offer rate.

faizan2590: I thought the target of alternate implementation is to give the exact clearing_rate with precision. I also tried to draw it on paper, supply/demand curve is intersecting at 2cts/kWh. So, it should return clearing_rate as 2 cts/kWh

spyrostz: Supply and demand curve do not intersect at 2, because we are not talking about linear curves here, but aggregated energy values. Therefore we cannot algorithmically talk about intersections, but only about pair of points that match the aforementioned criteria. In our case, the aggregated supply is greater than the aggregated demand for the pair Offer(2cts/kWh) , Bid (2.5 cts/kWh). This is the optimal set that match the criteria for market clearing. So the market is cleared, with a clearing price that depends on the algorithm we use (again, have a look at McAfee, since this algorithm deals exactly with selecting the optimal price in this range). Therefore there are 3 options here to select the clearing rate: 
- Offer rate 
- Bid rate 
- A rate between offer rate and bid rate (for instance their average)
In the former implementation, the bid rate is selected, and that is why I kept the same approach.  I could have used any of the 3 above, but the algorithm would still remain the same, and again, I would be very much in favor of using a variant of McAfee or a similar algorithm, to be game-theroetical safe. The reason why you see that the clearing rate is 2 on the former implementation is due to the integer relaxation, not due to interpolation.

faizan2590: If you try to plot these discrete offers/bids in terms of merit order. There is an intersection at 2cts/kWh

PR #432, opened 2019-04-24
Feature/d3 asim 1104 Added InfiniteBusStrategy

hannesdiedrich:


src/d3a/setup/two_sided_market/infinite_bus.py
```diff
@@ -0,0 +1,57 @@
...
+            ),
+            Area('Infinite Bus', strategy=InfiniteBusStrategy(energy_buy_rate=24,
+                                                              energy_sell_rate=25),
+                 appliance=SimpleAppliance()
+                 ),
+
+        ],
+        config=config
+    )
+    return area
```
faizan2590: I tested this setup file locally. What I see is Infinite bus is buying from PV and selling to load rather than allowing PV & Load to first trade mutually and then absorb the residual supply/demand. In my opinion, there should be a delay in InfiniteBus to post offers/bid so that it can only absorb residual supply/demand.

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -0,0 +1,74 @@
...
+
+        if ConstSettings.IAASettings.MARKET_TYPE == 1:
+            for market in self.area.all_markets:
+                self.buy_energy(market)
+
+    def event_market_cycle(self):
+        super().event_market_cycle()
+        if ConstSettings.IAASettings.MARKET_TYPE == 2:
+            for market in self.area.all_markets:
+                self.post_bid(market, self.energy_rate[market.time_slot] * INF_ENERGY, INF_ENERGY)
```
faizan2590: InfiniteBusStrategy should also be posting bids for PayAsClearMarket

src/d3a/setup/two_sided_market/infinite_bus.py
hannesdiedrich: You are right. The load should prefer the energy from the PV. But i dont know, whether implementing a delay would be a good solution. It should definitely be accessible for the user. I suggest for now to use the pricing configuration to achieve this. OK

src/d3a/models/strategy/infinite_bus.py
hannesdiedrich: You are right! I was confused by this comment: https://gridsingularity.atlassian.net/browse/D3ASIM-1104?focusedCommentId=13855&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13855 and only allowed the PAB. Will change this. Shall I add a separate integrationtest for that?

faizan2590: Or how about via Scenario Outline to test the same integration for 2-sided PAB & PAC?

hannesdiedrich: @spyrostz @faizan2590 But how can I set the MarketType outside of the config file? You mean by setting it in the ConstSettings?

spyrostz: Yes, you can run the same test 3 times by changing the ConstSettings.IAASettings.MARKET_TYPE parameter to run for different market types.

hannesdiedrich: @spyrostz but how can i change it in the setup file in an integration test?

spyrostz: you can import the ConstSettings, and change the value in a step implementation, after you have read the setup file (this is an example, there are other ways to do that, ask if it is not to your liking). That way the value that you have defined in the setup file will be 'overwritten', since the ConstSettings are global for the simulation.

hannesdiedrich:
I try to adapt to the PR comments. I only could make the behaviour mentioned by @faizan2590 work  in two separate setup files for the integration tests. The test for the PAC market is still due, but @spyrostz promised to deal with it in one of his stories. 
I realised another (unrelated) thing: If the bids/offers start/end with 0 Price, the price of an offer/bid can be negative. I added a check into the corresponding functions. Shall this also be unittested?

src/d3a/models/market/one_sided.py
```diff
@@ -59,7 +59,8 @@ def offer(self, price: float, energy: float, seller: str) -> Offer:
         if energy <= 0:
             raise InvalidOffer()
 
-        price = price * (1 + self.transfer_fee_ratio) + self.transfer_fee_const * energy
+        if price != 0:
```
spyrostz: I recall discussing about it at some point, but what is the reason for that change ? Even if a device is selling energy for free, I would expect that the IAAs would still need to get their fees.

hannesdiedrich: Yeah, but who is paying them? Do we allow negative rates?

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -51,8 +51,8 @@ def __repr__(self):  # pragma: no cover
     def bid(self, price: float, energy: float, buyer: str, seller: str, bid_id: str=None) -> Bid:
         if energy <= 0:
             raise InvalidBid()
-
-        price = price * (1 - self.transfer_fee_ratio) - self.transfer_fee_const * energy
+        if price != 0:
```
spyrostz: In this case, IMO we need to not create the new bid (and potentially warn the user with a warning message). The reason for that is that even if this bid is accepted, the grid fees would not be able to be paid.

src/d3a/models/market/one_sided.py
spyrostz: Ah I see, the new price is going to be 0, because it is relative to the initial price. In that case, I would suggest to leave it here as it is, and not forward the offer at all if its price is 0.

integration_tests/infinite_bus.feature
```diff
@@ -0,0 +1,13 @@
+Feature: Infinite Bus tests
```
faizan2590: Can you add one more integration_test to InfiniteBusStrategy for balancing_market?

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -0,0 +1,75 @@
...
+                continue
+            if (offer.price / offer.energy) <= self.energy_buy_rate[market.time_slot]:
+                try:
+                    self.accept_offer(market, offer)
+                except MarketException:
+                    # Offer already gone etc., try next one.
+                    continue
+
+    def event_tick(self, *, area):
+        super().event_tick(area=area)
```
faizan2590: Can you explain me this line. I couldn't find event_tick method being implemented in CommercialStrategy or BidEnabledStrategy

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -0,0 +1,75 @@
...
+        self.energy_per_slot_kWh = INF_ENERGY
+        energy_buy_rate = ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE \
+            if energy_buy_rate is None else energy_buy_rate
+        energy_sell_rate = ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE \
+            if energy_sell_rate is None else energy_sell_rate
+        self.energy_rate = read_arbitrary_profile(InputProfileTypes.IDENTITY, energy_sell_rate)
+        self.energy_buy_rate = read_arbitrary_profile(InputProfileTypes.IDENTITY, energy_buy_rate)
+
+    def offer_energy(self, market):
+        energy_rate = self.energy_rate[market.time_slot]
```
spyrostz: It is a bit of an overkill, but it might improve code reuse, therefore I won't hesitate to put it onto the table: instead of overriding the offer_energy CommercialProducer method with a slightly modified version of it here, how about refactoring the CommercialProducer class to parse the self.energy_rate as a profile (like you already do it in the constructor of the InfiniteBusStrategy). That way the offer_energy method can be deleted, and as an added bonus, parts of the InfiniteBusStrategy constructor will be handled automatically by the base class constructor (CommercialStrategy)

src/d3a/models/market/one_sided.py
hannesdiedrich: Sorry, can you please be more explicit? Shall this function not return an offer if the price is 0? Because offer() is not only used in the forward_offer method. 
I could add 
```
        if price <= 0:
            raise InvalidOffer()
```
What do you think?

src/d3a/models/market/two_sided_pay_as_bid.py
hannesdiedrich: Yes you are right! How about something like this:
```
If price < self.transfer_fee_const * energy :
     raise InvalidBid()
```
... or some other Expection that tells the user to set the initial_buy_rate higher?

src/d3a/models/market/one_sided.py
spyrostz: What I meant, is to add a separate check in the forward_offer, that will not forward the offer if the target market's percentage fee is greater than zero and the offer price is zero or less. 
Raising an exception is not the desired behavior IMO, because this would crash the simulation just because an offer could not be forwarded.

integration_tests/infinite_bus.feature
hannesdiedrich: Can you please specify what needs to be tested?

src/d3a/models/market/two_sided_pay_as_bid.py
spyrostz: I agree with the exception, I do not agree though on blowing it up to the user, because this is a perfectly valid use case to my eyes. I would instead try/catch for this exception in the forward_bid method, and in case the exception is raised the forwarding should be cancelled (potentially with a warning message).

src/d3a/models/strategy/infinite_bus.py
hannesdiedrich: You are right! This super() call is not needed. Removing it.

src/d3a/models/market/one_sided.py
hannesdiedrich: Understood. However, as mentioned, this method is not only called in the forward_offer method. So, even a storage can post an (initial) offer with price=0. Shall I add tests for that before each call? Sorry if i missed something.

integration_tests/infinite_bus.feature
faizan2590: Great, then there is not need for an integration tests, IMO.

src/d3a/models/strategy/area_agents/one_sided_engine.py
```diff
@@ -220,14 +225,14 @@ def event_offer_changed(self, *, market_id, existing_offer, new_offer):
...
             forwarded = self._forward_offer(new_offer, new_offer.id)
-
-            self.owner.log.info("Offer %s changed to residual offer %s",
-                                offer_info.target_offer,
-                                forwarded)
-
-            # Do not delete the forwarded offer entries for the case of residual offers
-            if existing_offer.seller != new_offer.seller:
-                self._delete_forwarded_offer_entries(offer_info.source_offer)
+            if forwarded:
```
spyrostz: In order to avoid nesting, you can early-return here by using : 
if not forwarded: return

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -40,6 +40,11 @@ def _forward_bid(self, bid):
             return
         if self.owner.name == self.markets.target.area.name:
             return
+        if bid.price < (self.markets.target.transfer_fee_const * bid.energy):
```
spyrostz: You need to also take into account the transfer_fee_pct here as well. Keep in mind that the end result in this check is to not have a negative price when forwarding the bid.

hannesdiedrich: Yes, you are right! Will change this

src/d3a/models/strategy/area_agents/one_sided_engine.py
hannesdiedrich: ok, will do!

PR #434, opened 2019-04-26
D3ASIM-1098: Adapted the energy_trade_profile endpoint.

hannesdiedrich:


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -84,7 +84,7 @@ def generate_json_report(self):
             "tree_summary": self.tree_summary,
             "status": self.status,
             "device_statistics": self.device_statistics_time_str_dict,
-            "energy_trade_profile": self.energy_trade_profile
+            "energy_trade_profile": self.energy_trade_profile_redis
```
spyrostz: Do we really want to expose the redis endpoint for the energy_trade_profile ?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -184,9 +184,10 @@ def _round_energy_trade_profile(cls, profile):
         for k in profile.keys():
             for sold_bought in ['sold_energy', 'bought_energy']:
                 for dev in profile[k][sold_bought].keys():
-                    for timestamp in profile[k][sold_bought][dev].keys():
-                        profile[k][sold_bought][dev][timestamp] = \
-                            round_floats_for_ui(profile[k][sold_bought][dev][timestamp])
+                    for target in profile[k][sold_bought][dev].keys():
+                        for timestamp in profile[k][sold_bought][dev][target].keys():
+                            profile[k][sold_bought][dev][target][timestamp] = round_floats_for_ui(
```
spyrostz: You can leave it as it is, but a JSON file with a 5-level nesting gives me the shivers. On the other hand if that's what the UI wants, they can have it.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: No, this is a test remnant. Will revert it. Thanks!

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: Yes, I understand. But I couldnt think of a different way. 
But when I think of it now, we could omit the "bought_energy" and "sold_energy" level and use negative values for showing sold energy. Why do you think @spyrostz and @faizan2590 ?

spyrostz: @hannesdiedrich I would suggest to be more explicit there, because it depends on the context of the application if the + price is referring to bought or to sold energy. 
Generally speaking it is ok for now, my point is that it would be much easier if the frontend did some calculations on its own and we did not have to send such a complex JSON file (for instance, if the UI was able to filter timestamps and check the energy from target devices on its own, we would not need such a complicated nesting). We can maybe keep that in mind for any potential refactoring in the future.

hannesdiedrich: OK, thanks!

PR #435, opened 2019-04-30
Bug/d3 asim 1109

faizan2590:


tests/test_inter_area_agent.py
```diff
@@ -120,16 +120,14 @@ def delete_offer(self, *args):
...
-            price = price * (1 + self.transfer_fee_ratio) + self.transfer_fee_const * energy
+        price = price * (1 + self.transfer_fee_ratio) + self.transfer_fee_const * energy
         self.forwarded_offer = Offer(self.forwarded_offer_id, price, energy, seller, market=self)
         return self.forwarded_offer
 
-    def bid(self, price, energy, buyer, seller, iaa_fee: bool = False):
-        if iaa_fee:
-            price = price * (1 - self.transfer_fee_ratio) + self.transfer_fee_const * energy
+    def bid(self, price, energy, buyer, seller):
+        price = price * (1 - self.transfer_fee_ratio) + self.transfer_fee_const * energy
```
spyrostz:
```suggestion
        price = price * (1 - self.transfer_fee_ratio) - self.transfer_fee_const * energy
```

tests/test_market.py
```diff
@@ -271,8 +280,10 @@ def test_market_trade_bid_not_found(market: TwoSidedPayAsBid):
 
 
 @pytest.mark.parametrize("market, offer, accept_offer", [
-    (OneSidedMarket(area=FakeArea("FakeArea")), "offer", "accept_offer"),
-    (BalancingMarket(area=FakeArea("FakeArea")), "balancing_offer", "accept_offer")
+    (OneSidedMarket(area=FakeArea("FakeArea", transfer_fee_pct=0)),
```
spyrostz: How about moving setting the transfer_fee_pct=0 directly on the FakeArea class? That way you would not have to make so many changes all over this file.

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -48,12 +48,11 @@ def __repr__(self):  # pragma: no cover
...
 
-    def bid(self, price: float, energy: float, buyer: str, seller: str, bid_id: str=None,
-            iaa_fee: bool = False) -> Bid:
+    def bid(self, price: float, energy: float, buyer: str, seller: str, bid_id: str=None) -> Bid:
         if energy <= 0:
             raise InvalidBid()
-        if iaa_fee:
-            price = price * (1 - self.transfer_fee_ratio) + self.transfer_fee_const * energy
+
+        price = price * (1 - self.transfer_fee_ratio) - self.transfer_fee_const * energy
```
hannesdiedrich: This change is correct and should have been implemented in my PR when adding the transfer_fee_const. 
IMO you should also change the adding of the constant grid fee in https://github.com/gridsingularity/d3a/blob/b8dc079dda704781b88e6d5a9a1b14ba72b17e7e/src/d3a/models/market/two_sided_pay_as_bid.py#L90 . Right?

hannesdiedrich:
A general comment: Is the default value of ConstSetting.IAASettings.FEE_PERCENTAGE still useful under the "new" way of thinking? Shouldnt it be 0 by default and the user should change it? @spyrostz @faizan2590

tests/test_market.py
faizan2590: I guess its would be better to keep it like this. In that way I can also test if grid fee is taken into account while immediately placing offers/bids. Like here: https://github.com/gridsingularity/d3a/pull/435/files#diff-9ad3b9e554e26bce69c8a134921b16f2R98

tests/test_inter_area_agent.py
faizan2590: The motivation to me for this change is if transfer_fee_ratio is reducing the bid_price, so should transfer_fee_const. Let me know if it makes sense now?

PR #438, opened 2019-05-03
D3ASIM-1110: Replaced grid fee calculation to take into account initi…

spyrostz:
…al offer/bid/trade price.

### DIFF
--- a/integration_tests/commercial_power_plant_tests.feature
+++ b/integration_tests/commercial_power_plant_tests.feature
@@ -18,6 +18,6 @@ Feature: Commercial Power Plant tests
      Given we have a scenario named strategy_tests/commercial_producer_market_maker_rate
      And d3a is installed
      And we have a profile of market_maker_rate for strategy_tests.commercial_producer_market_maker_rate
-     When we run the d3a simulation with config parameters [1, 1] and strategy_tests.commercial_producer_market_maker_rate
-     Then we test that config parameters are correctly parsed for strategy_tests.commercial_producer_market_maker_rate [1, 1]
+     When we run the d3a simulation with config parameters [1, 0] and strategy_tests.commercial_producer_market_maker_rate
+     Then we test that config parameters are correctly parsed for strategy_tests.commercial_producer_market_maker_rate [1, 0]
      And the Commercial Energy Producer always sells energy at the defined market maker rate
--- a/src/d3a/models/market/balancing.py
+++ b/src/d3a/models/market/balancing.py
@@ -26,7 +26,7 @@ from d3a.events.event_structures import MarketEvent
 from d3a.models.market.market_structures import BalancingOffer, BalancingTrade
 from d3a.d3a_core.exceptions import InvalidOffer, MarketReadOnlyException, \
     OfferNotFoundException, InvalidBalancingTradeException, \
-    DeviceNotInRegistryError, ChainTradeException
+    DeviceNotInRegistryError
 from d3a.d3a_core.device_registry import DeviceRegistry
 
 log = getLogger(__name__)
@@ -72,7 +72,7 @@ class BalancingMarket(OneSidedMarket):
                      energy: int = None, time: DateTime = None, already_tracked: bool = False,
                      trade_rate: float = None, iaa_fee: bool = False) -> BalancingTrade:
         if iaa_fee and trade_rate is None:
-            raise ChainTradeException()
+            raise Exception()
         if self.readonly:
             raise MarketReadOnlyException()
         if isinstance(offer_or_id, Offer):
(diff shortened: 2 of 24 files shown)


src/d3a/models/market/one_sided.py
```diff
@@ -90,23 +94,34 @@ def delete_offer(self, offer_or_id: Union[str, Offer]):
         # TODO: Once we add event-driven blockchain, this should be asynchronous
         self._notify_listeners(MarketEvent.OFFER_DELETED, offer=offer)
 
+    def _update_fee_and_calculate_final_price(self, energy, trade_rate,
+                                              energy_portion, original_price):
+        fees = self.transfer_fee_ratio * original_price * energy_portion
+        fees += self.transfer_fee_const * energy
+        self._grid_fee += fees
+        return energy * trade_rate + fees
```
hannesdiedrich: I like the level of verbosity of this function for understanding the code. However, it will be called quite a lot during a simulation, right? Wouldn't it be better ti reduce the number of calculations (lines) to reduce computational effort? Or doesn't it make a big difference?

src/d3a/models/market/one_sided.py
```diff
@@ -119,13 +119,13 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *, energy: in
...
             trade_rate = offer.price / offer.energy
 
-        energy_portion = energy / offer.energy
         self._sorted_offers = sorted(self.offers.values(),
                                      key=lambda o: o.price / o.energy)
         try:
             if time is None:
                 time = self._now
             if energy is not None:
+                energy_portion = energy / offer.energy
```
hannesdiedrich: This is just refactoring, right? The error was due to the deleted ChainTradeException right?

spyrostz: Actually this was the reason why the integration tests were failing just before this PR. It was introduced during a last careless refactoring before I opened this PR, which introduced moving this calculation outside of 'energy is not None' block. This had the effect that at some points, energy was indeed None, which crashed the whole simulation. That's why I fixed that and moved it back inside the if block.

hannesdiedrich: Ah, ok! Thanks!

src/d3a/models/market/one_sided.py
spyrostz: Hmm, I really do not know how the Python interpreter optimises arithmetic operations. If we were talking about compilers, gcc or clang automatically optimise and merge these 3 operations to one (by producing assembly code that sends these operations at the CPU on the same cycle) but I do not really know how the interpreter deals with this. Just to be on the safe side, I will follow your suggestion and calculate the fees variable on one line. Thanks!

integration_tests/commercial_power_plant_tests.feature
```diff
@@ -18,6 +18,6 @@ Feature: Commercial Power Plant tests
      Given we have a scenario named strategy_tests/commercial_producer_market_maker_rate
      And d3a is installed
      And we have a profile of market_maker_rate for strategy_tests.commercial_producer_market_maker_rate
-     When we run the d3a simulation with config parameters [1, 1] and strategy_tests.commercial_producer_market_maker_rate
-     Then we test that config parameters are correctly parsed for strategy_tests.commercial_producer_market_maker_rate [1, 1]
+     When we run the d3a simulation with config parameters [1, 0] and strategy_tests.commercial_producer_market_maker_rate
```
faizan2590: Any specific reason to set the grid fee as zero?

spyrostz: Yes, because due to the fact that the grid fee was 1, the grid fee was taken into account when trading, which resulted in an increased energy rate for the commercial producer (keep in mind that the purpose here is to test whether the market maker rate is followed, not the grid fee).

src/d3a/models/market/balancing.py
```diff
@@ -26,7 +26,7 @@
 from d3a.models.market.market_structures import BalancingOffer, BalancingTrade
 from d3a.d3a_core.exceptions import InvalidOffer, MarketReadOnlyException, \
     OfferNotFoundException, InvalidBalancingTradeException, \
-    DeviceNotInRegistryError, ChainTradeException
+    DeviceNotInRegistryError
```
faizan2590: https://github.com/gridsingularity/d3a/pull/438/files?file-filters%5B%5D=.py#diff-49ce5f7749215295a7b29d8030567663R49

I guess the balancing_offer should also consider the original rate

spyrostz: Thanks, I totally forgot about the balancing market. I will adapt it to match the one sided market case.

PR #443, opened 2019-05-07
Enabling forwarded offers to use source market's grid fee in order to…

faizan2590:
… avoid double spending while matching bid with offer

### DIFF
--- a/tests/test_area.py
+++ b/tests/test_area.py
@@ -61,16 +61,6 @@ class TestAreaClass(unittest.TestCase):
     def tearDown(self):
         GlobalConfig.market_count = 1
 
-    def test_respective_area_grid_fee_is_applied(self):
-        self.area = Area(name="Street", children=[Area(name="House")],
-                         config=GlobalConfig, transfer_fee_pct=5)
-        self.area.parent = Area(name="GRID")
-        self.area.config.market_count = 1
-        self.area.activate()
-        assert self.area.next_market.transfer_fee_ratio == 0.05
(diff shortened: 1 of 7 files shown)


tests/test_inter_area_agent.py
```diff
@@ -120,9 +120,11 @@ def delete_offer(self, *args):
     def delete_bid(self, *args):
         pass
 
-    def offer(self, price, energy, seller):
+    def offer(self, price, energy, seller, market=None):
         self.offer_count += 1
-        price = price * (1 + self.transfer_fee_ratio) + self.transfer_fee_const * energy
+        if market is not None:
+            price = \
+                price * (1 + self.transfer_fee_ratio) + self.transfer_fee_const * energy
```
hannesdiedrich:
```suggestion
                price * (1 + market.transfer_fee_ratio) + market.transfer_fee_const * energy
```

tests/test_area.py
```diff
@@ -61,16 +61,6 @@ def setUp(self):
     def tearDown(self):
         GlobalConfig.market_count = 1
 
-    def test_respective_area_grid_fee_is_applied(self):
```
hannesdiedrich: I guess this test has to be reimplemented in the test_market, right?

faizan2590: https://github.com/gridsingularity/d3a/pull/443/files#diff-f46f8a6e9b1f4757a0e767f3306499d2R154
I guess, its already taken care of in tests/test_inter_area_agent.py

PR #444, opened 2019-05-09
Bug/d3 asim 1138

hannesdiedrich:


src/d3a/models/strategy/storage.py
```diff
@@ -167,6 +171,16 @@ def _validate_break_even_points(break_even):
         if any(break_even_point.buy < 0 or break_even_point.sell < 0
                for _, break_even_point in break_even.items()):
             raise ValueError("Break even point should be positive energy rate values.")
+        if any(break_even_point.sell > initial_selling_rate
+               for break_even_point in break_even.values()):
+            raise ValueError(f"Break even point (selling) should be lower equal "
```
spyrostz: Trivial, but you could add at least the initial selling rate value in the exception text. That way the user will be able to see its value. Same for the initial buying rate exception below.

hannesdiedrich: I was thinking about that but got distracted. Will add it. Thanks!

src/d3a/models/strategy/storage.py
```diff
@@ -173,13 +173,13 @@ def _validate_break_even_points(break_even,
             raise ValueError("Break even point should be positive energy rate values.")
         if any(break_even_point.sell > initial_selling_rate
                for break_even_point in break_even.values()):
-            raise ValueError(f"Break even point (selling) should be lower equal "
-                             f"than initial_buying_rate")
+            raise ValueError(f"Break even point(s) (selling) should be lower equal "
```
faizan2590: Very Minor: Break even point(s) (selling) should be lower OR equal

hannesdiedrich: Sorry, I dont get it. Do you want an `or` between lower and equal?
If so, shall I change it for the selling_rate error message as well?

faizan2590: A ValueError is raised when break_even_point.sell is greater than initial_selling_rate. It means break_even_point.sell should either be less than or equal to initial_selling_rate.

hannesdiedrich: This is what `lower equal` means right?

faizan2590: IMO, it should be "lower or equal". Im not a linguist, so may be I'm wrong.

hannesdiedrich: Ok, I just tried to google it. It is indeed not used (apart from bash scripts, where i got it probably from ) , soo, changing it to `lower or equal`

PR #446, opened 2019-05-10
Feature/d3 asim 1127

spyrostz:


src/d3a/models/market/one_sided.py
```diff
@@ -90,92 +94,115 @@ def delete_offer(self, offer_or_id: Union[str, Offer]):
...
+            if energy == 0:
+                raise InvalidTrade("Energy can not be zero.")
+            # partial energy is requested
+            elif energy < offer.energy:
+                original_offer = offer
+                accepted_offer_id = offer.id \
+                    if self.area is None or self.area.bc is None \
+                    else offer.real_id
+
+                assert trade_rate + FLOATING_POINT_TOLERANCE >= (offer.price / offer.energy)
```
hannesdiedrich: I dont understand this assert at this point of the code. Is there any reason for not putting it just after  L 137?

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -69,11 +75,16 @@ def delete_bid(self, bid_or_id: Union[str, Bid]):
         log.info(f"[BID][DEL][{self.time_slot_str}] {bid}")
         self._notify_listeners(MarketEvent.BID_DELETED, bid=bid)
 
+    def _update_fee_and_calculate_final_price(self, energy, trade_rate,
```
hannesdiedrich: How about being consistent with the `_update_offer_fee_and_calculate_final_price` in the one-sided-market and renaming it to `_update_bid_fee_and_calculate_final_price`

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -83,49 +94,63 @@ def accept_bid(self, bid: Bid, energy: float = None,
...
-            self.market_fee += (trade_rate - source_rate) * energy
-        else:
-            source_rate = trade_rate
+
+        orig_price = bid.original_bid_price if bid.original_bid_price is not None else bid.price
+        residual = None
 
         if energy <= 0:
-            raise InvalidTrade("Energy cannot be zero.")
+            raise InvalidTrade("Energy cannot be less than zero.")
```
hannesdiedrich: `less than` --> `lower equal`  right?

src/d3a/models/market/one_sided.py
spyrostz: You are correct, it is better to place the assignment and validation close. I will move these probably to a separate function too, in order to reduce a bit the size of the accept_offer method.

src/d3a/models/market/two_sided_pay_as_bid.py
spyrostz: Agreed. Renaming, thanks

src/d3a/models/market/two_sided_pay_as_bid.py
spyrostz: Thanks, changing that.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -225,11 +227,13 @@ def reset_price_on_market_cycle(self, market):
                 continue
             try:
                 iterated_market.delete_offer(offer.id)
-
+                updated_price = offer.energy * \
+                    self.calculate_initial_sell_rate(iterated_market.time_slot)
```
hannesdiedrich: How about rounding here as well? Like in `_decrease_offer_price`?

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -97,17 +99,24 @@ def _delete_forwarded_bid_entries(self, bid):
     def _match_offers_bids(self):
         for bid, offer in self._perform_pay_as_bid_matching():
             selected_energy = bid.energy if bid.energy < offer.energy else offer.energy
+            original_bid_rate = bid.original_bid_price / bid.energy
+            clearing_rate = bid.price / bid.energy
```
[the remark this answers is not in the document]
spyrostz: Matched_rate sounds good to me, replacing it .

src/d3a/models/strategy/update_frequency.py
spyrostz: You are right, this should be rounded too, fixing it.

integration_tests/steps/integration_tests.py
```diff
@@ -792,3 +802,22 @@ def test_config_parameters(context):
     assert grid.config.iaa_fee_const == 1
     assert all([rate == 35
                 for rate in grid.config.market_maker_rate.values()])
+
+
+@then('trades on the {market_name} market clear with {clearing_rate} cents/kWh')
+def trades_clear_price(context, market_name, clearing_rate):
```
hannesdiedrich: Here again, `clear_price` or `clearing_rate` does not match with the test for the one-sided and pab market structure. How about using  `trade_rate` ? Again, minor! Thanks

spyrostz: Agreed, fixing it.

PR #447, opened 2019-05-13
D3ASIM-1129: Added real-time support for running the simulation.

spyrostz:
Added a sleep which synchronises the simulation runtime with the real time.

### DIFF
--- a/src/d3a/d3a_core/simulation.py
+++ b/src/d3a/d3a_core/simulation.py
@@ -281,6 +281,9 @@ class Simulation:
                     if console is not None:
                         self._handle_input(console, diff_slowdown)
 
+                if ConstSettings.GeneralSettings.RUN_REAL_TIME:
+                    sleep(tick_lengths_s - tick_length)
+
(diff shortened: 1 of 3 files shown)


src/d3a/d3a_core/simulation.py
```diff
@@ -281,6 +281,9 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
                     if console is not None:
                         self._handle_input(console, diff_slowdown)
 
+                if ConstSettings.GeneralSettings.RUN_REAL_TIME:
+                    sleep(tick_lengths_s - tick_length)
```
hannesdiedrich: took me a wehil to understand this line because of the naming of `tick_length` AFAIU this is the realtime time difference between the start and the end of the tick. How about renaming it to `realtime_tick_length`  or `real_tick_length` ?

spyrostz: Thanks, renaming it

PR #449, opened 2019-05-15
D3ASIM-1145: Added support for calculating unmatched loads incremen…

spyrostz:
…tally, by using only the current market (most recent past market).

### DIFF
--- a/integration_tests/steps/integration_tests.py
+++ b/integration_tests/steps/integration_tests.py
@@ -22,6 +22,8 @@ import glob
 from math import isclose
 from pendulum import duration, today, from_format
 from behave import given, when, then
+from deepdiff import DeepDiff
+from copy import deepcopy
 
 from d3a.models.config import SimulationConfig
 from d3a.models.read_user_profile import read_arbitrary_profile, InputProfileTypes
@@ -213,6 +215,11 @@ def two_sided_pay_as_bid_market(context):
     ConstSettings.IAASettings.MARKET_TYPE = 2
 
 
+@given('the past markets are kept in memory')
+def past_markets_in_memory(context):
+    ConstSettings.GeneralSettings.KEEP_PAST_MARKETS = True
+
+
 @when('the simulation is running')
 def running_the_simulation(context):
 
@@ -304,6 +311,18 @@ def run_d3a_with_settings_file(context):
               "--setup default_2a".format(export_path=context.export_path,
                                           settings_file=os.path.join(d3a_path, "setup",
                                                                      "d3a-settings.json")))
+
+
+@when('the reported unmatched loads are saved')
+def save_reported_unmatched_loads(context):
+    context.unmatched_loads = deepcopy(context.simulation.endpoint_buffer.unmatched_loads)
+    context.unmatched_loads_redis = \
+        deepcopy(context.simulation.endpoint_buffer.unmatched_loads_redis)
+
+
+@when('the past markets are not kept in memory')
+def past_markets_not_in_memory(context):
+    ConstSettings.GeneralSettings.KEEP_PAST_MARKETS = False
 
 
 @then('we test the export functionality of {scenario}')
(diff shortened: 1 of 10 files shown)


src/d3a/models/area/__init__.py
```diff
@@ -293,7 +293,7 @@ def next_market(self):
         except IndexError:
             return None
 
-    @cached_property
+    @property
```
hannesdiedrich: There are a number of places in the code where @cached_property is used. Why changing it (only) here?

integration_tests/steps/jira_issues_tests.py
```diff
@@ -141,6 +141,7 @@ def pv_produces_same_amount_of_energy_day(context):
                                         for trade in market.trades
                                         if trade.seller == "H2 PV")
                                     for market in same_time_markets]
+        print(same_time_markets_energy)
```
hannesdiedrich: Please remove. Thanks.

integration_tests/steps/integration_tests.py
```diff
@@ -821,3 +840,17 @@ def assert_trade_rates(context, market_name, trade_rate):
...
+
+@then('the unmatched loads are identical no matter if the past markets are kept')
+def identical_unmatched_loads(context):
+    unmatched_loads = context.simulation.endpoint_buffer.unmatched_loads
+    unmatched_loads_redis = context.simulation.endpoint_buffer.unmatched_loads_redis
+
+    assert len(DeepDiff(unmatched_loads, context.unmatched_loads)) == 0
+    # The 2 simulation runs do not assign the same uuids to the same areas
+    # therefore we have to search whether there are elements with the same values
+
```
hannesdiedrich: Please move the blank line to the top of this comment (was confused where to connect this comment to) Thanks!

src/d3a/models/area/__init__.py
spyrostz: Thanks for pointing it out, I just saw that it was used in other places in the same file, and it probably needs to be removed from there too. The problem was that, due to caching, the current_market property did not always return the last past market, but one-before-the-last at some points. I suspect that the same can happen for the next_market for instance, that's why I think we need to removed the cached_property from all the properties of this file.

integration_tests/steps/jira_issues_tests.py
spyrostz: Thanks, removing

integration_tests/steps/integration_tests.py
spyrostz: Will do, thanks!

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
```diff
@@ -194,3 +212,81 @@ def arrange_output(self, indict, area):
...
+                            if ts not in target_ul[timestamp]['unmatched_times'][device]:
+                                target_ul[timestamp]['unmatched_times'][device].append(ts)
+
+                unm_count = 0
+                for _, hours in target_ul[timestamp]['unmatched_times'].items():
+                    unm_count += len(hours)
+
+                    target_ul['unmatched_count'] = unm_count
+
+    def _merge_accumulated_unmatched_loads(self, accumulated_results, current_results, area):
```
hannesdiedrich: Just minor, but doesnt this function only copies the value from the current to the accumulated_results? 
The name merge confused me a bit. Again, just personal taste. If you dont agree, just leave it.

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
```diff
@@ -75,20 +85,28 @@ def find_unmatched_loads(self, area, indict):
             self.name_uuid_map[child.name] = child.uuid
             self.name_type_map[child.name] = child.display_type
             if child.children:
-                indict[area.name] = self.find_unmatched_loads(child, indict[area.name])
+                indict[area.name] = self.find_unmatched_loads(
+                    child, indict[area.name], all_past_markets
+                )
             else:
                 if isinstance(child.strategy, LoadHoursStrategy):
+                    curr_market = [child.parent.current_market] \
```
hannesdiedrich: Sorry for nitpicking, IMO `current_market` would be a better name.

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
spyrostz: Agreedm changing it to copy_accumulated_unmatched_loads

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
spyrostz: agreed, changing that

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
hannesdiedrich: I tried to leave at least very short doc strings, about what the individual functions do in the ExportUnmatchedLoads class. In my experience these can help later when understanding the code, especially if it is for these recursive structures. I suggest to do it also for the MarketUnmatchedLoads. Thanks a lot!

spyrostz: ok, in that case I will add some docstrings for this class too, thanks!

PR #452, opened 2019-05-17
D3ASIM-1110: Adapted two sided pay as clear engine to the new grid fe…

spyrostz:
…e calculation model.

### DIFF
--- a/src/d3a/d3a_core/simulation.py
+++ b/src/d3a/d3a_core/simulation.py
@@ -138,14 +138,12 @@ class Simulation:
         self.pause_after = pause_after
         self.slowdown = slowdown
 
-        # TODO: An issue was experienced when comparing random behaviour of the simulation between
-        # cli-call and behave integration tests
-        # Furthermore, random.seed() is also called before this following block by setuptools
         if seed is not None:
             random.seed(int(seed))
         else:
             random_seed = random.randint(0, 1000000)
             random.seed(random_seed)
+            self.initial_params["seed"] = random_seed
             log.error("Random seed: {}".format(random_seed))
 
         self.area = self.setup_module.get_setup(self.simulation_config)
--- a/src/d3a/d3a_core/simulation.py
+++ b/src/d3a/d3a_core/simulation.py
@@ -142,7 +142,7 @@ class Simulation:
         # cli-call and behave integration tests
         # Furthermore, random.seed() is also called before this following block by setuptools
         if seed is not None:
-            random.seed(seed)
+            random.seed(int(seed))
         else:
             random_seed = random.randint(0, 1000000)
             random.seed(random_seed)
--- a/src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
+++ b/src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
@@ -21,7 +21,7 @@ from d3a.models.const import ConstSettings
 import math
 from collections import OrderedDict
 from logging import getLogger
-
+from d3a.constants import FLOATING_POINT_TOLERANCE
 
 BidInfo = namedtuple('BidInfo', ('source_bid', 'target_bid'))
 
@@ -62,15 +62,15 @@ class TwoSidedPayAsClearEngine(TwoSidedPayAsBidEngine):
 
     def _smooth_discrete_point_curve(self, obj, limit, asc_order=True):
         if asc_order:
-            for i in range(limit+1):
-                obj[i] = obj.get(i, 0) + obj.get(i-1, 0)
+            for i in range(limit + 1):
+                obj[i] = obj.get(i, 0) + obj.get(i - 1, 0)
         else:
             for i in range((limit), 0, -1):
-                obj[i] = obj.get(i, 0) + obj.get(i+1, 0)
+                obj[i] = obj.get(i, 0) + obj.get(i + 1, 0)
         return obj
 
     def _get_clearing_point(self, max_rate):
-        for i in range(1, max_rate+1):
+        for i in range(1, max_rate + 1):
             if self.markets.source.state.cumulative_offers[self.owner.owner.now][i] >= \
                     self.markets.source.state.cumulative_bids[self.owner.owner.now][i]:
                 return i, self.markets.source.state.cumulative_bids[self.owner.owner.now][i]
@@ -132,15 +132,15 @@ class TwoSidedPayAsClearEngine(TwoSidedPayAsBidEngine):
             self.markets.source.state.clearing[time] = (clearing_rate, clearing_energy)
 
         cumulative_traded_bids = 0
+        accepted_bids = []
         for bid in self.sorted_bids:
-            bid_buyer_name = bid.buyer
             original_bid_rate = bid.original_bid_price / bid.energy
             if cumulative_traded_bids >= clearing_energy:
                 break
-            elif (bid.price/bid.energy) >= clearing_rate and \
+            elif (bid.price / bid.energy) >= clearing_rate and \
                     (clearing_energy - cumulative_traded_bids) >= bid.energy:
                 cumulative_traded_bids += bid.energy
-                self.markets.source.accept_bid(
+                trade = self.markets.source.accept_bid(
                     bid=bid,
                     energy=bid.energy,
                     seller=self.owner.name,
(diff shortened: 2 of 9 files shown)


faizan2590:
A general comment, would be better to use the same ticket id to commit messages and branch name.

src/d3a/d3a_core/simulation.py
```diff
@@ -142,7 +142,7 @@ def _init(self, slowdown, seed, paused, pause_after):
         # cli-call and behave integration tests
         # Furthermore, random.seed() is also called before this following block by setuptools
         if seed is not None:
-            random.seed(seed)
+            random.seed(int(seed))
```
hannesdiedrich: When I was testing the export in another branch, I saw that the seed is not reported, if not set explicitly.
So, can you please add an update of the `self.initial_params.seed` in order for the d3a to also export the seed, if it is determined by the simulation itself (else block)? Thanks a lot!

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -165,12 +169,17 @@ def _match_offers_bids(self):
                 break
             elif (math.floor(offer.price/offer.energy)) <= clearing_rate and \
                     (clearing_energy - cumulative_traded_offers) >= offer.energy:
+                # TODO: Used the clearing_rate as the original_trade_rate for the offers, because
```
hannesdiedrich: I think I need a little bit of background of the difference between aggregated market and p2p market. I couldnt find a good explanation from uncle google. If you have a short explanation, it would be great, but i am also happy to discuss offline.

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -127,12 +127,14 @@ def _match_offers_bids(self):
         clearing_rate, clearing_energy = clearing
         if clearing_energy > 0:
             self.owner.log.info(f"Market Clearing Rate: {clearing_rate} "
-                                f"||| Clearing Energy: {clearing_energy} ")
+                                f"||| Clearing Energy: {clearing_energy} "
+                                f"||| Market: {self.markets.source.area.name}")
```
faizan2590: I guess, adding market's area name is redundant as _self.owner.log.info()_ already shows the area description at which the clearing is done.

faizan2590:
Can you add an integration test to test if grid fee are also calculated as expected for 2-sided-pay-as-clear market?

spyrostz:
> Can you add an integration test to test if grid fee are also calculated as expected for 2-sided-pay-as-clear market?

@faizan2590 thanks for reminding me I totally forgot about that. Adding one.

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
spyrostz: You are right, I will remove this, thanks

src/d3a/d3a_core/simulation.py
spyrostz: I'll add it, thanks!

spyrostz: @hannesdiedrich  btw, is the comment above still relevant? If it is not let me know to remove it, thanks!

hannesdiedrich: @spyrostz Not that I know of. This is probably solved by the changes you propose here. But I am not sure. IMO this comment does not help a lot. It just tells us that there might be an issue (experienced 1 year ago or so) but does not tell exactly what. So i propose to delete it.

spyrostz: Thanks, I also think we should delete that since it is not clear enough, removing it

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
spyrostz: On second thought this is not redundant, because these are 2 separate markets: the one that is reported on the info log output is the area name of the agent, whilst the market the log is referring to here is the market in which the clearing is performed. These 2 could be the same, but (depending on the order which the random function will call the tick method across IAA engines) 50% of the time it is not. Therefore since I find it useful to see in this log in which market did the clearing happen, I am voting to keep this log.

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
spyrostz: It is a short explanation but I think it is better to discuss offline to clear any misunderstandings

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -127,63 +127,115 @@ def _match_offers_bids(self):
...
                 break
-            elif (math.floor(offer.price/offer.energy)) <= clearing_rate and \
+            elif (math.floor(offer.price / offer.energy)) <= clearing_rate and \
                     (clearing_energy - cumulative_traded_offers) >= offer.energy:
+                # TODO: Used the clearing_rate as the original_trade_rate for the offers, because
+                # currently an aggregated market is used. If/once a peer-to-peer market is
+                # implemented, we should use the original bid rate for calculating the fees
+                # on the source offers, similar to the two sided pay as bid market.
+
+                # energy == None means to use the bid energy instead of this
```
hannesdiedrich: I propose to replace `this` by `offer.energy`

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -127,63 +127,115 @@ def _match_offers_bids(self):
...
                 break
-            elif (bid.price/bid.energy) >= clearing_rate and \
+            elif (bid.price / bid.energy) >= clearing_rate and \
                     (clearing_energy - cumulative_traded_bids) >= bid.energy:
                 cumulative_traded_bids += bid.energy
-                self.markets.source.accept_bid(
+                trade = self.markets.source.accept_bid(
                     bid=bid,
                     energy=bid.energy,
                     seller=self.owner.name,
```
faizan2590: Sorry if I misunderstood. But isn't it that the bid's seller are still the clearing_engine owner's name. Meaning that bids are accepted in an aggregated fashion rather than being in P2P fashion

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
faizan2590: Thanks for the explanation. I would then vouch to simply use _log.warn_ rather than _self.owner.log_

faizan2590:
I have left a comment. Will start the review again once addressed. And sorry if I sound like being nitpicking, please harmonise the branch name with PR's header name

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
spyrostz: Agreed, fixing

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
spyrostz: The bids are never tracked (have a look at the already_tracked=True parameter). Therefore the bids are not being tracked, and the matching of the bids-offers take place exclusively on the offer side, by defining the offer.buyer according to the matched bid. Hope this makes it clearer.

spyrostz:
> I have left a comment. Will start the review again once addressed. And sorry if I sound like being nitpicking, please harmonise the branch name with PR's header name

Comment addressed, I will not harmonise now the branch name with the PR header's name though, because that would oblige me to close the PR, thus losing any discussion taking place here. Therefore I will open it once the pending discussions are resolved. 
On the other hand, why do you need that? The branch names do not appear in the commit history (only the commit text is displayed) therefore the branch name is invisible once merged AFAIK. Is there any specific reason why you are adamant in changing that?

faizan2590:
> > I have left a comment. Will start the review again once addressed. And sorry if I sound like being nitpicking, please harmonise the branch name with PR's header name
> 
> Comment addressed, I will not harmonise now the branch name with the PR header's name though, because that would oblige me to close the PR, thus losing any discussion taking place here. Therefore I will open it once the pending discussions are resolved.
> On the other hand, why do you need that? The branch names do not appear in the commit history (only the commit text is displayed) therefore the branch name is invisible once merged AFAIK. Is there any specific reason why you are adamant in changing that?

It was just a suggestion. Please ignore it then

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
faizan2590: Can we discuss offline? May be I would better be able to explain?

PR #453, opened 2019-05-17
Feature/d3 asim 1161

hannesdiedrich:


src/d3a/d3a_core/util.py
```diff
@@ -409,3 +410,27 @@ def convert_datetime_to_str_keys(indict, outdict, ui_format=False):
...
+
+def add_or_create_key(dict, key, value):
+    if key in dict:
+        dict[key] += value
+    else:
+        dict[key] = value
+    return dict
+
+
+def substract_or_create_key(dict, key, value):
```
spyrostz: Trivial typo, substract -> subtract

src/d3a/models/strategy/__init__.py
```diff
@@ -108,6 +116,7 @@ def remove(self, offer):
                 return True
         except KeyError:
             self.strategy.log.warning("Could not find offer to remove")
+            return True
```
spyrostz: I am trying to understand a bit more this function, if I understand correctly it returns either True in some of the branches or None (if the offer.id is not in the sold dict), shouldn't it return False to match the return value ?

hannesdiedrich: This was a quickfix for one unittest. The reason for this:
If market not in self.sold, a KeyError is raised (before it was not, because the DefaultDict just created the market in self.sold). But you are right, this should nit return True in this case. I will change (new)Line 112 to `if market in self.sold and offer.id in self.sold[market]:` to tackle this. OK?
Thanks!

hannesdiedrich: Wanted to wait for more comments first but for some reason committed and pushed it already. Sorry!!!

src/d3a/d3a_core/util.py
```diff
@@ -297,15 +297,16 @@ def update_nested_settings(class_object, class_name, settings_dict):
         update_nested_settings(setting_class, settings_class_name, advanced_settings)
 
 
-def generate_market_slot_list(area):
+def generate_market_slot_list(area=None):
     """
     Returns a list of all slot times
     """
+    config = GlobalConfig if area is None else area.config
```
faizan2590: Can you help me understand, how could it help reduce the RAM requirement?

src/d3a/models/strategy/pv.py
```diff
@@ -106,7 +106,7 @@ def produced_energy_forecast_kWh(self):
         # This forecast ist based on the real PV system data provided by enphase
         # They can be found in the tools folder
         # A fit of a gaussian function to those data results in a formula Energy(time)
-        for slot_time in generate_market_slot_list(self.area):
+        for slot_time in generate_market_slot_list(area=self.area):
```
faizan2590: Does preferring kwargs over args also help in reducing RAM requirement?

tests/test_market.py
```diff
@@ -637,31 +630,13 @@ def check_avg_trade_price(self):
     @precondition(lambda self: self.market.traded_energy)
     @rule()
     def check_acct(self):
-        actor_sums = defaultdict(int)
```
spyrostz: Good that you replaced it here too, hope it did not give you a hard time to fix the 'hypothesis' tests.

src/d3a/models/strategy/pv.py
hannesdiedrich: No, this is just because I changed the implementation of `generate_market_slot_list`.

src/d3a/d3a_core/util.py
hannesdiedrich: This does not reduce the requirement of RAM. This change was only done in order to have a list of time_slots for the creation of the states (PV, Load, Storage), when there is not a `config` member in self.area yet.
In this case we use the GlobalConfig now, which is updated before the creation of the device instances.
Does this make sense?

tests/test_market.py
hannesdiedrich: No, no problems actually.

PR #454, opened 2019-05-20
Feature/d3 asim 1150

hannesdiedrich:


hannesdiedrich:
@spyrostz I was thiking the same: "Why dont I use the KEEP_PASTMARKETS" ? I though it is an advantage, but you are right. For the reason mentioned, i will add it. Thanks!

PR #455, opened 2019-05-21
D3ASIM-1177: Implemented removing of past markets.

hannesdiedrich:


src/d3a/models/area/markets.py
```diff
@@ -60,6 +60,10 @@ def _market_rotation(self, current_time, markets, past_markets, area_agent):
             if timeframe < current_time:
                 market = markets.pop(timeframe)
                 market.readonly = True
+                if ConstSettings.GeneralSettings.KEEP_PAST_MARKETS is False:
+                    for pm in past_markets:
```
spyrostz:
```suggestion
                    past_markets = [pm for pm in past_markets if abs(pm - timeframe) < GlobalConfig.slot_length]
```

spyrostz: I added a suggested change for the aforementioned for loop, because this implementation has the disadvantage that the past_markets list is mutated at the same time as it is iterated over. The reason why this does not crash is that always the current element of the list is deleted (or not), if another element was deleted an exception "RuntimeError: list changed size during iteration" would be raised.

src/d3a/models/area/markets.py
```diff
@@ -60,6 +60,10 @@ def _market_rotation(self, current_time, markets, past_markets, area_agent):
             if timeframe < current_time:
                 market = markets.pop(timeframe)
                 market.readonly = True
+                if ConstSettings.GeneralSettings.KEEP_PAST_MARKETS is False:
```
spyrostz: You can use the 'not' operator instead of checking if it is False.

spyrostz:
Added some comments, LGTM once these are dealt with.

src/d3a/models/area/markets.py
hannesdiedrich: @spyrostz : This does not seam to work.
1. Your proposal writes just the timeslots to a list. However, past_markets is a OrderedDict.
2. I tried to add it accordingly:
```
if not ConstSettings.GeneralSettings.KEEP_PAST_MARKETS:
    past_markets = OrderedDict((slot_time, pm) for slot_time, pm in past_markets.items() if
                    abs(slot_time - timeframe) < GlobalConfig.slot_length)
```
However, now the connection to the self.past_market member is lost and the dict is not persistent. Is there any other way to deal with this? I am out of ideas. 

Actually, in my former implementation, the "RuntimeError: list changed size during iteration" wasnt shown, even for multiple future markets.

Thanks in advance!

spyrostz: Ah I see, in this case the value is passed by reference and it gets modified in place. I did not notice that this is an ordered dict that you are iterating over, so obviously my change is incorrect. I would still suggest though to not mutate the dictionary while iterating it, so I would create a temporary list that holds the keys of the dict keys that are eligible for deletion. Once the loop is over, you can use the list to delete all the dict elements that you want to remove from the dict. I know I am overly careful about this, the reason for that is that this is not an allowed operation by Python.

hannesdiedrich: Yes, I understand! Thanks @spyrostz! I adapted it: https://github.com/gridsingularity/d3a/pull/455/commits/379c7ca8f9334a3a39b1244af78f3057e3c0594c

src/d3a/models/area/markets.py
faizan2590: @spyrostz Are there any specific advantage of using _not_ rather than _False_ ?

spyrostz: There is a slight difference, `not` catches both `False` and `None`. So what I wanted to indicate (nitpicking, I know) is that if a user assigns something like `KEEP_PAST_MARKETS=None` out of ignorance of the parameter type, the program will treat it as False.

PR #457, opened 2019-05-21
D3ASIM-1149: Using current_markets to accumulate grid trades

faizan2590:


src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -176,34 +190,45 @@ def _accumulate_house_trades(house, grid, accumulated_trades, past_market_types)
...
+                if area_name_from_area_or_iaa_name(trade.seller) == \
+                        house.name and trade.buyer not in child_names:
+                    accumulated_trades[house.name]["consumedFromExternal"] += trade.offer.energy
+                    accumulated_trades[house.name]["spentToExternal"] += trade.offer.price
+                elif area_name_from_area_or_iaa_name(trade.buyer) == \
+                        house.name and trade.seller not in child_names:
+                    accumulated_trades[house.name]["producedForExternal"] -= trade.offer.energy
+                    accumulated_trades[house.name]["earnedFromExternal"] += trade.offer.price
+
+    if grid_markets is not None:
```
hannesdiedrich: You can specify a default value when running `getattr()` (in line 194) and could save both if clauses here.

src/d3a/models/area/__init__.py
```diff
@@ -301,6 +301,14 @@ def current_market(self):
         except IndexError:
             return None
 
+    @property
```
hannesdiedrich: This property will be handy for my branch es well. Will copy it. Thanks!

src/d3a/d3a_core/sim_results/area_statistics.py
faizan2590: I tired, but didn't.May be I can show you offline?

faizan2590:
@hannesdiedrich Please have a look at here:https://github.com/gridsingularity/d3a/pull/457/files#diff-5e8e25677e3663098aa5bdd932bd1f8cR113
Im using KEEP_PAST_MARKET flag

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -110,14 +110,19 @@ def update_stats(self, area, simulation_status):
             "load-unit": "kWh",
             "price-energy-day": export_price_energy_day(area)
         }
+        market_type = \
```
spyrostz: I would recommend to create a separate private method called "_update_cumulative_grid_trades" which will handle the update of this stat, in order to not clutter up the update_stats function. It is already done in the unmatched loads case, and the PR for the price energy day follows the same pattern.

integration_tests/integration_tests.feature
```diff
@@ -129,14 +129,14 @@ Feature: Run integration tests
      |  default_3a  |
      |  default_3b  |
 
-  Scenario: Cumulative Grid Trades is the same with and without keeping the past markets
-     Given we have a scenario named default_2a
+  Scenario: Cumulative Grid Trades are the same with and without keeping the past markets
+     Given we have a scenario named balancing_market.default_2a
```
hannesdiedrich: Why changing to the balancing_market.default_2a ?

faizan2590: I want to test if cumulative_grid_balancing_trades too are indifferent of using past OR current market for calculating results

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -132,49 +133,61 @@ def _is_prosumer_node(area):
...
+                    sell_id = area_name_from_area_or_iaa_name(trade.seller)
+                    accumulated_trades[load.name]["consumedFrom"] = add_or_create_key(
+                        accumulated_trades[load.name]["consumedFrom"], sell_id, trade.offer.energy)
+                    accumulated_trades[load.name]["spentTo"] = add_or_create_key(
+                        accumulated_trades[load.name]["spentTo"], sell_id, trade.offer.price)
+        return accumulated_trades
+
+
+def _accumulate_producer_trades(producer, grid, accumulated_trades, past_market_types):
+    if producer.name not in accumulated_trades:
```
spyrostz: I think we discussed it offline too, but why does the assert need to be converted to an if ?

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -186,40 +199,52 @@ def _accumulate_house_trades(house, grid, accumulated_trades, past_market_types)
...
-                    trade.buyer not in child_names:
-                accumulated_trades[house.name]["consumedFromExternal"] += trade.offer.energy
-                accumulated_trades[house.name]["spentToExternal"] += trade.offer.price
-            elif area_name_from_area_or_iaa_name(trade.buyer) == house.name and \
-                    trade.seller not in child_names:
-                accumulated_trades[house.name]["producedForExternal"] -= trade.offer.energy
-                accumulated_trades[house.name]["earnedFromExternal"] += trade.offer.price
+    house_markets = getattr(house, past_market_types)
+    grid_markets = getattr(grid, past_market_types)
+    if house_markets is not None:
```
spyrostz: Since you are already in the process of refactoring this, I would like to see this function split into 2 separate functions: 
1. Accumulate house market trades 
2. Accumulate grid markets trades 
The first can be responsible for the if house_markets is not None block, and the second for the if grid_markets is not None block

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -186,40 +199,52 @@ def _accumulate_house_trades(house, grid, accumulated_trades, past_market_types)
...
-            if area_name_from_area_or_iaa_name(trade.seller) == house.name and \
-                    trade.buyer not in child_names:
-                accumulated_trades[house.name]["consumedFromExternal"] += trade.offer.energy
-                accumulated_trades[house.name]["spentToExternal"] += trade.offer.price
-            elif area_name_from_area_or_iaa_name(trade.buyer) == house.name and \
-                    trade.seller not in child_names:
-                accumulated_trades[house.name]["producedForExternal"] -= trade.offer.energy
-                accumulated_trades[house.name]["earnedFromExternal"] += trade.offer.price
+    house_markets = getattr(house, past_market_types)
+    grid_markets = getattr(grid, past_market_types)
```
spyrostz: Another question that just came to mind, by grid_markets I assume we are talking about the parent market of the Area, right ? Because if it is really the grid, then this code is only working for 2-level hierarchies, like default_2a, and will not work as soon as Streets are included. I am confident that it is referring to the parent market, but since the naming is misleading here I would like to check it with you too.

src/d3a/d3a_core/sim_results/area_statistics.py
faizan2590: Since, accumulated trades are recalculated if KEEP_PAST_MARKETS flag is set to True, Im deleting the old results at ever update_stats
https://github.com/gridsingularity/d3a/pull/457/files#diff-5e8e25677e3663098aa5bdd932bd1f8cR124

src/d3a/d3a_core/sim_results/area_statistics.py
faizan2590: I tested with default_5 by adding one additional hierarchy and it works for more than two hierarchy too. In order to avoid confusion for future renaming 'grid' -> 'parent'

PR #459, opened 2019-05-23
D3ASIM-1170: Improved AreaSerialiser.

hannesdiedrich:
This is not the fix of the bug (its done in d3a-web).
This is just an add on, that enables us to use area_serializer.area_to_string() to use for debugging.

### DIFF
--- a/src/d3a/models/strategy/storage.py
+++ b/src/d3a/models/strategy/storage.py
@@ -60,6 +60,10 @@ class StorageStrategy(BidEnabledStrategy, OfferUpdateFrequencyMixin, BidUpdateFr
                  cap_price_strategy: bool=False,
                  min_allowed_soc=None):
 
+        self.initial_capacity_kWh = initial_capacity_kWh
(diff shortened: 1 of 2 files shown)


src/d3a/models/strategy/storage.py
```diff
@@ -60,6 +60,10 @@ def __init__(self, risk: int = GeneralSettings.DEFAULT_RISK,
                  cap_price_strategy: bool=False,
                  min_allowed_soc=None):
 
+        self.initial_capacity_kWh = initial_capacity_kWh
```
spyrostz: Why are these needed here? I would suggest to leave them only in the state class. The disadvantage that the duplication introduces is that any change that happens to the variables of the StorageStrategy class will be independent from any change that happens to their clones in StorageState class.

hannesdiedrich: Makes sense. The lack of these self members threw an error in the areaSerialiser so i added them without thinking. Will remove them here and add a catch in the serializer.

PR #462, opened 2019-05-28
Added script for converting energy profiles with arbitrary resolution…

spyrostz:
… to power profiles with market slot length resolution.

### DIFF
--- a/src/d3a/resources/convert_energy_to_power_profile.py
+++ b/src/d3a/resources/convert_energy_to_power_profile.py
@@ -1,6 +1,5 @@
 from d3a.models.const import GlobalConfig
-from d3a.models.read_user_profile import _readCSV, _eval_time_period_consensus, \
-    default_profile_dict
+from d3a.models.read_user_profile import _readCSV, default_profile_dict
 from pendulum import duration
 import csv
 import argparse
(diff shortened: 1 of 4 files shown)


src/d3a/resources/convert_energy_to_power_profile.py
```diff
@@ -0,0 +1,71 @@
...
+        if time in input_profile.keys():
+            current_val = input_profile[time]
+        out_profile[time] = current_val
+
+    return out_profile
+
+
+def convert_energy_profile_to_power(input_profile, output_file):
+    profile = _readCSV(input_profile)
+    _eval_time_period_consensus(profile)
```
hannesdiedrich: Just minor. But as this is a preprocessing step, the user might want to not care about the time period consensus.
But i dont have a strong opinion.

src/d3a/resources/convert_energy_to_power_profile.py
```diff
@@ -0,0 +1,71 @@
...
+        # Average market slot values
+        iter_duration = duration(minutes=0)
+        averaged_value = 0
+        while iter_duration < GlobalConfig.slot_length:
+            averaged_value += profile[k + iter_duration]
+            iter_duration += duration(minutes=1)
+        averaged_value /= GlobalConfig.slot_length.minutes
+        output_dict[k] = averaged_value
+
+    def convert_energy_to_power(e):
```
hannesdiedrich: Minor: Could be moved outside of the parent function.

spyrostz: Will do, thanks!

src/d3a/resources/convert_energy_to_power_profile.py
spyrostz: The thing is that Fatma does, she wanted to reinterpolate the data in 15 minutes intervals (for the market slot length). If you agree I can refactor this function to accept the time period as an explicit input argument, and not to rely on the const settings for that.

PR #464, opened 2019-05-29
D3ASIM-1199: Added price_energy_day results per area, instead of glob…

spyrostz:
…al results that were used until now. Removed the pv and storage statistics. Removed timeslot index. Added uuid output to be compatible with d3a-web/redis.

### DIFF
--- a/src/d3a/d3a_core/sim_results/endpoint_buffer.py
+++ b/src/d3a/d3a_core/sim_results/endpoint_buffer.py
@@ -16,8 +16,8 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from d3a.d3a_core.sim_results.area_statistics import export_cumulative_grid_trades, \
-    export_cumulative_grid_trades_redis, export_cumulative_loads, export_price_energy_day, \
-    generate_inter_area_trade_details, MarketPriceEnergyDay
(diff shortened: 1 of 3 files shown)


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -168,12 +151,17 @@ def update_stats(self, area, simulation_status):
             self.file_export_endpoints.traded_energy_profile_redis)
 
     def _update_tree_summary(self, area):
-        price_energy_list = export_price_energy_day(area)
+        price_energy_list = self.price_energy_day.csv_output
 
         def calculate_prices(key, functor):
             # Need to convert to euro cents to avoid having to change the backend
             # TODO: Both this and the frontend have to remove the recalculation
```
hannesdiedrich: Should be use this opportunity to communicate with the frontend to remove the rounding and euro-converting on their side (if not already happend) and remove this comment?

spyrostz: I agree, and I also suggest to use Euro cents for that. Right now the only displayed values will be trade rates, which make more sense to me in cents/kWh rather than Euro/kWh. I would suggest to also remove conversion on our side too and remain consistent to the other graphs. What do you think ?

hannesdiedrich: Perfect, thanks!

PR #465, opened 2019-05-29
Changed local bills export.

hannesdiedrich:


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -234,28 +238,31 @@ def _round_energy_trade_profile(cls, profile):
...
     def _round_area_bill_result_redis(cls, results):
-        for i, _ in enumerate(results):
-            for k in results[i].keys():
-                results[i][k]['bought'] = round_floats_for_ui(results[i][k]['bought'])
-                results[i][k]['sold'] = round_floats_for_ui(results[i][k]['sold'])
-                results[i][k]['spent'] = round_floats_for_ui(results[i][k]['spent'])
-                results[i][k]['earned'] = round_floats_for_ui(results[i][k]['earned'])
-                results[i][k]['total_energy'] = round_floats_for_ui(results[i][k]['total_energy'])
-                results[i][k]['total_cost'] = round_floats_for_ui(results[i][k]['total_cost'])
+        for k in results.keys():
```
spyrostz: Out of curiosity, why did we need the index for in the former implementation? I cannot recall the reason.

hannesdiedrich: The `results` was a list of areas/devices before. Now it is a dictionary (please see the change in line 218 (new)).

spyrostz: That explains the former reason for the index, thanks a lot for explaining!

PR #467, opened 2019-05-31
D3ASIM-1158: Renaming cumulative_grid_trades labels

faizan2590:


src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -419,11 +419,9 @@ def _external_trade_entries(child, accumulated_trades):
             earned = round_floats_for_ui(area_data["earnedFromExternal"][k])
             results["bars"].append({
                 "energy": outgoing_energy,
-                "targetArea": child.name,
-                "energyLabel": f"{child.name} sold {abs(outgoing_energy)} kWh "
-                               f"of {k} to external consumers",
-                "priceLabel": f"{child.name} earned {earned} cents  "
-                              f"of {k} from external consumers."
+                "targetArea": k,
+                "energyLabel": f"External sources sold {abs(outgoing_energy)} kWh ",
```
spyrostz: Same here, external sources are not selling anything, the root area is selling to external sources.

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -406,10 +406,10 @@ def _external_trade_entries(child, accumulated_trades):
...
             results["bars"].append({
                 "energy": incoming_energy,
-                "targetArea": child.name,
-                "energyLabel": f"{child.name} bought {abs(incoming_energy)} "
-                               f"kWh from external sources for {k}",
-                "priceLabel": f"{child.name} spent {abs(spent)} cents for {k} to external sources"
+                "targetArea": k,
+                "energyLabel": f"External sources bought {abs(incoming_energy)} "
+                               f"kWh from {k}",
+                "priceLabel": f"External sources spent {abs(spent)} cents"
```
spyrostz: You need to check with @sarahhambridge if this is what she really wants here. This label implies that the External sources are buying something from the child, and that the external sources spend money, while it is the other way round, the root area is spending money to buy energy from the external .

spyrostz:
Have left 2 comments, will check again once they are dealt with.

PR #469, opened 2019-06-03
D3ASIM-1225: Added support for handling job termination not only via …

spyrostz:
…the stop redis message, but also via the 1. job metadata 2. absence of job in the redis queue.

### DIFF
--- a/src/d3a/d3a_core/redis_communication.py
+++ b/src/d3a/d3a_core/redis_communication.py
@@ -20,7 +20,8 @@ import json
 from logging import getLogger
 from redis import StrictRedis
 from redis.exceptions import ConnectionError
-
+from rq import get_current_job
+from rq.exceptions import NoSuchJobError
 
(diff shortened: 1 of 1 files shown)


src/d3a/d3a_core/redis_communication.py
```diff
@@ -87,11 +88,23 @@ def _slowdown_callback(self, message):
             return
         self._simulation.slowdown = slowdown
 
+    def _handle_redis_job_metadata(self):
+        should_exit = True
+        try:
+            job = get_current_job()
+            job.refresh()
+            should_exit = "terminated" in job.meta and job.meta["terminated"]
+        except NoSuchJobError:
```
hannesdiedrich: Awesome Exception naming! :-)

spyrostz: Indeed

PR #473, opened 2019-06-04
D3ASIM-1203: Added BC flag for running without Blockchain.

hannesdiedrich:


src/d3a/d3a_core/simulation.py
```diff
@@ -44,7 +44,9 @@
 from d3a.d3a_core.redis_communication import RedisSimulationCommunication
 from d3a.models.const import ConstSettings
 from d3a.d3a_core.exceptions import D3AException
-if platform.python_implementation() != "PyPy":
+
+if platform.python_implementation() != "PyPy" and \
+        ConstSettings.BlockchainSettings.ENABLE_BC is True:
```
spyrostz: I would change it to something like BC_INSTALLED in order to recall that this does not really enable blockchain, just denotes whether blockchain modules are installed or not.

faizan2590: @spyrostz Sorry, I didn't get it. This flag is set in ConstSetting, not returned when blockchain modules are installed . So, I couldn't get it how this flag ensures if blockchain modules are installed or not.

spyrostz: @faizan2590 it does not ensure if the modules are installed or not. It is a way though for the user to configure and select if he has installed the web3 module, and, more importantly, to unselect in if for some reason he cannot install the blockchain modules.

hannesdiedrich: no ;-)

PR #476, opened 2019-06-07
D3ASIM-1167: Added check for load count.

hannesdiedrich:


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -90,12 +91,21 @@ def generate_json_report(self):
             "energy_trade_profile": self.energy_trade_profile
         }
 
+    def _write_none_to_unmatched_loads(self, area):
```
spyrostz: I think that it is better to move this function inside the MraketUnmatchedLoads class, since it is relevant only to the unmatched loads.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -59,6 +59,7 @@ def __init__(self, job_id, initial_params):
         self.energy_trade_profile = {}
         self.energy_trade_profile_redis = {}
         self.file_export_endpoints = FileExportEndpoints()
+        self.load_count = load_count
```
spyrostz: Same as below, the load count could be part of the MarketUnmatchedLoads class, and could also potentially even be calculated in its constructor.

src/d3a/d3a_core/simulation.py
```diff
@@ -97,6 +98,17 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig =
         self._init(**self.initial_params)
         self._init_events()
         validate_const_settings_for_simulation()
+        self.count_load_devices_in_setup(self.area, self.load_count)
+        self.endpoint_buffer = SimulationEndpointBuffer(redis_job_id, self.initial_params,
+                                                        self.load_count)
+
+    @classmethod
+    def count_load_devices_in_setup(cls, area, load_count):
```
spyrostz: This function could also be part of the MarketUnmatchedLoad class, and to be called in its constructor.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: Ok, yes. Will do.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: But I like to add to consideration, that if `ConstSettings.GeneralSettings.KEEP_PAST_MARKETS = True` this function will mot be applied. Or shall I also add it to the ExportUnmatchedLoads class?

spyrostz: Hmm, I see the issue now. I would prefer it to add it to the ExportUnmatchedLoads class too if possible. That way the unmatched loads calculations will be isolated in their own module.

PR #477, opened 2019-06-11
D3ASIM-1167: Fixed bug.

hannesdiedrich:


hannesdiedrich:
@spyrostz Thanks for you review. Yes, self.latest_time_slot was updated every market slot before this market slot because ExportUnmatchedLoads was initiated every market slot. The integrationtests did not fail because both ways used the same code base and the same value of self.latest_time_slot for getting the unmatched loads. They bot returned the results mentioned here:
https://gridsingularity.atlassian.net/browse/D3ASIM-1167?focusedCommentId=14469&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-14469
Let me know if this is still unclear.

PR #479, opened 2019-06-20
Issue 478: Fixed pysolc dependency for d3a python3 and pypy dockerfiles.

spyrostz:


Dockerfile_pypy
```diff
@@ -9,5 +9,6 @@ RUN pip install -r/app/requirements/dev.txt && \
     pip install -e . && \
     pip install git+https://github.com/Jonasmpi/py-solc.git
 
-ENTRYPOINT ["pypy3", "src/d3a/d3a_core/launcher.py"]
+RUN which d3a
+ENTRYPOINT ["d3a"]
```
faizan2590: Can you explain me these two lines?

spyrostz: Thanks for noticing that, 'which d3a' is just a debugging command (in order to detect where d3a was installed). Therefore I will remove this. 
As for the entrypoint, the purpose is to use the CLI interface (via the 'd3a' command, and not the launcher script which is responsible for communicating via Redis)

hannesdiedrich:
LGTM

Dockerfile_pypy
faizan2590: Thanks a lot for the explanation!

PR #482, opened 2019-07-03
D3ASIM-1182: Added handler for correctly populating the spot market t…

spyrostz:
…ype in d3a. Fixed 1000 houses setup file to use the updated strategy parameters.

### DIFF
--- a/src/d3a/d3a_core/d3a_jobs.py
+++ b/src/d3a/d3a_core/d3a_jobs.py
@@ -29,7 +29,7 @@ from d3a.d3a_core.util import available_simulation_scenarios
 from d3a.d3a_core.util import update_advanced_settings
 from d3a.d3a_core.simulation import run_simulation
 from d3a.constants import TIME_ZONE, DATE_FORMAT
-from d3a.models.const import GlobalConfig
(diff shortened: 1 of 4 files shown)


src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -48,6 +48,13 @@ def start(scenario, settings):
     if advanced_settings is not None:
         update_advanced_settings(ast.literal_eval(advanced_settings))
 
+    spot_market_type = settings.get('spot_market_type', None)
+    if spot_market_type is not None:
+        if not 1 <= spot_market_type <= 3:
```
hannesdiedrich: This reminds me that we wanted to create a common interface between d3a and d3a-web. Shall we tackle this before moving forward and adding features? IMO yes. What do you think?

spyrostz: I agree, but I think that it has to be done in parallel. One thing that comes to mind is that we need to create a separate 'common' repository that will keep this common interface. Nevertheless, I am opening [this task](https://gridsingularity.atlassian.net/browse/D3ASIM-1337) and creating corresponding subtasks in order to tackle this issue incrementally.

PR #483, opened 2019-07-04
D3ASIM-1329: Added expansion of user one-day-profile for multiple days.

hannesdiedrich:


integration_tests/steps/integration_tests.py
```diff
@@ -950,3 +950,13 @@ def identical_energy_bills(context):
...
                    for _, old_area_results in context.energy_bills_redis.items())
+
+
+@then("the load profile schould be identical on each day")
+def identical_profiles(context):
+    device_stats_dict = context.simulation.endpoint_buffer.device_statistics.device_stats_dict
+    load_profile = device_stats_dict['House 1']['H1 DefinedLoad']['load_profile_kWh']
+    for time_slot, value in load_profile.items():
+        day_plus_one_key = time_slot.add(days=1)
+        if day_plus_one_key in load_profile:
```
faizan2590: This conditional statement could by-pass the below assert. What Im concerned is that it should always assert the current and next day profile. Does it makes sense?

hannesdiedrich: I know what you mean. The problem here is that e.g. 2019-07-01T00:00 + 1 day will not be part of the device stats (neither the load_profile). So i suggest I replace this if clause by something like:
`if not today().add(days=2):` ok?

hannesdiedrich: I already adapted. Please have a look. Thanks!

faizan2590: if Im right you had to compare this(not today().add(days=2)) with other date object too

src/d3a/models/read_user_profile.py
```diff
@@ -289,12 +304,17 @@ def read_arbitrary_profile(profile_type: InputProfileTypes,
...
     or a dict with arbitrary time data (Dict[str, float])
     or a string containing a serialized dict of the aforementioned structure
+    :param copy:
     :return: a mapping from time to profile values
     """
 
     profile = _read_from_different_sources_todict(input_profile)
+    profile_time_list = list(profile.keys())
+    profile_duration = profile_time_list[-1] - profile_time_list[0]
+    if GlobalConfig.sim_duration > duration(days=1) and profile_duration <= duration(days=1):
```
spyrostz: I can understand the reason for the less than or equal (instead of just equal) in the last block, but I am afraid it is going to bite us in the future. Can you please explain what happens if the profile_duration is less than a day ? Because I do not recall what happens in that occasion.

hannesdiedrich: If the profile is provided only e.g. until 12 o'clock, the last value will be used for the rest of the day (12:00-23:45). 
If a user provides profile less than a day, this behaviour should also be applied to the following days. (IMO, what do you think?)

integration_tests/steps/integration_tests.py
hannesdiedrich: Sorry, i dont really understand the question. 
Just do explain a bit more:
If user sets the sim_duration to 2 days, the time slots will be:
`2019-07-01T00:00` until `2019-07-02T23:45`
If you see, there is no `2019-07-03T00:00`. But if in the check you add two days to the first time step, then it will look for exactly this time stamp and fail. 
Is this clearer?

faizan2590: What I understood here is you are using a library method _today()_ to get current day and then adding some two days into it, this would always be successful. IMO, you should use _GlobalConfig.sim_duration_ to check if there should be a next day or not

hannesdiedrich: @faizan2590 you are totally right! The if statement is wrong. I added a fix here: 
https://github.com/gridsingularity/d3a/pull/483/commits/b916eda08e678a0bc04637fdb75d3738f08eb737

PR #486, opened 2019-07-10
D3ASIM-1328: Added deserialization functionality on d3a to handle inc…

spyrostz:
…oming area and settings events via Redis job. Added unit tests.

### DIFF
--- a/src/d3a/d3a_core/simulation.py
+++ b/src/d3a/d3a_core/simulation.py
@@ -44,6 +44,7 @@ from d3a.d3a_core.sim_results.endpoint_buffer import SimulationEndpointBuffer
 from d3a.d3a_core.redis_communication import RedisSimulationCommunication
 from d3a.models.const import ConstSettings
 from d3a.d3a_core.exceptions import D3AException
+from d3a.models.area.event_deserializer import deserialize_events_to_areas
 
 if platform.python_implementation() != "PyPy" and \
         ConstSettings.BlockchainSettings.BC_INSTALLED is True:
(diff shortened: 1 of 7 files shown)


src/d3a/d3a_core/simulation.py
```diff
@@ -94,7 +96,12 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig =
 
         self._load_setup_module()
         self._init(**self.initial_params)
+        # TODO: Deprecated event mechanism since D3ASIM-1328 is about to be merged.
+        # Candidate for deletion.
```
hannesdiedrich: Dont we need that when running locally setup files that contain events?

src/d3a/models/area/event_deserializer.py
```diff
@@ -0,0 +1,58 @@
...
+                  'transfer_fee_pct': event.get("iaa_fee", None),
+                  'market_maker_rate': market_maker_rate}
+        output_events.append(ConfigEvents(event_time, params))
+    return output_events
+
+
+def deserialize_events_to_areas(events, root_area):
+    if not events:
+        return
+    event_list = generate_settings_events(events.get("settings_events", [])) \
```
hannesdiedrich: Could you please rename `event_list` to `settings_event_list`  for better understandability? It took me a while. ;-)

src/d3a/models/area/event_deserializer.py
```diff
@@ -0,0 +1,58 @@
...
+    area.set_events(event_list)
+
+    if not area.children:
+        return
+    for child in area.children:
+        assign_events_to_areas(child, area_events, settings_events)
+
+
+def generate_settings_events(settings_events):
+    output_events = []
```
hannesdiedrich: And maybe `output_events` to `settings_event_list`, too? 
Very minor of corse! Thanks a lot!

src/d3a/d3a_core/simulation.py
spyrostz: No, this is from the event mechanism that Brainbot had introduced before taking over the project. Unfortunately it was partially implemented, that's why we introduced the new event mechanism. As far as I understand, the old event mechanism is not used at all anymore (apart from, maybe, unit tests) and we can safely remove it.

src/d3a/models/area/event_deserializer.py
spyrostz: I agree the naming is not clear, will do , thanks!

src/d3a/d3a_core/simulation.py
hannesdiedrich: Ok, understood. So please go ahead and delete it. Thanks a lot!

src/d3a/models/area/event_deserializer.py
spyrostz: You are right again, will rename this. Thanks!

PR #495, opened 2019-07-19
Feature/d3 asim 1338

hannesdiedrich:


hannesdiedrich:
@spyrostz Yes, i was thinking the same. Lets discus!

PR #498, opened 2019-07-24
Feature/d3 asim 1342

hannesdiedrich:


src/d3a/is_installed.py
```diff
@@ -0,0 +1,7 @@
+
+def is_installed(module_name):
```
faizan2590: I couldn't find its usage.

src/d3a/d3a_core/util.py
```diff
@@ -477,6 +474,14 @@ def create_subdict_or_update(indict, key, subdict):
     return indict
 
 
+def is_installed(module_name):
```
faizan2590: I guess its duplicated here _src/d3a/is_installed.py_

src/d3a/is_installed.py
hannesdiedrich: Weird, I thought I deleted it. You are right, it can be removed.

PR #500, opened 2019-07-30
D3ASIM-1400: Fixed Energy trade price and unmatched loads graphs.

hannesdiedrich:


src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -95,17 +95,17 @@ def _serialize_traded_energy_lists(self, area):
...
         for row in data.rows():
             for ii, label in enumerate(data.labels()):
-                out_dict[label].append(row[ii])
-        return out_dict
+                out_dict[area.slug][label].append(row[ii])
 
     def update_plot_stats(self, area):
-        self.plot_stats[area.slug] = self._get_stats_from_market_data(area, False)
-        self.plot_balancing_stats[area.slug] = self._get_stats_from_market_data(area, True)
+        self._get_stats_from_market_data(self.plot_stats, area, False)
```
faizan2590: Sorry, may be I have misunderstood it but it will always return None and neither this method call updates any instance variable. So, what exactly is the use of this method call now?

hannesdiedrich: Sorry, i dont understand. What returns none? Neither `self.update_plot_stats` nore `self._get_stats_from_market_data` returns any data. They both deal with adding information to `self.plot_stats` and `self.plot_balancing_stats` . Let me know if you need more information. We can discuss offline.

PR #501, opened 2019-07-30
D3ASIM-929: Adding min_offer_age as const parameter as well as recalc…

faizan2590:
…ulating MAX_OFFER_TRAVERSAL_LENGTH based on min_offer_age

### DIFF
--- a/src/d3a/d3a_core/simulation.py
+++ b/src/d3a/d3a_core/simulation.py
@@ -104,8 +104,9 @@ class Simulation:
                                                         self.area)
 
     def _set_traversal_length(self):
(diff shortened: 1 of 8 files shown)


src/d3a/d3a_core/simulation.py
```diff
@@ -104,8 +104,9 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig =
                                                         self.area)
 
     def _set_traversal_length(self):
+        from d3a.models.const import ConstSettings
```
hannesdiedrich: You can remove this line because ConstSetttings is imported already.

PR #504, opened 2019-08-13
D3ASIM-1440: Appending last market data in CSV

faizan2590:


src/d3a/d3a_core/export.py
```diff
@@ -244,13 +259,14 @@ def _export_area_stats_csv_file(self, area: Area, directory: dir, balancing: boo
             area_name += "-balancing"
         data = self.export_data.generate_market_export_data(area, balancing)
         rows = data.rows()
-        if not rows:
+        if not rows and not is_first:
```
hannesdiedrich: Can you please explain, why you added `and not is_first`  here? Thanks!

src/d3a/d3a_core/export.py
```diff
@@ -132,6 +131,11 @@ def export(self):
                 ConstSettings.GeneralSettings.SUPPLY_DEMAND_PLOTS:
             self.plot_supply_demand_curve(self.area, self.plot_dir)
         self.move_root_plot_folder()
+        self.export_json_data(self.directory)
+
+    def data_to_csv(self, area, is_first):
+        with page_lock:
```
hannesdiedrich: Is the page_lock really necessary? Doesnt `with open() as file` deals with that?

src/d3a/d3a_core/export.py
faizan2590: Without this additional check, it will return without adding a header in CSV.

src/d3a/d3a_core/export.py
faizan2590: Yes, you are right. Removing it!

src/d3a/d3a_core/export.py
```diff
@@ -147,40 +148,46 @@ def move_root_plot_folder(self):
...
-        self._export_area_stats_csv_file(area, directory, balancing=True)
         if area.children:
-            self.kpi.update_kpis_from_area(area)
-            self._export_trade_csv_files(area, directory, balancing=False)
-            self._export_trade_csv_files(area, directory, balancing=True)
-            self._export_area_offers_bids_csv_files(area, directory, "offers",
-                                                    Offer, "offer_history", area.past_markets)
-            self._export_area_offers_bids_csv_files(area, directory, "bids",
-                                                    Bid, "bid_history", area.past_markets)
+            # self.kpi.update_kpis_from_area(area) To be uncommented once KPI is implemented
```
spyrostz: Please prepend the comment with a TODO in order to stand out on Pycharm

PR #505, opened 2019-08-21
Feature/d3 asim 1139

spyrostz:


src/d3a/models/market/one_sided.py
```diff
@@ -67,13 +67,18 @@ def offer(self, price: float, energy: float, seller: str,
 
         offer_id = self.bc_interface.create_new_offer(energy, price, seller)
         offer = Offer(offer_id, price, energy, seller, original_offer_price)
-        self.offers[offer.id] = offer
+        from copy import deepcopy
```
hannesdiedrich: Please move to top.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -203,13 +202,19 @@ def _calculate_price_decrease_rate(self, market):
             reduced_price = \
                 self.calculate_initial_sell_rate(market.time_slot) - \
                 price_dec_per_update * self._price_update_interval
+            if reduced_price < self.final_selling_rate:
```
hannesdiedrich: Weird that this did not come up earlier...

src/d3a/models/market/market_structures.py
```diff
@@ -36,6 +36,16 @@ def __str__(self):
         return "{{{s.id!s:.6s}}} [{s.seller}]: {s.energy} kWh @ {s.price} @ {rate}"\
             .format(s=self, rate=self.price / self.energy)
 
+    def __hash__(self):
+        return hash(self.id)
+
+    def __eq__(self, other):
```
hannesdiedrich: Can you please tell me where this is used?

src/d3a/models/market/one_sided.py
spyrostz: Will do, thanks!

src/d3a/models/market/market_structures.py
spyrostz: Mostly in tests for now, but I thought it was a good addition in general. The root cause is the deepcopy call that creates a clone of the original offer object. The clone and the original offers are not the same objects, therefore the == comparison failed in the unit tests. In order to fix this behavior, the most straightforward way is to override the == operator (using the Python magic method __eq__) which defines the conditions on which an object is equal to another. This could of course be solved directly by changing the tests, but IMO this change is also nice to have to the production code too, because that way offer object comparisons will not have false negatives when comparing an offer with its clone, and that's the reason I added it globally.

src/d3a/models/strategy/update_frequency.py
spyrostz: That is indeed something I did not figure out too, why it remained undetected.

src/d3a/models/market/market_structures.py
hannesdiedrich: Ah, nice! Thanks for the info!

PR #506, opened 2019-08-23
Feature/d3 asim 1356

spyrostz:


src/d3a/d3a_core/util.py
```diff
@@ -51,12 +49,47 @@
 
 _CONTRACT_CACHE = {}
 
+TRACE = 5
```
faizan2590: Sorry, I didn't get it

spyrostz: In order to understand it you need to have a look at the logging module source code, and how to add custom levels. Each of the logging levels are associated to a unique integer: 
https://docs.python.org/3/library/logging.html#logging-levels
Therefore for our custom level we need to associate it with an integer lower than 10 (which is the debug log level). The reason why it is 5 and not, say, 9, is purely speculative. It could be any value between 1 and 9.

PR #507, opened 2019-08-26
D3ASIM-1429: Changed offer/bid update frequency

faizan2590:


src/d3a/models/strategy/__init__.py
```diff
@@ -297,6 +297,22 @@ def are_bids_posted(self, market_id):
...
+        # should be only bid from a device to a market at all times, which will be replaced if
+        # it needs to be updated. If this check is not there, the market cycle event will post
+        # one bid twice, which actually happens on the very first market slot cycle.
+        if not all(bid.buyer != self.owner.name for bid in market.bids.values()):
+            self.owner.log.warning(f"There is already another bid posted on the market, therefore"
+                                   f" do not repost another first bid.")
+            return None
+        return self.post_bid(
+            market,
+            energy_Wh * self.bid_update.initial_rate / 1000.0,
```
spyrostz: Pass initial_rate as method argument.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -15,237 +15,86 @@
...
-                 final_rate_profile):
-        self._initial_rate_profile = initial_rate_profile
-        self._final_rate_profile = final_rate_profile
-        self._increase_rate_timepoint_s = {}
-
-    @cached_property
-    def _increase_frequency_s(self):
-        return self.area.config.tick_length.seconds * \
+    def update_on_activate(self, strategy):
+        assert self.update_interval * 60 >= strategy.area.config.tick_length.seconds * \
```
spyrostz: Should always be >= to the constant one

src/d3a/models/strategy/update_frequency.py
```diff
@@ -15,237 +15,86 @@
...
-            return reduced_price
+    @property
+    def _get_updated_rate(self):
+        updated_rate = self.initial_rate - self.energy_rate_change_per_update * self.update_counter
+        return updated_rate
+
+    def get_price_update_point(self, strategy):
+        current_tick_number = strategy.area.current_tick % strategy.area.config.ticks_per_slot
+        elapsed_seconds = current_tick_number * strategy.area.config.tick_length.seconds
+        if elapsed_seconds >= self.update_interval * 60 * (self.update_counter+1):
```
spyrostz: update_interval is better fit to be a timedelta.

src/d3a/models/const.py
```diff
@@ -29,7 +29,9 @@ class GeneralSettings:
         # Max energy price (market maker rate) in ct / kWh
         DEFAULT_MARKET_MAKER_RATE = 30  # 0.3 Eur
         # Number of ticks, an offer needs to be able to travel to reach each part of the setup
-        MAX_OFFER_TRAVERSAL_LENGTH = 6
+        MAX_OFFER_TRAVERSAL_LENGTH = 2
+        # number of time OFFER/BID needs to be updated
```
hannesdiedrich: Please remove "number of"

src/d3a/models/strategy/pv.py
```diff
@@ -19,88 +19,98 @@
...
+                 float=ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE,
+                 final_selling_rate:
+                 float=ConstSettings.PVSettings.FINAL_SELLING_RATE,
+                 fit_to_limit: bool=True,
+                 update_interval=timedelta(minutes=ConstSettings.GeneralSettings.UPDATE_RATE),
+                 energy_rate_change_per_update:
+                 float=ConstSettings.GeneralSettings.ENERGY_RATE_DECREASE_PER_UPDATE,
+                 max_panel_power_W: float=ConstSettings.PVSettings.MAX_PANEL_OUTPUT_W):
+        """
+        :param panel_count: Number of solar panels for this PV plant
```
hannesdiedrich: I'm a big fan of these doc strings! Thanks for adding!

src/d3a/models/strategy/update_frequency.py
```diff
@@ -15,247 +15,141 @@
...
+        number_of_available_updates = \
+            int(strategy.area.config.slot_length.seconds / self.update_interval.seconds) - 1
+        return number_of_available_updates
+
+    def update_on_activate(self, strategy):
+        assert self.update_interval.seconds >= \
+               ConstSettings.GeneralSettings.UPDATE_RATE * 60
+        self.number_of_available_updates = \
+            self._calculate_number_of_available_updates_per_slot(strategy)
+        self.energy_rate_change_per_update = \
```
hannesdiedrich: I suggest to modify the `energy_rate_change_per_update` member inside the `_get_energy_rate_change_per_update()` function.
```
        self.energy_rate_change_per_update = read_arbitrary_profile(InputProfileTypes.IDENTITY,
                                                                    energy_rate_change_per_update)
```
could also be moved from the __init__ into _get_energy_rate_change_per_update(). To have everything in the same place.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -15,247 +15,141 @@
...
+            self._get_energy_rate_change_per_update()
+
+    def reset_on_market_cycle(self):
+        self.update_counter = 0
```
hannesdiedrich: Does this one command really need a separate function? I see it only used in two places. Additionally, the function name does not really describe what is done here.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -15,247 +15,141 @@
...
+
+    def get_updated_rate(self, market):
```
hannesdiedrich: As you only need the time_slot here, please omit to pass a market object here (remember the reducing RAM story). Thanks!

src/d3a/models/strategy/update_frequency.py
```diff
@@ -15,247 +15,141 @@
...
-        elif self.initial_rate_option is InitialRateOptions.MARKET_MAKER_RATE:
-            return self.area.config.market_maker_rate[current_time_h]
-        elif self.initial_rate_option is InitialRateOptions.CUSTOM_RATE:
-            return self.initial_selling_rate
+                energy_rate_change_per_update[slot] = self.energy_rate_change_per_update[slot]
+        return energy_rate_change_per_update
+
+    def _calculate_number_of_available_updates_per_slot(self, strategy):
+        number_of_available_updates = \
+            int(strategy.area.config.slot_length.seconds / self.update_interval.seconds) - 1
```
hannesdiedrich: Did you check whether with this calculation the final_rate is really reached (even for combinations of e.g. slot_length=60s and update_interval=7s)?

src/d3a/models/strategy/update_frequency.py
```diff
@@ -15,247 +15,141 @@
...
+        for market in strategy.area.all_markets[:-1]:
+            self.update_energy_price(market, strategy)
+
+    def update_offer(self, strategy):
+        if self.get_price_update_point(strategy):
+            for market in strategy.area.all_markets:
+                self.update_energy_price(market, strategy)
+
+    #############
+    # BID_METHOD#
```
hannesdiedrich: Is this comment block necessary?

src/d3a/models/strategy/storage.py
```diff
@@ -15,208 +15,197 @@
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-from typing import Union
+# from typing import Union
```
hannesdiedrich: Please remove if not needed.

src/d3a/models/strategy/storage.py
```diff
@@ -15,208 +15,197 @@
...
-    BidUpdateFrequencyMixin
-from d3a.models.read_user_profile import read_arbitrary_profile
-from d3a.models.read_user_profile import InputProfileTypes
+from d3a.models.strategy.update_frequency import UpdateFrequencyMixin
+from d3a.models.read_user_profile import read_arbitrary_profile, InputProfileTypes
 from d3a.d3a_core.device_registry import DeviceRegistry
 
 BalancingRatio = namedtuple('BalancingRatio', ('demand', 'supply'))
-BreakEven = namedtuple('BreakEven', ('buy', 'sell'))
+# BreakEven = namedtuple('BreakEven', ('buy', 'sell'))
```
hannesdiedrich: Same here

src/d3a/models/strategy/storage.py
```diff
@@ -15,208 +15,197 @@
...
+            raise ValueError("Initial selling rate must be greater than final selling rate.")
+        if initial_buying_rate is not None and initial_buying_rate < 0:
+            raise ValueError("Initial buying rate must be greater equal 0.")
+        if final_buying_rate is not None and final_buying_rate < 0:
+            raise ValueError("Final buying rate must be greater equal 0.")
+        if initial_buying_rate is not None and final_buying_rate is not None and \
+                initial_buying_rate > final_buying_rate:
+            raise ValueError("Initial buying rate must be less than final buying rate.")
+        if final_selling_rate is not None and final_buying_rate is not None and \
+                final_selling_rate < final_buying_rate:
```
hannesdiedrich: Cant they be equal as well? Then there is no revenue for the owner, but it should be possible right?
Also, please make the error message more verbose here. Thanks!

src/d3a/models/strategy/update_frequency.py
```diff
@@ -15,247 +15,141 @@
...
+                                                   initial_rate)
+        self.final_rate = read_arbitrary_profile(InputProfileTypes.IDENTITY,
+                                                 final_rate)
+        self.energy_rate_change_per_update = read_arbitrary_profile(InputProfileTypes.IDENTITY,
+                                                                    energy_rate_change_per_update)
+        self.update_interval = update_interval
+        self.update_counter = 0
+        self.number_of_available_updates = 0
+
+    def reassign_mixin_arguments(self, strategy, market, initial_rate=None, final_rate=None,
```
hannesdiedrich: Here again, you only need the time_slot, not the market object.

src/d3a/models/strategy/storage.py
```diff
@@ -291,8 +284,7 @@ def event_market_cycle(self):
         if ConstSettings.IAASettings.MARKET_TYPE == 2 or \
            ConstSettings.IAASettings.MARKET_TYPE == 3:
             self.state.clamp_energy_to_buy_kWh([current_market.time_slot])
-            self.update_market_cycle_bids(final_rate=self.break_even[
-                self.area.now].buy)
+            self.bid_update.update_market_cycle_bids(self)
```
hannesdiedrich: You are calling this line twice in this function (also in line 272). But really, you only want to call it for the two sided markets, so ONLY here, right?

src/d3a/models/strategy/storage.py
```diff
@@ -354,7 +347,7 @@ def sell_energy(self):
         energy_sell_dict = self.state.clamp_energy_to_sell_kWh(
             [ma.time_slot for ma in markets_to_sell])
         for market in markets_to_sell:
-            self.set_initial_selling_rate_alternative_pricing_scheme(market)
+            # self.set_initial_selling_rate_alternative_pricing_scheme(market)
```
hannesdiedrich: please remove

src/d3a/models/const.py
```diff
@@ -57,16 +59,16 @@ class StorageSettings:
         RATE_DECREASE_OPTION = 1
         # Energy sell break-even point, storage never sells for less than this value.
         # Unit is ct/kWh.
-        BREAK_EVEN_SELL = 25
+        BREAK_EVEN_SELL = 25.1
```
hannesdiedrich: Can you please elaborate the reasons for changing the default values here?
Also: As in the storage strategy, the BREAK_EVEN_SELL was renamed to final_selling_rate and BREAK_EVEN_BUY to final_buying_rate, how about renaming it here as well!?

src/d3a/models/strategy/pv.py
```diff
@@ -186,3 +194,21 @@ def event_offer(self, *, market_id, offer):
         # if offer was deleted but not traded, free the energy in state.available_energy_kWh again
         if offer.seller == self.owner.name:
             self.state.available_energy_kWh[market.time_slot] -= offer.energy
+
+    def _set_alternative_pricing_scheme(self, market):
```
hannesdiedrich: Cant this be done only once at the __init__ or when activating? Why does it have to be done for each market? Th same applies to the storage and load.

src/d3a/models/strategy/load_hours.py
```diff
@@ -19,13 +19,13 @@
 from pendulum import duration
 from typing import Union
 from collections import namedtuple
-
+from datetime import timedelta
```
hannesdiedrich: I know, it is in the update_frequency as well, but please use pendulum for managing times and durations (here you could use `from pendulum import duration`. 
At some point we decided to use pendulum instead of datetime. Thanks!

src/d3a/d3a_core/simulation.py
```diff
@@ -106,7 +106,7 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig =
     def _set_traversal_length(self):
         no_of_levels = self._get_setup_levels(self.area) + 1
         num_ticks_to_propagate = no_of_levels * 2
-        ConstSettings.GeneralSettings.MAX_OFFER_TRAVERSAL_LENGTH = int(num_ticks_to_propagate)
+        ConstSettings.GeneralSettings.MAX_OFFER_TRAVERSAL_LENGTH = 2
```
hannesdiedrich: If you use the default value, you can omit this line.

hannesdiedrich: Or omit the setting in const.py.

src/d3a/setup/default_2a.py
```diff
@@ -55,7 +55,8 @@ def get_setup(config):
                                                                            range(12, 16)),
                                                                        final_buying_rate=35),
                          appliance=SwitchableAppliance()),
-                    Area('H2 PV', strategy=PVStrategy(4, 80),
+                    Area('H2 PV', strategy=PVStrategy(panel_count=4, initial_selling_rate=30,
```
hannesdiedrich: Nice! The old why of setting already resulted in some questions.

src/d3a/setup/default_2b.py
```diff
@@ -36,11 +36,11 @@ def get_setup(config):
                                                                        hrs_per_day=6,
                                                                        hrs_of_day=list(
                                                                            range(12, 18)),
-                                                                       final_buying_rate=25),
+                                                                       final_buying_rate=26),
```
hannesdiedrich: Why was this change needed?

src/d3a/setup/jira/custom_storage_strategy.py
```diff
@@ -56,8 +56,8 @@ def decrease_energy_price_over_ticks(self, market):
         # example for determining the current tick number:
         current_tick_number = self.area.current_tick % self.area.config.ticks_per_slot
         if current_tick_number >= 0:
-            if reduced_rate < self.final_selling_rate:
```
hannesdiedrich: Why did you rename only this file (there is also test_strategy_custom_p.py and *_load.py).
IMO this does not have to be renamed.

src/d3a/setup/strategy_tests/pv_const_price_decrease.py
```diff
@@ -37,16 +37,7 @@
...
-market_maker_rate = {
-    0: 35, 1: 35, 2: 35, 3: 35, 4: 35, 5: 35, 6: 35, 7: 35,
-    8: 35, 9: 35, 10: 35, 11: 35, 12: 35, 13: 35, 14: 35,
-    15: 35, 16: 35, 17: 35, 18: 35, 19: 35, 20: 35, 21: 35,
-    22: 35, 23: 35
-}
-
-
 def get_setup(config):
-    config.read_market_maker_rate(market_maker_rate)
```
hannesdiedrich: Isn't it needed here or did you remove it in order to fix the integrationtest?

src/d3a/setup/strategy_tests/storage_strategy_break_even_hourly.py
```diff
@@ -1,122 +0,0 @@
-"""
```
hannesdiedrich: Can you please explain, why you deleted this? Also src/d3a/setup/strategy_tests/storage_strategy_break_even_range.py

src/d3a/setup/two_sided_market/user_min_rate_profile_load_dict.py
```diff
@@ -50,12 +50,12 @@ def get_setup(config):
                     Area('H1 General Load 1',
                          strategy=LoadHoursStrategy(avg_power_W=200, hrs_of_day=list(range(0, 24)),
                                                     initial_buying_rate=user_profile_int,
-                                                    final_buying_rate=35),
+                                                    final_buying_rate=user_profile_int),
```
hannesdiedrich: Can you please explain why you changed it from 35 to the user_profile_int ?

tests/test_area_serializer.py
```diff
@@ -59,17 +59,16 @@ def test_raises_unknown_class():
...
     area_str = area_to_string(area)
-    assert json.loads(area_str)['strategy']['kwargs']['risk'] == 1
     recovered = area_from_string(area_str)
     assert recovered.strategy.panel_count == 42
 
 
 def test_non_attr_param():
     area1 = Area('area1', [], None, PVStrategy())
     recovered1 = area_from_string(area_to_string(area1))
-    assert recovered1.strategy.final_selling_rate == ConstSettings.PVSettings.FINAL_SELLING_RATE
```
hannesdiedrich: This should still work, right? Why did you change it?

tests/test_strategy_storage.py
```diff
@@ -339,14 +339,14 @@ def storage_strategy_test5(area_test5, called):
     s.offers.post(area_test5.past_market.offers['id3'], area_test5.past_market.id)
     s.offers.post(area_test5.past_market.offers['id2'], area_test5.past_market.id)
     s.offers.sold_offer('id2', area_test5.past_market)
-    assert s.state.used_storage == 5
+    assert(isclose(s.state.used_storage, 5, rel_tol=1e-05))
```
hannesdiedrich: I would also be fine with changing the initial_soc and battery_capacity_kWh in order to not need `isclose` (this applies to the whole test suite in this file.

integration_tests/steps/two_sided_market.py
```diff
@@ -134,20 +140,19 @@ def energy_rate_average_between_min_and_max_ess_pv(context):
...
     for market in house2.past_markets:
         for trade in market.trades:
             if trade.seller == pv.name:
-                pv_rates_set.add(trade.offer.price / trade.offer.energy)
-
-    assert all([RATE_LOWER_THRESHOLD < rate < RATE_UPPER_THRESHOLD for rate in storage_rates_set])
-    assert all([RATE_LOWER_THRESHOLD < rate < RATE_UPPER_THRESHOLD for rate in pv_rates_set])
+                initial = storage.strategy.bid_update.initial_rate[market.time_slot]
+                final = storage.strategy.bid_update.final_rate[market.time_slot]
+                assert initial <= (trade.offer.price / trade.offer.energy) <= final
```
hannesdiedrich: IMO this is a too unconstrained assertion. Yes, it should be in between initial and final, but arent there rates that are in a certain range? 
Of course we will probably change this behaviour when we implement the changes tot the 2 sided market behaviour. But still, should be the trade rate be somewhere around 15?

integration_tests/steps/two_sided_market.py
```diff
@@ -74,10 +79,11 @@ def device_partially_fulfill_bid(context, device):
                    for trade in house2.get_past_market(slot).trades)
         assert all(trade.seller in pv_names
                    for trade in house2.get_past_market(slot).trades)
-        assert all(RATE_LOWER_THRESHOLD <
-                   trade.offer.price / trade.offer.energy <
-                   RATE_UPPER_THRESHOLD
-                   for trade in market.trades)
+
+        # assert all(pv_names[0].strategy.offer_update.ini <
```
hannesdiedrich: You can remove this.

integration_tests/steps/two_sided_market.py
```diff
@@ -22,6 +22,7 @@
 from d3a.models.const import ConstSettings
 from d3a.d3a_core.util import make_iaa_name
 from d3a import limit_float_precision
+from d3a.constants import DEFAULT_PRECISION
 
 RATE_LOWER_THRESHOLD = 13
```
hannesdiedrich: If you dont need these, please remove.

integration_tests/steps/pv_tests.py
```diff
@@ -36,59 +37,39 @@ def storages_pv_final_selling_rate(context):
...
 
-    slot_length = context.simulation.simulation_config.slot_length.seconds
-    tick_length = context.simulation.simulation_config.tick_length.seconds
-    wait_time = tick_length * ConstSettings.GeneralSettings.MAX_OFFER_TRAVERSAL_LENGTH + 1
-    number_of_updates_per_slot = int(slot_length/wait_time)
+    number_of_available_updates = \
+        int(pv.config.slot_length.seconds / pv.strategy.offer_update.update_interval.seconds) - 1
+    energy_rate_change_per_update = 4
+    rate_list = []
+    for i in range(number_of_available_updates):
```
hannesdiedrich: Minor, but this can be done in one line.

integration_tests/alternative_pricing.feature
```diff
@@ -11,9 +11,9 @@ Feature: Alternative Pricing
     And d3a is installed
     When we run the d3a simulation with <scenario> [24, 60, 60]
     Then average trade rate is the MMR until <time>
-    And average trade rate after <time> is <trade_rate>
+    And average trade rate between <time> & <time2> is <trade_rate>
```
hannesdiedrich: Can you please rename `time` to `time1` ? Thanks!

src/d3a/d3a_core/simulation.py
faizan2590: Yes but the purpose of this method is to calculate MAX_OFFER_TRAVERSAL_LENGTH which is not used any where

integration_tests/integration_tests.feature
```diff
@@ -91,38 +91,39 @@ Feature: Run integration tests
      When we run the d3a simulation on console with default_2a
      Then aggregated result files are exported
 
-  Scenario: Grid fees are calculated based on the original offer rate
```
spyrostz: Why do these need to be disabled ? Can you explain how does this story affects the grid fees ?

integration_tests/steps/load_tests.py
```diff
@@ -98,14 +98,18 @@ def check_min_user_rate_profile_dict(context):
     load2 = next(filter(lambda x: x.name == "H1 General Load 2", house.children))
 
     for market in house.past_markets:
+        print(f"NAME: {market.area.name}")
```
spyrostz: Reminder before closing the PR, please remove the prints

src/d3a/models/strategy/storage.py
```diff
@@ -15,208 +15,202 @@
...
-                             "MIN_ALLOWED_SOC and 100.")
-        if risk and not 0 <= risk <= 100:
-            raise ValueError("Risk is a percentage value, should be between 0 and 100.")
-        if initial_capacity_kWh and battery_capacity_kWh and min_allowed_soc and \
-           not min_allowed_soc * battery_capacity_kWh <= \
-           initial_capacity_kWh <= battery_capacity_kWh:
-            raise ValueError(f"Initial capacity should be between min_allowed_capacity and "
-                             "battery_capacity_kWh parameter.")
-        if initial_selling_rate < 0:
+        if max_abs_battery_power_kW is not None and max_abs_battery_power_kW < 0:
```
spyrostz: Probably not part of this story, but these validation methods would be nice to be on the d3a-interface, so that d3a-web can take advantage of these too. Not necessary for now, if you agree (also @hannesdiedrich ) we can open a new story for merging the validation functions for these parameters in d3a-interface.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -15,247 +15,138 @@
...
             else:
-                return self.area.historical_avg_rate
-        elif self.initial_rate_option is InitialRateOptions.MARKET_MAKER_RATE:
-            return self.area.config.market_maker_rate[current_time_h]
-        elif self.initial_rate_option is InitialRateOptions.CUSTOM_RATE:
-            return self.initial_selling_rate
+                energy_rate_change_per_update[slot] = self.energy_rate_change_per_update[slot]
+        self.energy_rate_change_per_update = energy_rate_change_per_update
+
+    def _calculate_number_of_available_updates_per_slot(self, strategy):
```
spyrostz: Trivial, but I would convert it to a @property

src/d3a/models/strategy/storage.py
```diff
@@ -15,208 +15,202 @@
...
+                               fit_to_limit=None, update_interval=None,
+                               energy_rate_change_per_update=None):
+
+        self._validate_constructor_arguments(
+            initial_selling_rate=initial_selling_rate,
+            final_selling_rate=final_selling_rate,
+            initial_buying_rate=initial_buying_rate,
+            final_buying_rate=final_buying_rate,
+            energy_rate_change_per_update=energy_rate_change_per_update)
+        if cap_price_strategy is not None:
```
spyrostz: Because this is getting lengthy, I would suggest to gather all these read_arbitrary_profile calls in a separate method, named e.g. populate_offer_bid_update

src/d3a/models/strategy/storage.py
```diff
@@ -276,7 +270,10 @@ def event_bid_traded(self, *, market_id, bid_trade):
 
     def event_market_cycle(self):
         super().event_market_cycle()
-        self.update_market_cycle_offers(self.break_even[self.area.now].sell)
+
+        self.offer_update.update_market_cycle_offers(self)
+        for market in self.area.all_markets[:-1]:
```
spyrostz: Isn't this the same as past_markets ? Moreover, won't this fail in case we have more than one future market?

faizan2590: Not really, all_markets returns the list of all active_markets, whereas past_markets as the name says returns the list of all closed past_markets.

And sorry,  I didn't get it, why would it fail if we have more than one future markets. We already had an integration_test to ensure to have the same results in both cases(either one future market OR multiple future_markets).

Let me know if I haven't understood you correctly or making any mistake in terms of understanding.

src/d3a/models/strategy/update_frequency.py
faizan2590: Sorry for the broken link, have a look at this link (https://github.com/gridsingularity/d3a/blob/0f81e5fdd70d5502498c9c2dc8d7147d71a04049/src/d3a/models/strategy/update_frequency.py#L78). If I recall correctly, @spyrostz  suggested to have an additional assertion to avoid too low `update_interval` in order to avoid spamming the blockchain with very high frequency of offers/bid update.

We can also discuss offline too

integration_tests/integration_tests.feature
faizan2590: Sorry, you are right, re-adding it

hannesdiedrich: Please also delete the comment then. Thanks!

integration_tests/steps/two_sided_market.py
hannesdiedrich: Actually, mean mean the middle here, so initially, the rates were in the middle between initial and final. I remember when I first changed this, was the trade rate 15 because the initial was 30 and the final rate was 15. 
Now you just check whether the rate is between the two, so it could be 1 or also 29. 
Maybe the behaviour of the trade rate changed and the trade rate is fluctuating a lot, but then, please rename the step function. Thanks!

src/d3a/models/strategy/storage.py
hannesdiedrich: Yeah, its weird that the integrationtest does not fail. I also would vote for removing the `[:-1]` and reset the counter for all active markets.

src/d3a/models/strategy/update_frequency.py
hannesdiedrich: It looks ugly. :-) Thanks for removing!

src/d3a/models/strategy/pv.py
```diff
@@ -188,3 +196,25 @@ def event_offer(self, *, market_id, offer):
         # if offer was deleted but not traded, free the energy in state.available_energy_kWh again
         if offer.seller == self.owner.name:
             self.state.available_energy_kWh[market.time_slot] -= offer.energy
+
+    def _set_alternative_pricing_scheme(self):
```
hannesdiedrich: Correct me if I am wrong, but do we really need this function? Isnt the rate dictated by the sellers in case of the alternative pricing? So IMO you can omit this function.

hannesdiedrich: EDIT: Ahhh, this is the PV. Of course we need it then. Sorry!

src/d3a/models/strategy/storage.py
faizan2590: The reason I omit the most recent active_market here is there is no bid placed in this market. So there is no reason to update it here. Let me know if its clear.

hannesdiedrich: Yes! I got it wrong. You are right. It doesnt make sense to set the counter in the most recent market.

integration_tests/steps/two_sided_market.py
hannesdiedrich: EDIT: There is an error in my comment. This is the correct one:
Actually, mean mean the middle here, so initially, the rates were in the middle between initial and final. I remember when I first changed this, the trade rate was 15 because the initial was 30 and the final rate was 0.
Now you just check whether the rate is between the two, so it could be 1 or also 29.
Maybe the behaviour of the trade rate changed and the trade rate is fluctuating a lot, but then, please rename the step function. Thanks!

tests/test_strategy_storage.py
hannesdiedrich: I saw that you didnt change this. But ok. Leave it now.
But please delete the unneeded brackets. Thanks!

integration_tests/steps/two_sided_market.py
faizan2590: May be I have misunderstood it in place but the scenario is testing two sided market for one ESS and one PV `Scenario: One storage, one pv`. It was previously also checking if ESS not violating its buying THRESHOLD.
What I tried to do here is to rather than comparing with hard coded THRESHOLD, Im now  comparing with threshold(i.e. initial and final buying rate) that is set at the time of initialising ESS strategy.

I hope I didn't deviated the initial target of this integration test

tests/test_strategy_storage.py
faizan2590: Please look into the commit after your comment. It has already been incorporated. Thanks!

hannesdiedrich: Sorry! I is was in the wrong timezone ;-)

integration_tests/steps/two_sided_market.py
hannesdiedrich: Well, you kind of did. 
The step function is called: "the energy rate for all the trades is the mean of max and min pv/storage rate" 
As you said, now it is checking, whether it is not violating the initial-final limit. These two are different. 
So, either you rename the function OR you look in which range the trade_rates are and readjust the RATE_LOWER_THRESHOLD and RATE_UPPER_THRESHOLD .

tox.ini
```diff
@@ -87,7 +87,7 @@ commands =
     pip install -e .
     # TODO: Workaround for https://github.com/ethereum/py-solc/issues/64
     pip install git+https://github.com/Jonasmpi/py-solc.git
-    pip install git+https://github.com/gridsingularity/d3a-interface.git
+    pip install /Users/muhammad.faizan/d3a-interface
```
spyrostz: Please revert that before merging, thanks!

integration_tests/steps/two_sided_market.py
faizan2590: Sorry, I couldn't find any `mean` in the diff. It is accumulating the rate at which pv_traded_energy and ess_traded_energy. And the asserting if all of these trades are in between hard-coded RATE_LOWER_THRESHOLD & RATE_UPPER_THRESHOLD.

BTW: its been renamed to `the energy rate for all trades are in between initial and final buying rate of storage`

src/d3a/models/strategy/storage.py
faizan2590: @spyrostz Exactly!

src/d3a/d3a_core/simulation.py
hannesdiedrich: Yes, and to inform the user how long the bids and offers need to propagate. If this is not needed any more, feel free to remove this function.

PR #508, opened 2019-09-09
Bug/d3 asim 1475

spyrostz:


integration_tests/jira_issues_tests.feature
```diff
@@ -75,4 +75,9 @@ Feature: Jira Issues Tests
...
      And d3a is installed
      When we run the simulation with setup file jira.d3asim_1139 and parameters [24, 60, 60, 0, 1]
-     Then there should be no unmatched loads
\ No newline at end of file
+     Then there should be no unmatched loads
+
+   Scenario: D3ASIM-1475, default_2off finishes successfully for two-sided pay as clear market
+     Given we have a scenario named jira/default_2off_d3asim_1475
+     And d3a is installed
+     When we run the simulation with setup file jira.default_2off_d3asim_1475 and parameters [24, 60, 60, 0, 1]
```
hannesdiedrich: Rather minor, but:
I guess if there is an exception, this step does not finalize, right?
Just for the readability of this integrationtest, can you please add a THEN step or rename the last WHEN step to a THEN?

spyrostz: If an exception is raised, the when step will crash, therefore the test will be marked as a failure. I can rename the when to a then if it makes the test more readable. Changing it.

PR #511, opened 2019-09-13
Feature/d3 asim 1391

spyrostz:


src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
```diff
@@ -0,0 +1,96 @@
+
+
+def update_incoming_bid_with_fee(source_bid, original_bid, tax_percentage=None):
```
hannesdiedrich: `tax_percentage` is not used here. Is it kind of a placeholder?

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -119,30 +129,35 @@ def accept_bid(self, bid: Bid, energy: float = None,
             changed_bid = Bid(
                 str(uuid.uuid4()), residual_price, residual_energy,
                 buyer, seller,
-                original_bid_price=energy_portion * orig_price
+                original_bid_price=(1 - energy_portion) * orig_price
```
hannesdiedrich: This change is not due to the new implementation of the grid fees, right? This was a bug, right?

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
```diff
@@ -0,0 +1,96 @@
...
+def calculate_trade_price_and_fees(trade_bid_info, tax_percentage):
+    original_bid_rate, bid_rate, original_offer_rate, \
+        offer_rate, trade_rate_source = trade_bid_info
+
+    revenue, fees, trade_price = calculate_fee_revenue_from_clearing_trade(
+        bid_rate, original_bid_rate,
+        offer_rate, original_offer_rate,
+        trade_rate_source, tax_percentage
+    )
+    return revenue, fees, trade_price
```
hannesdiedrich: Please correct me if I am wrong, but I dont see the revenue output used anywhere when executing this function.

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
```diff
@@ -0,0 +1,96 @@
...
+
+def propagate_offer_trade_original_info(trade_original_info, tax_percentage):
+    if trade_original_info is None:
+        return None
+    original_bid_rate, bid_rate, _, _, trade_rate_source = trade_original_info
+    bid_rate = bid_rate - original_bid_rate * tax_percentage
+    return [original_bid_rate, bid_rate, trade_rate_source]
+
+
+def propagate_bid_trade_original_info(trade_original_info, tax_percentage):
```
hannesdiedrich: There seams to be a type in this and the previous function names. This one should be named `propagate_offer_trade_original_info` and the former `propagate_bid_trade_original_info` right?

hannesdiedrich: EDIT: Not really, but please rename the variables inside the functions (bid to bid and offer to offers, currently its exchanged) Thanks!

src/d3a/models/strategy/__init__.py
```diff
@@ -204,14 +204,14 @@ def is_eligible_for_balancing_market(self):
 
     def accept_offer(self, market: Market, offer, *, buyer=None, energy=None,
                      already_tracked=False, trade_rate: float = None,
-                     original_trade_rate: float = None):
+                     trade_bid_info: float = None):
```
hannesdiedrich: Minor: How about renaming this parameter to `trade_offer_bid_info` ?

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
```diff
@@ -0,0 +1,96 @@
...
+    demand_side_tax = 0 if bid_original == 0 else 1 - bid_propagated / bid_original
+    supply_side_tax = 0 if offer_original == 0 else offer_propagated / offer_original - 1
+    total_tax_percentage = demand_side_tax + supply_side_tax
+    revenue = trade_rate_source / (1 + total_tax_percentage)
+    fee_n = revenue * tax_percentage_n
+    trade_price = revenue + revenue * supply_side_tax
+    return revenue, fee_n, trade_price
+
+
+def calculate_fee_revenue_from_propagated_trade(
```
hannesdiedrich: This function doesnt seam to be used anywhere in the code.

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
```diff
@@ -0,0 +1,96 @@
...
+    return source_offer_price + original_offer_price * tax_percentage
+
+
+# Protocol B
+def calculate_fee_revenue_from_clearing_trade(
+        bid_propagated, bid_original,
+        offer_propagated, offer_original,
+        trade_rate_source, tax_percentage_n
+):
+    demand_side_tax = 0 if bid_original == 0 else 1 - bid_propagated / bid_original
```
hannesdiedrich: I know, you tried to follow the naming convention introduced in the ticket, but still, I am voting to attach the type of the variable. So, instead of `bid_propagated` I ask to rename to `bid_propagated_rate` .
Or do you think this will clatter the code?

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -40,19 +44,19 @@ def _forward_bid(self, bid):
             return
         if self.owner.name == self.markets.target.area.name:
             return
-        if (bid.price * (1 - self.markets.target.transfer_fee_ratio)
-                - self.markets.target.transfer_fee_const * bid.energy) <= 0:
-            self.owner.log.debug("Bid is not forwarded because bid price lower "
-                                 "than transfer_fee_const")
-            return
+        # if (bid.price * (1 - self.markets.target.transfer_fee_ratio)
```
hannesdiedrich: Yeah, this shouldn't be needed any more, right? Are you keeping it just in case?

tests/test_inter_area_agent.py
```diff
@@ -399,7 +399,7 @@ def test_iaa_event_trade_buys_accepted_bid(iaa_double_sided):
     assert len(iaa_double_sided.lower_market.calls_energy_bids) == 1
 
     expected_price = 10 * (1 - iaa_double_sided.lower_market.transfer_fee_ratio)
-
+    print(expected_price)
```
hannesdiedrich: Please remove

hannesdiedrich: There are some other prints in the tests, too.

spyrostz: Thanks, removing all prints

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
spyrostz: You are right, this should be removed, deleting this

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
spyrostz: That was what I was intending to do. For now this is not useful, but I wanted to have a future-proof interface that would allow this module to be replaced with a custom fee calculation model. In this case, the tax_percentage would be needed. 
I did not finish implementing this though, since I was rushing to wrap up this story. But if you also agree, I think it would be nice to have a proper interface (using this https://docs.python.org/3/library/abc.html for instance) that would oblige the one who replaces the market fee to follow the same interface as here. Let me know if you agree, so that I can add it in this PR.

src/d3a/models/market/two_sided_pay_as_bid.py
spyrostz: You are right, this is a bug that has been lingering there for quite a while, and I wonder why did we not find it until now.

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
spyrostz: Same reason as before, revenue will not be needed now, but it will probably be needed as an extra statistic in the future, therefore I wanted to make it part of the interface.

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
spyrostz: I think there is a misunderstanding here. This function is propagating the offer info in case we have a bid trade. Looking at it now, the naming of this function could be better. What do you think about naming it like propagate_original_offer_info_on_bid_trade ? Same goes for propagate_offer_trade_original_info -> propagate_original_bid_info_on_offer_trade

src/d3a/models/strategy/__init__.py
spyrostz: Because this should contain only the bid info. The offer info have to be retrieved from this market.

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
spyrostz: You are right, this is not needed here. Removing

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
hannesdiedrich: Yes, please!

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -119,30 +129,35 @@ def accept_bid(self, bid: Bid, energy: float = None,
...
-            final_price = self._update_bid_fee_and_calculate_final_price(
-                energy, trade_rate, 1, orig_price
-            ) if already_tracked is False else energy * trade_rate
-
-            bid = bid._replace(price=energy * trade_rate)
+            revenue, fees, trade_price = calculate_trade_price_and_fees(
+                trade_offer_info, self.transfer_fee_ratio
+            )
+            self.market_fee += fees
+            final_price = energy * trade_price
```
hannesdiedrich: Sorry, here I ym confused over the naming again. I followed the calculation of the trade_price and see that is indeed a price, not a rate. Why are you multiplying with the energy here? Or am I wrong and it is really a rate and just needs to renamed?

spyrostz: Sorry, it really is a rate here. Will rename this

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
```diff
@@ -0,0 +1,96 @@
+
```
faizan2590: Wouldn't it be better to have all these methods of this module part of a class called something like GridFeeCalculator and then make them like an staticmethod?

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
```diff
@@ -0,0 +1,96 @@
...
+        offer_propagated, offer_original, tax_percentage_n
+):
+    calculate_fee_revenue_from_clearing_trade(
+        bid_propagated, bid_original,
+        offer_propagated, offer_original,
+        bid_original, tax_percentage_n
+    )
+
+
+def update_forwarded_bid_with_fee(source_bid, original_bid, tax_percentage_n):
```
faizan2590: To my understanding _tax_percentage_n_ is a ratio. If yes, then how about having a better name like _tax_ratio_?

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -102,21 +101,26 @@ def _match_offers_bids(self):
             original_bid_rate = bid.original_bid_price / bid.energy
             matched_rate = bid.price / bid.energy
 
+            trade_bid_info = [
+                original_bid_rate, bid.price/bid.energy,
+                offer.original_offer_price/offer.energy, offer.price/offer.energy,
+                original_bid_rate]
```
faizan2590: Correct me if Im wrong, shouldn't the last value of _trade_bid_info_ list be _matched_rate_ rather than _original_bid_rate_ ?

faizan2590:
I have left some comments, will start re-review later once addressed.

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
spyrostz: It will clutter it a bit, but it is good to have info about the unit of the variable. Fixing

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
spyrostz: To clone the naming of the documented formula from here https://gridsingularity.atlassian.net/wiki/spaces/D3A/pages/1225654275/Grid+Fees. If you would prefer it that way, I can change it to reflect the ratio unit.

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -102,21 +106,26 @@ def _match_offers_bids(self):
...
```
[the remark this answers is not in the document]
spyrostz: I was thinking a lot about this, since you also agree with that I'll change it. Fixing.

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
spyrostz: Not according to the documentation:

> determine trade_price to be original_bid (according PAB);

if market is clearing market (i.e. the match was made and cleared at this market / trade was created here): trade_price is original_bid. save this trade_price in newly created trade object. 

if market received a trade object from adjacent market: use trade_price provided in trade object

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
```diff
@@ -0,0 +1,85 @@
...
+    return revenue, fee_n, trade_price
+
+
+def update_forwarded_bid_with_fee(source_bid, original_bid, tax_percentage_n):
+    if source_bid is None:
+        return original_bid * (1 - tax_percentage_n)
+    return source_bid - original_bid * tax_percentage_n
+
+
+def update_forwarded_offer_with_fee(source_offer, original_offer):
```
hannesdiedrich: If this is just a placeholder to have the same interface as `update_forwarded_bid_with_fee`, please also include `tax_percentage_n` as parameter. Or is there another motivation in this function?

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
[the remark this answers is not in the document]
hannesdiedrich: I would like that, yes.

src/d3a/models/area/event_dispatcher.py
```diff
@@ -101,7 +101,8 @@ def _should_dispatch_to_strategies_appliances(self, event_type):
             return self.area.events.is_connected and self.area.events.is_enabled
 
     def event_listener(self, event_type: Union[MarketEvent, AreaEvent], **kwargs):
-        if event_type is AreaEvent.TICK:
+        if event_type is AreaEvent.TICK and \
+                self._should_dispatch_to_strategies_appliances(event_type):
```
hannesdiedrich: Also something weird, why didnt this fail before, right?

src/d3a/setup/area_events/pv_event.py
```diff
@@ -32,7 +32,7 @@ def get_setup(config):
                 children=[
                     Area('H2 PV', strategy=PVStrategy(1, 80),
                          appliance=PVAppliance(),
-                         event_list=[StrategyEvents(12, {'panel_count': 10,
+                         event_list=[StrategyEvents(12, {'panel_count': 5,
```
hannesdiedrich: Sorry, can not see from the diff, why this change was necessary. Can you help me please? Thanks!

integration_tests/steps/area_events.py
```diff
@@ -179,39 +178,34 @@ def both_sunny_profile(context):
...
 
 @then('House 1 PV follows the partially cloudy profile after 12:00')
 def house1_partially_cloudy(context):
-    slot_length = context.simulation.area.config.slot_length
     partial_cloud = _read_solar_profile(d3a_path + '/resources/Solar_Curve_W_partial.csv')
     house1 = next(filter(lambda x: x.name == "House 1", context.simulation.area.children))
 
     for time_slot, market in house1._markets.past_markets.items():
-        if time_slot.hour > 12:
+        if time_slot.hour < 12:
```
hannesdiedrich: Now the step description is not match any more, right?

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
spyrostz: You are correct, now that I created the abc it will be quite helpful to have it there even if it is not used. Adding it.

src/d3a/models/area/event_dispatcher.py
spyrostz: That surprised me too. I guess the reason for not finding it out until now is that this is useful only for the area events, and the integration tests for these were running only for one hour, therefore they were rendered useless. Now that the tests pass, the issue was revealed.

src/d3a/setup/area_events/pv_event.py
spyrostz: It was not that necessary. The reason for this change was to expect a value multiplied by *10 and not by *20 after 12:00 during the test. That way it was easier for me to see that the traded energy after 12:00 was related to the traded energy before 12:00

integration_tests/steps/area_events.py
spyrostz: This test was not running up until now, and we had not figured out that the step definition was incorrect :(

hannesdiedrich: Ah right! NOW it is correct. I confused it. Thanks!

src/d3a/models/market/grid_fees/base_model.py
```diff
@@ -0,0 +1,92 @@
+from d3a.models.market.grid_fees import GridFees
+
+
+class BaseModel(GridFees):
```
hannesdiedrich: IMO BaseModel is a bid generic. When calling one of the function one could easily get confused. How about renaming it to e.g. GridFees and rename the Baseclass in __init__.py to BaseClassGridFees ?

spyrostz: Agreed, fixing that. Thanks!

src/d3a/models/strategy/area_agents/two_sided_revenue_fee.py
faizan2590: Its okay to reflect the documentation but simultaneously could also lead to confusion in future. So, I would personally prefer to have a variable reflecting to what its value.

spyrostz: Cool, then I will rename it now. Thanks!

faizan2590: Great, there are some occurrences other then this method parameter too where it would be cool to replace `tax_percentage` with `tax_ratio`

spyrostz: Yup, it will be changed globally

src/d3a/models/market/grid_fees/base_model.py
faizan2590: Just for my better understanding would you re-write all of GridFees class method for PACGridFee calculation?

spyrostz: I do not plan to actually. PAC has mostly been dealt with in the context of this story, but not completely yet (it will be done in the context of [this story](https://gridsingularity.atlassian.net/browse/D3ASIM-1449) . However, from a quick look on Dirk's documentation, the fee calculation on the PAC market is almost identical to PAB. Therefore I do not think that this will be needed. Despite that, if the needed calculation is entirely different, then I will rewrite or override the class methods for PAC, yes. 
The main rational behind making this an interface though, is for creating completely different grid fee models. For instance, if a customer wants something like that, we can write another class that supports a combination of constant fees and ratios, with minimal changes on IAA or market class code.

PR #512, opened 2019-09-13
D3ASIM-1378: Added global max_panel_power_W setting.

hannesdiedrich:


src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -71,7 +71,8 @@ def start(scenario, settings, events):
             "market_count": settings.get('market_count', GlobalConfig.market_count),
             "cloud_coverage": settings.get('cloud_coverage', GlobalConfig.cloud_coverage),
             "pv_user_profile": settings.get('pv_user_profile', None),
-            "iaa_fee": settings.get('iaa_fee', GlobalConfig.iaa_fee)
+            "iaa_fee": settings.get('iaa_fee', GlobalConfig.iaa_fee),
+            "max_panel_power_W": settings.get('max_panel_power_W', GlobalConfig.iaa_fee)
```
faizan2590: Ummm,  correct me If Im wrong. Shouldn't it get the **MAX_PANEL_OUTPUT_W** as default ?

spyrostz: Nice catch. Fixing it

PR #514, opened 2019-09-19
D3ASIM-1497: Added support for sending zipped version of the d3a resu…

spyrostz:
…lts to d3a-web via Redis.


src/d3a/d3a_core/export.py
```diff
@@ -114,6 +115,17 @@ def _file_path(directory: dir, slug: str):
         file_name = ("%s.csv" % slug).replace(' ', '_')
         return directory.joinpath(file_name).as_posix()
 
+    def export_to_zip_file(self):
+        self.export()
```
hannesdiedrich: As discussed offline, what has to be decided is whether we want to export the htmls, too. They take a lot of space in the database and can be omitted IMO.

src/d3a/d3a_core/simulation.py
```diff
@@ -289,12 +296,11 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
                     sleep(abs(tick_lengths_s - realtime_tick_length))
 
             self._update_and_send_results()
-            if self.export_on_finish:
+            if self.export_on_finish or self.redis_connection.is_enabled():
```
hannesdiedrich: Can you please share the reason for this change? The final version of the files is already send to d3a-web by `_update_and_send_results`, right?

spyrostz: The reason for that is that the csv files need to be generated in addition to when the export_on_finish is set to true, also whenever there is a Redis connection and we want to send the zipped version of the files. In the old state of the if clause, the json files were not generated at all, because the parameter export_on_finish was set to false.

hannesdiedrich: Ah yes! You are right. Now I understand. Thanks!

src/d3a/d3a_core/export.py
spyrostz: Removed it already, discussion still pending.

src/d3a/d3a_core/redis_communication.py
```diff
@@ -101,17 +103,36 @@ def _handle_redis_job_metadata(self):
...
+        if not self.is_enabled():
+            return
+
+        fp = open(zip_results, 'rb')
+        zip_data = fp.read()
+        fp.close()
+
+        zip_results_key = ZIP_RESULTS_KEY + str(self._simulation_id)
+        # Write results to a separate Redis key
+        self.redis_db.set(zip_results_key, zip_data)
```
faizan2590: I guess it could be available to all channels. Right?

spyrostz: Sorry, same as the other PR, this is not a channel but a Redis key :)

faizan2590: Thanks for an offline explanation 👍

PR #516, opened 2019-09-25
D3ASIM-1173: Fixed output of energy bills csv file..

spyrostz:


src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -111,27 +114,34 @@ def energy_bills(self, area, past_market_types):
...
-                result[child.name]['market_fee'] = child_result.pop("market_fee")
                 result[child.name]['children'] = child_result
+                result[child.name]['market_fee'] = self.market_fees[child.name]
 
         return result
 
+    def _accumulate_market_fees(self, area, past_market_types):
+        if area.name not in self.market_fees:
+            self.market_fees[area.name] = 0.0
+        for market in self._get_past_markets_from_area(area, past_market_types):
```
hannesdiedrich: This only works, if ConstSettings.GeneralSettings.KEEP_PAST_MARKETS is True, right?

spyrostz: you are right, the self.market_fees dictionary needs to be reset if the KEEP_PAST_MARKETS is true. Fixing that

PR #517, opened 2019-09-26
Feature/d3 asim 1478

hannesdiedrich:


src/d3a/events/__init__.py
```diff
@@ -57,7 +57,7 @@ def event_listener(self, event_type: Union[AreaEvent, MarketEvent], **kwargs):
         self.log.trace("Dispatching event %s", event_type.name)
         self._event_mapping(event_type)(**kwargs)
 
-    def event_tick(self, *, area):
+    def event_tick(self):
```
spyrostz: I like it a lot now that the strategy events are not dependent on the area object, thanks!

src/d3a/models/area/__init__.py
```diff
@@ -128,7 +127,8 @@ def activate(self, bc=None):
             self.log.debug("No strategy. Using inter area agent.")
         self.log.debug('Activating area')
         self.active = True
-        self.dispatcher.broadcast_activate()
+        if len(self.children) != 0:
```
spyrostz: Is this still needed ? I think that broadcast_activate deals already with areas that do not have children.

src/d3a/models/area/__init__.py
```diff
@@ -184,7 +183,8 @@ def tick(self, is_root_area=False):
         self.events.update_events(self.now)
         if self.current_tick % self.config.ticks_per_slot == 0 and is_root_area:
             self._cycle_markets()
-        self.dispatcher.broadcast_tick(area=self)
+        if len(self.children) != 0:
```
spyrostz: Same as above

src/d3a/models/area/event_dispatcher.py
```diff
@@ -27,11 +32,15 @@
...
+from d3a.d3a_core.redis_communication import REDIS_URL
+
+log = getLogger(__name__)
+
+EVENT_DISPATCHING_VIA_REDIS = ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS
 
 
 class AreaDispatcher:
     def __init__(self, area):
-        self.listeners = []
```
spyrostz: Thank you!

src/d3a/models/area/event_dispatcher.py
```diff
@@ -205,3 +209,102 @@ def _delete_past_agents(self, area_agent_member):
                     agent.higher_market = None
                     agent.lower_market = None
                 del area_agent_member[pm]
+
+
+class RedisAreaDispatcher(AreaDispatcher):
+    def __init__(self, area, redis_comm):
+        super().__init__(area)
+        self.redis_comm = redis_comm
```
spyrostz: You can rename redis_comm to just redis IMO, this will make the code much more concise and equally readable.

src/d3a/models/area/event_dispatcher.py
```diff
@@ -205,3 +209,102 @@ def _delete_past_agents(self, area_agent_member):
...
+        self.subscribe_to_response_channel()
+        self.subscribe_to_area_event_channel()
+        self.str_area_events = [event.name.lower() for event in AreaEvent]
+
+    def subscribe_to_response_channel(self):
+        channel = f"{self.area.slug}/area_event_response"
+        self.redis_comm.sub_to_response(channel, self.response_callback)
```
spyrostz: Trivial, but I would prefer to name the method subscribe_to_response

src/d3a/models/area/event_dispatcher.py
```diff
@@ -205,3 +209,102 @@ def _delete_past_agents(self, area_agent_member):
...
+
+    def subscribe_to_area_event_channel(self):
+        channel = f"{self.area.slug}/area_event"
+        self.redis_comm.sub_to_area_event(channel, self.event_listener_redis)
```
spyrostz: Same as above, subscribe_to_area_event

tests/test_balancing_agent.py
```diff
@@ -91,6 +91,7 @@ def test_baa_event_trade(baa):
     fake_spot_market.set_time_slot(baa.lower_market.time_slot)
     baa.event_trade(trade=trade,
                     market_id=fake_spot_market.id)
+    print(baa.lower_market.__dict__)
```
spyrostz: Trivial, make sure to remove the print statement before you merge the PR. Thanks

src/d3a/models/area/__init__.py
hannesdiedrich: Not really, but it saves two function calls for each leaf area. Do you think this is worth it to keep it? In case of the tick event this could make a difference in runtime, right?

spyrostz: Not really, 2 function calls should have really negligible performance impact here. It is better IMO to go for readability in this one, and to not distract the reader of the method by additional checks.

PR #518, opened 2019-09-27
D3ASIM-1524: Adapting to d3a-inteface const parameters

faizan2590:


src/d3a/setup/two_sided_market/one_storage_5_pv_partial.py
```diff
@@ -40,9 +40,9 @@ def get_setup(config):
     ConstSettings.IAASettings.MARKET_TYPE = 2
     ConstSettings.PVSettings.FINAL_SELLING_RATE = 0
     ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE = 30
-    ConstSettings.StorageSettings.BUYING_RANGE[0] = 0
-    ConstSettings.StorageSettings.BUYING_RANGE[1] = 29.9
-    ConstSettings.StorageSettings.SELLING_RANGE[1] = 30
+    ConstSettings.StorageSettings.BUYING_RANGE._replace(min=0)
```
spyrostz: I would very much prefer to change these in one go and use something like ConstSettings.StorageSettings.BUYING_RANGE = Ranges(0, 29.9) 
Same goes for the SELLING_RANGE and other occurrences in the PR.

PR #519, opened 2019-09-27
Deleted traces of deprecated device parameters.

hannesdiedrich:


integration_tests/steps/integration_tests.py
```diff
@@ -899,6 +885,7 @@ def assert_trade_rates(context, market_name, trade_rate):
 
     for market in markets:
         for t in market.trades:
+            print(t.offer.price / t.offer.energy, float(trade_rate))
```
faizan2590: Please remove before merging

src/d3a/models/strategy/pv.py
```diff
@@ -32,25 +31,21 @@
...
-                 float=ConstSettings.PVSettings.FINAL_SELLING_RATE,
-                 fit_to_limit: bool=True,
+                 float = ConstSettings.PVSettings.FINAL_SELLING_RATE,
+                 fit_to_limit: bool = True,
                  update_interval=duration(minutes=ConstSettings.GeneralSettings.UPDATE_RATE),
                  energy_rate_decrease_per_update:
-                 float=ConstSettings.GeneralSettings.ENERGY_RATE_DECREASE_PER_UPDATE,
-                 max_panel_power_W: float=None):
+                 float = ConstSettings.GeneralSettings.ENERGY_RATE_DECREASE_PER_UPDATE,
+                 max_panel_power_W: float = None):
```
faizan2590: Sorry, I didn't get it. Are there any specific rules to add space in between `=` while defining default values?

src/d3a/setup/strategy_tests/home_hybrid_pv_ess_load.py
```diff
@@ -1,72 +0,0 @@
...
-
-                ]
-            ),
-            Area('Commercial Energy Producer',
-                 strategy=CommercialStrategy(energy_rate=15),
-                 appliance=SimpleAppliance()),
-        ],
-        config=config
-    )
-    return area
```
faizan2590: I would suggest to keep this setup file, as it depicts an smart home self balancing itself and then purchasing the residual electricity from utility.

src/d3a/setup/strategy_tests/pv_initial_pv_rate_option.py
```diff
@@ -58,7 +58,7 @@ def get_setup(config):
                          appliance=SwitchableAppliance()),
                     # The default value is 1, for historical average price
                     # Here a value of 2 is used, which is using the market maker rate
-                    Area('H1 PV', strategy=PVStrategy(4, 80, initial_rate_option=2),
+                    Area('H1 PV', strategy=PVStrategy(4),
```
faizan2590: I think this setup file could be deleted as we have deprecated initial_rate_option.

src/d3a/models/strategy/pv.py
hannesdiedrich: Yes. Pycharm also plains about that. It is not set as crucial in flake8. It is minor indeed.

src/d3a/setup/strategy_tests/home_hybrid_pv_ess_load.py
hannesdiedrich: Ok, but how should we tackle this? The setup file is not up to date (some parameter have to be removed) but i dont know the expected outcome. Do YOU know?

src/d3a/setup/strategy_tests/pv_initial_pv_rate_option.py
hannesdiedrich: Happy to remove it! Thanks!

src/d3a/models/strategy/pv.py
faizan2590: Now, that I noticed, Pycharm is also suggesting to have classes arguments to be lower case. For Eg: [https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/state.py#L69](url)

src/d3a/setup/strategy_tests/home_hybrid_pv_ess_load.py
faizan2590: Im not sure either, should there be a rule of deleting the setup file which are not part of any integration test?
May be @spyrostz has better opinion on it

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -115,8 +115,8 @@ class PVUserProfileStrategy(PVPredefinedStrategy):
     """
         Strategy responsible for reading a profile in the form of a dict of values.
     """
-    parameters = ('panel_count', 'risk', 'final_selling_rate', 'energy_rate_decrease_option',
-                  'energy_rate_decrease_per_update', 'power_profile', 'initial_selling_rate')
+    parameters = ('panel_count', 'final_selling_rate', 'energy_rate_decrease_per_update',
+                  'power_profile', 'initial_selling_rate')
```
faizan2590: @spyrostz Would there be any harm if simply this class variable is removed? Or is there any advantage of having it?

spyrostz: This needs to stay here, otherwise this would cause these parameters to not be read when sent by d3a-web. IIRC the area deserializer requires this 'parameter' class variable which contains serialized versions of the constructor arguments.

src/d3a/setup/1000_houses.py
```diff
@@ -37,7 +37,7 @@ def get_setup(config):
                  strategy=PVStrategy(6, 80),
                  appliance=PVAppliance()),
             Area(f'H{i} Storage',
-                 strategy=StorageStrategy(initial_capacity_kWh=0.6),
+                 strategy=StorageStrategy(),
```
faizan2590: Better replace `initial_capacity_kWh` with parameter `initial_soc` having reflected value

src/d3a/setup/strategy_tests/home_hybrid_pv_ess_load.py
spyrostz: I would suggest to keep this and document this better (maybe via an integration test). The reason for that lies on the comments of the setup file. The reason why the one that created this setup file has it here, is because the commercial producer offers arrive to the load sooner than the PVs. This needs to be described in Gherkin language in an integration test, so that at least we can understand why this setup file is there. Since this is not in the context of this story, you can reset the file and open a new enhancement story for adding an integration test for this file.

hannesdiedrich: Ok! Will include an integrationtest.

hannesdiedrich: @faizan2590 @spyrostz 
I am currently trying to reenact this setup file with the new parameter set. I dont understand the break_even set to (12, 17.01) This way the storage never sells to the load.
I still need a motivation for this setup file. Spyros wrote: "because the commercial producer offers arrive to the load sooner than the PVs" I dont know how to reenact this. Can you help me?
Faizan was writing that it should be an example of a "smart home self balancing itself" I can do that. 
I just realised that Spyros wrote that this should be done in a different story. IMO we should deal with it now or delete the setup file. Otherwise this story will never be dealt with.

hannesdiedrich: @faizan2590 @spyrostz Sorry, will skip this one and keep the old version of the setup file in here (commented out with a TODO message) IMO we have more pressing stuff to do. Thanks anyway!

spyrostz: @hannesdiedrich regarding this:
> Otherwise this story will never be dealt with.
every story that ends up in the backlog should be dealt with eventually. Even if it takes a long time, this can act as documentation so that we know what this setup file is about. 

The issue described in the comments is quite clear. According to that, Load buys energy only from the Commercial Producer and not from the PV, even though the PV is cheaper. Did you notice where does the load buy its energy from ? If it is the Commercial Producer, then this "bug/behavior" is still valid. The reason why this behavior appears is again easy to understand, because the Commercial Producer will place an offer before the PV places one. Due to the fact that offers propagate immediately, Commercial Producer's offer will reach the Load before PV even manages to place an offer. Therefore, the behavior can be explained. Hope this is clear.

hannesdiedrich: Thanks @spyrostz!
Yes, this makes sense, BUT. I can not reproduce with the current master version.
For making the setup file run, I have to delete all deprecated parameters and exchange break_even with final_selling_rate and final_buying_rate. When I run this setup file, the load buys energy from the PV perfectly and the storage charges only with the pv energy. No behaviour seen, like mentioned in the comment on the top. 
Makes sense, right? 
Shall I introduce an integrationtest that checks for this behaviour?
I still not think that this was the purpose of the original setup file (judging from the comment in it).

spyrostz: The more we discuss about it the more I tend to think that we should open a separate small story for that, that clarifies the expected behavior and writes an integration test for this. So, feel free to ignore this setup file in your PR, and I will open a story that deals with this.

src/d3a/setup/sam_config.py
```diff
@@ -38,8 +38,7 @@ def get_setup(config):
                          appliance=PVAppliance()),
                     Area('H1 Storage1',
                          strategy=StorageStrategy(battery_capacity_kWh=12 * 1.2,
-                                                  max_abs_battery_power_kW=12 * 1.2,
-                                                  initial_capacity_kWh=0.6 * 12 * 1.2),
+                                                  max_abs_battery_power_kW=12 * 1.2),
```
faizan2590: Please incorporate this comment [here](https://github.com/gridsingularity/d3a/pull/519#discussion_r329089443) in rest of PR too. Thanks!

hannesdiedrich: ok, will do!

src/d3a/setup/sam_config_const_load.py
```diff
@@ -38,8 +38,7 @@ def get_setup(config):
                          appliance=PVAppliance()),
                     Area('H1 Storage1',
                          strategy=StorageStrategy(battery_capacity_kWh=12 * 1.2,
-                                                  max_abs_battery_power_kW=12 * 1.2,
-                                                  initial_capacity_kWh=0.6 * 12 * 1.2),
+                                                  max_abs_battery_power_kW=12 * 1.2),
```
faizan2590: [It](https://github.com/gridsingularity/d3a/pull/519#discussion_r330047979) needs to adapted here as well.

tests/test_strategy_storage.py
```diff
@@ -447,8 +447,7 @@ def test_calculate_initial_sell_energy_rate_upper_bound(storage_strategy_test7_1
 
 @pytest.fixture()
 def storage_strategy_test7_2(area_test7):
-    s = StorageStrategy(initial_capacity_kWh=3.0, battery_capacity_kWh=3.01,
-                        max_abs_battery_power_kW=5.21, initial_rate_option=2, break_even=(16, 17))
+    s = StorageStrategy(battery_capacity_kWh=3.01, max_abs_battery_power_kW=5.21)
```
faizan2590: I just noticed, this pytest fixture hasn't been used, better remove it then.

PR #520, opened 2019-09-30
D3ASIM-1520: Added market maker strategy, with unit and integration t…

spyrostz:
…ests.


src/d3a/models/read_user_profile.py
```diff
@@ -322,3 +322,9 @@ def read_arbitrary_profile(profile_type: InputProfileTypes,
             return _calculate_energy_from_power_profile(filled_profile, GlobalConfig.slot_length)
         else:
             return filled_profile
+
+
+def read_and_convert_identity_profile_to_float(profile):
```
hannesdiedrich: Can you please tell me the reason for this additional function? Shouldn't `read_arbitrary_profile deal` (or rather its child _read_from_different_sources_todict ) deal with the conversion from json to float dict?

spyrostz: Yes, but it did it only for the SimulationConfig class, and specifically for its market_maker_rate member. I needed it though as a free function, in order to apply the same conversion on the market maker rate on the constructor of the MarketMakerStrategy class, and I did not want to duplicate the code. Hence, this calculation had to be refactored to a free function.

hannesdiedrich: Ok, thanks! Makes sense.

spyrostz:
@hannesdiedrich you are correct, we need to also add the communication with d3a-web in the context of this story. Adding this.

tests/test_strategy_commercial_producer.py
```diff
@@ -270,3 +270,11 @@ def testing_event_market_cycle(commercial_test3, area_test3):
...
     with pytest.raises(ValueError):
         CommercialStrategy(energy_rate=-1)
+
+
+def test_market_maker_strategy_constructor_modifies_global_market_maker_rate():
+    import d3a_interface.constants_limits
+    from d3a.models.strategy.market_maker_strategy import MarketMakerStrategy
+    MarketMakerStrategy(energy_rate=22)
+    assert all(v == 22
+               for v in d3a_interface.constants_limits.GlobalConfig.market_maker_rate.values())
```
faizan2590: Can you add a unittest, ensuring `MarketMakerStrategy` must never have any child device? Or route me there, if there is already any such kind of test.

spyrostz: Sorry, I do not get what you want to test. Children are held in the Area class, therefore all strategies do not have anything to do with child management. 
In addition to that, same goes for the rest of the strategies. I need to try it out first, but I do not think there is such a check yet on the production code. We have to rely on the user not configuring an area with children and adding a strategy at the same time. If you really think this is important I will add it, but I will open a small story that deals directly with that, because it is a totally different matter from the market maker.

faizan2590: Yes please make a separate story

PR #523, opened 2019-10-01
D3ASIM-1524: Moving strategy parameters' validation to d3a-interface.

faizan2590:


src/d3a/models/strategy/storage.py
```diff
@@ -74,13 +77,25 @@ def __init__(self, initial_soc: float = StorageSettings.MIN_ALLOWED_SOC,
                                  fit_to_limit=fit_to_limit,
                                  energy_rate_change_per_update=energy_rate_decrease_per_update,
                                  update_interval=update_interval)
+        for time_slot in generate_market_slot_list():
+            result = validate_storage_device(
```
hannesdiedrich: How about renaming the `result` variable to `is_valid` (every time you loop over device parameters, also in PV strategy)?

src/d3a/d3a_core/util.py
```diff
@@ -319,6 +319,10 @@ def update_nested_settings(class_object, class_name, settings_dict):
             elif isinstance(set_val, dict):
                 nested_class = getattr(class_object, set_var)
                 update_nested_settings(nested_class, set_var, settings_dict[class_name])
+            elif isinstance(set_val, list):
+                from d3a_interface.constants_limits import RangeLimit
```
hannesdiedrich: Please move this lazy import to the top. Thanks!

src/d3a/models/strategy/load_hours.py
```diff
@@ -57,6 +59,17 @@ def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None,
                                  fit_to_limit=fit_to_limit,
                                  energy_rate_change_per_update=energy_rate_increase_per_update,
                                  update_interval=update_interval)
+        for time_slot in generate_market_slot_list():
```
spyrostz: No need to change this now, since there are no performance concerns yet, but if we run multiyear profiles we might consider to optimise this method by splitting the validation in 2 parts: 
1. call validate_load_device outside of the for loop with all the arguments EXCEPT initial and final buying rate
2. inside the for loop, call validate_load_device ONLY with initial and final buying rate
That way we would avoid validating all the parameters as many times as the market slots, and only the 2 parameters that are really needed will be validated for every market slot. 

Nevertheless, I am writing this here as a mental note, since it is premature optimization there is no need to do anything else now, maybe add a TODO comment in order not to forget this.

PR #527, opened 2019-10-07
Feature/d3 asim 1431

faizan2590:


src/d3a/models/strategy/pv.py
```diff
@@ -163,7 +164,8 @@ def event_market_cycle(self):
                     offer_price,
                     self.state.available_energy_kWh[market.time_slot],
                     self.owner.name,
-                    original_offer_price=offer_price
+                    original_offer_price=offer_price,
+                    energy_origin=deepcopy(self.owner.name)
```
hannesdiedrich: You dont need the deepcopy here.

src/d3a/models/market/market_structures.py
```diff
@@ -20,20 +20,22 @@
 
 
 class Offer:
-    def __init__(self, id, price, energy, seller, original_offer_price=None):
+    def __init__(self, id, price, energy, seller, original_offer_price=None, energy_origin=None):
```
hannesdiedrich: As discussed briefly offline. How about renaming the `energy_origin` to `seller_origin` in the offer case and `buyer_origin` the bid case. This would also be consistent with the parameter naming in the Trade object. 
@spyrostz What do you think?

src/d3a/models/market/market_structures.py
```diff
@@ -20,20 +20,22 @@
...
         self.seller = seller
+        self.energy_origin = energy_origin
 
     def __repr__(self):
         return "<Offer('{s.id!s:.6s}', '{s.energy} kWh@{s.price}', '{s.seller} {rate}'>"\
             .format(s=self, rate=self.price / self.energy)
 
     def __str__(self):
-        return "{{{s.id!s:.6s}}} [{s.seller}]: {s.energy} kWh @ {s.price} @ {rate}"\
+        return "{{{s.id!s:.6s}}} [ORIGIN: {s.energy_origin}] " \
```
spyrostz: Can you please convert origin to lowercase? IMO uppercase words in logs are not ideal.

src/d3a/models/market/market_structures.py
spyrostz: @hannesdiedrich @faizan2590 I am not sure, but I think I would prefer Hannes's suggestion too. The reason for that is that the for the bid case this name is misleading - there is no energy produced for the bid, therefore energy_origin looks weird for this case. If it is desirable to keep the same name for both bid and offer, I would suggest renaming both to something like 'origin', 'source_area' for instance. If this is not a problem, then Hannes's suggestion looks good to me.

integration_tests/steps/two_sided_market.py
```diff
@@ -185,3 +185,17 @@ def cep_offer_residual_offer_rate(context):
             for id, offer in market.offers.items():
                 assert isclose((offer.price / offer.energy),
                                cep.strategy.energy_rate[market.time_slot])
+
+
+@then('Energy producer is {producer} & consumer is {consumer}')
+def energy_origin(context, producer, consumer):
+    trade_count = 0
+
+    for market in context.simulation.area.past_markets:
```
hannesdiedrich: Are you sure that you are looping over all past markets? Is KEEP_PAST_MARKETS set to True?

spyrostz: @hannesdiedrich it is set to true by default for the integration tests, in order to simplify checks. Have a look at environment.py -> before_scenario and before_all methods.

hannesdiedrich: Ah, Thanks!

PR #528, opened 2019-10-07
Bug/d3 asim 1531

hannesdiedrich:


src/d3a/setup/jira/d3asim_1531.py
```diff
@@ -0,0 +1,56 @@
+"""
```
hannesdiedrich: @spyrostz @faizan2590 I kept this here for now. I am not sure whether an integrationtest that checks for trades happening in the only root market is an overkill or not. What do you think?

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_engine.py
```diff
@@ -33,225 +29,10 @@ class TwoSidedPayAsClearEngine(TwoSidedPayAsBidEngine):
     def __init__(self, name: str, market_1, market_2, min_offer_age: int,
                  owner: "InterAreaAgent"):
         super().__init__(name, market_1, market_2, min_offer_age, owner)
-        self.forwarded_bids = {}  # type: Dict[str, BidInfo]
-        self.sorted_bids = []
-        self.sorted_offers = []
 
     def __repr__(self):
-        return "<TwoSidedPayAsClearEngine [{s.owner.name}] {s.name} " \
+        return "<TwoSidedPayAsClearE`gine [{s.owner.name}] {s.name} " \
```
spyrostz: Slight typo

src/d3a/setup/jira/d3asim_1531.py
spyrostz: I would actually suggest to add an integration test that trades indeed happen. The reason I mention that is because in this case you have only one market.

src/d3a/setup/jira/d3asim_1531.py
```diff
@@ -0,0 +1,56 @@
...
+                 appliance=SwitchableAppliance()),
+            Area('Load', strategy=LoadHoursStrategy(avg_power_W=100, hrs_per_day=9,
+                                                    hrs_of_day=list(range(8, 18))),
+                 appliance=SwitchableAppliance()),
+        ],
+        config=config
+    )
+    return area
+
+# This works:
```
spyrostz: Please delete this

tests/test_inter_area_agent.py
```diff
@@ -368,8 +366,8 @@ def iaa_double_sided():
     owner = FakeArea('owner')
     iaa = TwoSidedPayAsBidAgent(owner=owner, lower_market=lower_market,
                                 higher_market=higher_market)
-    iaa.engines[0]._match_offers_bids = lambda: None
-    iaa.engines[1]._match_offers_bids = lambda: None
+    iaa.engines[0].match_offers_bids = lambda: None
+    iaa.engines[1].match_offers_bids = lambda: None
```
faizan2590: Since the matching is performed in market and triggered from area, I guess this check is not necessary inside inter_area_engine and above two lines could be omitted.

integration_tests/steps/jira_issues_tests.py
```diff
@@ -258,3 +258,16 @@ def device_statistics(context):
                     assert len(list(output_dict[house][device]["min_" + stats_name])) == 48
                     assert len(list(output_dict[house][device]["max_" + stats_name])) == 48
     assert counter == 12
+
+
+@then('trades happen when the load seeks energy')
+def trades_happen(context):
+    from pendulum import today
+    from d3a.constants import TIME_ZONE
```
faizan2590: Can we get rid of this lazy imports?

faizan2590:
Have left some comments, will approve once addressed

tests/test_inter_area_agent.py
hannesdiedrich: Yes you are right, removing it.

src/d3a/models/market/two_sided_pay_as_clear.py
```diff
@@ -29,6 +34,11 @@ def __init__(self, time_slot=None, area=None,
                  notification_listener=None, readonly=False):
         super().__init__(time_slot, area, notification_listener, readonly)
         self.state = MarketClearingState()
+        self.sorted_bids = []
+        self.sorted_pac_offers = []
```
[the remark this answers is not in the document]
hannesdiedrich: Ah no, I just see that this is "heritage" code. Meaning this has been like this sins the PAC was implemented. Will try to change it anyway, ok @faizan2590 ?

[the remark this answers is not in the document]
hannesdiedrich: Sorry @faizan2590 : YES, it needs to be named like this (or at least not sorted_offers) because `sorted_offers` is a property by the Market base class. OK?

integration_tests/steps/jira_issues_tests.py
faizan2590: Can you tell me the reason of importing inside the step. As far as I see all the [constants](https://github.com/gridsingularity/d3a/blob/master/integration_tests/environment.py#L51) are reset after every scenario. Let me know if I have misunderstood.

src/d3a/models/market/two_sided_pay_as_clear.py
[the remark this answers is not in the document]
faizan2590: even better, you don't even have to resort the already sorted_offers. Does it makes sense?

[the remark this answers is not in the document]
hannesdiedrich: You are right! Changing it!

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -168,3 +173,58 @@ def accept_bid(self, bid: Bid, energy: float = None,
...
+
+    def _perform_pay_as_bid_matching(self):
+        # Pay as bid first
+        # There are 2 simplistic approaches to the problem
+        # 1. Match the cheapest offer with the most expensive bid. This will favor the sellers
+        # 2. Match the cheapest offer with the cheapest bid. This will favor the buyers,
+        #    since the most affordable offers will be allocated for the most aggressive buyers.
+
+        # Sorted bids in descending order
+        sorted_bids = list(reversed(sorted(
```
spyrostz: Move these sort operations to a separate method on the base class. That way they can be reused by both classes.

PR #530, opened 2019-10-08
Feature/d3 asim 1547

faizan2590:


src/d3a/setup/two_sided_market/one_pv_one_load.py
```diff
@@ -42,15 +44,18 @@ def get_setup(config):
                         hrs_per_day=6,
                         hrs_of_day=list(range(9, 15)),
                         initial_buying_rate=ConstSettings.LoadSettings.INITIAL_BUYING_RATE,
-                        final_buying_rate=ConstSettings.LoadSettings.FINAL_BUYING_RATE
+                        final_buying_rate=ConstSettings.LoadSettings.FINAL_BUYING_RATE,
+                        fit_to_limit=True, update_interval=duration(minutes=14)
```
spyrostz: I assume this 14 is here, because there is a test that validates that the price reaches final_buying_rate? In this case, won't 15 work ?

spyrostz: Answered offline, it is due to the assertion that the update_interval should never surpass slot length (15 mins). Thanks for clarifying, all is good.

PR #534, opened 2019-10-10
D3ASIM-1449: Converted 2 sided PAC market from aggregated to peer-to-…

spyrostz:
…peer market. Added unit tests. Fixed grid fees calculation.


src/d3a/models/market/two_sided_pay_as_clear.py
```diff
@@ -133,126 +133,99 @@ def match_offers_bids(self):
...
+        for bid in bid_list:
+            bid_energy = bid.energy
+            while bid_energy > FLOATING_POINT_TOLERANCE:
+                # Get the first offer from the list
+                offer = offer_list.pop(0)
+                # See if this offer has been matched with another bid beforehand.
+                # If it has, fetch the offer energy from the residual dict
+                # Otherwise, use offer energy as is.
+                offer_energy = residual_offer_energy.get(offer.id, offer.energy)
+                if offer_energy - bid_energy > FLOATING_POINT_TOLERANCE:
```
faizan2590:
```suggestion
                if offer_energy - bid_energy >= FLOATING_POINT_TOLERANCE:
```

spyrostz: That would be good, but I actually want the equality to be handled on the else block. The reason for that is that this if block is responsible for cases where the bid does not exhaust the offer. For the case of equality, the offer will be exhausted, therefore the else block should handle it.

src/d3a/models/market/two_sided_pay_as_clear.py
```diff
@@ -133,126 +133,99 @@ def match_offers_bids(self):
...
+                    # Bid energy completely covered by offer energy
+                    # Update the residual offer energy to take into account the matched offer
+                    residual_offer_energy[offer.id] = offer_energy - bid_energy
+                    # Place the offer at the front of the offer list to cover following bids
+                    # since the offer still has some energy left
+                    offer_list.insert(0, offer)
```
faizan2590: Shouldn't this ´offer´ be updated with residual energy? In case if it is and I missed it, please point me there.

spyrostz: The offer and bid objects are never updated here. Instead, a more verbose approach was followed, where the offer_energy is kept on a residual_offer_energy dict and is updated every time an offer gets accepted partially. You can see the updated energy in line 183

spyrostz:
@faizan2590 I am not sure if this answers exactly your question, but [here](https://github.com/gridsingularity/d3a/pull/534/files#diff-5645bb872ea2b06a6652ac8b54ab7c23R153). This line is responsible for sending the clearing rate as the trade rate to the propagated trade event.

PR #536, opened 2019-10-15
Bug/d3 asim 1565

faizan2590:


src/d3a/models/strategy/load_hours.py
```diff
@@ -58,8 +58,8 @@ def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None,
         if use_market_maker_rate:
             final_buying_rate = GlobalConfig.market_maker_rate
 
-        if isinstance(update_interval, int):
-            update_interval = duration(minutes=update_interval)
+        if isinstance(update_interval, (int, float)):
```
hannesdiedrich: Hmm, why are you allowing floats first but converting them to int anyways?

src/d3a/models/config.py
```diff
@@ -58,6 +58,9 @@ def __init__(self, sim_duration: duration, slot_length: duration, tick_length: d
         self.cloud_coverage = cloud_coverage
 
         self.read_pv_user_profile(pv_user_profile)
+        if isinstance(market_maker_rate, dict):
+            if "global" in market_maker_rate:
```
hannesdiedrich: IMO this shouldnt be dealt with in d3a, but only in d3a-web

src/d3a/models/strategy/load_hours.py
spyrostz: I assume this is because the 'minutes' field probably does not accept floats. My question is, where does the float come from? Update interval units in d3a-web is minutes-int, right? If yes, why do you even need to check it with the isinstance here. If not, where is it generated from? It would be much better to cast it to duration at the point where this variable is converted to float, than cast it in d3a.

faizan2590: You are right, I was being extra cautious. Reverting it

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -56,11 +56,15 @@ def __init__(
...
         :param final_selling_rate: Lower Threshold for PV offers
         :param cloud_coverage: cloud conditions. 0=sunny, 1=partially cloudy, 2=cloudy
-        :param fit_to_limit: Linear curve following initial_selling_rate & initial_selling_rate
+        :param fit_to_limit: Linear curve following initial_selling_rate & final_selling_rate
         :param update_interval: Interval after which PV will update its offer
         :param energy_rate_decrease_per_update: Slope of PV Offer change per update
-        :param max_panel_power_W:
+        :param max_panel_power_W: Peak power per panel
         """
+        if update_interval is None:
```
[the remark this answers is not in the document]
faizan2590: Sorry, I was playing around to know the reason of failed simulation launch. The actual reason was `update_interval` was not part of `parameters`. So, removing it now

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -35,8 +35,9 @@ class PVPredefinedStrategy(PVStrategy):
     """
         Strategy responsible for using one of the predefined PV profiles.
     """
-    parameters = ('panel_count', 'final_selling_rate', 'energy_rate_decrease_per_update',
-                  'initial_selling_rate', 'use_market_maker_rate')
+    parameters = ('panel_count', 'initial_selling_rate', 'final_selling_rate', 'cloud_coverage',
```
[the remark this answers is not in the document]
faizan2590: @spyrostz Sorry, I didn't get you. `cloud_coverage` is in strategy parameter.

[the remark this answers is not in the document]
spyrostz: Sorry, my bad, I did not see the constructor correctly. It is correct

PR #538, opened 2019-10-17
D3ASIM-1584: Fixed read of enery_rate in MMR strategy.

hannesdiedrich:


src/d3a/models/strategy/commercial_producer.py
```diff
@@ -26,8 +26,9 @@ class CommercialStrategy(BaseStrategy):
     parameters = ('energy_rate',)
 
     def __init__(self, energy_rate=None):
-        if energy_rate is not None and energy_rate < 0:
-            raise ValueError("Energy rate should be positive.")
+        # TODO: add this check to the device validator in d3a-interface:
+        # if energy_rate is not None and energy_rate < 0:
+        #     raise ValueError("Energy rate should be positive.")
```
faizan2590: Why not take the opportunity now rather than later? It should be trivial.

PR #550, opened 2019-10-21
D3ASIM-1605: Adding missing parameters to DefinedLoadStrategy.

faizan2590:


src/d3a/models/strategy/predefined_load.py
```diff
@@ -67,6 +75,7 @@ def event_activate(self):
         for each slot.
         :return: None
         """
+        self.bid_update.update_on_activate()
```
spyrostz: Why is it needed here?

faizan2590: This method is overwritten by base class method. Therefore it is needed here to explicitly call it here to calculate `energy_rate_change_per_update` if `fit_to_limit` is set to `True`.

PR #555, opened 2019-10-28
D3ASIM-1593: Adding device validation for CEP/MMR/InfinitePower.

faizan2590:


src/d3a/models/strategy/commercial_producer.py
```diff
@@ -20,15 +20,18 @@
...
     parameters = ('energy_rate',)
 
     def __init__(self, energy_rate=None):
-        # TODO: add this check to the device validator in d3a-interface:
-        # if energy_rate is not None and energy_rate < 0:
-        #     raise ValueError("Energy rate should be positive.")
+        try:
+            validate_commercial_producer(energy_rate=energy_rate)
+        except D3ADeviceException as e:
+            raise D3ADeviceException(str(e))
```
hannesdiedrich: Sorry, i just realised this when browsing through the code. Maybe you answered this question already, but: Why raising the same Exception tape again? Isn't it enough to call validate_commercial_producer and let this function raise the error?

PR #558, opened 2019-10-30
D3ASIM-1637: Added exhaustive matching on PAB clearing. Added integra…

spyrostz:
…tion test that validates that the setup file used in the UI runs without unmatched loads.


src/d3a/setup/jira/d3asim_1637.json
```diff
@@ -0,0 +1,287 @@
+{
```
hannesdiedrich: Nice!

src/d3a/setup/json_file.py
```diff
@@ -25,7 +25,7 @@ def get_setup(config):
         with open(setup_path, 'r') as area_file:
             area_str = area_file.read().replace('\n', '')
         recovered_area = area_from_string(area_str)
-        recovered_area.config = config
+        recovered_area._config = config
```
hannesdiedrich: This got through because nobody used this ever, right?

spyrostz: Exactly, this hasn't been used for a while now. I find it convenient for replicating errors in the production environment though, that's why I thought to reintroduce it.

PR #561, opened 2019-11-05
D3ASIM-1675: Infinite bus to buy energy at energy_buy_rate.

faizan2590:


src/d3a/models/strategy/infinite_bus.py
```diff
@@ -49,7 +49,7 @@ def buy_energy(self, market):
                 continue
             if (offer.price / offer.energy) <= self.energy_buy_rate[market.time_slot]:
                 try:
-                    self.accept_offer(market, offer)
+                    self.accept_offer(market, offer, buyer_origin=self.owner.name)
```
hannesdiedrich: @faizan2590 I think I need some context here. Tracking the buyer_origin doesnt seem to have an impact on the buying rate. Can we please discuss offline? Thanks!

spyrostz: Me too. I agree with the change, it should be here, I just do not understand how can this be the root cause of the problem.

faizan2590: @spyrostz The root cause of the bug is this [line](https://github.com/gridsingularity/d3a/pull/561/files#diff-a3d85ae365c0819646839ae7ef75bc23R68) where `InfiniteBus` was placing bid at sell_offered_rate

spyrostz: Thanks for the clarification!

PR #564, opened 2019-11-08
D3ASIM-1688: Removed area reference in market class.

hannesdiedrich:


src/d3a/models/area/__init__.py
```diff
@@ -190,6 +190,8 @@ def tick(self, is_root_area=False):
             self._cycle_markets()
         self.dispatcher.broadcast_tick()
         self.current_tick += 1
+        for market in self._markets.markets.values():
+            market.update_clock()
```
faizan2590: Can you help me understand why `current_tick` has to be re-incremented when its already been incremented two line above?

src/d3a/models/market/balancing.py
```diff
@@ -104,17 +105,15 @@ def accept_offer(self, offer_or_id: Union[str, BalancingOffer], buyer: str, *,
                                      key=lambda o: o.price / o.energy)
         try:
             if time is None:
-                time = self._now
+                time = self.now
```
faizan2590: Thanks for adapting the right convention.

src/d3a/models/area/__init__.py
hannesdiedrich: The self.current_tick in this function only increases the number in the area class, but we also want to increase the current_tick member of the market class.

hannesdiedrich:
@faizan2590 The reason for this change is to remove the circular dependency of market and area objects. 
Areas are responsible for creating markets. They are connected to them via the `self._markets member`. When creating the markets, the area objects is also parsed and the market holds a reference to the area in `self.area` this is not needed and is most likely making our life in the redis stories.
Does it make sense?

faizan2590:
> @faizan2590 The reason for this change is to remove the circular dependency of market and area objects.
> Areas are responsible for creating markets. They are connected to them via the `self._markets member`. When creating the markets, the area objects is also parsed and the market holds a reference to the area in `self.area` this is not needed and is most likely making our life in the redis stories.
> Does it make sense?

Thanks for the explanation, it makes sense to me now.

src/d3a/models/area/markets.py
```diff
@@ -68,7 +69,7 @@ def _market_rotation(self, current_time, markets, past_markets, area_agent):
                 else:
                     first = False
                 self.log.trace("Moving {t:%H:%M} {m} to past"
-                               .format(t=timeframe, m=past_markets[timeframe].area.name))
+                               .format(t=timeframe, m="past_markets[timeframe].area.name"))
```
spyrostz: You can circumvent the "" by using past_markets[timeframe].name (I assume the name of the market is the same as the name of the area, please correct me if my assumption is wrong).

src/d3a/models/market/__init__.py
```diff
@@ -63,7 +71,7 @@ def __init__(self, time_slot=None, area=None, notification_listener=None, readon
         self.accumulated_trade_energy = 0
         if notification_listener:
             self.notification_listeners.append(notification_listener)
-
+        self.current_tick = 0
```
spyrostz: What you could do, is instead of increasing the current_tick in parallel to the area, increase it only in the area and just assign it in the markets. That way there will be only one place where the current_tick is increased, and we are going to be saved by any potential errors that the current_tick of the market is out-of-sync with the current_tick of the area.

src/d3a/models/area/markets.py
hannesdiedrich: yes, sorry. This was temporary. I forgot to change this. Will do thanks!

src/d3a/models/market/__init__.py
hannesdiedrich: Ok, this means though, that we again have a connection between area and market. Or do you mean that i parse it in the constructor? But the we have to increase it in the market, too.

spyrostz: I meant to pass it as an argument of the update_clock method. In that case, this will be a simple setter (only sets the current_tick of the market to the current tick of the area)

hannesdiedrich: ahh, yes. makes sense. Cool, will implement it.

PR #565, opened 2019-11-11
Feature/d3 asim 915

faizan2590:


src/d3a/models/strategy/infinite_bus.py
```diff
@@ -11,7 +11,7 @@
 but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.
-
+x
```
hannesdiedrich: Probably only a typo. Please remove. Thanks!

src/d3a/d3a_core/simulation.py
```diff
@@ -149,6 +150,9 @@ def _init(self, slowdown, seed, paused, pause_after):
             log.info("Random seed: {}".format(random_seed))
 
         self.area = self.setup_module.get_setup(self.simulation_config)
+        if GlobalConfig.POWER_FLOW:
+            self.power_flow = PowerFlow(self.area)
+            self.power_flow.run_power_flow()
```
hannesdiedrich: I think you can write this in one line:
```
self.power_flow = PowerFlow(self.area).run_power_flow()
```

src/d3a/d3a_core/export.py
```diff
@@ -122,10 +122,13 @@ def delete_exported_files(self):
             os.remove(zip_file_with_ext)
         shutil.rmtree(str(self.directory))
 
-    def export(self, export_plots=True):
+    def export(self, export_plots=True, power_flow=None):
```
hannesdiedrich: Why not using False as default value here?

hannesdiedrich: I see now. It snot a flag. But why parsing as a parameter and not saving it as a member of the simulation class like all other stats as well?

src/d3a/models/powerflow/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+from d3a.models.strategy.pv import PVStrategy
+from d3a.models.strategy.finite_power_plant import FinitePowerPlant
+from d3a.models.strategy.infinite_bus import InfiniteBusStrategy
+
+
+class PowerFlow:
+    def __init__(self, root_area, root_voltage=400):
+        """
+        :param root_area: It contains tree structured hierarchical energy grid
+        :param root_voltage: It contains the voltage of the top most hierarchy
```
hannesdiedrich: Please remove "it contains" in both lines, thanks!

src/d3a/models/powerflow/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+        min_energy_mwh = area.strategy.state.min_allowed_soc_ratio * battery_capacity_mwh
+        return pp.create_storage(self.network, bus=area.parent.bus, p_mw=battery_capacity_mwh,
+                                 max_e_mwh=battery_capacity_mwh, min_e_mwh=min_energy_mwh,
+                                 soc_percent=soc_ratio, name=area.name)
+
+    def _add_line(self, source_area, target_area):
+        """
+        :param source_area: parent area of referenced area
+        :param target_area: child area of referenced area
+        :return: It will add an electrical line connecting parent and child area
```
hannesdiedrich: Minor, but the "return" member of the doc string is for explaining what is returned and not what the function does. Please move the "It will ..." sentences in this (and other) functions to the top of the doc string, Thanks!

src/d3a/models/powerflow/__init__.py
```diff
@@ -0,0 +1,132 @@
...
+    def _grid_forming(self, area):
+        """
+        :param area: Takes area object and convert it to specific energy device
+        :return: It will return pandapower compliant connected grid
+        """
+        if area.strategy is None:
+            area.bus = None
+            area.bus = self._create_bus(area, self.root_voltage)
+            if area.parent is not None:
+                area.line = None
```
hannesdiedrich: Do we really need these lines here? Can you please explain why. Thanks

src/d3a/d3a_core/export.py
faizan2590: Im not sure if I have exactly understood, `power_flow` is an instance variable that belongs to `Simulation` class and `export` is also an instance variable of `Simulation` class. So, I had to pass the instance variable `self.power_flow` of `Simulation` class as a parameter to a method that belongs to other instance. We can also discuss for better clarification.

src/d3a/models/powerflow/__init__.py
faizan2590: You are right, removing it.

src/d3a/d3a_core/simulation.py
faizan2590: `self.power_flow` is an instance variable that has to be used later to access other methods too, so I think your suggestion wont work.

src/d3a/models/powerflow/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+        :param peak_power_kw: Takes into account power rating of the generating device
+        :return: add a generating device to the specified area node
+        """
+        pass
+
+    @abstractmethod
+    def add_storage_device(self, area):
+        """
+        :param area: Takes area that is using Storage as a strategy
+        :return: add an Storage device to the specified area node
```
hannesdiedrich: Sorry to nitpick here. Can you please put `Add a storage device to the specified area node` (there is also a typo) in the first line of the docstring and either remove the return line completely or describe what is returned (in my opinion not feasible).

src/d3a/d3a_core/export.py
hannesdiedrich: Yes, you are right. Didnt understand it before. Thanks!

src/d3a/d3a_core/simulation.py
hannesdiedrich: Yes, you are right! Thanks

src/d3a/models/powerflow/__init__.py
hannesdiedrich: BTW: I mean this for all the docstrings in this class, sorry to write this so late.

src/d3a/models/powerflow/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+class PowerFlowBase(ABC):
+    def __init__(self, root_area, root_voltage):
+        """
+        :param root_area: contains tree structured hierarchical energy grid
+        :param root_voltage: contains the voltage of the top most hierarchy
+        """
+        self.root_voltage = root_voltage
+        self._grid_forming(root_area)
+
+    def _grid_forming(self, area):
```
spyrostz: Very minor, but I would prefer to name this function _form_grid_from_area, if you prefer it like it is now feel free to leave it as is.

src/d3a/models/powerflow/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+        elif isinstance(area.strategy, InfiniteBusStrategy):
+            area.ext_grid = self.add_external_grid(area)
+        elif isinstance(area.strategy, LoadHoursStrategy):
+            area.load_device = self.add_load_device(area, area.strategy.avg_power_W)
+        elif isinstance(area.strategy, StorageStrategy):
+            area.ess_device = self.add_storage_device(area)
+        elif isinstance(area.strategy, PVStrategy):
+            area.pv_device = \
+                self.add_generation_device(area, area.strategy.max_panel_power_W * 1e-3)
+        elif isinstance(area.strategy, FinitePowerPlant):
```
spyrostz: Great job, the only thing that I am missing here is the commercial producer. It can also be added here by creating a generation device with infinite energy (sys.maxsize).

src/d3a/models/powerflow/pandapower.py
```diff
@@ -0,0 +1,61 @@
...
+        super().__init__(root_area, root_voltage)
+
+    def create_bus(self, area, voltage):
+        return pp.create_bus(self.network, vn_kv=voltage, name=area.name)
+
+    def add_external_grid(self, area):
+        return pp.create_ext_grid(self.network, bus=area.parent.bus, vm_pu=1.0, name=area.name)
+
+    def add_load_device(self, area, avg_power_w):
+        return pp.create_load(self.network, bus=area.parent.bus, p_mw=avg_power_w * 1e-6,
```
spyrostz: It would be nice to have 1e-6 and 1e-3 as constants or small functions (namely convert_w_to_mw and convert_kw_to_mw respectively). That way we will be able to get rid of the hardcoded conversions, and to reuse them in other places too.

src/d3a/models/powerflow/pandapower.py
```diff
@@ -0,0 +1,61 @@
...
+                              name=area.name)
+
+    def add_generation_device(self, area, peak_power_kw):
+        return pp.create_gen(self.network, bus=area.parent.bus, p_mw=peak_power_kw * 1e-3,
+                             name=area.name)
+
+    def add_storage_device(self, area):
+        soc_ratio = area.strategy.initial_soc/100
```
spyrostz: Same as above, a small function like convert_percent_to_fraction would be great.

src/d3a/d3a_core/util.py
```diff
@@ -492,3 +492,19 @@ def str_to_pendulum(input_str: str):
...
+def convert_unit_to_kilo(unit):
+    return unit * 1e-3
+
+
+def convert_kilo_to_mega(unit_k):
+    return unit_k * 1e-3
+
+
+def convert_percent_to_ratio(factor):
+    return factor / 100
```
hannesdiedrich: Minor, but why "factor?" How about "unit_percent" or just "percent" 
Apart from that, nice!

PR #567, opened 2019-11-13
Feature/d3 asim 1480 buffer

spyrostz:


src/d3a/models/strategy/external_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+
+from d3a.models.strategy import BaseStrategy
+
+from redis import StrictRedis
+import json
+from d3a.d3a_core.redis_communication import REDIS_URL
+from d3a.models.market.market_structures import trade_bid_info_from_JSON_string
+
+
+class RedisExternalConnection:
```
spyrostz: Note, need to merge this class with the MarketRedisAPI class.

src/d3a/models/area/redis_external_connection.py
```diff
@@ -0,0 +1,57 @@
...
+class RedisExternalConnection:
+    def __init__(self, area):
+        self.area = area
+        self.redis_db = StrictRedis.from_url(REDIS_URL)
+        self.pubsub = self.redis_db.pubsub()
+        self.sub_to_area_event()
+        self.areas_to_register = []
+
+    def register_new_areas(self):
+        if not self.areas_to_register:
```
hannesdiedrich: First time I learn that an empty list is a Falso in logical space. Thanks!

src/d3a/models/market/one_sided.py
```diff
@@ -133,6 +133,7 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *, energy: in
         if energy is None:
             energy = offer.energy
 
+        # TODO: can this line be removed due to duplication with line 231 ???
```
hannesdiedrich: Yeah, I was meant to ask you that (or to try it out).

src/d3a/models/area/__init__.py
```diff
@@ -85,17 +88,22 @@ def __init__(self, name: str = None, children: List["Area"] = None,
...
         self.transfer_fee_const = transfer_fee_const
         self.display_type = "Area" if self.strategy is None else self.strategy.__class__.__name__
+        self.redis_ext_conn = RedisExternalConnection(self) \
+            if external_connection_available is True else None
 
     def set_events(self, event_list):
         self.events = Events(event_list, self)
 
     def activate(self, bc=None):
+        # TODO: Validate whether this is right:
```
hannesdiedrich: yeah, I moved this from the constructor to the activate method because we only need the markets then. Also: create markets inly for none-leaf areas.

src/d3a/models/area/__init__.py
```diff
@@ -190,8 +201,9 @@ def tick(self, is_root_area=False):
             self._cycle_markets()
         self.dispatcher.broadcast_tick()
         self.current_tick += 1
-        for market in self._markets.markets.values():
-            market.update_clock(self.current_tick)
+        if self._markets:
```
hannesdiedrich: Ahh, yes, Thanks!

src/d3a/models/area/event_dispatcher.py
```diff
@@ -152,10 +152,18 @@ def create_area_agents(self, is_spot_market, market):
                 return
             # Only connect an InterAreaAgent if we have a parent, a corresponding
             # timeframe market exists in the parent and we have no strategy
+            if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
```
hannesdiedrich: Oh, sorry, this was still in there. But we decided against just parsing the market_ids, because we said that the area and IAAS are always connected to the markets and we deal with the communication between market and area later (when deploying). Now I am confused why you let it in.

src/d3a/models/strategy/external_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+    def _serialize_offer_list(offer_list):
+        return json.dumps([offer.to_JSON_string() for offer in offer_list])
+
+    @staticmethod
+    def _serialize_offer_dict(offer_dict):
+        return json.dumps({k: v.to_JSON_string() for k, v in offer_dict.items()})
+
+    def _offer_lists(self, payload):
+        try:
+            return_data = self._serialize_offer_dict(self.area.next_market.offers)
```
hannesdiedrich: OK! Cool! Agreed, that the connected area can serve the offer list. 
But wouldn't it also be helpful in the future, if the market itselfs provide an endpoint that provides these information?
I mean we could even put it in the smart contract, can't we?

src/d3a/models/strategy/external_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @staticmethod
+    def _parse_payload(payload):
+        data_dict = json.loads(payload["data"])
+        if isinstance(data_dict, (str, bytes)):
+            data_dict = json.loads(data_dict)
+        if "trade_bid_info" in data_dict and data_dict["trade_bid_info"] is not None:
+            data_dict["trade_bid_info"] = \
+                trade_bid_info_from_JSON_string(data_dict["trade_bid_info"])
+        # if "offer_or_id" in data_dict and data_dict["offer_or_id"] is not None:
```
hannesdiedrich: We need this in order to parse the offer, don't we?

hannesdiedrich: ... so I wonder why you commented it out.

src/d3a/models/strategy/external_strategy.py
hannesdiedrich: Ahhhh, ok. So we agree. This is the answer to my comment here   https://github.com/gridsingularity/d3a/pull/567/files#r346193322  right?

src/d3a/models/strategy/__init__.py
```diff
@@ -202,19 +218,117 @@ def is_eligible_for_balancing_market(self):
                 ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET:
             return True
 
-    def accept_offer(self, market: Market, offer, *, buyer=None, energy=None,
+    # def sorted_offers_market(self, market_or_id):
```
hannesdiedrich: If you want to go the way of communicating between IAAs and markets via redis, you need this, because the Strategies "need" market.sorted_offers. But if the market is disconnected, it will fail.

src/d3a/d3a_core/simulation.py
```diff
@@ -380,7 +380,7 @@ def _handle_input(self, console, sleep: float = 0):
                     self.stop()
                 elif cmd == '+':
                     v = 5
-                    if self.slowdown <= 95:
+                    if self.slowdown <= 9995:
```
faizan2590: Can you tell the reason for this change?

src/d3a/models/area/__init__.py
```diff
@@ -62,7 +64,8 @@ def __init__(self, name: str = None, children: List["Area"] = None,
                  balancing_spot_trade_ratio=ConstSettings.BalancingSettings.SPOT_TRADE_RATIO,
                  event_list=[],
                  transfer_fee_pct: float = None,
-                 transfer_fee_const: float = None):
+                 transfer_fee_const: float = None,
+                 external_connection_available=False):
```
faizan2590: Can you add any integration/unittest showcasing how external agents could be connected using this flag?

src/d3a/models/strategy/external_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+                          "error_message": str(e)})
+
+
+class ExternalStrategy(BaseStrategy):
+    def __init__(self, area):
+        super().__init__()
+        self.redis = RedisExternalConnection(area)
+
+    def event_market_cycle(self):
+        pass
```
faizan2590: Shouldn't there also be `event_tick`  method here in case user wants to update his/her offer/bid?

src/d3a/models/area/event_dispatcher.py
```diff
@@ -175,6 +174,7 @@ def create_area_agents(self, is_spot_market, market):
             if market.time_slot in self.balancing_agents or \
                     market.time_slot not in self.area.parent._markets.balancing_markets:
                 return
+            # TODO: don't forget the balancing market!!!
```
hannesdiedrich: You can remove this comment.

src/d3a/models/market/__init__.py
```diff
@@ -15,25 +15,117 @@
...
+    offer_from_JSON_string
+from d3a_interface.constants_limits import ConstSettings
 from d3a_interface.constants_limits import GlobalConfig
 
 log = getLogger(__name__)
 
 TransferFees = namedtuple("TransferFees", ('transfer_fee_pct', 'transfer_fee_const'))
 
 
+# TODO: this will live somewhere else after Spyros' PR
```
hannesdiedrich: Plase also remove this comment, thanks!

src/d3a/models/strategy/area_agents/one_sided_agent.py
```diff
@@ -26,15 +26,20 @@ class OneSidedAgent(InterAreaAgent):
...
+                         lower_market=lower_market,
+                         min_offer_age=min_offer_age)
         self.name = make_iaa_name(owner)
 
     def usable_offer(self, offer):
         """Prevent IAAEngines from trading their counterpart's offers"""
         return all(offer.id not in engine.forwarded_offers.keys() for engine in self.engines)
 
     def _get_market_from_market_id(self, market_id):
+        if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
```
hannesdiedrich: This is not needed here as well (probably leading to an error)

src/d3a/models/strategy/external_strategy.py
hannesdiedrich: Are you planning to move this as mentioned or only after tomorrow?

src/d3a/models/strategy/external_strategy.py
hannesdiedrich: Ok, understood, it is in the market API. Please remove it here. Thanks!

src/d3a/models/strategy/external_strategy.py
spyrostz: Actually, let me do it now, better to have it in the first place than forgetting it later.

src/d3a/models/strategy/__init__.py
```diff
@@ -161,9 +163,21 @@ def __init__(self):
...
         self.enabled = True
+        if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
+            self.redis = RedisMarketCommunicator()
+        self.trade_buffer = None
+        self.offer_buffer = None
 
     parameters = None
 
+    def _send_events_to_market(self, event_type, market_id, data, callback):
+        # TODO: This could be moved to the RedisMarketCommunicator
```
hannesdiedrich: Please remove this comment. Thanks!

src/d3a/models/strategy/__init__.py
```diff
@@ -161,9 +163,21 @@ def __init__(self):
         super(BaseStrategy, self).__init__()
         self.offers = Offers(self)
         self.enabled = True
+        if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
+            self.redis = RedisMarketCommunicator()
+        self.trade_buffer = None
+        self.offer_buffer = None
```
hannesdiedrich: Please indent (didnt think about this back then)

src/d3a/models/area/__init__.py
spyrostz: No problem for me, everything works good even with this change.

src/d3a/models/strategy/external_strategy.py
spyrostz: As discussed offline, let's leave this in the markets for now, and we can think about refactoring it in the future, if performance impact is not that severe.

src/d3a/models/strategy/external_strategy.py
spyrostz: As discussed offline, external API does not accept offer objects but offer_ids.

src/d3a/models/strategy/__init__.py
spyrostz: Resolved offline

src/d3a/models/strategy/external_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+        return json.dumps([offer.to_JSON_string() for offer in offer_list])
+
+    @staticmethod
+    def _serialize_offer_dict(offer_dict):
+        return json.dumps({k: v.to_JSON_string() for k, v in offer_dict.items()})
+
+    def _offer_lists(self, payload):
+        try:
+            return_data = self._serialize_offer_dict(self.area.next_market.offers)
+            print(return_data)
```
faizan2590: Please remove before merging.

src/d3a/d3a_core/simulation.py
spyrostz: The reason for this change is because the max slowdown is 10000, not 100. This had the effect that I could not slow down the runtime of the simulation enough. If you agree though, I can change the max slowdown to 100 globally, so that it can be associated with a ratio of delay (0-no delay, 100-realtime).

src/d3a/models/area/__init__.py
spyrostz: The tricky part is that real integration tests actually require Redis for that. I could add a small script though that exhibits the behavior we want, though I do not know how helpful it will be if  it does not run on Travis. Let me try adding unit tests, maybe these will be enough.

src/d3a/models/strategy/external_strategy.py
spyrostz: Not really, because the user does not care about ticks in this case. The user can place offers to the market even in-between ticks, since the redis connection is running on a thread. Therefore, the user does not need to get notified when to place offers, one can put the offers at will, without waiting for a tick.

src/d3a/models/area/event_dispatcher.py
spyrostz: Removing it, thanks!

src/d3a/models/market/__init__.py
spyrostz: Removing it, thanks

src/d3a/models/strategy/area_agents/one_sided_agent.py
spyrostz: Thanks, removing this

src/d3a/setup/redis_communication_default_2a.py
```diff
@@ -47,17 +49,16 @@ def get_setup(config):
...
-                                                                           range(12, 16)),
+                                                                           range(0, 24)),
                                                                        final_buying_rate=35),
                          appliance=SwitchableAppliance()),
                     Area('H2 PV', strategy=PVStrategy(4, 80),
                          appliance=PVAppliance()),
 
                 ],
                 transfer_fee_pct=0, transfer_fee_const=0,
-
```
faizan2590: Any particular reason to change parameters in this setup file?

hannesdiedrich: No, Just to have loads trying to buy at the first market slot, not on midday.

src/d3a/models/strategy/__init__.py
spyrostz: Adding it here.

src/d3a/models/strategy/external_strategy.py
spyrostz: Removing it, thanks

src/d3a/setup/redis_communication_default_2a.py
spyrostz: @hannesdiedrich any reason for that? Otherwise I will revert this setup file to its original version.

src/d3a/models/strategy/external_strategy.py
```diff
@@ -0,0 +1,125 @@
...
+        accept_offer_channel = f"{self.area.parent.slug}/{self.area.slug}/accept_offer"
+        list_offers_channel = f"{self.area.parent.slug}/{self.area.slug}/offers"
+
+        self.pubsub.subscribe(**{
+            offer_channel: self._offer,
+            delete_offer_channel: self._delete_offer,
+            accept_offer_channel: self._accept_offer,
+            list_offers_channel: self._offer_lists
+        })
+        self.pubsub.run_in_thread(daemon=True)
```
faizan2590: Are we only allowing external agents to interact with one sided market?

src/d3a/setup/redis_communication_default_2a.py
hannesdiedrich: @spyrostz please see my comment above.

src/d3a/models/strategy/external_strategy.py
faizan2590: Are we only allowing external agent to interact with single sided market?

spyrostz: This is an initial implementation for the one-sided market indeed. You are right though, we can add the corresponding 2-sided market changes here. Adding these, thanks.

src/d3a/setup/redis_communication_default_2a.py
spyrostz: @hannesdiedrich sorry, just saw your comment. I will revert this file to its original state if it is fine for you then.

src/d3a/models/area/__init__.py
faizan2590: That would be great, if its too much of a trouble, could be skipped for later.

src/d3a/models/area/redis_external_connection.py
```diff
@@ -0,0 +1,57 @@
...
+            return
+
+        for new_area in self.areas_to_register:
+            area_object = self.area.__class__(name=new_area)
```
faizan2590: I guess here you are instantiating the user defined area class? Correct me if Im wrong.

spyrostz: Exactly. The reason I did not use the 'normal' way of constructing an are object, is circular references, if I imported Area class in this file this would result in circular import references.

faizan2590: Wouldn'T it be a good approach to implement `abc` module to define minimum requirements for user defined classes?

spyrostz: Actually for the strategies, there are none. The 'interface' is defined in the BaseStrategy class, which provides default implementations for the necessary methods for every strategy. Therefore, even if a new strategy just inherits from the BaseStrategy class, it is going to have the minimum functionality of it. That's why the ExternalStrategy class is so small, because the BaseStrategy provides default implementations for every required method.

src/d3a/models/area/__init__.py
spyrostz: @hannesdiedrich on second thought, many unit tests fail just because of that. I will revert this change if it is not necessary with a TODO comment to refactor it in isolation.

hannesdiedrich: OK! Agreed!

src/d3a/models/area/redis_external_connection.py
```diff
@@ -0,0 +1,56 @@
+from redis import StrictRedis
+import json
+from d3a.d3a_core.redis_communication import REDIS_URL
+from d3a.models.strategy.external_strategy import ExternalStrategy
+
+
+class RedisExternalConnection:
+    def __init__(self, area):
```
faizan2590: One more clarification, shouldn't there needs to have an implementation to unsubscribe external connection like we have for [RedisMarketCommunicator](https://github.com/gridsingularity/d3a/pull/567/files#diff-326ea26bc50c7cb2ae739e6ac2e2774bR65) ?

tests/test_strategy_base.py
```diff
@@ -184,9 +187,10 @@ def base():
 
 def test_accept_offer(base, offer_to_accept):
     market = FakeMarket(raises=False)
+    print(offer_to_accept)
```
hannesdiedrich: please remove

spyrostz: Just did, thanks!

src/d3a/models/area/redis_external_connection.py
spyrostz: That's a good question. In reality, we probably need an API endpoint that unregisters the user from the simulation (even though we do not need it immediately, I am almost sure that this is going to be a requirement for the future). I will try to expose another channel on the area redis subscribe class, so that a user can not only register to a simulation, but unregister from that too. Thanks!

faizan2590: On second thought, its not an immediate requirement, so better create a separate enhancement story else this PR may unnecessary be delay to be resolved :-)

src/d3a/models/strategy/external_strategy.py
```diff
@@ -0,0 +1,150 @@
...
+    def _offer_channel(self):
+        return f"{self.area.parent.slug}/{self.area.slug}/offer"
+
+    @property
+    def _delete_offer_channel(self):
+        return f"{self.area.parent.slug}/{self.area.slug}/delete_offer"
+
+    @property
+    def _accept_offer_channel(self):
+        return f"{self.area.parent.slug}/{self.area.slug}/accept_offer"
```
faizan2590: May be offline if you can help me understand the reason to overwrite these method of base class __MarketRedisApi__

src/d3a/models/market/__init__.py
```diff
@@ -73,6 +230,9 @@ def __init__(self, time_slot=None, bc=None, notification_listener=None, readonly
             self.notification_listeners.append(notification_listener)
         self.current_tick = 0
         self.device_registry = DeviceRegistry.REGISTRY
+        if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
+            self.redis_api = MarketRedisApi(self)
```
faizan2590: I couldn't the usage of __self.redis_api__ instance variable. May be I have missed something. .

src/d3a/models/strategy/external_strategy.py
spyrostz: Simple, the external connection needs to expose different channels than the simulation itself (eg. from an Area to its market). We can discuss this offline too if you want

src/d3a/models/market/__init__.py
spyrostz: You are right, it is not used anywhere in the code. It needs to be there though for the subscription. In order for the subscribe thread to keep on running, there has to be a reference to this class, because if all references of this object get lost, the object gets destroyed, and the subscribe thread terminates accordingly. Even though run_in_thread(daemon=True) does its best to avoid this situation, it is not guaranteed that it will not get deleted when it goes out of scope, therefore it is good for this to stay as a class member.

faizan2590: Thanks for the explanation. Makes sense to me now.

PR #570, opened 2019-11-19
D3ASIM-1701: Only import pandapower if interpreter is not pypy.

hannesdiedrich:


requirements/pandapower.txt
```diff
@@ -0,0 +1,70 @@
+#
```
spyrostz: I would say to start off slow, and not include all these requirements from the beginning. Despite that, this does not affect the functionality of the PR.

PR #572, opened 2019-11-19
Feature/d3 asim 1427: Self_sufficiency KPI

faizan2590:


src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -370,50 +371,83 @@ def _accumulate_energy(self, area, child, is_cep: bool):
...
+
+    def _accumulate_self_consumption(self, trade):
+        if trade.seller_origin in self.producer_list and trade.buyer_origin in self.consumer_list:
+            self.total_self_consumption += trade.offer.energy * 1000
+
+    def _accumulate_self_consumption_buffer(self, trade):
+        if trade.seller_origin in self.producer_list and trade.buyer_origin in self.ess_list:
+            self.self_consumption_buffer += trade.offer.energy * 1000
+
+    def _dissipate_self_consumption_buffer(self, trade):
```
hannesdiedrich: I think I need an offline walkthrough through this PR. I am not sure what the role of  _accumulate_self_consumption_buffer is. It seams that this is the self consumption of all storages in an area!?
Let me know when you are free to meet. Thanks!

src/d3a/models/strategy/load_hours.py
```diff
@@ -244,7 +244,9 @@ def event_market_cycle(self):
             if self.hrs_per_day[current_day] <= FLOATING_POINT_TOLERANCE:
                 self.energy_requirement_Wh[market.time_slot] = 0.0
                 self.state.desired_energy_Wh[market.time_slot] = 0.0
-
+                for trade in market.trades:
+                    if trade.buyer_origin == self.owner.name:
+                        self.state.desired_energy_Wh[market.time_slot] += trade.offer.energy * 1000
```
hannesdiedrich: You are increasing the desired energy when a trade had happened? Shouldn't this value be decreased then?

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -370,50 +371,83 @@ def _accumulate_energy(self, area, child, is_cep: bool):
                     if trade.buyer is child.name:
```
hannesdiedrich: In this function an above, a distinction between "nromal" trades and trades including ceps is done. AFAIS, this is not needed any more. So, IMO you can (and should) remove the self._cep_energy and self._cep_device members. Or did i miss something?

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -370,50 +371,83 @@ def _accumulate_energy(self, area, child, is_cep: bool):
...
+                    self.total_energy_demanded += energy
+            elif _is_prosumer_node(child):
+                self.ess_list.append(child.name)
+
+            if child.children:
+                self._accumulate_devices(child)
+
+    def _accumulate_self_consumption(self, trade):
+        if trade.seller_origin in self.producer_list and trade.buyer_origin in self.consumer_list:
+            self.total_self_consumption += trade.offer.energy * 1000
```
hannesdiedrich: Can you please explain, why you are converting to Wh here (and everywhere)?
Is it because of the loads' desired_energy_Wh ? If true, please keep it, but please append a unit to all energy variables (e.g. `_Wh` Thanks!

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -370,50 +371,83 @@ def _accumulate_energy(self, area, child, is_cep: bool):
...
+
+    def _dissipate_self_consumption_buffer(self, trade):
+        if trade.seller_origin in self.ess_list and \
+                (trade.buyer_origin not in self.consumer_list or
+                 trade.buyer_origin not in self.ess_list) and \
+                self.self_consumption_buffer > 0:
+            if (self.self_consumption_buffer - trade.offer.energy * 1000) > 0:
+                self.self_consumption_buffer -= trade.offer.energy * 1000
+            else:
+                self.self_consumption_buffer = 0
```
hannesdiedrich: AFAIU, this branch should never happen, right? Or are you just catching the values close to FLOATING_POINT_TOLERANCE ?

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -370,50 +371,83 @@ def _accumulate_energy(self, area, child, is_cep: bool):
...
+    def _area_self_sufficiency(self, area):
+        self.producer_list = list()
+        self.consumer_list = list()
+        self.consumer_area_list = list()
+        self.ess_list = list()
+        self.total_energy_demanded = 0
+        self.total_self_consumption = 0
+        self.self_consumption_buffer = 0
+
+        self._accumulate_devices(area)
```
hannesdiedrich: AFAIU, you dont need to get all the children for every market slot, right? IMO you can derive the lists in the constructor.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -370,50 +371,83 @@ def _accumulate_energy(self, area, child, is_cep: bool):
...
-            self_sufficiency = traded_from_pv / total_energy_bought
-        else:
-            self_sufficiency = 0
+        """
+        In case when the ess' SOC at start of simulation was more than the end of simulation
+        due to which it will end-up discharging more than its initial state.
+        """
+        if (self.total_self_consumption - self.self_consumption_buffer) < 0:
+            self.self_sufficiency[area.name] = 1.0
+            return {"self_sufficiency": 1.0}
```
hannesdiedrich: Now (after thiking again after the offline discussion) I get it, if  `(self.total_self_consumption - self.self_consumption_buffer) < 0`, you are setting self.total_self_consumption to 0 and consequently, self_sufficiency is 1, per definition. Makes sense. Please keep it. Thanks

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -370,50 +371,83 @@ def _accumulate_energy(self, area, child, is_cep: bool):
...
-            self_consumption_within_house = traded_from_pv / trade_by_pv
-        else:
-            self_consumption_within_house = 0
-        if total_energy_bought != 0:
-            self_sufficiency = traded_from_pv / total_energy_bought
-        else:
-            self_sufficiency = 0
+        """
+        In case when the ess' SOC at start of simulation was more than the end of simulation
+        due to which it will end-up discharging more than its initial state.
```
hannesdiedrich: Sorry, but this comment doesnt fit to the if block after (IMO).
If self_consumption_buffer > total_self_consumption, the Storage was charging, right?

src/d3a/d3a_core/sim_results/file_export_endpoints.py
hannesdiedrich: Please ignore this comment. I just realised that the self sufficiency is not populated for each market but for the whole simulation.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
faizan2590: You should ignore it, as its only used in `_accumulated_trade_energy` method which is not used anywhere anymore and I will remove it to avoid confusion.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
faizan2590: If the the ESS sold energy to any external load more than what was recorded in `self_consumption_buffer`, in such a case it means there is no energy in ess which was originated from internal producer and has to be set to zero.

hannesdiedrich: Ah, yes, makes sense! Thanks!

src/d3a/models/strategy/load_hours.py
faizan2590: Removing it, as it no more needed for total_energy_demanded.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
faizan2590: Improved the formula, have a look at later commit [at the lines below](https://github.com/gridsingularity/d3a/pull/572/commits/74d1bb3e0b76bd23583e1355d94c9dda363d731b#diff-6a3a50e2566925be315bd7728a1e6629R373).

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -344,76 +342,83 @@ def _specific_row(self, slot, market):
...
-        if total_energy_bought != 0:
-            self_sufficiency = traded_from_pv / total_energy_bought
-        else:
-            self_sufficiency = 0
-
-        return {"self_consumption_within_house": self_consumption_within_house,
-                "self_sufficiency": self_sufficiency}
+                    self.total_self_consumption_wh += self.self_consumption_buffer_wh
+                    self.self_consumption_buffer_wh = 0
+            elif (trade.buyer_origin not in self.consumer_list or
```
hannesdiedrich: This elif statement isnt captioning a different case than the if branch. 
What I mean is: if `trade.buyer_origin in self.consumer_list and self.self_consumption_buffer_wh > 0` is True, both branched can be entered. I think the first one will be taken, but i think you can simplify this function with less if then else statements.

integration_tests/steps/kpi.py
```diff
@@ -0,0 +1,38 @@
...
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+# from behave import then
+# from math import isclose
+# from pendulum import duration
+# from d3a_interface.constants_limits import ConstSettings
+# from d3a.constants import DEFAULT_PRECISION
```
hannesdiedrich: If not needed, please remove. Thanks

integration_tests/kpi.feature
```diff
@@ -0,0 +1,26 @@
...
+     And d3a is installed
+     When we run the d3a simulation on console with kpi.house_partially_self_sufficient for 24 hrs
+     Then the export functionality of kpi of {'Grid': 1.0, 'House 1': 0.34925} are correctly reported
+
+
+  Scenario: House fully self_sufficiency with storage
+     Given we have a scenario named kpi.house_self_sufficient_with_ess
+     And d3a is installed
+     When we run the d3a simulation on console with kpi.house_self_sufficient_with_ess for 24 hrs
+     Then the export functionality of kpi of {'Grid': 1.0, 'House 1': 1.0} are correctly reported
```
hannesdiedrich: Nit picking, but only the SS of the House is changing, so I suggest to only parse this value.

faizan2590: Yes, but in future there could be more integration tests checking `self_sufficiency` of multiple areas al together.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
faizan2590: Umm, I don't think so, first `if` checks the `trade.buyer_origin in self.consumer_list` and the later checks if `trade.buyer_origin not in self.consumer_list`

hannesdiedrich: Ah yes! Thanks! This was a test ;-) . Haha, joking!

hannesdiedrich: Buut, `(trade.buyer_origin not in self.consumer_list or trade.buyer_origin not in self.ess_list)` is True for all devices right?

faizan2590: And thats why this statement is tested later not at the first if clause

integration_tests/two_sided_pay_as_clear_market.feature
```diff
@@ -24,11 +24,11 @@ Feature: Two sided pay_as_clear market tests
   Scenario: Test offers, bids and market-clearing-rate are exported
      Given we have a scenario named two_sided_pay_as_clear/default_2a
      And d3a is installed
-     When we run the d3a simulation on console with two_sided_pay_as_clear.default_2a
+     When we run the d3a simulation on console with two_sided_pay_as_clear.default_2a for 24 hrs
```
hannesdiedrich: Minor, but how about decreasing the simulation runtime here as well to save time in the tests?

integration_tests/steps/kpi.py
```diff
@@ -0,0 +1,38 @@
...
+from behave import then
+
+
+@then('the export functionality of kpi of {areas} are correctly reported')
+def test_export_of_kpi_result(context, areas):
+    sim_data_csv = glob.glob(os.path.join(context.export_path, "*",
+                                          "aggregated_results", "KPI.json"))
+    with open(sim_data_csv[0], "r") as sf:
+        kpi_data = json.load(sf)
+    areas = ast.literal_eval(areas)
```
hannesdiedrich: It should work without this additional parsing, schouldn't it?

integration_tests/jira_issues_tests.feature
```diff
@@ -61,7 +61,7 @@ Feature: Jira Issues Tests
   Scenario: D3ASIM-891, bids are reported correctly in csv export files
      Given we have a scenario named two_sided_pay_as_clear.default_2a
      And d3a is installed
-     When we run the d3a simulation on console with two_sided_pay_as_clear.default_2a
+     When we run the d3a simulation on console with two_sided_pay_as_clear.default_2a for 2 hrs
```
spyrostz: How come you changed the duration here ? To decrease the integration tests' runtime ?

faizan2590: If you have look at this [line](https://github.com/gridsingularity/d3a/pull/572/files#diff-0e919c3f7c79c6df9954c6557f70efa5L261), It was already running for 2 hrs. Just to use the same step for `self_sufficiency` test to be tested for one day, I made simulation runtime as a variable and set `2 hrs` for old tests.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
spyrostz: In this block, I assume that you want to check if an external device buys energy from the storage, correct ? If this is the case, I think that 'or' statement is not appropriate, 'and' should be the correct one. I have probably misunderstood something here though.

integration_tests/jira_issues_tests.feature
spyrostz: Thanks for clarifying this, looks great.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -344,76 +342,83 @@ def _specific_row(self, slot, market):
...
+    def _accumulate_energy_trace(self):
+        for c_area in self.consumer_area_list:
+            for market in c_area.past_markets:
+                for trade in market.trades:
+                    self._accumulate_self_consumption(trade)
+                    self._accumulate_self_consumption_buffer(trade)
+                    self._dissipate_self_consumption_buffer(trade)
+
+    def _area_self_sufficiency(self, area):
+        self.producer_list = list()
```
spyrostz: Well, I am not sure whether it is that safe to create member variables outside of the constructor. We can discuss also offline about the potential structure, but I would feel much better if all self-sufficiency is encapsulated in a class. The reason I am asking this is because there are going to be more KPIs quite soon and I think there could be conflicts between these variables, which could be avoided if all these were inside a class.

src/d3a/models/strategy/predefined_load.py
```diff
@@ -99,10 +99,9 @@ def _update_energy_requirement(self):
                             for day in range(self.area.config.sim_duration.days + 1)}
         for slot_time in generate_market_slot_list(area=self.area):
             if self._allowed_operating_hours(slot_time.hour):
-                self.energy_requirement_Wh[slot_time] = \
-                    self.load_profile[slot_time] * 1000
-                self.state.desired_energy_Wh[slot_time] = \
-                    self.load_profile[slot_time] * 1000
+                self.energy_requirement_Wh[slot_time] = self.load_profile[slot_time] * 1000
+                self.state.desired_energy_Wh[slot_time] = self.load_profile[slot_time] * 1000
+                self.state.total_energy_demanded_wh += self.load_profile[slot_time] * 1000
```
spyrostz: Does not hurt, but I think that you could avoid adding the total_energy_demanded_wh to the state class, by adding a property to the state class that accumulates the values of the desired_energy_Wh dict. Optional, if you prefer it this way keep it, should not hurt.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
faizan2590: `OR` statement is inside the brackets that is testing if the internal ESS` energy buyer is not either of internal consumer OR internal storage?

We can discuss offline too for better understanding!

integration_tests/steps/kpi.py
faizan2590: Its an string object, so had to convert it to dict

hannesdiedrich: Hmm, doesnt deal `json.load()` with that already?

src/d3a/d3a_core/sim_results/file_export_endpoints.py
faizan2590: Resolved after offline discussion.

src/d3a/models/strategy/predefined_load.py
faizan2590: Well the issue here is `class LoadState` is used by both `LoadHoursStrategy` and `DefinedLoadStrategy`. 

- For LoadHoursStrategy, `total_energy_demanded_wh = avg_power_W * hrs_per_day`
- For DefinedLoadStrategy, `total_energy_demanded_wh` is accumulated via iterating over load_profiles.

The reason im skeptical of using `desired_energy_Wh` to accumulate `total_energy_demanded_wh` is due to its different  behaviour in both strategy. In LoadHoursStrategy, it is populated for all the hrs_of_day due which it gets accumulated more than the required operating hours.

We can also discuss offline, if it is unclear.

integration_tests/steps/kpi.py
faizan2590: `json.load()` is used for `kpi_data` which is an actual kpi after running the simulation, whereas `areas` variable contains the expected kpi

src/d3a/models/strategy/predefined_load.py
spyrostz: All clear, agreed. Thanks

integration_tests/steps/kpi.py
hannesdiedrich: ok, resolved offline

hannesdiedrich: @faizan2590 but you could rename areas to "expected_kpis" in order to make it more readable (and to prevent this confusion).

faizan2590: Doing it so!

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -408,16 +413,23 @@ def _area_self_sufficiency(self, area):
...
         if self.total_energy_demanded_wh <= 0:
-            self.self_sufficiency[area.name] = None
             return {"self_sufficiency": None}
 
         self_sufficiency = self.total_self_consumption_wh / self.total_energy_demanded_wh
-        self.self_sufficiency[area.name] = self_sufficiency
         return {"self_sufficiency": self_sufficiency}
 
+
+class KPI:
```
hannesdiedrich: Ah yes, nice!

PR #576, opened 2019-11-25
D3ASIM-1697: Added the unregister functionality on the redis external…

spyrostz:
… connection for areas. Added test coverage for RedisExternalConnection class.


src/d3a/models/area/redis_external_connection.py
```diff
@@ -11,8 +11,32 @@ def __init__(self, area):
         self.pubsub = self.redis_db.pubsub()
         self.sub_to_area_event()
         self.areas_to_register = []
+        self.areas_to_unregister = []
+
+    def market_cycle_event(self):
+        self.unregister_pending_areas()
```
hannesdiedrich: I guess that you dont need this here as it is already part of `register_new_areas`. right?

spyrostz: You are right, I forgot to remove it from the register_new_areas method, doing it. Thanks for pointing it out.

tests/test_external_connection.py
```diff
@@ -0,0 +1,120 @@
...
+                   f"base-area/{slugify(created_area)}/offers",
+                   f"base-area/{slugify(created_area)}/offer/response",
+                   f"base-area/{slugify(created_area)}/offer_delete/response",
+                   f"base-area/{slugify(created_area)}/offer_accept/response"
+                ]})
+
+    def test_unregister_areas_deletes_children_from_area(self):
+        area_list = ["kreuzberg", "friedrichschain", "prenzlauer berg"]
+        self.external_connection.areas_to_register = area_list
+        self.external_connection.register_new_areas()
```
hannesdiedrich: Very minor, but this code is called 3 times in this class. You could refactor this into a separate function.

tests/test_external_connection.py
```diff
@@ -0,0 +1,120 @@
...
+        assert set(ch.name for ch in self.area.children) == set(area_list)
+
+        self.external_connection.areas_to_unregister = ["schoneberg"]
+        self.external_connection.unregister_pending_areas()
+        assert len(self.area.children) == 3
+        self.external_connection.redis_db.publish.assert_called_with(
+            "base-area/unregister_participant/response", json.dumps({"response": "failed"}))
+
+    def test_subscribe_channel_list_fetches_correct_channel_names(self):
```
hannesdiedrich: Dont you already test this implicitly in  test_register_new_areas_creates_new_child_areas_from_buffer ?

tests/test_external_connection.py
spyrostz: You are right, I'll create a new method and call it at the start of the tests. Thanks

tests/test_external_connection.py
spyrostz: True, I wanted to have a specific test though that validates that the method that fetches the channels works as expected. If you think this is an overkill I can remove this test.

src/d3a/models/area/redis_external_connection.py
```diff
@@ -11,6 +11,31 @@ def __init__(self, area):
...
+        for area in self.areas_to_unregister:
+            try:
+                area_object = next(child for child in self.area.children if child.name == area)
+                area_object.deactivate()
+                self.area.children.remove(area_object)
+            except Exception as e:
+                self.area.log.error(f"Unsubscribing of area {area} failed with error {str(e)}.")
+                self.publish(f"{self.area.slug}/unregister_participant/response",
+                             json.dumps({"response": "failed"}))
+            else:
```
faizan2590: Shouldn't here be an `if` block first before this `else` block?

spyrostz: This is not an if/else block, but a try/except/else block. Have a look at [the Python documentation](https://docs.python.org/3/tutorial/errors.html#handling-exceptions).

PR #577, opened 2019-11-27
Feature/d3 asim 1727

spyrostz:


src/d3a/d3a_core/exceptions.py
```diff
@@ -63,3 +63,7 @@ class InvalidBalancingTradeException(MarketException):
 
 class DeviceNotInRegistryError(MarketException):
     pass
+
+
+class D3ARedisException(D3AException):
```
hannesdiedrich: Cool! Can you please also use this exception type for the exceptions added in 1479? Thanks!

spyrostz: Will do, thanks!

src/d3a/models/market/market_structures.py
```diff
@@ -161,10 +181,12 @@ def to_JSON_string(self):
 
 def trade_from_JSON_string(trade_string):
     trade_dict = json.loads(trade_string)
-    trade_dict['offer'] = offer_from_JSON_string(trade_dict['offer'])
+    trade_dict['offer'] = offer_or_bid_from_JSON_string(trade_dict['offer'])
     if 'residual' in trade_dict and trade_dict['residual'] is not None:
-        trade_dict['residual'] = offer_from_JSON_string(trade_dict['residual'])
+        trade_dict['residual'] = offer_or_bid_from_JSON_string(trade_dict['residual'])
     trade_dict['time'] = pendulum.parse(trade_dict['time'])
+    # if 'offer_bid_trade_info' in trade_dict:
```
hannesdiedrich: Any reason for this to be removed?

spyrostz: Yes, the reason for that is because this is a list with not always 5 arguments, but also 3. If you want me to give it a deeper look on that let me know and I will find out the reason.

src/d3a/models/area/__init__.py
```diff
@@ -192,8 +192,7 @@ def _cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fa
     def tick(self, is_root_area=False):
         if ConstSettings.IAASettings.MARKET_TYPE == 2 or \
                 ConstSettings.IAASettings.MARKET_TYPE == 3:
-            for market in self._markets.markets.values():
-                market.match_offers_bids()
```
hannesdiedrich: Dont we still need this? And only publish_market_clearing when redis is enabled?

spyrostz: You are totally right, we need this. What is more alarming here is why the integration tests pass despite this change. I will check it out.

src/d3a/models/market/market_structures.py
hannesdiedrich: Sorry, I am confused. Which list are you referring to? But i just realise, the we dont serialise the `offer_bid_trade_info` member of the trade. It should be added in the to_JSON_string method, right?

spyrostz: Not really. The type of `trade_dict['offer_bid_trade_info']` is not a TradeBidInfo, but a list with 3 elements (have a look at grid_fees/base_model.py, method propagate_original_bid_info_on_offer_trade or propagate_original_offer_info_on_bid_trade). The reason it was implemented that way was that either the bid or the offer members need to be populated according to the trade (if it is an accept_offer, the offer members will be populated, if it is an accept_bid the bid members will be populated). Even though this works, I will refactor it to use the same type and to do sanity checks whether the correct members are None.

hannesdiedrich: Ah, yes! Sorry. If this works for now, i susggest: dont do it in this story but open another story please.

spyrostz: Enhancement story created https://gridsingularity.atlassian.net/browse/D3ASIM-1742

src/d3a/models/area/__init__.py
```diff
@@ -193,8 +193,11 @@ def _cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fa
...
+            if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
+                self.dispatcher.publish_market_clearing()
```
faizan2590: Correct me if Im wrong but this line calls this [method](https://github.com/gridsingularity/d3a/pull/577/files#diff-ad05f547b22979aeada7dea2b3262509R27) which would is only taking care single-sided market.

spyrostz: The method you link does not handle one-sided market, only two-sided markets. The first line in this method is a check, that returns immediately if the market is an one-sided market. Generally, one sided markets do not need to clear, therefore this method needs to be called only by the two sided markets.

faizan2590: Ah sorry, my bad, you are totally right, I completely overlooked it.

src/d3a/models/area/redis_dispatcher/market_event_dispatcher.py
```diff
@@ -21,6 +22,9 @@ def __init__(self, area, root_dispatcher, redis):
             MarketEvent.OFFER.value: Event(),
             MarketEvent.OFFER_DELETED.value: Event(),
             MarketEvent.OFFER_CHANGED.value: Event(),
+            MarketEvent.BID_TRADED.value: Event(),
```
faizan2590: Shouldn't there be a MarketEvent for posting bid like `MarketEvent.BID` ?

spyrostz: There is no BID event in the MarketEvent enum. There is no need for this event in the current implementation.

PR #579, opened 2019-11-28
D3ASIM-1714: Moved adapting of the initial/final_buy/sell_rate to MMR…

hannesdiedrich:
… if use_market_maker_rate is True. into activate function.

### DIFF
--- a/src/d3a/models/strategy/load_hours.py
+++ b/src/d3a/models/strategy/load_hours.py
@@ -70,9 +70,7 @@ class LoadHoursStrategy(BidEnabledStrategy):
             update_interval = \
                 duration(minutes=ConstSettings.GeneralSettings.DEFAULT_UPDATE_INTERVAL)
 
-        # If use_market_maker_rate is true, overwrite final_buying_rate to market maker rate
-        if use_market_maker_rate:
-            final_buying_rate = GlobalConfig.market_maker_rate
+        self.use_market_maker_rate = use_market_maker_rate
 
         if isinstance(update_interval, int):
             update_interval = duration(minutes=update_interval)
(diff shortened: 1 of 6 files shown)


src/d3a/models/strategy/load_hours.py
```diff
@@ -109,6 +107,9 @@ def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None,
 
     @property
     def active_markets(self):
+        # If use_market_maker_rate is true, overwrite final_buying_rate to market maker rate
```
spyrostz: Why is it needed here ? It should suffice to be only in the event_activate method

hannesdiedrich: OMG, you are totally right! I was too focused on the word `activate` and confused active_markets with event_activate. Thanks for catching this!

tests/test_strategy_load_hours.py
```diff
@@ -433,6 +433,11 @@ def test_balancing_offers_are_created_if_device_in_registry(
 def test_use_market_maker_rate_parameter_is_respected(use_mmr, expected_rate):
     GlobalConfig.market_maker_rate = 9
     load = LoadHoursStrategy(200, final_buying_rate=33, use_market_maker_rate=use_mmr)
+
+    load.area = MagicMock(spec=Area)
+    load.area.now = now()
```
spyrostz: I assume that you need the 'now' area property just because it is used in the activate method, right ? Can a mock not work instead of the pendulum function ?

hannesdiedrich: Yeah, I could mock it. Can you tell me the advantage? Is mocking using less resources?

spyrostz: Well, in this specific case it does not hurt of course, but in general it is an antipattern to use the current time in unit tests. The reason for that is because there is no way to reproduce the exact same test run in 2 different now calls, because the different test runs will take place in different points in time. This makes the tests way less predictable (e.g. one test might fail today but succeed tomorrow, which makes the tests unpredictable). I would like to repeat though, that this is not such a case, therefore feel free to leave it as is if you prefer it.

hannesdiedrich: Thanks for the extensive explanation! Totally makes sense. Will change this here (and on the other occasions) .

PR #581, opened 2019-11-29
D3ASIM-1714: Moved device_validator into event_activate for load and …

hannesdiedrich:
…pv strategy.


src/d3a/models/strategy/load_hours.py
```diff
@@ -146,6 +139,12 @@ def event_activate(self):
         # If use_market_maker_rate is true, overwrite final_buying_rate to market maker rate
         if self.use_market_maker_rate:
             self.area_reconfigure_event(final_buying_rate=GlobalConfig.market_maker_rate)
+        for time_slot in generate_market_slot_list():
```
spyrostz: I think that the validation should also be applied to the `area_reconfigure_event` method, because final rate and initial rate could also potentially change from there.

src/d3a/models/strategy/pv.py
```diff
@@ -104,6 +102,10 @@ def event_activate(self):
         if self.use_market_maker_rate:
             self.area_reconfigure_event(initial_selling_rate=GlobalConfig.market_maker_rate,
                                         validate=False)
+        for time_slot in generate_market_slot_list():
```
spyrostz: Same here, should be refactored to its own function and called from the `area_reconfigure_event` method too.

PR #583, opened 2019-12-02
D3ASIM-1690: Improving 2-sided PAC Algorithm.

faizan2590:


src/d3a/models/market/two_sided_pay_as_clear.py
```diff
@@ -74,7 +76,10 @@ def _get_clearing_point(self, max_rate):
         for i in range(1, max_rate + 1):
             if self.state.cumulative_offers[self.now][i] >= \
                     self.state.cumulative_bids[self.now][i]:
-                return i, self.state.cumulative_bids[self.now][i]
+                if self.state.cumulative_bids[self.now][i] == 0:
+                    return i-1, self.state.cumulative_offers[self.now][i-1]
+                else:
+                    return i, self.state.cumulative_bids[self.now][i]
```
hannesdiedrich: I think i need an offline explanation for this part.

integration_tests/two_sided_pay_as_clear_market.feature
```diff
@@ -27,8 +27,14 @@ Feature: Two sided pay_as_clear market tests
...
      Then there are files with offers, bids & market_clearing_rate for every area
 
-   Scenario: Supply Demand Curve
+  Scenario: Supply Demand Curve
      Given we have a scenario named two_sided_pay_as_clear/test_clearing_energy
      And d3a is installed
      When we run the d3a simulation on console with two_sided_pay_as_clear.test_clearing_energy for 24 hrs
      Then the export functionality of supply/demand curve is tested
+
+  Scenario: No unmatched load
```
hannesdiedrich: Can you please also  prepend the ticket number here?

integration_tests/steps/two_sided_pay_as_clear.py
```diff
@@ -99,3 +99,11 @@ def test_offer_bid_market_clearing_rate_files(context):
...
 
     assert all(len(glob.glob(f)) == 1 for f in file_list)
+
+
+@then('the unmatched loads are tested')
+def test_export_supply_demand_curve(context):
+    sim_data_csv = glob.glob(os.path.join(context.export_path, "*", "plot",
+                                          "unmatched_loads_grid.html"))
+    if len(sim_data_csv) != 0:
+        raise FileExistsError("Not found in {path}".format(path=context.export_path))
```
hannesdiedrich: This is ok, bu there is also an function `no_unmatched_loads` that can be used. With this you can omit exporting the csv files.

src/d3a/models/market/two_sided_pay_as_clear.py
faizan2590: Sure, ping me when you are free

integration_tests/two_sided_pay_as_clear_market.feature
spyrostz: I would also find it better to move this on the jira_issues_tests feature file.

src/d3a/models/market/two_sided_pay_as_clear.py
```diff
@@ -27,6 +27,8 @@
 
 log = getLogger(__name__)
 
+Clearing = namedtuple('Clearing', ('rate', 'energy'))
```
spyrostz: How about moving this to the market_structures module? That way all market structures will be on the same file.

src/d3a/models/market/two_sided_pay_as_clear.py
hannesdiedrich: Discussed it offlline. Makes sense now. Please rename i to something like `rate`, Thanks!

PR #584, opened 2019-12-05
D3ASIM-1756: Added dispatching of tick event from bottom to top.

hannesdiedrich:


src/d3a/models/area/__init__.py
```diff
@@ -199,9 +199,10 @@ def tick(self, is_root_area=False):
                 for market in self.all_markets:
                     market.match_offers_bids()
         self.events.update_events(self.now)
-        if self.current_tick % self.config.ticks_per_slot == 0 and is_root_area:
-            self._cycle_markets()
-        self.dispatcher.broadcast_tick()
+        if not ConstSettings.GeneralSettings.DISPATCH_EVENTS_BOTTOM_TO_TOP:
```
spyrostz: I would like to try to use this code for both cases (bottom-to-top and top-to-bottom). Therefore, I would like to see if it is feasible to first cycle all markets in both cases, and then start dispatching the tick events. If you decide to try and do this, the code related to cycling markets in the tick event will be removed, and the broadcast_tick call for both cases will reside in the event_listener. If it is not clear let's discuss this offline.

src/d3a/d3a_core/simulation.py
```diff
@@ -290,6 +290,12 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
                     (tick_no + 1) / config.ticks_per_slot * 100,
                 )
 
+                if ConstSettings.GeneralSettings.DISPATCH_EVENTS_BOTTOM_TO_TOP:
+                    # trigger market_cycle:
+                    if self.area.current_tick % self.area.config.ticks_per_slot == 0:
```
faizan2590: This check could be avoided and this code block could be moved out of this for loop too.

src/d3a/models/area/__init__.py
hannesdiedrich: Sorry, just saw your comment @spyrostz .
Ok, I will try that. But wasn't it part of the story, that we get rid of the connection between tick and market cycle events?
So NOT doing this at all?

src/d3a/d3a_core/simulation.py
hannesdiedrich: True that! Will try it.

src/d3a/models/area/__init__.py
hannesdiedrich: Sorry, just understood what you meant. Trying it.

src/d3a/models/area/__init__.py
```diff
@@ -190,23 +191,29 @@ def _cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fa
...
         self.events.update_events(self.now)
-        if self.current_tick % self.config.ticks_per_slot == 0 and is_root_area:
-            self._cycle_markets()
-        self.dispatcher.broadcast_tick()
         self.current_tick += 1
         if self._markets:
             for market in self._markets.markets.values():
                 market.update_clock(self.current_tick)
 
+    def tick_and_dispatch(self):
```
spyrostz: This looks much better now (in comparison to calling the same code in the tick method), thanks for changing that.

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -206,13 +206,15 @@ def _perform_pay_as_bid_matching(self):
         return offer_bid_pairs
 
     def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selected_energy):
+        already_tracked = bid.buyer == offer.seller
         trade = self.accept_offer(offer_or_id=offer,
                                   buyer=bid.buyer,
                                   energy=selected_energy,
                                   trade_rate=clearing_rate,
-                                  already_tracked=False,
+                                  already_tracked=already_tracked,
```
spyrostz: Why is it needed here ? The reason I am asking is because if `already_tracked` is true, this will lead to tracking the same trade twice.

hannesdiedrich: This is needed for the case dispatching bottom-> up and offers and bid matching doesnt happen in direct line fo connection. Meaning in the default_2a case if the Cell Tower bid gets matched with the CEP offer in  House 1 (and not in grid or House 2, as expected).

hannesdiedrich: There is a Trade from IAA House 1 --> IAA House 1 without this change.

hannesdiedrich: An offline discussion revealed that this change indeed makes sense. The trade from and to House 1 market is not tracked with this change. 
@faizan2590 If you need more information, please ping me and we can discuss offline.

faizan2590: Yes, please. Let me know when you are free

src/d3a/models/area/__init__.py
```diff
@@ -190,23 +191,29 @@ def _cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fa
                 and _trigger_event and ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET:
             self.dispatcher.broadcast_balancing_market_cycle()
 
-    def tick(self, is_root_area=False):
+    def tick(self):
```
faizan2590: Was `is_root_area` ever used?

hannesdiedrich: yes, some months ago. But when not needed any more, it wasnt deleted here.

PR #585, opened 2019-12-05
D3ASIM-1760: Adding validate_rate check in DefinedLoadStrategy.

faizan2590:


src/d3a/models/strategy/update_frequency.py
```diff
@@ -67,7 +67,12 @@ def _set_or_update_energy_rate_change_per_update(self):
                     (self.initial_rate[slot] - self.final_rate[slot]) / \
                     self.number_of_available_updates
             else:
-                energy_rate_change_per_update[slot] = self.energy_rate_change_per_update[slot]
+                if self.rate_limit_object is min:
```
spyrostz: So, to understand the sign convention here, for the case of bids the `energy_rate_change_per_update` is negative, and for the offers is positive, right ? If I understood it correctly, it is a bit counter-intuitive, but ok for me.

faizan2590: Yes, you are right. If you have a look at this [method](https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/strategy/update_frequency.py#L87), updated rate is found via subtracting `energy_rate_change_per_update` from `initial_rate`. In case of load, rate has to increase increase so, sign of `energy_rate_change_per_update` has to be reversed. I hope I didn't confused you more

hannesdiedrich: I agree with Spyros, that this is counter intuitive . You could also change the sign in your mentioned method, Faizan. But ok, leave it like this for now.

src/d3a/setup/jira/d3asim_1637.json
```diff
@@ -113,16 +109,15 @@
...
               "display_type": "Load"
             },
             {
               "name": "PV 3",
-              "type": "PredefinedPV",
+              "type": "PV",
               "uuid": "f04cf867-eeac-48a0-89d2-23695340e268",
               "libraryUUID": null,
               "panel_count": 3,
-              "fit_to_limit": false,
```
hannesdiedrich: SO you moving from setting the energy_rate change per update to "fit_to_limit" . Any reason for this?

faizan2590: Reverting it to original state, and actually the original json setup file revealed an issue now that `Storage` strategy was not using `energy_rate_increase_per_update`, `energy_rate_decrease_per_update` since they were not part of parameters. Have added it in storage strategy parameters in my latest commit. Thanks for nitpicking.

PR #586, opened 2019-12-05
D3ASIM-1426: Adding self_consumption kpi.

faizan2590:


src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -339,16 +339,21 @@ def _specific_row(self, slot, market):
         return []
 
 
-class SelfSufficiency:
+class KPI:
```
spyrostz: How about moving this whole class to its own module? The reason I mention this is because this class is not directly relevant to the file export endpoints, but it is about calculating area statistics/kpis

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -339,16 +339,21 @@ def _specific_row(self, slot, market):
...
     def __init__(self):
+        self.performance_index = dict()
```
hannesdiedrich: Very minor, how about renaming it to `performance_indices`  As it contains multiple indices now.

src/d3a/d3a_core/sim_results/kpi.py
```diff
@@ -0,0 +1,145 @@
...
+    def area_performance_indices(self, area):
+        self.producer_list = []
+        self.consumer_list = []
+        self.consumer_area_list = []
+        self.ess_list = []
+        self.buffer_list = []
+        self.total_energy_demanded_wh = 0
+        self.total_energy_produced_wh = 0
+        self.total_self_consumption_wh = 0
+        self.self_consumption_buffer_wh = 0
```
hannesdiedrich: Do you think you can make these variables NOT part of self?

src/d3a/d3a_core/sim_results/kpi.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def __repr__(self):
+        return f"KPI: {self.performance_indices}"
+
+    def _accumulate_devices(self, area):
+        for child in area.children:
+            if type(child.strategy) in [PVStrategy, PVUserProfileStrategy,
```
spyrostz: This case does not catch FinitePowerPlant, and I think it is needed here, please correct me if I am wrong.

faizan2590: [InfiniteBus](https://github.com/gridsingularity/d3a/pull/586/files#diff-82afe03c9b73bbef99c1ff6b646c24ddR52) is being separately dealt.

src/d3a/d3a_core/sim_results/kpi.py
```diff
@@ -0,0 +1,145 @@
...
+        self.buffer_list = list()
+        self.total_energy_demanded_wh = 0
+        self.total_energy_produced_wh = 0
+        self.total_self_consumption_wh = 0
+        self.self_consumption_buffer_wh = 0
+
+    def __repr__(self):
+        return f"KPI: {self.performance_indices}"
+
+    def _accumulate_devices(self, area):
```
spyrostz: What this function is also not handling is ExternalStrategy, but due to the fact that such a strategy can be either a producer or a consumer, while at the same time not being a storage, I would just leave a TODO comment here and wait for more concrete requirements.

src/d3a/d3a_core/sim_results/kpi.py
spyrostz: InfiniteBus i a different strategy than FinitePowerPlant and MarketMakerStrategy though. The difference is that infinite bus is also a consumer, while the 2 other strategies are only producers.

src/d3a/d3a_core/sim_results/kpi.py
```diff
@@ -0,0 +1,145 @@
...
+        if trade.seller_origin in self.buffer_list and trade.buyer_origin in self.consumer_list:
+            self.total_self_consumption_wh += trade.offer.energy * 1000
+
+    def _dissipate_infinite_consumption(self, trade):
+        if trade.buyer_origin in self.buffer_list and trade.seller_origin in self.producer_list:
+            self.total_self_consumption_wh += trade.offer.energy * 1000
+
+    def _accumulate_energy_trace(self):
+        for c_area in self.consumer_area_list:
+            for market in c_area.past_markets:
```
spyrostz: Are you sure the code is using all past markets ? I see that this function is used only when exporting the csv files. During that time, the area.past_markets list will include only the very last market (except if the value of KEEP_PAST_MARKETS is set to True). In case I did not understand something wrong let me know, but I have the impression that this stat should either be incrementally calculated or not exposed to the d3a-web/UI.

faizan2590: Yes, you are right, I didn't any effort into dealing with it as part of this story. Since the KPI requirements were constantly changing. 

Since, now they seems to have stabilised and this story has been dragged a lot, I would propose to not do it now and create a small story to deal with it later. 

Let me know if you agree?

spyrostz: Sure, just create a new story, as long as it is tracked it is fine by me.

src/d3a/d3a_core/sim_results/kpi.py
faizan2590: Sorry, mis-read in first place. You are right adding `FinitePowerPlant` too

src/d3a/d3a_core/sim_results/kpi.py
faizan2590: Then it has to be part of method parameter and then catch it as part of function return value.

@spyrostz Do you have any opinion on that?

spyrostz: It annoys me too actually, because there is no clear way to see how these values are populated, and makes it harder to see where these variables are actually used. I would suggest to at least add a separate class that can hold these parameters, and then you can simply pass an object of this class to the methods. That way you will at least avoid storing these as members of the KPI class.

src/d3a/d3a_core/sim_results/kpi.py
```diff
@@ -0,0 +1,147 @@
...
+        self.performance_indices = dict()
+        self.state = None
+
+    def __repr__(self):
+        return f"KPI: {self.performance_indices}"
+
+    def area_performance_indices(self, area):
+        self.state = KPIState()
+
+        self.state.accumulate_devices(area)
```
spyrostz: I would expose one single `update(area)` method on the KPIState class, that would be only responsible for calling the `accumulate_devices` and `accumulate_energy_trace` methods. That way there is only one call for updating the statistics in the KPIState class.

PR #589, opened 2019-12-10
D3ASIM-1696: Added 2-sided market functionality in the external strat…

spyrostz:
…egy API. Added unit tests for the external strategy.


src/d3a/models/strategy/external_strategy.py
```diff
@@ -141,11 +156,38 @@ def _delete_offer(self, payload):
...
+
+    def _delete_bid(self, payload):
+        try:
+            arguments = self._parse_payload(payload)
+            assert set(arguments.keys()) == {'bid'}
+            arguments['bid_or_id'] = arguments.pop('bid')
+        except Exception:
+            self.publish(
+                self._delete_bid_response_channel,
+                {"error": "Incorrect offer request. Available parameters: (bid)."}
```
hannesdiedrich: Very minor but you could rename "Incorrect offer request" to "Incorrect bid request"  (also in _bid).

hannesdiedrich: Edit: Here you can even use "delete_offer request"

src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -73,6 +73,8 @@ def bid(self, price: float, energy: float, buyer: str, seller: str,
         if energy <= 0:
             raise InvalidBid()
 
+        if original_bid_price is None:
```
hannesdiedrich: Just for my understanding: This case is for the original bid, right?

src/d3a/models/market/market_redis_connection.py
```diff
@@ -265,7 +265,7 @@ def _accept_bid_impl(self, arguments):
                          {"status": "ready", "trade": trade.to_JSON_string(),
                           "transaction_uuid": transaction_uuid})
         except Exception as e:
-            logging.error(f"Error when handling accept_bid on market {self.market_object.name}: "
+            logging.error(f"Error when handling accept_bid on market {self.market.name}: "
```
hannesdiedrich: Upsi. This got throught the last time, right?

tests/test_strategy_external.py
```diff
@@ -0,0 +1,134 @@
...
+        )
+
+    def _assert_dict_is_the_same_as_offer(self, offer_dict, offer):
+        assert offer.id == offer_dict["id"]
+        assert offer.real_id == offer_dict["real_id"]
+        assert offer.price == offer_dict["price"]
+        assert offer.energy == offer_dict["energy"]
+        assert offer.seller == offer_dict["seller"]
+
+    def test_list_offers(self):
```
hannesdiedrich: When I am reading this. I guess we dont need a list_bids functionality for the external strategy?

src/d3a/models/strategy/external_strategy.py
```diff
@@ -75,14 +75,29 @@ def _accept_offer_response_channel(self):
...
+                self._delete_offer_channel: self._delete_offer,
+                self._accept_offer_channel: self._accept_offer,
+                self._list_offers_channel: self._offer_lists,
+            })
+        else:
+            self.redis_db.sub_to_multiple_channels({
+                self._offer_channel: self._offer,
+                self._delete_offer_channel: self._delete_offer,
+                self._bid_channel: self._bid,
+                self._delete_bid_channel: self._delete_bid
```
faizan2590: Correct me if Im wrong but shouldn't there be one more channel for `accepting bid`

src/d3a/models/strategy/external_strategy.py
spyrostz: Well, there are different parameters for offer and bid, therefore we cannot reuse the same method. Adapted the offer -> bid rename though

src/d3a/models/market/two_sided_pay_as_bid.py
spyrostz: Exactly, this is the case where the bid is first placed in a market.

src/d3a/models/market/market_redis_connection.py
spyrostz: No, we were obliged to revert this comment actually. The reason for that is because there needs to be a property in order to get overriden, which conflicts to the 'market' instance variable name, thus causing errors.

tests/test_strategy_external.py
spyrostz: Actually we do not. For two-sided market, the external device does neither list nor accept the bids, the device only manages its own bids (via placing and deleting a bid). The clearing of bids and offers is done exclusively on the markets, that's why there is no need for external connections to see the available bids and offers to the market.

src/d3a/models/strategy/external_strategy.py
hannesdiedrich: Ah sorry, i misspelled. I meant "delete_bid request.

src/d3a/models/strategy/external_strategy.py
spyrostz: Like my comment to Hannes above, the external connection should not be able to accept bids for 2-sided markets, since the market is responsible for the market clearing (accepting bids and offers). The external connection is only responsible for placing offers and bids to a market, the market itself performs the trade.

faizan2590: ah okay makes sense, thanks for explaining.

src/d3a/models/strategy/external_strategy.py
spyrostz: Ahhh I see now, thanks a lot. Will fix this

PR #600, opened 2020-01-09
D3ASIM-972: Adapted Redis endpoints to send only the results from the…

spyrostz:
…  last market. For now this is controlled by a constant, in order not to break the current d3a-web results format. Once the d3a-web has changed to support recomposing results from individual markets, the variable values needs to change to update the API.


src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -549,21 +550,34 @@ def update(self, area):
...
+            use_last_past_market=not ConstSettings.GeneralSettings.KEEP_PAST_MARKETS
+        )
+
+        # Calculate aggregated results for all markets
+        self._convert_output_format(self._price_energy_day, self.csv_output, self.redis_output)
+
+        # Calculate results for one market
+        self._convert_output_format(latest_output_area_names, {}, self.latest_output)
+
+        return self.csv_output, self.redis_output
```
hannesdiedrich: AFAIK, wherever "update" is used, the return values are not used. So you could remove this line, right?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -91,7 +107,9 @@ def generate_json_report(self):
             "tree_summary": self.tree_summary,
             "status": self.status,
             "device_statistics": self.device_statistics.device_stats_time_str,
-            "energy_trade_profile": self.energy_trade_profile
+            "energy_trade_profile": self.energy_trade_profile,
+            "last_unmatched_loads": self.last_unmatched_loads,
+            "last_energy_trade_profile": self.last_energy_trade_profile
```
hannesdiedrich: Why are you adding these to the local export? Is this a remnant of testing?

spyrostz: Indeed remnant of testing, removing these. Thanks!

src/d3a/d3a_core/sim_results/area_statistics.py
spyrostz: Right, deleting this

PR #606, opened 2020-01-16
Bug/d3 asim 1862

hannesdiedrich:


integration_tests/steps/jira_issues_tests.py
```diff
@@ -284,3 +284,16 @@ def trades_happen(context):
 @given('the file {setup_json} is used for the area setup')
 def json_setup_file(context, setup_json):
     os.environ["D3A_SETUP_PATH"] = os.path.join(d3a_path, 'setup', setup_json)
+
+
+@then('the DSO should only pay 30 cents')
+def dso_pays_certain_price(context):
```
spyrostz: I would also add an initial check in this function, that validates that there is at least one trade in any market. That way it is going to be checked that at least one trade has been performed and the test will be executed correctly.

hannesdiedrich: Ok, will do!

faizan2590: May be pass rate as a parameter. Not necessary now but would be handy later for using the same method for grid fee testing.

hannesdiedrich: Hm, this test is really special. To make it more general, more stuff has to be added (to also take other grid structures). Do you still want me to change it?

faizan2590: Hmm, okay leave it as it is for now. If needed, could be modified later too.

PR #608, opened 2020-01-16
D3ASIM-1705: Added missing endpoint for the bid listing. Differentiat…

spyrostz:
…ed behavior of offer listing according to whether the market is one sided or two sided.

### DIFF
--- a/src/d3a/models/strategy/external_strategy.py
+++ b/src/d3a/models/strategy/external_strategy.py
@@ -60,6 +60,10 @@ class RedisMarketExternalConnection(TwoSidedMarketRedisEventSubscriber):
         return f"{self.area.parent.slug}/{self.area.slug}/delete_bid"
 
     @property
+    def _list_bids_channel(self):
+        return f"{self.area.parent.slug}/{self.area.slug}/bids"
+
+    @property
     def _offer_response_channel(self):
         return f"{self._offer_channel}/response"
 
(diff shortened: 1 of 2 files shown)


src/d3a/models/strategy/external_strategy.py
```diff
@@ -113,9 +123,17 @@ def _serialize_offer_dict(offer_dict):
 
     def _offer_lists(self, payload):
         try:
-            return_data = self._serialize_offer_dict(self.market.offers)
-            self.publish(self._list_offers_response_channel,
-                         {"status": "ready", "offer_list": return_data})
+            if ConstSettings.IAASettings.MARKET_TYPE == 1:
+                filtered_offers = [{"id": v.id, "price": v.price, "energy": v.energy}
+                                   for _, v in self.market.offers.items()]
```
hannesdiedrich: So for the one sided market the user can see ALL offers, not only just the one she posted?

spyrostz: Yes, because on one-sided markets a user will need to select from available offers in the market and accept it via the 'accept_offer' endpoint. If the user was not able to see other people's offers, he would not know which offer to accept.

hannesdiedrich: You are right of course! Thanks for explaining again!

PR #613, opened 2020-01-22
Feature/d3 asim 1867

spyrostz:


src/d3a/models/market/__init__.py
```diff
@@ -224,3 +224,12 @@ def total_spent(self, buyer):
...
         return sum(trade.offer.price for trade in self.trades if trade.seller == seller)
+
+    @property
+    def info(self):
+        return {
+            "name": self.name,
+            "id": self.id,
+            "start_time": self.time_slot_str,
+            "duration_min": GlobalConfig.slot_length.minutes
+        }
```
hannesdiedrich: Where is this used? I neither see it in d3a-interface nor in the client.

spyrostz: It is a bit inconspicuous, but it is used here: https://github.com/gridsingularity/d3a/pull/613/files#diff-4227e14dfec828d6e1d346e65d981d94R36 . Essentially I wanted a way to report market information for the new market, that's why this method was introduced.

hannesdiedrich: Ah yes, makes sense! Thanks!

spyrostz:
Closing the PR for now to integrate the changes we discussed offline. Reopening once these are ready.

PR #614, opened 2020-01-22
Feature/d3 asim 1865

hannesdiedrich:


src/d3a/setup/kpi/root_area_self_sufficient_with_infinitebus.py
```diff
@@ -14,7 +14,6 @@
 import os
 
 current_dir = os.path.dirname(__file__)
-print(current_dir)
 
 user_profile_path1 = os.path.join(current_dir, "../resources/House_1_Winter_Berlin.csv")
```
faizan2590: Below four lines could be deleted too

PR #615, opened 2020-01-23
D3ASIM-1896: Removed increment_update_counter from 2 sided market in …

hannesdiedrich:
…storage buy method as it is doen inplicitly by update_posted_bids_over_ticks.


src/d3a/models/strategy/storage.py
```diff
@@ -261,8 +261,8 @@ def _validate_constructor_arguments(initial_soc=None, min_allowed_soc=None,
     def event_tick(self):
         self.state.clamp_energy_to_buy_kWh([ma.time_slot for ma in self.area.all_markets])
         for market in self.area.all_markets:
-            self.bid_update.increment_update_counter(self, market.time_slot)
             if ConstSettings.IAASettings.MARKET_TYPE == 1:
+                self.bid_update.increment_update_counter(self, market.time_slot)
```
spyrostz: Instead of moving it from here, why not removing this from the `update_posted_bids_over_ticks` method ? That way both markets will have a common way to update the counter, and also the `update_posted_bids_over_ticks` will have the single responsibility of only updating the bids.

PR #616, opened 2020-01-23
Feature/d3 asim 1867

spyrostz:


tox.ini
```diff
@@ -79,17 +76,14 @@ commands =
 
 [testenv:travis]
 basepython = python3.6
-passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH SOLC_BINARY GANACHE_BINARY
+passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH SOLC_BINARY GANACHE_BINARY BRANCH
```
hannesdiedrich: Which library needs the BRANCH env variable?

spyrostz: No library, have a look at the installation of d3a-interface on the setup.py file though. The reason for that is for the dockerfile to use an environment variable to select the correct d3a-interface branch (master, production or live). The change in Travis is needed in order to test against the correct d3a-interface branch according to merge target.

hannesdiedrich: Ahhh. ok! Now with your d3a-web PR it makes sense. Thanks!

tox.ini
```diff
@@ -79,17 +76,14 @@ commands =
...
 deps =
 	pip-tools
 	coverage
 commands =
 	python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
-    pip-sync requirements/base.txt requirements/tests.txt
+    pip-sync requirements/base.txt requirements/tests.txt requirements/pandapower.txt
```
faizan2590: Correct me if Im wrong, didn't we skipped it in first place due to `pandapower` incompatability with `pypy` ?

spyrostz: Yes, but we do not use pypy when running tests AFAIK. Pandapower is needed for the tests due to its integration test, that's why it is included here.

PR #618, opened 2020-01-23
D3ASIM-1878: Enabling grid_fee_percentage to be accepted from UI.

faizan2590:


src/d3a/d3a_core/area_serializer.py
```diff
@@ -135,8 +135,9 @@ def optional(attr):
             children = [area_from_dict(child) for child in description['children']]
         else:
             children = None
+        transfer_fee_pct = description.get('grid_fee_percentage', None)
```
spyrostz: It is no longer a transfer fee, therefore I would recommend to rename it to 'market_fee_pct' or 'area_fee_pct'

src/d3a/d3a_core/area_serializer.py
```diff
@@ -135,8 +135,9 @@ def optional(attr):
             children = [area_from_dict(child) for child in description['children']]
         else:
             children = None
+        grid_fee_percentage = description.get('grid_fee_percentage', None)
```
spyrostz: I would call the `validate_area` method here instead of the constructor. That way we are going to keep the constructor of the class smaller, and the Area object will not even be created.

faizan2590: After offline discussion, leaving it as it is.

PR #619, opened 2020-01-23
Bug/d3 asim 1862

faizan2590:


integration_tests/steps/integration_tests.py
```diff
@@ -945,13 +945,15 @@ def _filter_markets_by_market_name(context, market_name):
...
-def assert_trade_rates(context, market_name, trade_rate):
+@then('trades on the {market_name} market clear with {trade_rate} cents/kWh and '
+      'at grid_fee_rate with {grid_fee_rate} cents/kWh')
+def assert_trade_rates(context, market_name, trade_rate, grid_fee_rate=0):
     markets = _filter_markets_by_market_name(context, market_name)
 
     for market in markets:
         for t in market.trades:
             assert isclose(t.offer.price / t.offer.energy, float(trade_rate))
+            assert isclose(t.fee_price / t.offer.energy, float(grid_fee_rate), rel_tol=1e-01)
```
spyrostz: Isn't 0.1 a bit too low for a tolerance level ? I would decrease it by a couple more factors (0.00001 would be safer to my eyes)

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -119,14 +122,14 @@ def energy_bills(self, area, past_market_types):
...
                 if seller in result:
-                    self._store_sold_trade(result[seller], trade.offer)
+                    self._store_sold_trade(result[seller], trade)
         for child in area.children:
-            child_result = self.energy_bills(child, past_market_types)
+            child_result = self._energy_bills(child, past_market_types)
             if child_result is not None:
                 result[child.name]['children'] = child_result
-                result[child.name]['market_fee'] = self.market_fees[child.name]
+                # result[child.name]['market_fee'] = self.market_fees[child.name]
```
spyrostz: Could you also add either to the unit tests or to the integration tests an assert that fees are populated on the energy_bills? I am not sure, but it would be nice to have an extra check that the grid fees end up in the flattened results of the energy bills.

src/d3a/models/market/market_structures.py
```diff
@@ -143,22 +143,22 @@ def trade_bid_info_from_JSON_string(info_string):
...
                                          already_tracked, offer_bid_trade_info, seller_origin,
-                                         buyer_origin)
+                                         buyer_origin, fee_price)
 
     def __str__(self):
         return (
             "{{{s.id!s:.6s}}} [origin: {s.seller_origin} -> {s.buyer_origin}] "
             "[{s.seller} -> {s.buyer}] {s.offer.energy} kWh @ {s.offer.price} {rate} "
-            "{s.offer.id}".
+            "{s.offer.id} [FEE: {s.fee_price} cts.]".
```
spyrostz: Can you lower-case the fee? It will look better on the log.

src/d3a/models/market/one_sided.py
```diff
@@ -173,12 +174,12 @@ def determine_offer_price(self, energy_portion, energy, trade_rate,
...
         else:
-            revenue, fees, trade_rate_incl_fees = \
+            revenue, grid_fee_rate, trade_rate_incl_fees = \
                 GridFees.calculate_trade_price_and_fees(
                     trade_bid_info, self.transfer_fee_ratio
                 )
-            self.market_fee += fees
-            return energy * (trade_rate_incl_fees - fees) \
+            self.market_fee += grid_fee_rate * energy_portion * energy
+            return grid_fee_rate, energy * (trade_rate_incl_fees - grid_fee_rate) \
```
spyrostz: The return values in the if block are not consistent, if the if block is True 2 values are returned, if it is False only one value is returned.

src/d3a/models/market/one_sided.py
```diff
@@ -173,12 +174,12 @@ def determine_offer_price(self, energy_portion, energy, trade_rate,
                 energy, trade_rate, energy_portion, orig_offer_price
             )
         else:
-            revenue, fees, trade_rate_incl_fees = \
+            revenue, grid_fee_rate, trade_rate_incl_fees = \
```
spyrostz: Looking at the diff below, I think it would be much simpler and more consistent if the `determine_offer_price` function returned grid_fee_price instead of grid_fee_rate, since you need only the fee price after this function returns.

src/d3a/models/market/one_sided.py
```diff
@@ -245,11 +247,13 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *, energy: in
 
         # Delete the accepted offer from self.offers:
         self.offers.pop(offer.id, None)
-
+        fee_price = fee_rate * offer.energy
```
spyrostz: I think that `self.market_fee` should be updated here, in order to cover all market types. Needs to also take into account the `already_tracked` flag. Also please clone the changes to the 2 sided market too, thanks!

src/d3a/setup/non_compounded_grid_fees.py
```diff
@@ -20,9 +20,11 @@
 from d3a.models.strategy.load_hours import LoadHoursStrategy
 from d3a.models.appliance.pv import PVAppliance
 from d3a.models.strategy.pv import PVStrategy
+from d3a_interface.constants_limits import ConstSettings  # NOQA
 
 
 def get_setup(config):
+    # ConstSettings.IAASettings.MARKET_TYPE = 3
```
spyrostz: Please remove this

tests/test_stats.py
```diff
@@ -196,38 +197,36 @@ def test_energy_bills_ensure_device_types_are_populated(grid2):
 def grid_fees():
     house1 = FakeArea('house1',
                       children=[FakeArea("testPV")],
-                      past_markets=[FakeMarket([], name='house1', fees=2.0),
-                                    FakeMarket([], name='house1', fees=6.0)])
+                      past_markets=[FakeMarket([], name='house1', fees=0.0),
```
spyrostz: Why disabling the fees here ?

src/d3a/models/market/one_sided.py
faizan2590: The reason I did it this way is to take into account the both cases, i.e. accepting full as well as partial offers. We can discuss offline too if this still doesn't seems like a good approach to you.

src/d3a/d3a_core/sim_results/stats.py
faizan2590: Correct me if Im wrong but isn't this [test](https://github.com/gridsingularity/d3a/blob/bug/D3ASIM-1862/tests/test_stats.py#L214) is already taking care of it?

spyrostz: I am not sure if it is really needed, but we do not see the market fee for the grid market in the test you have linked to.

src/d3a/models/market/one_sided.py
spyrostz: Ah you are right, the if block applies to the second argument, not the tuple. All good

src/d3a/models/market/one_sided.py
spyrostz: Since you have the energy_portion here, you can calculate the fee price for both cases. The reason why I am insisting in this is because the same function in the balancing market returns fees as price (if I understand the PR correctly), and not as rate. Also, having a price function here would allow us to only use energy rates in the context of this function, and once the `determine_offer_price` is called the market is only going to use prices as a unit.

faizan2590: Okay, I have adapted this comment, now `determine_offer_price` will return grid_fee in price.

PR #620, opened 2020-01-24
D3ASIM-1867: Fixed channel list reporting to the client.

spyrostz:


src/d3a/setup/kpi/root_area_self_sufficient_with_infinitebus.py
```diff
@@ -1,75 +1,75 @@
...
+                    ),
+                ],),
+
+            Area('DSO', strategy=InfiniteBusStrategy(energy_buy_rate=5),
+                 appliance=SimpleAppliance()),
+
+        ],
+        config=config, transfer_fee_pct=0, transfer_fee_const=0,
+    )
+    return area
```
hannesdiedrich: What happened here? 
Its the same file, isnt it?
Please add our file header concerning the licence though.  Thanks!

spyrostz: That strikes me as odd too. I can revert this file again, the only change I actually did was removing a print (which was done at the master branch and caused a merge conflict that I had to resolve)

PR #624, opened 2020-01-28
Adding integration tests to validate clearing is equal to the last ma…

faizan2590:
…tched bid_rate


integration_tests/two_sided_pay_as_clear_market.feature
```diff
@@ -33,3 +33,15 @@ Feature: Two sided pay_as_clear market tests
      When we run the d3a simulation on console with two_sided_pay_as_clear.test_clearing_energy for 24 hrs
      Then the export functionality of supply/demand curve is tested
 
+
+  Scenario: One Offer and One Bid are matched at the Bid Rate
+     Given we have a scenario named two_sided_pay_as_clear/one_offer_one_bid
```
spyrostz: How about doing something more radical, in order to validate that this works for multiple offers too, why not changing the setup file to incorporate multiple PVs? Then you can rename the scenario to `multiple_offers_one_bid` and validate that this does not work just for one offer.

PR #626, opened 2020-01-30
Feature/d3 asim 1907

spyrostz:
Initial implementation, added external connections for load and pv devices. Added market and device statistics on the external connections. 
Still missing:
- delete bid and offer functionality on the respective strategies
- storage strategy external connection  
- tests
- Refactoring to diminish code duplication, which will become more evident by the introduction of the storage strategy external connection.


src/d3a/models/strategy/load_hours.py
```diff
@@ -173,6 +142,18 @@ def area_reconfigure_event(self, avg_power_W=None, hrs_per_day=None,
...
             self.assign_energy_requirement(avg_power_W)
 
+        self._area_reconfigure_prices(final_buying_rate)
+
+    def event_activate_price(self):
+        # If use_market_maker_rate is true, overwrite final_buying_rate to market maker rate
+        if self.use_market_maker_rate:
+            self.area_reconfigure_event(final_buying_rate=GlobalConfig.market_maker_rate)
+        self._validate_rates()
+        self.bid_update.update_on_activate()
```
faizan2590: Correct me if Im wrong, what I have understood so far is that this bid_price_curve would also be formed for  [LoadExternalStrategy](https://github.com/gridsingularity/d3a/pull/626/files#diff-df43fe29b55cd7a82421a0c25a3f60dfR25) which I think should be avoided as `External Agent` would be having their own algorithm to buy energy

spyrostz: The whole method `event_activate_price` is overridden in the `LoadHoursExternalStrategy` , and the base class method is invoked only if the `connected` parameter is true, essentially only when a device is connected to the strategy. The reason it is implemented this way (I would prefer to skip it altogether) is because there is a requirement for the device to follow the normal LoadHoursStrategy until the external client is registered/connected to the device.

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -0,0 +1,146 @@
...
+                {"status": "error",
+                 "error_message": f"Error reporting stats for area {self.device_name}."})
+
+    @property
+    def market(self):
+        return self.area.next_market
+
+    @property
+    def device_name(self):
+        return self.owner.name
```
faizan2590: I think these two method could be moved to some Class ExternalStrategyBase as they are reimplemented in `PV` as well as in `Load`

spyrostz: I suggest to postpone this for the following PR for the same story, because there is a lot of code duplication that can be avoided in these 2 classes. The reason why it is not that easy to do though, is because these classes do not inherit from ExternalStrategyBase, they inherit only from the LoadHours and PVStrategy.

src/d3a/setup/strategy_tests/external_devices.py
```diff
@@ -0,0 +1,71 @@
...
+                                                                    hrs_per_day=4,
+                                                                    hrs_of_day=list(
+                                                                        range(12, 16)),
+                                                                    final_buying_rate=35),
+                         appliance=SwitchableAppliance()),
+                    Area('pv', strategy=PVExternalStrategy(panel_count=4),
+                         appliance=PVAppliance()),
+
+                ],
+                transfer_fee_pct=0, transfer_fee_const=0,
```
faizan2590: Please change  `transfer_fee_pct` to `grid_fee_percentage`

src/d3a/setup/strategy_tests/external_devices.py
```diff
@@ -0,0 +1,71 @@
...
+            ),
+            Area(
+                'House 2',
+                [
+                    Area('load', strategy=LoadHoursExternalStrategy(avg_power_W=200,
+                                                                    hrs_per_day=4,
+                                                                    hrs_of_day=list(
+                                                                        range(12, 16)),
+                                                                    final_buying_rate=35),
```
faizan2590: Shouldn't it be throwing an exception when user wants to use `LoadHoursExternalStrategy` and simultaneously also tries to set price preference?

spyrostz: That's again the confusing requirement we discussed in the comments above. We need to be able to set the price preference for the default strategies, the external strategy only takes over after a client is connected/registered.

faizan2590: Ohkay, I think this needs to be added in test somehow, as I fear strategy placing default offer/bid in addition to external agent's offer/bid, leading to double counting

spyrostz: I think there is a misunderstanding here. This happens in either/or fashion. Either the default strategy is used or the external agent. There can never be a case where the default strategy will place offers at the same time as the external agent.

src/d3a/setup/strategy_tests/external_devices.py
spyrostz: Will do, thanks!

PR #627, opened 2020-01-31
Feature/d3 asim 1907 2

spyrostz:
Deals with delete offer and delete bid functionality

### DIFF
--- a/Dockerfile_pypy
+++ b/Dockerfile_pypy
@@ -11,7 +11,6 @@ RUN apt-get update && apt-get install build-essential libatlas3-base libatlas-ba
 RUN pypy3 -m pip install numpy==1.16.4
 RUN pypy3 -m pip install scipy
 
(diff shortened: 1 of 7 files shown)


Dockerfile_pypy
```diff
@@ -13,8 +13,7 @@ RUN pypy3 -m pip install scipy
 
 # TODO: Workaround for pysolc due to https://github.com/ethereum/py-solc/issues/64
```
faizan2590: Probably this line could be deleted too if that [line](https://github.com/gridsingularity/d3a/pull/627/files#diff-43324f89e338058c3de62665c5d1802dL17) is being removed.

spyrostz: You are right, deleting this. Thanks

spyrostz:
> I assume there would be one more version to incorporate storage with external strategy?

Precisely, I am summarising the remaining work in the subtasks of this story https://gridsingularity.atlassian.net/browse/D3ASIM-1907

PR #629, opened 2020-02-03
Added new endpoint for the current market in the sim results.

spyrostz:


Dockerfile
```diff
@@ -4,8 +4,8 @@ ADD . /app
 
 WORKDIR /app
 
-RUN pip install --upgrade pip
-RUN pip install --upgrade setuptools
+RUN pip install --upgrade pip ptpython
```
hannesdiedrich: Why would one need a REPL in a docker container with an entrypoint, meaning a non-interactive container?

spyrostz: My ubuntu machine at home was complaining about this when building the image. I do not know if this is necessary when building via mac or travis, however I will add a comment on [this issue](https://gridsingularity.atlassian.net/browse/TESTS-4) which will deal with it, in order to check if it is really necessary. Thanks for pointing it out.

PR #630, opened 2020-02-03
D3ASIM-1946: Added get_bids method that gets protected by concurrent …

spyrostz:
…thread access using a lock and a copy of the data that are accessed. Changed the list_bids response to serialize the output of the bids to a JSON string.

### DIFF
--- a/src/d3a/models/strategy/external_strategies/load.py
+++ b/src/d3a/models/strategy/external_strategies/load.py
@@ -47,7 +47,7 @@ class LoadHoursExternalStrategy(LoadHoursStrategy):
             filtered_bids = [{"id": v.id, "price": v.price, "energy": v.energy}
                              for _, v in self.market.get_bids().items()
                              if v.buyer == self.device_name]
-            self.redis.publish(
+            self.redis.publish_json(
                 list_bids_response_channel,
                 {"status": "ready", "bid_list": filtered_bids})
         except Exception as e:
--- a/tests/test_inter_area_agent.py
+++ b/tests/test_inter_area_agent.py
@@ -83,6 +83,9 @@ class FakeMarket:
     def sorted_offers(self):
         return list(sorted(self.offers.values(), key=lambda b: b.price / b.energy))
 
+    def get_bids(self):
+        return self.bids
+
     def set_time_slot(self, timeslot):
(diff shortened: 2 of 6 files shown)


src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -47,7 +47,7 @@ def _list_bids(self, payload):
             filtered_bids = [{"id": v.id, "price": v.price, "energy": v.energy}
                              for _, v in self.market.get_bids().items()
                              if v.buyer == self.device_name]
-            self.redis.publish(
+            self.redis.publish_json(
```
hannesdiedrich: This is not part of the fixing the bug, or is it? 
Is there any advantage in comparison to "publish" ?

tests/test_inter_area_agent.py
```diff
@@ -83,6 +83,9 @@ def __init__(self, offers, bids=[], m_id=123, transfer_fees=transfer_fees, name=
     def sorted_offers(self):
         return list(sorted(self.offers.values(), key=lambda b: b.price / b.energy))
 
+    def get_bids(self):
+        return self.bids
```
hannesdiedrich: Can you use the deepcopy here to, please? Just to be consistent? Thanks!

src/d3a/models/strategy/external_strategies/load.py
spyrostz: Unfortunately our Jira version does not allow to create hyperlinks to comments, but have a look at the first comment of [the story](https://gridsingularity.atlassian.net/browse/D3ASIM-1946). It is a separate issue, but since it was mentioned in the same bug I tried to solve it in the context of this PR. 
The advantage is that you do not have to wrap the response with json.dumps, which has to be done when issuing `publish`. `publish_json` handles it automatically by doing it inside the method body. 
Of course this could have been avoided if I had more time to write tests that validated the endpoints. Hopefully other bugs like this will be revealed once https://gridsingularity.atlassian.net/browse/TESTS-4 is implemented.

tests/test_inter_area_agent.py
spyrostz: Will do, thanks

PR #631, opened 2020-02-04
D3ASIM-1923: Changed the external PV and Load classes to Mixins, in o…

spyrostz:
…rder to be able to inject the external connection behavior to the superclasses of the PV and LoadHoursStrategy. Refactored device and market area name to be more easily readable.

### DIFF
--- a/src/d3a/models/strategy/predefined_load.py
+++ b/src/d3a/models/strategy/predefined_load.py
@@ -85,7 +85,6 @@ class DefinedLoadStrategy(LoadHoursStrategy):
         for each slot.
         :return: None
         """
-        # If use_market_maker_rate is true, overwrite final_buying_rate to market maker rate
         self.load_profile = read_arbitrary_profile(
             InputProfileTypes.POWER,
             self.daily_load_profile)
--- a/src/d3a/models/strategy/predefined_load.py
+++ b/src/d3a/models/strategy/predefined_load.py
@@ -23,7 +23,6 @@ from d3a.d3a_core.util import generate_market_slot_list
 from d3a.models.strategy.load_hours import LoadHoursStrategy
 from d3a.models.read_user_profile import read_arbitrary_profile
 from d3a.models.read_user_profile import InputProfileTypes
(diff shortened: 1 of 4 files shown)


src/d3a/models/strategy/predefined_load.py
```diff
@@ -80,17 +79,13 @@ def __init__(self, daily_load_profile,
...
         """
         Runs on activate event. Reads the power profile data and calculates the required energy
         for each slot.
         :return: None
         """
         # If use_market_maker_rate is true, overwrite final_buying_rate to market maker rate
-        if self.use_market_maker_rate:
-            self.area_reconfigure_event(final_buying_rate=GlobalConfig.market_maker_rate)
-        self._validate_rates()
-        self.bid_update.update_on_activate()
```
hannesdiedrich: This functionality is not needed any more? If so, i can not see the connection between this removal and the story context. Can you please explain?

spyrostz: The method name has also changed in this diff. Essentially in a previous PR last week, I refactored the event_activate method to 2 separate methods: event_activate_energy for energy-related characteristics that the event needs to handle, and event_activate_price for price-related characteristics of the event. The deleted lines are already handled by the base class implementation of event_activate_price, therefore these do not need to be duplicated in this class anymore, since this class does not alter the price characteristics, but only deals with energy profiles (hence energy characteristics).

The rationale behind the original refactoring was during the implementation of the external connection. I needed to isolate price-related methods of the class, because the external connection should deal with these and not the base strategy. If you need me to elaborate more on that ping me to discuss offline

hannesdiedrich: I see now, thanks (wasn't aware of this change, had a look now)!
And as DefinedLoadStrategy inherits from Loadhours, it also inherits event_activate_price. nice, ok!
Please remove the comment in line 88 here though, please. Thanks!

spyrostz: will do , thanks

PR #632, opened 2020-02-05
D3ASIM-1862: Ledgering grid fee only on buyers' side to avoid double …

faizan2590:
…counting of market_fee.

### DIFF
--- a/tests/test_stats.py
+++ b/tests/test_stats.py
@@ -209,7 +209,10 @@ def grid_fees():
         'street',
         children=[house1, house2],
         past_markets=[FakeMarket((_trade(2, make_iaa_name(house1), 3, make_iaa_name(house2),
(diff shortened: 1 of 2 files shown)


tests/test_stats.py
```diff
@@ -235,7 +235,7 @@ def test_energy_bills_report_correctly_market_fees(grid_fees):
     m_bills.update(grid_fees)
     result = m_bills.bills_results
     assert result["street"]["house1"]["market_fee"] == 0.04
-    assert result["street"]["house2"]["market_fee"] == 0.04
+    assert result["street"]["house2"]["market_fee"] == 0.0
```
spyrostz: I understand why you had to change this, but I think it would be better if you added one more trade that would change the market_fee of house2. That way you would test that the behavior is consistent for both markets.

PR #633, opened 2020-02-05
Feature/tests 4

spyrostz:


src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -182,6 +186,15 @@ def _area_reconfigure_prices(self, validate=True, **kwargs):
...
             super().event_tick()
+        else:
+            while len(self.pending_requests) > 0:
+                req = self.pending_requests.pop()
+                if req.request_type == "offer":
+                    self._offer_impl(req.arguments, req.response_channel)
+                elif req.request_type == "delete_offer":
+                    self._delete_offer_impl(req.arguments, req.response_channel)
+                else:
+                    assert False, f"Incorrect incoming request name: {req}"
```
hannesdiedrich: It looks like you reengineered the redis queue. Obviously I am wrong Can you please explain why you needed to add this? Thanks!

spyrostz: Thanks for asking this, because it is kind of important for the implementation here. Before this change, I encountered many weird errors which were caused by deleting an offer or bid on a market while a trade that was destined for this offer/bid happened on another market. This was causing many trades being interrupted (essentially trades were performed in one market but did not complete). The way that I selected to circumvent this, is not to handle offer/bid event mutations in real-time, but to accumulate these and handle them during the event_tick. That way we are sure that all offer and bid commands will be placed when it is possible and not during a trade.

PR #635, opened 2020-02-07
D3ASIM-1906: Moved price assertion from accept_bid accept_offer to th…

hannesdiedrich:
…e strategies.

### DIFF
--- a/src/d3a/models/strategy/__init__.py
+++ b/src/d3a/models/strategy/__init__.py
@@ -22,7 +22,7 @@ from uuid import uuid4
 
 from d3a.d3a_core.exceptions import SimulationException
 from d3a.models.base import AreaBehaviorBase
-from d3a.models.market.market_structures import Offer
+from d3a.models.market.market_structures import Offer, Bid
 from d3a_interface.constants_limits import ConstSettings
 from d3a.constants import REDIS_PUBLISH_RESPONSE_TIMEOUT
 from d3a.d3a_core.device_registry import DeviceRegistry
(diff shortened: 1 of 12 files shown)


src/d3a/models/strategy/storage.py
```diff
@@ -285,6 +287,16 @@ def event_tick(self):
...
         super().event_trade(market_id=market_id, trade=trade)
+
+        if isinstance(trade.offer, Offer) and trade.offer.seller == self.owner.name:
+            offer = [o for o in self.offers.sold[market_id] if o.id == trade.offer.id][0]
+            assert trade.offer.price / trade.offer.energy >= \
+                offer.price / offer.energy - FLOATING_POINT_TOLERANCE
+        elif isinstance(trade.offer, Bid) and trade.offer.buyer == self.owner.name:
+            bid = [b for b in self.get_posted_bids(market) if b.id == trade.offer.id][0]
+            assert trade.offer.price / trade.offer.energy <= \
+                bid.price/bid.energy + FLOATING_POINT_TOLERANCE
```
faizan2590: Would be better to abstract these checks into separate methods under `class BaseStrategy` since they are reimplemented for PV, Load and Storage strategy.

hannesdiedrich: Yes, you're right! I was thinking the same and will change this.

src/d3a/models/strategy/__init__.py
```diff
@@ -32,10 +32,25 @@
 from d3a.d3a_core.util import append_or_create_key
 from d3a.models.market.market_structures import trade_from_JSON_string, offer_from_JSON_string
 from d3a.d3a_core.redis_connections.redis_area_market_communicator import BlockingCommunicator
+from d3a.constants import FLOATING_POINT_TOLERANCE
 
 log = getLogger(__name__)
 
 
+def assert_if_trade_bid_price_is_too_high(strategy, market, trade):
```
spyrostz: I think you can make this function and the next as methods of the BaseStrategy class. That way you do not have to pass the self object as a first argument.

hannesdiedrich: You are right! Will do it.

PR #638, opened 2020-02-17
Feature/d3 asim 1979

hannesdiedrich:


src/d3a/models/area/redis_external_connection.py
```diff
@@ -70,7 +70,22 @@ def channel_unregister_callback(self, payload):
...
+        channel_market_stats = f"{self.area.slug}/market_stats"
 
         self.pubsub.subscribe(**{channel: self.channel_register_callback,
-                                 channel_unregister: self.channel_unregister_callback})
+                                 channel_unregister: self.channel_unregister_callback,
+                                 channel_market_stats: self.market_stats_callback})
         self.pubsub.run_in_thread(daemon=True)
+
+    def market_stats_callback(self, payload):
+        market_stats_respsonse_channel = f"{self.area.slug}/market_stats/response"
```
spyrostz: Trivial typo, should be `market_stats_response_channel `

src/d3a/models/area/stats.py
```diff
@@ -90,3 +94,26 @@ def cheapest_offers(self):
...
         return cheapest_offers
+
+    def min_max_avg_rate_market(self, time_slot):
+        out_dict = {"min_trade_rate": None,
+                    "max_trade_rate": None,
+                    "avg_trade_rate": None}
+        for market in self._markets.all_spot_markets:
+            if market.time_slot == time_slot and len(market.trades) > 0:
+                trade_rates = [trade.offer.price/trade.offer.energy for trade in market.trades]
+                out_dict["min_trade_rate"] = round(min(trade_rates), 6)
```
spyrostz: If these are calculated only for the external API, I would recommend either using our default precision constant from constants.py, or not round it at all.

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -141,15 +141,15 @@ def _area_stats(self, payload):
                                              self.connected):
             return
         try:
-            device_stats = {k: v for k, v in self.device.stats.aggregated_stats.items()
-                            if v is not None}
-            market_stats = {k: v for k, v in self.market_area.stats.aggregated_stats.items()
-                            if v is not None}
+            device_stats = self.device.stats.aggregated_stats["bills"]
+            market_stats = self.market_area.stats.min_max_avg_rate_market(
```
spyrostz: I am afraid there is a chance for concurrent access to the market stats, because this is executed in the Redis thread. Therefore if the stats are modified by the main thread while the Redis thread is reading them, an error will be raised. Despite that, let's not overengineer and wait for stress test to validate if I am right.

src/d3a/models/area/stats.py
hannesdiedrich: OK, using DEFAULT_PRECISION then. Thanks!

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: Yes, I was thinking about deleting the _area_stats (and the corresponding list_(device)stats function on the d3a-api-client, because these numbers are pushed via the market_cycle event anyways. 
The min/max/avg prices can be acquired via the list_market_stats anyways. What do you think about this?

hannesdiedrich: @spyrostz

src/d3a/models/area/redis_external_connection.py
```diff
@@ -70,7 +70,22 @@ def channel_unregister_callback(self, payload):
...
+        payload_data = json.loads(payload["data"])
+        ret_val = {"status": "ready",
+                   "market_stats":
+                       self.area.stats.get_market_price_stats(payload_data["market_slots"])}
+        self.redis_db.publish(market_stats_respsonse_channel, json.dumps(ret_val))
+
+    # def sub_to_market_stats(self):
+    #     channel = f"{self.area.slug}/market_stats"
+    #     self.pubsub.subscribe(**{channel: self.market_stats_callback})
+    #     self.pubsub.run_in_thread(daemon=True)
```
faizan2590: Are you planning to use it somewhere?

src/d3a/models/area/stats.py
faizan2590: We also have a util function `limit_float_precision` that might be better to maintain code uniformity.

src/d3a/models/area/redis_external_connection.py
hannesdiedrich: No, remnant. Already removed it locally.

src/d3a/models/strategy/external_strategies/load.py
spyrostz: @hannesdiedrich I agree, feel free to remove this. It makes sense IMO too, to read data directly from the market and not from the device, since we followed that approach.

src/d3a/models/area/__init__.py
```diff
@@ -313,7 +313,7 @@ def next_market(self):
 
     @property
     def current_market(self):
-        """Returns the 'current' market (i.e. the one currently 'running')"""
+        """Returns the 'last' market (i.e. the one that has be finished last)"""
```
faizan2590: To me personally only word `last` confuses me, method `current_market` also doesn't explains what would be returned. I would vouch to be a bit more verbose here, for instance rename the method `current_market` to something like `most_recent_past_market`

hannesdiedrich: Although i agree that current_market is not the right wording, I dont want to rename the function. I just wanted to adapt the doc string because this is a copy paste from "next_market".
I will change it to:
"""Returns the 'most recent past market' market (i.e. the one that has been finished last)""" 
OK with you @faizan2590 ?

hannesdiedrich:
@faizan2590 It does not really matter in which order i adapt it. ATM it is not crucia IMO. Will get to 1980 right after this.

src/d3a/models/area/__init__.py
faizan2590: But I think it is trivial to rename the method `current_market` --> `most_recent_past_market`. Or if you think it needs to be done more carefully, please create a small sub task for future.

hannesdiedrich: @spyrostz do you agree with the renaming? If so, I will add it in this PR.

spyrostz: Frankly, I am by now used to the term current_market. And it kind of makes sense to my eyes, because all markets refer to the future, thus the current market (last past market) refers to the current point in time. I predict that even if we rename it to last_past_market, we are going to have the opposite problem, meaning that one might expect that the last past market is not the current time, but one slot before that.
Despite that, if you both agree that the renamed version looks better feel free to do this.

PR #639, opened 2020-02-17
Feature/d3 asim 1922

faizan2590:


src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -50,3 +51,55 @@ def unregister_area(redis, device_name, is_connected):
             {"status": "error",
              "error_message": f"Error when unregistering from area {device_name}."})
         return is_connected
+
+
+class ExternalMixin:
```
hannesdiedrich: Thanks for abstracting this! I was about to ask this todo it in my story.

src/d3a/models/state.py
```diff
@@ -175,6 +174,7 @@ def clamp_energy_to_sell_kWh(self, market_slot_time_list):
                                                             energy / len(market_slot_time_list),
                                                             self.max_offer_energy_kWh(time_slot),
                                                             self._battery_energy_per_slot))
+            self.energy_to_sell_dict[time_slot] = storage_dict[time_slot]
```
hannesdiedrich: I dont see another place where this dict is used. Is there a plan to use it a future ticket?

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -126,6 +105,7 @@ def _bid_impl(self, arguments, bid_response_channel):
             )
             self.redis.publish_json(bid_response_channel,
                                     {"status": "ready", "bid": bid.to_JSON_string()})
+            print(f"bid : {bid}")
```
hannesdiedrich: Please remove this and all other prints before merging. Thanks!

src/d3a/setup/strategy_tests/external_devices.py
```diff
@@ -1,17 +1,14 @@
...
 This program is free software: you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
 the Free Software Foundation, either version 3 of the License, or
 (at your option) any later version.
-
 This program is distributed in the hope that it will be useful,
 but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.
-
```
hannesdiedrich: Please revert this to be consistent in all files. Thanks!

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -0,0 +1,261 @@
...
+    """
+    Mixin for enabling an external api for the storage strategies.
+    Should always be inherited together with a superclass of StorageStrategy.
+    """
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.pending_requests = []
+
+    def event_activate(self):
+        super().event_activate()
```
hannesdiedrich: ExternalMixin class does not have a event_activate() method. Or am i missing something?

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -0,0 +1,261 @@
...
+        except Exception as e:
+            logging.error(f"Error when handling offer create on area {self.device.name}: "
+                          f"Exception: {str(e)}, Offer Arguments: {arguments}")
+            self.redis.publish_json(
+                response_channel,
+                {"status": "error",
+                 "error_message": f"Error when handling offer create "
+                                  f"on area {self.device.name} with arguments {arguments}."})
+
+    def _list_bids(self, payload):
```
hannesdiedrich: I wonder whether we add the list_offers and list_bids also to the ExternalMixin as they are used on multiple occasions (i know, a load should not be able to call list_offers, but we could disable this by checking for the type. 
What you you think. Also tagging @spyrostz

hannesdiedrich: If you continue this point, functions like _delete_offer, _offer etc, too actually.

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -0,0 +1,261 @@
...
+            self.state.market_cycle(self.market_area.current_market.time_slot,
+                                    self.market.time_slot)
+            self.state.clamp_energy_to_sell_kWh([self.market.time_slot])
+            market_event_channel = f"{self.device.name}/market_event"
+            current_market_info = self.market.info
+            current_market_info["free_storage"] = self.state.free_storage(self.market.time_slot)
+            current_market_info["used_storage"] = self.state.used_storage
+            current_market_info["min_allowed_soc_ratio"] = self.state.min_allowed_soc_ratio
+            current_market_info["capacity"] = self.state.capacity
+            current_market_info["max_abs_battery_power_kW"] = self.state.max_abs_battery_power_kW
```
hannesdiedrich: These values should be known beforehand right? IMO we shouldnt clamp our message queue with this redundant stuff.

src/d3a/models/strategy/external_strategies/storage.py
spyrostz: This class is a mixin too though, and it depends on a normal strategy class. Therefore, this is not the event_activate of the EventMixin, but the event_activate of the strategy.

src/d3a/models/strategy/external_strategies/storage.py
spyrostz: I would recommend having 2 different mixins, one for the bids and one for the offers, if we plan to go this way. If 2 separate mixins are used, then the load can use only the bid-related one, the pv the offer-related one, and storage both. However, if it feels like an overkill feel free to skip it, it does not look that important to my eyes

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -0,0 +1,261 @@
...
+            )
+        else:
+            self.pending_requests.append(
+                IncomingRequest("bid", arguments, bid_response_channel))
+
+    def _bid_impl(self, arguments, bid_response_channel):
+        try:
+            self.state.clamp_energy_to_buy_kWh([self.market.time_slot])
+            max_energy = self.state.energy_to_buy_dict[self.market.time_slot]
+            assert arguments["energy"] <= max_energy
```
spyrostz: Please do not assert here, and move this check in the _bid method, and also validate that the error response message is sent to the user. The general pattern here is that, validitation of the api arguments should be performed in the _bid, _offer .... methods (and respond with an error to the user), and handle errors when performing market operations in the *_impl methods.

src/d3a/models/strategy/external_strategies/storage.py
spyrostz: In addition to that, there are 2 stats that would be very useful and are currently missing, the amount of available energy to buy and sell in this market slot. I would recommend to add these 2 values and remove min_allowed_soc_ratio, capacity, max_abs_battery_power_kW

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -0,0 +1,261 @@
...
+        else:
+            self.pending_requests.append(
+                IncomingRequest("offer", arguments, offer_response_channel))
+
+    def _offer_impl(self, arguments, response_channel):
+        try:
+            energy_sell_dict = self.state.clamp_energy_to_sell_kWh([self.market.time_slot])
+            assert energy_sell_dict[self.market.time_slot] >= arguments['energy']
+            expected_energy = self.state.used_storage * (1 - self.state.min_allowed_soc_ratio)
+            assert arguments['energy'] <= expected_energy
```
spyrostz: Same as below, these assertions should be moved in the _offer method.

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -0,0 +1,261 @@
...
+            self.redis.publish_json(
+                bid_response_channel,
+                {"status": "error",
+                 "error_message": f"Error when handling bid create "
+                                  f"on area {self.device.name} with arguments {arguments}."})
+
+    def event_market_cycle(self):
+        self.register_on_market_cycle()
+        if self.connected:
+            self.state.market_cycle(self.market_area.current_market.time_slot,
```
spyrostz: I hope that these are the only things that are needed here, and we have not missed anything else. For future reference, and as a potential improvement in the future, it would be much better to split the market_cycle event in the storage class to price/energy functions, to mimic the pattern of the pv and the load. Not required for now though

src/d3a/models/strategy/external_strategies/storage.py
hannesdiedrich: Please skip the refactoring then. Thanks!

src/d3a/models/strategy/external_strategies/storage.py
faizan2590: Okay, then I assume, I also had to delete the offer that is posted [here](https://github.com/gridsingularity/d3a/pull/639/files#diff-1d553c99bd95475906818e1177a93f29R114) too.

src/d3a/models/strategy/external_strategies/storage.py
faizan2590: To my eyes this assert protects the [bid](https://github.com/gridsingularity/d3a/pull/639/files#diff-1d553c99bd95475906818e1177a93f29R201) to be posted in first place. Following the other approach would also need to delete the posted bid too.

spyrostz: I do not want to delete the bid actually. I would just like to move this check on the _bid method, instead of the _bid_impl method, in order to not even try to post an offer if the assert is violated.

src/d3a/models/strategy/external_strategies/storage.py
spyrostz: I think you misunderstood. There is no reason to delete any offer, since the offer is not posted yet. As mentioned below, what I suggest is to move this assert to the _offer method, which does not place an offer to the market, but only creates an incoming request that invokes _offer_impl on the event_tick.

src/d3a/models/state.py
faizan2590: Its needed to be sure if external agent is not over-selling. Also added an assert in recent [commit](https://github.com/gridsingularity/d3a/pull/639/files#diff-1d553c99bd95475906818e1177a93f29R95)

src/d3a/setup/strategy_tests/external_ess_offers.py
```diff
@@ -33,7 +33,7 @@ def get_setup(config):
                 'House 1',
                 [
                     Area('load', strategy=LoadHoursStrategy(
-                        hrs_of_day=list(range(24)), hrs_per_day=24, avg_power_W=4000,
+                        hrs_of_day=list(range(2, 24)), hrs_per_day=20, avg_power_W=4000,
```
hannesdiedrich: I guess this it to account for the integrationtest in d3a-api-client?

faizan2590: Exactly, as I only want it to be tested with external_ess_agent's offer

PR #642, opened 2020-02-19
Feature/d3 asim 1959

spyrostz:


src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -126,3 +127,14 @@ def _dispatch_event_tick_to_external_agent(self):
...
             self.redis.publish_json(tick_event_channel, current_tick_info)
+
+    def event_trade(self, market_id, trade):
+        super().event_trade(market_id=market_id, trade=trade)
+        if self.connected:
+            trade_dict = json.loads(trade.to_JSON_string())
+            trade_dict.pop('already_tracked', None)
+            trade_dict.pop('offer_bid_trade_info', None)
+            trade_dict.pop('seller_origin', None)
+            trade_dict.pop('buyer_origin', None)
```
faizan2590: Could there be any harm if the trade info is reported with `seller_origin` & `buyer_origin` ?

spyrostz: IMO yes, because you would reveal who you bought the energy from, which IMO should not be revealed for the hackathon.

PR #644, opened 2020-02-20
Feature/d3 asim 1991

spyrostz:
Better to check this PR per-commit. First commit deals with refactoring of the channel structure, in order to be more conveniently subscribed in d3a-web. Second commit deals with converting the strategy to external strategy types.

### DIFF
--- a/src/d3a/constants.py
+++ b/src/d3a/constants.py
@@ -36,7 +36,7 @@ DISPATCH_EVENTS_BOTTOM_TO_TOP = True
 # 20% of the slot length
 DISPATCH_EVENT_TICK_FREQUENCY_PERCENT = 20
 
-REDIS_JOB_ID = ""
+COLLABORATION_ID = ""
 # Controls whether the external connection is for use with the redis api client
 # or with th3 d3a-web.
 # False -> Redis api client
--- a/src/d3a/constants.py
+++ b/src/d3a/constants.py
@@ -35,3 +35,10 @@ DISPATCH_EVENTS_BOTTOM_TO_TOP = True
 # Controls how often will event tick be dispatched to external connections. Defaults to
 # 20% of the slot length
 DISPATCH_EVENT_TICK_FREQUENCY_PERCENT = 20
+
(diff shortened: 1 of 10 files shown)


src/d3a/constants.py
```diff
@@ -35,3 +35,10 @@
 # Controls how often will event tick be dispatched to external connections. Defaults to
 # 20% of the slot length
 DISPATCH_EVENT_TICK_FREQUENCY_PERCENT = 20
+
+COLLABORATION_ID = ""
+# Controls whether the external connection is for use with the redis api client
+# or with th3 d3a-web.
```
hannesdiedrich: slight typo 3->e

src/d3a/constants.py
```diff
@@ -35,3 +35,10 @@
...
+# False -> Redis api client
+# True -> d3a-web
+EXTERNAL_CONNECTION_REDIS_OR_WEB = False
```
hannesdiedrich: Minor but EXTERNAL_CONNECTION_WEB would fit better IMO, because it is a boolean variable.

src/d3a/constants.py
```diff
@@ -35,3 +35,10 @@
 # Controls how often will event tick be dispatched to external connections. Defaults to
 # 20% of the slot length
 DISPATCH_EVENT_TICK_FREQUENCY_PERCENT = 20
+
+COLLABORATION_ID = ""
```
faizan2590: Can we move these constants to `d3a-interface` too?

src/d3a/constants.py
spyrostz: Thanks, fixing that.

src/d3a/constants.py
spyrostz: Agreed, fixing this

src/d3a/constants.py
spyrostz: These constants only apply to d3a, therefore I think it is wrong to move them to d3a-interface

PR #645, opened 2020-02-21
Feature/d3 asim 1979

hannesdiedrich:


src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -0,0 +1,28 @@
...
+        self.redis_db.publish(channel, data)
+
+    def sub_to_area_event(self):
+        channel_market_stats = f"{self.area.slug}/market_stats"
+
+        self.pubsub.subscribe(**{channel_market_stats: self.market_stats_callback})
+        self.pubsub.run_in_thread(daemon=True)
+
+    def market_stats_callback(self, payload):
+        market_stats_response_channel = f"{self.area.slug}/market_stats/response"
```
spyrostz: This will fail with the new channel structure, but don't worry about it, I will fix it in the context of 1943

src/d3a/models/area/stats.py
```diff
@@ -90,3 +95,39 @@ def cheapest_offers(self):
...
+        return out_dict
+
+    @property
+    def current_market(self):
+        past_markets = list(self._markets.past_markets.values())
+        return past_markets[-1] if len(past_markets) > 0 else None
+
+    def get_market_price_stats(self, market_slot_list):
+        if self.current_market is None:
+            return {"INFO": "No market stats available yet"}
```
spyrostz: I would prefer if, instead of info, there was the source of the message, `market_stats`

src/d3a/models/area/stats.py
```diff
@@ -90,3 +95,39 @@ def cheapest_offers(self):
...
+                trade_volumes = [trade.offer.energy for trade in market.trades]
+                trade_rates = [trade.offer.price/trade.offer.energy for trade in market.trades]
+                out_dict["min_trade_rate"] = limit_float_precision(min(trade_rates))
+                out_dict["max_trade_rate"] = limit_float_precision(max(trade_rates))
+                out_dict["avg_trade_rate"] = limit_float_precision(mean(trade_rates))
+                out_dict["total_traded_energy_kWh"] = limit_float_precision(sum(trade_volumes))
+        return out_dict
+
+    @property
+    def current_market(self):
```
spyrostz: It is a bummer that this property is on the area, and we have to duplicate it here. It's ok for now though.

hannesdiedrich: Yeah, I was thinking about this a lot, but i could not come up with a different approach. We could skip the tests in get_market_price_stats and just dont return anything if there is no data, but i thought it might be helpful for the user.

src/d3a/models/area/stats.py
hannesdiedrich: in RedisMarketExternalConnection.market_stats_callback the ret val of this function is already assigne tot he member "market_stats". IMO your proposal would duplicate this. 
But i am happy about other renaming proposals.

PR #646, opened 2020-02-23
D3ASIM-1995: Added restrictions on max amount of posted energy for bi…

spyrostz:
…ds and offers for loads and pvs respectively.


src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -87,6 +87,15 @@ def _bid(self, payload):
             arguments = json.loads(payload["data"])
             assert set(arguments.keys()) == {'price', 'energy'}
             arguments['buyer_origin'] = self.device.name
+            posted_energy = (arguments['price'] + self.posted_bid_energy(self.market.id)) * 1000.0
```
hannesdiedrich: I think you meant `arguments['energy']` here, right?

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -92,6 +92,16 @@ def _offer(self, payload):
             assert set(arguments.keys()) == {'price', 'energy'}
             arguments['seller'] = self.device.name
             arguments['seller_origin'] = self.device.name
+
+            posted_energy = (arguments['price'] + self.posted_offer_energy(self.market.id))
```
hannesdiedrich: Same as above.

src/d3a/models/strategy/external_strategies/load.py
spyrostz: Should be fixed now, thanks!

src/d3a/models/strategy/external_strategies/pv.py
spyrostz: Fixed, thanks

PR #651, opened 2020-02-25
D3ASIM-1943: Fixed slowdown of simulation when triggered from d3a-web…

spyrostz:
…. Fixed available energy reporting of PV external interface. Added command name to register and unregister commands.

### DIFF
--- a/src/d3a/d3a_core/simulation.py
+++ b/src/d3a/d3a_core/simulation.py
@@ -304,7 +304,7 @@ class Simulation:
                     if console is not None:
                         self._handle_input(console, diff_slowdown)
                     else:
-                        sleep(diff_slowdown / 100.0)
+                        sleep(diff_slowdown)
 
                 if ConstSettings.GeneralSettings.RUN_REAL_TIME:
                     sleep(abs(tick_lengths_s - realtime_tick_length))
--- a/src/d3a/d3a_core/simulation.py
+++ b/src/d3a/d3a_core/simulation.py
@@ -303,6 +303,8 @@ class Simulation:
                     log.trace("Slowdown: %.4f", diff_slowdown)
                     if console is not None:
                         self._handle_input(console, diff_slowdown)
+                    else:
+                        sleep(diff_slowdown / 100.0)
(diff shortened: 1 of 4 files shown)


src/d3a/d3a_core/simulation.py
```diff
@@ -303,6 +303,8 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
                     log.trace("Slowdown: %.4f", diff_slowdown)
                     if console is not None:
                         self._handle_input(console, diff_slowdown)
+                    else:
+                        sleep(diff_slowdown / 100.0)
```
hannesdiedrich: I was looking into the code again, and I dont think that the `/100` is right here. When you look  into the formula of diff_slowdown: `diff_slowdown = tick_diff * self.slowdown / SLOWDOWN_FACTOR` where SLOWDOWN_FACTOR=100, its perfectly fine to sleep for diff_slowdown seconds. Dont you think?

spyrostz: division by100 is not for the slowdown factor, but for matching the timeout in this line https://github.com/gridsingularity/d3a/pull/651/files#diff-fa25e2850120fece94bc9fbdfbd1266eR354

hannesdiedrich: Sorry, the link you posted it reffering to this comment. 
What i mean is, diff_slowdown is already the sleeptime, no need to devide it by 100.

spyrostz: You are right, apparently the line URL failed in this case https://github.com/gridsingularity/d3a/blob/master/src/d3a/d3a_core/simulation.py#L352 here is the correct URL. You can see here that the diff_slowdown is divided by 100 once more.

hannesdiedrich: yes, i dont understand it either, and i still dont understand how the slowdown works now. Still not understanding the /100 . Can we breifly discuss offline? 
I know, it seams minor, but i have the feeling that this will come up again.

PR #653, opened 2020-02-26
D3ASIM-1995: Reversed the logic on the if block which allows an offer…

spyrostz:
… or bid to be posted. Forced rejection of all pending requests of the past market on the market_cycle event.

### DIFF
--- a/src/d3a/models/strategy/external_strategies/pv.py
+++ b/src/d3a/models/strategy/external_strategies/pv.py
@@ -13,7 +13,6 @@ class PVExternalMixin(ExternalMixin):
     """
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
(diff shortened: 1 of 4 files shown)


src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -13,7 +13,6 @@ class PVExternalMixin(ExternalMixin):
     """
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.pending_requests = []
```
hannesdiedrich: Why not  initiating the list here?

hannesdiedrich: I see now, you are calling _reject_all_pending_requests on every market cycle. Thanks!

PR #657, opened 2020-02-28
D3ASIM-2010: Added penalty energy and penalty rate calculation to the…

spyrostz:
… Load and PV strategy energy bills. Added integration test that validates that the Redis endpoints are actually containing the correct value.

### DIFF
--- a/integration_tests/steps/kpi.py
+++ b/integration_tests/steps/kpi.py
@@ -47,7 +47,6 @@ def device_reports_penalties(context, device_name, penalty_energy):
     house1 = list(filter(lambda x: x.name == "House 1", context.simulation.area.children))[0]
     device = list(filter(lambda x: device_name in x.name, house1.children))[0]
     bills = context.simulation.endpoint_buffer.market_bills.bills_redis_results
-    print(bills[str(device.uuid)]["penalty_energy"])
     assert isclose(bills[str(device.uuid)]["penalty_energy"], float(penalty_energy))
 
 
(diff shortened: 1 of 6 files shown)


integration_tests/steps/kpi.py
```diff
@@ -21,6 +21,7 @@
 import ast
 from behave import then
 from math import isclose
+from d3a.constants import DEVICE_PENALTY_RATE
```
faizan2590: Added a a constant [parameter](https://github.com/gridsingularity/d3a-interface/blob/master/d3a_interface/constants_limits.py#L61) in d3a-interface lately

spyrostz: Why percentage ? I expected that the penalty is just an energy rate based on the energy deficit / surplus of the load/ pv respectively.

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -107,6 +110,26 @@ def _get_child_data(self, area):
...
+                area.strategy.state.available_energy_kWh.get(market.time_slot, 0)
+                for market in self._get_past_markets_from_area(area.parent, "past_markets"))
+        else:
+            return
+        if "penalty_energy" not in results_dict:
+            results_dict["penalty_energy"] = 0.0
+            results_dict["penalty_cost"] = 0.0
+
+        results_dict["penalty_energy"] += penalty_energy
+        results_dict["penalty_cost"] += penalty_energy * DEVICE_PENALTY_RATE
```
hannesdiedrich: I think the bills are accumulated in Euro or Dollar not in cents. But DEVICE_PENALTY_RATE is in cents AFAIU.
Right?

spyrostz: Correct! This calculation is currently returning cents, therefore I will convert it to Euro. Thanks!

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -107,6 +110,26 @@ def _get_child_data(self, area):
                         for child in area.children}
             return self.bills_results[area.name]
 
+    def _store_area_penalties(self, results_dict, area):
```
faizan2590: Correct me if Im wrong but this method would be called at every_tick?

spyrostz: No, this is called on market cycle

integration_tests/steps/kpi.py
spyrostz: @faizan2590  since it was clarified by Sarah that device penalty is not a percentage but just an energy rate, I will remove the `PENALTY_PERCENTAGE ` parameter from the d3a-interface.

PR #663, opened 2020-03-06
fixed fabfile.

hannesdiedrich:


fabfile.py
```diff
@@ -161,4 +160,5 @@ def reqs():
 @task
 @hosts('localhost')
 def write_default_settings_file():
+    from default_settings_to_json_file import export_default_settings_to_json_file
```
faizan2590: Sorry, can you help me understand why lazy import is important here?

hannesdiedrich: We had this already like this, but somebody put it up to the top again.
The problem is, when you run `fab sync` export_default_settings_to_json_file is also imported , that need pendulum. So either you install pendulum yourself before running fab sync or you only import export_default_settings_to_json_file when you need it, namely only when you want to run `fab write_default_settings_file` Does it make sense?

faizan2590: Please leave a comment above this line as it has to be moved to the top before doing `fab write_default_settings_file `

PR #665, opened 2020-03-10
D3ASIM-2087: Adding assertion meesage when external client places bid…

faizan2590:
… after surpasing its ess_limit for that market_slot


src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -215,6 +211,12 @@ def _bid(self, payload):
 
     def _bid_impl(self, arguments, bid_response_channel):
         try:
+            max_energy = self.state.energy_to_buy_dict[self.market.time_slot]
+            cumulative_allowed = \
+                arguments["energy"] + self.state.pledged_buy_kWh[self.market.time_slot] + \
+                self.state.offered_buy_kWh[self.market.time_slot]
+            assert cumulative_allowed <= max_energy, \
+                "ESS limit for this market_slot has been surpassed"
```
hannesdiedrich: How about changing "limit" to "energy limit" ?

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -176,8 +176,9 @@ def _delete_bid(self, payload):
 
     def _delete_bid_impl(self, arguments, response_channel):
         try:
+            self.state.offered_buy_kWh[self.market.time_slot] -= \
+                self.market.bids[arguments['bid']].energy
             self.remove_bid_from_pending(arguments["bid"], self.market.id)
-            self.state.offered_buy_kWh[self.market.time_slot] -= arguments["bid"]["energy"]
```
spyrostz: I think the fix needs to be ported to the offer validation too, because I see in the delete_offer_implementation an `arguments["offer"]["energy"]`, which is obviously incorrect.

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -73,9 +73,10 @@ def _delete_offer(self, payload):
 
     def _delete_offer_impl(self, arguments, response_channel):
         try:
+            self.state.offered_sell_kWh[self.market.time_slot] -= \
```
spyrostz: Outside of the context of this story, but we had a discussion with @hannesdiedrich in his PR that it would be better to recalculate the offered_sell_kWh from the energy of the posted offers after the offer was deleted. I think that he is dealing with it in his PR though, correct me if I am wrong.

hannesdiedrich: Yes, I am deeling with this here: https://github.com/gridsingularity/d3a/pull/666

spyrostz: LGTM then :)

PR #666, opened 2020-03-10
D3ASIM-2084: Added possibility to remove all posted bids via external…

hannesdiedrich:
… load and storage strategies.

### DIFF
--- a/src/d3a/models/strategy/__init__.py
+++ b/src/d3a/models/strategy/__init__.py
@@ -397,6 +397,9 @@ class BidEnabledStrategy(BaseStrategy):
         posted_energy = (bid_energy + self.posted_bid_energy(market.id))
         return posted_energy <= required_energy_kWh
 
+    def is_bid_posted(self, market, bid_id):
+        return bid_id in [bid.id for bid in self.get_posted_bids(market)]
+
     def posted_bid_energy(self, market_id):
         if market_id not in self._bids:
             return 0.0
@@ -406,21 +409,19 @@ class BidEnabledStrategy(BaseStrategy):
         market = self.area.get_future_market_from_id(market_id)
         if market is None:
             return
+        posted_bid_ids = [bid.id for bid in self.get_posted_bids(market)]
+        if posted_bid_ids == []:
+            return []
         if bid_id is None:
-            # delete all bids associated with this strategy
-            deleted_bids_dict = {b_id: bid.energy for b_id, bid in market.bids.items()
-                                 if bid.buyer == self.owner.name}
-            for b_id in deleted_bids_dict:
-                market.delete_bid(b_id)
-            self._bids[market.id] = []
+            deleted_bid_ids = posted_bid_ids
         else:
-            # delete only one bid
-            deleted_bids_dict = {b_id: bid.energy for b_id, bid in market.bids.items()
-                                 if bid.id == bid_id}
-            if bid_id in market.bids.keys():
-                market.delete_bid(bid_id)
-            self._bids[market.id] = [bid for bid in self._bids[market.id] if bid.id != bid_id]
-        return deleted_bids_dict
+            deleted_bid_ids = [bid_id]
+        for b_id in deleted_bid_ids:
+            if b_id in market.bids.keys():
+                market.delete_bid(b_id)
+        self._bids[market.id] = [bid for bid in self._bids[market.id]
+                                 if bid.id not in deleted_bid_ids]
+        return deleted_bid_ids
 
     def add_bid_to_posted(self, market_id, bid):
         if market_id not in self._bids.keys():
@@ -464,7 +465,7 @@ class BidEnabledStrategy(BaseStrategy):
 
     def get_posted_bids(self, market):
         if market.id not in self._bids:
-            return {}
+            return []
         return self._bids[market.id]
 
     def event_bid_deleted(self, *, market_id, bid):
--- a/src/d3a/models/strategy/__init__.py
+++ b/src/d3a/models/strategy/__init__.py
@@ -402,13 +402,25 @@ class BidEnabledStrategy(BaseStrategy):
             return 0.0
         return sum(b.energy for b in self._bids[market_id])
 
-    def remove_bid_from_pending(self, bid_id, market_id):
+    def remove_bid_from_pending(self, market_id, bid_id=None):
         market = self.area.get_future_market_from_id(market_id)
         if market is None:
             return
-        if bid_id in market.bids.keys():
-            market.delete_bid(bid_id)
-        self._bids[market.id] = [bid for bid in self._bids[market.id] if bid.id != bid_id]
+        if bid_id is None:
+            # delete all bids associated with this strategy
+            deleted_bids_dict = {b_id: bid.energy for b_id, bid in market.bids.items()
+                                 if bid.buyer == self.owner.name}
+            for b_id in deleted_bids_dict:
+                market.delete_bid(b_id)
+            self._bids[market.id] = []
+        else:
+            # delete only one bid
+            deleted_bids_dict = {b_id: bid.energy for b_id, bid in market.bids.items()
+                                 if bid.id == bid_id}
+            if bid_id in market.bids.keys():
+                market.delete_bid(bid_id)
+            self._bids[market.id] = [bid for bid in self._bids[market.id] if bid.id != bid_id]
+        return deleted_bids_dict
 
     def add_bid_to_posted(self, market_id, bid):
         if market_id not in self._bids.keys():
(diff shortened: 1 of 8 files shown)


src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -176,11 +176,14 @@ def _delete_bid(self, payload):
 
     def _delete_bid_impl(self, arguments, response_channel):
         try:
-            self.remove_bid_from_pending(arguments["bid"], self.market.id)
-            self.state.offered_buy_kWh[self.market.time_slot] -= arguments["bid"]["energy"]
+            deleted_bids_dict = self.remove_bid_from_pending(self.market.id, arguments["bid"])
```
spyrostz: I think this is not enough. You need to at least handle the lack of the `bid` argument, which will implicitly be treated as  'delete all bids'.

src/d3a/models/strategy/__init__.py
```diff
@@ -402,13 +402,25 @@ def posted_bid_energy(self, market_id):
             return 0.0
         return sum(b.energy for b in self._bids[market_id])
 
-    def remove_bid_from_pending(self, bid_id, market_id):
+    def remove_bid_from_pending(self, market_id, bid_id=None):
```
spyrostz: I assume the reversal of the  argument order is not really necessary, mostly cosmetics for matching the order of the EventMixin methods argument order (the market precedes the other arguments in these methods IIRC), or am I mistaken?

src/d3a/models/strategy/__init__.py
```diff
@@ -402,13 +402,25 @@ def posted_bid_energy(self, market_id):
...
         market = self.area.get_future_market_from_id(market_id)
         if market is None:
             return
-        if bid_id in market.bids.keys():
-            market.delete_bid(bid_id)
-        self._bids[market.id] = [bid for bid in self._bids[market.id] if bid.id != bid_id]
+        if bid_id is None:
+            # delete all bids associated with this strategy
+            deleted_bids_dict = {b_id: bid.energy for b_id, bid in market.bids.items()
+                                 if bid.buyer == self.owner.name}
```
spyrostz: I know that this has a similar effect, but I would very much prefer to use the bid cache here and not access the market (self._bids are the bids posted by this strategy on a market). Second, the energy should not be needed as value of the dictionary.

src/d3a/models/strategy/__init__.py
```diff
@@ -402,13 +402,25 @@ def posted_bid_energy(self, market_id):
...
+            for b_id in deleted_bids_dict:
+                market.delete_bid(b_id)
```
spyrostz:
```suggestion
            for bid in self._bids[market_id].keys():
                market.delete_bid(bid.id)
```

src/d3a/models/strategy/__init__.py
```diff
@@ -402,13 +402,25 @@ def posted_bid_energy(self, market_id):
...
+            self._bids[market.id] = []
+        else:
+            # delete only one bid
+            deleted_bids_dict = {b_id: bid.energy for b_id, bid in market.bids.items()
+                                 if bid.id == bid_id}
+            if bid_id in market.bids.keys():
```
spyrostz: Same here, you should use the bid cache instead of the market bids IMO

src/d3a/models/strategy/external_strategies/storage.py
hannesdiedrich: The lack of the bid argument is handled in remove_bid_from_pending (case when bid is None).
Or do you mean something else?

src/d3a/models/strategy/__init__.py
hannesdiedrich: I wanted to be explicit and emphasise with the optional argument "bid" that there is an option to dont parse it or set it to None. For that i had to change the order.

src/d3a/models/strategy/external_strategies/storage.py
spyrostz: I meant that, if the endpoint receives a request with `arguments = {}` (without the bid argument at all), the endpoint should delete all bids. I know the function handles it, but only if we pass arguments["bid"] = None, not if we do not pass any argument at all (in this case it will crash with a `KeyError` in this line)

src/d3a/models/strategy/__init__.py
hannesdiedrich: I agree with you that using the `_bids` member for accessing energy (sorry, didnt think).
But we need the energy in the external storage to account for the already offered energy: https://github.com/gridsingularity/d3a/blob/af70782338868c2fb978367c6e820ae4c9e447ae/src/d3a/models/strategy/external_strategies/storage.py#L181

src/d3a/models/strategy/external_strategies/storage.py
hannesdiedrich: Ahh, ok. Will handle this.

spyrostz: Thanks!

src/d3a/models/strategy/__init__.py
spyrostz: I see now why the energy is required to be returned. In that case, I would prefer to make this calculation a bit more stateless, by exposing another function called `total_bid_energy` , which would automatically calculate the `self.state.offered_buy_kWh` by summing up the energy of all posted bids (`_bids`). That way you will not have to keep track of the state by decrementing the energy of the deleted bid, but it can be recalculated every time a bid is posted or deleted. If it is not that clear let's discuss offline

hannesdiedrich: Ok, I like that idea! Didnt like the extraction of energy in the `remove_bid_from_pending` function. 
So, will proceed your proposed way. I guess that this `total_bid_energy ` could live in the BidEnabledStrategy class right?

spyrostz: Exactly, it would be great to be a small function in this class, thanks

hannesdiedrich: OMG! there is already such function: `posted_bid_energy` Sorry!

src/d3a/models/strategy/__init__.py
```diff
@@ -397,18 +397,31 @@ def can_bid_be_posted(self, bid_energy, required_energy_kWh, market):
...
         if market is None:
             return
-        if bid_id in market.bids.keys():
-            market.delete_bid(bid_id)
-        self._bids[market.id] = [bid for bid in self._bids[market.id] if bid.id != bid_id]
+        posted_bid_ids = [bid.id for bid in self.get_posted_bids(market)]
+        if posted_bid_ids == []:
+            return []
+        if bid_id is None:
+            deleted_bid_ids = posted_bid_ids
```
spyrostz:
```suggestion
        if bid_id is None:
            deleted_bid_ids = [bid.id for bid in self.get_posted_bids(market)]
```

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -56,22 +56,27 @@ def _delete_bid(self, payload):
         try:
             arguments = json.loads(payload["data"])
             assert set(arguments.keys()) == {'bid'}
```
spyrostz: Actually what I meant before is that this clause should not be needed anymore. The endpoint should be able to receive an empty dict as argument (even without including the 'bid' member)

hannesdiedrich: Ok, ATM i set the bid=None in the delete_bid method in the api_client. 
What you mean is that stetting a bit when wanting to delete all bids is not logical, that is why you want to be able to send delete_bid without any arguments, right?

src/d3a/models/strategy/__init__.py
```diff
@@ -397,18 +397,31 @@ def can_bid_be_posted(self, bid_energy, required_energy_kWh, market):
         posted_energy = (bid_energy + self.posted_bid_energy(market.id))
         return posted_energy <= required_energy_kWh
 
+    def is_bid_posted(self, market, bid_id):
+        return bid_id in [bid.id for bid in self.get_posted_bids(market)]
+
     def posted_bid_energy(self, market_id):
```
faizan2590: Out of context of this PR but I just realised that this method should have a better name like `posted_bid_cumulative_energy `

hannesdiedrich: Hmm, I am not convinced. But if @spyrostz agrees, will rename it.

spyrostz: I also disagree. I think the name is concise and quite understandable. `Cumulative` is a term that I , in general, do not like the way we use it, because it implies that there is a reduce function, or that more results will be added in subsequent steps. I think `accumulated` or even better `sum`, is the correct term when we want to display the final result of an addition. But for this case, the function name is clear (to my eyes at least): energy of the posted bids, therefore no need to rename that.

src/d3a/models/strategy/__init__.py
```diff
@@ -118,7 +122,19 @@ def post(self, offer, market_id):
...
+        else:
+            to_delete_offers = [o for o, m in self.posted.items() if o.id == offer_id]
+        deleted_offer_ids = []
+        for offer in to_delete_offers:
+            market.delete_offer(offer.id)
+            self.remove(offer)
+            deleted_offer_ids.append(offer.id)
+        return deleted_offer_ids
+
+    def remove_offer_by_id(self, market_id, offer_id=None):
```
hannesdiedrich: I can delete this function now i just realised

PR #667, opened 2020-03-11
D3ASIM-2086: Exposing min/max/avg trade price to external agent.

faizan2590:


src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -98,11 +98,14 @@ def _area_stats(self, payload):
                             if v is not None}
             market_stats = {k: v for k, v in self.market_area.stats.aggregated_stats.items()
                             if v is not None}
+            trade_stats = {k: v for k, v in self.market_area.stats.rate_stats_market.items()
```
hannesdiedrich: I have two general comments:
1. I thought i deprecated _area_stats . It is not used any more (from the client side) 
2. AFAIU, the target of the story was just to add the median to min_max_avg_rate_current_market (in src/d3a/models/area/stats.py) 
Am I right?

PR #670, opened 2020-03-11
D3ASIM-2082: Added endpoints for device stats.

spyrostz:


src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -88,6 +88,28 @@ def _unregister(self, payload):
     def register_on_market_cycle(self):
         self.connected = self._connected
 
+    def _device_stats(self, _):
```
hannesdiedrich: I think we have to be careful now with the naming.
On several point in the code we have device_stats, device_statistics and similar. 
IMO at this point here (and for the endpoint channel) a name like device_info would fit more. But i am of course open for other names two. What do you think?

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -24,7 +24,8 @@ def event_activate(self):
             f'{self.channel_prefix}/bid': self._bid,
             f'{self.channel_prefix}/delete_bid': self._delete_bid,
             f'{self.channel_prefix}/list_bids': self._list_bids,
-            f'{self.channel_prefix}/stats': self._area_stats
+            f'{self.channel_prefix}/stats': self._area_stats,
```
hannesdiedrich: As discussed in the @faizan2590 s PR. I think we deprecated this endpoint.

spyrostz: Great, I'll delete it and the corresponding methods in the context of this PR.

src/d3a/models/strategy/external_strategies/__init__.py
spyrostz: I agree, and you are right, this is not device_statistics that we are sending here. It is actually the current device information (essentially information about the current available/required energy). Changing this endpoint to device_info in d3a-web too. Thanks

PR #675, opened 2020-03-16
D3ASIM-2100: Filtered out offers that need to be rejected for forward…

spyrostz:
…ing due to offer age, and being already part of forwarded offers. Optimised duration check by precalculating simulation end date. Added condition for importing the pandapower module only when the power flow global config is set to true.


integration_tests/power_flow_tests.feature
```diff
@@ -1,5 +1,6 @@
 Feature: Run power flow integration tests
 
+  @disabled
```
hannesdiedrich: This is temporary, right?

src/d3a/models/area/__init__.py
```diff
@@ -332,10 +332,7 @@ def current_balancing_market(self):
             return None
 
     def get_future_market_from_id(self, _id):
-        try:
-            return [m for m in self._markets.markets.values() if m.id == _id][0]
-        except IndexError:
-            return None
+        return self._markets.indexed_future_markets.get(_id, None)
```
hannesdiedrich: Ahh, nice! I guess that a dict lookup is less expensive than looping in a list, right?

src/d3a/models/market/__init__.py
```diff
@@ -123,7 +123,7 @@ def _update_stats_after_trade(self, trade, offer, buyer, already_tracked=False):
         self._update_accumulated_trade_price_energy(trade)
         self.traded_energy = add_or_create_key(self.traded_energy, offer.seller, offer.energy)
         self.traded_energy = subtract_or_create_key(self.traded_energy, buyer, offer.energy)
-        self._update_min_max_avg_trade_prices(offer.price / offer.energy)
+        self._update_min_max_avg_trade_prices(offer.energy_rate)
```
hannesdiedrich: Again, out of curiosity: Do you know how much resources are saved when one-time precalculating the offer rate in the offer instance  instead of doing it here all the time?

spyrostz: 3-4 seconds on a total of 67 seconds in one tick, in a 150-house simulation, which is not extreme, but contributes to the slowdown.

src/d3a/models/market/one_sided.py
```diff
@@ -82,7 +84,7 @@ def offer(self, price: float, energy: float, seller: str, seller_origin,
         offer = Offer(offer_id, price, energy, seller, original_offer_price,
                       seller_origin=seller_origin)
 
-        self.offers[offer.id] = deepcopy(offer)
+        self.offers[offer.id] = offer
```
hannesdiedrich: Hmm, but arent we falling into the memory leak? I dont remember, but are we deleting the connection from market to the offers before deleting the markets too?

src/d3a/models/strategy/__init__.py
```diff
@@ -185,6 +185,7 @@ def __init__(self):
         super(BaseStrategy, self).__init__()
         self.offers = Offers(self)
         self.enabled = True
+        self._allowed_disable_events = [AreaEvent.ACTIVATE, MarketEvent.TRADE]
```
hannesdiedrich: Very minor, but a name like "_allowed_events_when_disabled" would be more descriptive, IMO. What do you think?

src/d3a/models/market/one_sided.py
spyrostz: This was mostly a precaution, especially after removing the markets/agents offers when recycling the market. I did not observe any increased memory when removing this, but I'll give it one more check to be on the safe side. Thanks

src/d3a/models/area/__init__.py
spyrostz: Exactly. Even creating a list takes more time than accessing a dict key.

src/d3a/models/strategy/area_agents/one_sided_engine.py
```diff
@@ -91,12 +97,41 @@ def _delete_forwarded_offer_entries(self, offer):
...
+            return
+        if offer.id in self.forwarded_offers:
+            return
+        if offer.id in self.offer_age:
+            return
+        if offer.id not in self.markets.source.offers:
+            # TODO: Potential further optimisation here
+            # This case handles trade events that we receive but are not contained in any of
+            # the source and target market. This happens when another device accepts the offer
+            # before the offer event is dispatched to this market. Current solution is to
```
hannesdiedrich: But his only in a split event, right?

spyrostz: Indeed, but one can squeeze off some more performance from here. In general, the IAA is a place where it can become way way faster. I just did the obvious fixes, but did not dare to go further after one day of debugging to fix all integration tests.

src/d3a/models/strategy/__init__.py
spyrostz: Will do, thanks!

integration_tests/power_flow_tests.feature
spyrostz: The problem here is that power flow is controlled by a config, namely POWER_FLOW, and the Python interpreter (which is pointless as a distinction IMO, since there might be people that want to disable power flow, but still use Python interpreter). The fact that this is a config has the disadvantage that it is not controllable when running console integration tests (because there we do not have access to the config). Also, the default value of this is true, which is surprising considering that this feature is not complete. This leads to this integration test failing, because pandapower was removed as a dependency a while ago. 
This was some background on the issue, and the reason why I decided to ignore the mess and just disable the test. If we want to resolve this properly (which should probably happen in the context of the power flow subtasks) the power flow config needs to be moved to ConstantsLimits and this test needs to be converted to a non-console integration test, so that we can enable the power flow by changing the ConstantsLimits.

src/d3a/models/strategy/load_hours.py
```diff
@@ -309,8 +312,13 @@ def event_activate_energy(self):
...
     def active_markets(self):
-        return [market for market in self.area.all_markets
-                if self._is_market_active(market)]
+        return self._active_markets
+
+    def _calculate_active_markets(self):
+        self._active_markets = [
+            market for market in self.area.all_markets
+            if self._is_market_active(market)
+        ] if self.area else []
```
hannesdiedrich: Is this ever going to happen, that self.area is None?

src/d3a/models/strategy/area_agents/one_sided_agent.py
```diff
@@ -68,5 +68,9 @@ def event_offer_split(self, *, market_id,  original_offer, accepted_offer, resid
                                      accepted_offer=accepted_offer,
                                      residual_offer=residual_offer)
 
+    def event_offer(self, *, market_id, offer):
+        for engine in sorted(self.engines, key=lambda _: random()):
+            engine.event_offer(market_id=market_id, offer=offer)
+
```
hannesdiedrich: Hmm, i am confused. We never dispatched the offer event to the engines before?

src/d3a/models/strategy/storage.py
```diff
@@ -399,8 +401,10 @@ def buy_energy(self, market):
                 # Don't buy our own offer
                 continue
             # Check if the price is cheap enough
-            if (offer.price / offer.energy) > max_affordable_offer_rate:
-                continue
+            if offer.energy_rate > max_affordable_offer_rate:
+                # Can early return here, because the offers are sorted according to energy rate
+                # therefore the following offers will be more expensive
```
hannesdiedrich: Thanks!

src/d3a/models/strategy/area_agents/one_sided_agent.py
spyrostz: We used to send the offer event, and trigger automatic offer forwarding whenever the offer event was received. However we removed that some time ago.

tests/test_inter_area_agent.py
```diff
@@ -183,8 +183,8 @@ def bid(self, price: float, energy: float, buyer: str, seller: str,
         bid = Bid(bid_id, price, energy, buyer, seller,
                   original_bid_price=original_bid_price,
                   buyer_origin=buyer_origin)
-        self._bids.append(deepcopy(bid))
-        self.forwarded_bid = deepcopy(bid)
```
hannesdiedrich: Please make sure that these are also only precautions and dont safe us from a memeroy leak. Thanks a lot!

src/d3a/models/strategy/load_hours.py
spyrostz: This is more of a failsafe, I do not think that the active_markets are accessed during activate event.

integration_tests/power_flow_tests.feature
hannesdiedrich: Ok, thank you! Makes sense. Lets get to it when adding proper power flow.

src/d3a/models/market/one_sided.py
hannesdiedrich: ahh, wow. Ok, I hope that it is indeed only a precaution and we can omit the deepcopy. Thanks for checking again!

src/d3a/models/market/market_structures.py
```diff
@@ -33,6 +33,7 @@ def __init__(self, id, price, energy, seller, original_offer_price=None, seller_
         self.energy = energy
         self.seller = seller
         self.seller_origin = seller_origin
+        self.energy_rate = price / energy
```
faizan2590: Could be used [here](https://github.com/gridsingularity/d3a/pull/675/files#diff-286711868906f1b84a065750ab4e674fR40) and [here](https://github.com/gridsingularity/d3a/pull/675/files#diff-286711868906f1b84a065750ab4e674fR45) ?

src/d3a/models/market/market_structures.py
```diff
@@ -72,18 +74,21 @@ def offer_from_JSON_string(offer_string):
...
 
 
 class Bid(namedtuple('Bid', ('id', 'price', 'energy', 'buyer', 'seller',
-                             'original_bid_price', 'buyer_origin'))):
+                             'original_bid_price', 'buyer_origin', 'energy_rate'))):
     def __new__(cls, id, price, energy, buyer, seller, original_bid_price=None,
-                buyer_origin=None):
+                buyer_origin=None, energy_rate=None):
+        if energy_rate is None:
+            energy_rate = price / energy
```
faizan2590: Could be used in [line](https://github.com/gridsingularity/d3a/pull/675/files#diff-286711868906f1b84a065750ab4e674fR96)  and [line](https://github.com/gridsingularity/d3a/pull/675/files#diff-286711868906f1b84a065750ab4e674fR102)

faizan2590: Also [here](https://github.com/gridsingularity/d3a/pull/675/files#diff-286711868906f1b84a065750ab4e674fR102)

src/d3a/models/market/market_structures.py
faizan2590: [here](https://github.com/gridsingularity/d3a/pull/675/files#diff-286711868906f1b84a065750ab4e674fR68) too

src/d3a/models/market/one_sided.py
```diff
@@ -199,7 +201,7 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *, energy: in
         residual_offer = None
 
         if trade_rate is None:
-            trade_rate = offer.price / offer.energy
+            trade_rate = offer.energy_rate
```
faizan2590: I guess could be used [here](https://github.com/gridsingularity/d3a/pull/675/files#diff-5c5fbb99e0155bbce18f8ca72347fd58R388-R389) too

src/d3a/models/strategy/__init__.py
faizan2590: A bit out of of context but aren'T we sticking to small letters for [channel_names](https://github.com/gridsingularity/d3a/pull/675/files#diff-5c5fbb99e0155bbce18f8ca72347fd58R200)

spyrostz: These are not channel names, but event enums.

src/d3a/models/strategy/storage.py
```diff
@@ -388,6 +388,8 @@ def event_balancing_market_cycle(self):
                                                                               self.owner.name)
 
     def buy_energy(self, market):
+        if self.state.has_battery_reached_max_power(-0.0001, market.time_slot):
```
faizan2590: Wouldn'T it be better to use a FLOATING_POINT_TOLERANCE then a constant value here ?

src/d3a/models/strategy/__init__.py
faizan2590: Sorry for not clear in first place, I meant about the link I used in my former comment

spyrostz:
> Thanks a lot for taking care of this. One more potential improvement could be to check if methods related to populating `BALANCING_MARKET` is not called when it is disabled.
> Do you agree?

Balancing markets are not created when disabled as far as I checked, therefore I think we are safe.

src/d3a/models/strategy/storage.py
spyrostz: Indeed, fixing this

src/d3a/models/area/markets.py
```diff
@@ -54,20 +56,19 @@ def rotate_markets(self, current_time, stats, dispatcher):
                                   past_markets=self.past_balancing_markets,
                                   area_agent=dispatcher.balancing_agents)
         stats.update_accumulated()
+        self.indexed_future_markets = {
```
hannesdiedrich: How about adding this as a property  as it is used twice in this class?

src/d3a/models/market/market_structures.py
```diff
@@ -67,23 +69,31 @@ def _to_csv(self):
         return self.id, rate, self.energy, self.price, self.seller
 
 
+def copy_offer(offer):
```
hannesdiedrich: Can you please tell the reason for this? Dont you trust deepcopy any more ;-) ?

hannesdiedrich: Explained offline, deppcopy is slow.

tests/test_inter_area_agent.py
spyrostz: Checked it out, on the terminal memory does not seem to increase significantly with a large number of devices.

src/d3a/models/market/market_structures.py
spyrostz: I cannot see where the links refer to, but I did a global search and replaced the division with the energy_rate parameter. Thanks

src/d3a/models/market/market_structures.py
spyrostz: As discussed offline, its performance was far far slower than creating manually the same object.

src/d3a/models/area/markets.py
spyrostz: I'll do it, thanks!

src/d3a/models/area/event_dispatcher.py
```diff
@@ -162,9 +162,6 @@ def create_area_agents(self, is_spot_market, market):
                 min_offer_age=ConstSettings.IAASettings.MIN_OFFER_AGE
             )
 
-            self._delete_past_agents(self._inter_area_agents)
```
hannesdiedrich: So this is not needed any more and was just a precaution?
I see that you moved the deletion of the dispatchers into the area. But who deleted the `_inter_area_agents` ?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -116,7 +116,8 @@ def update_stats(self, area, simulation_status, progress_info):
         self.cumulative_grid_trades.update(area)
 
         self.market_bills.update(area)
-        self.balancing_bills.update(area)
+        if ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET:
+            self.balancing_bills.update(area)
```
hannesdiedrich: Nice!

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -166,4 +164,4 @@ def _gather_device_statistics(cls, area: Area, subdict: Dict, flat_result_dict:
             _create_or_append_dict(subdict, "production_kWh",
                                    {market.time_slot: area.strategy.energy_per_slot_kWh})
 
-        flat_result_dict[area.uuid] = deepcopy(subdict)
+        flat_result_dict[area.uuid] = subdict.copy()
```
hannesdiedrich: And this makes a significant difference?

faizan2590:
> > Thanks a lot for taking care of this. One more potential improvement could be to check if methods related to populating `BALANCING_MARKET` is not called when it is disabled.
> > Do you agree?
> 
> Balancing markets are not created when the corresponding config is disabled as far as I checked, therefore I think we are safe.

What I meant is results' reporting related to `balancing_market` even when the simulation is executed with ENABLE_BALANCING_MARKET=False.
Like for instance the `csvs` related to balancing market are still generated in case of  ENABLE_BALANCING_MARKET=False

spyrostz:
> > > Thanks a lot for taking care of this. One more potential improvement could be to check if methods related to populating `BALANCING_MARKET` is not called when it is disabled.
> > > Do you agree?
> > 
> > 
> > Balancing markets are not created when the corresponding config is disabled as far as I checked, therefore I think we are safe.
> 
> What I meant is results' reporting related to `balancing_market` even when the simulation is executed with ENABLE_BALANCING_MARKET=False.
> Like for instance the `csvs` related to balancing market are still generated in case of ENABLE_BALANCING_MARKET=False

Not relevant to performance, since the balancing markets will be empty anyways, therefore it does not spend CPU time populating them, and the purpose of this story is not to optimize how fast the csv files are generated, but the actual runtime of the simulation. Despite that, I added the fix to the bug of the balancing market that you describe.

src/d3a/d3a_core/sim_results/device_statistics.py
spyrostz: Indeed it does. As we discussed in the former comment, shallow copy is much much faster than deepcopy.

src/d3a/models/area/markets.py
```diff
@@ -54,20 +60,19 @@ def rotate_markets(self, current_time, stats, dispatcher):
                                   past_markets=self.past_balancing_markets,
                                   area_agent=dispatcher.balancing_agents)
         stats.update_accumulated()
+        self._indexed_future_markets = {
+            m.id: m for m in self.all_spot_markets
+        }
```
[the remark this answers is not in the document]
spyrostz: Ah I see what you mean now. Adapting it, thanks.

src/d3a/models/area/event_dispatcher.py
spyrostz: The deletion of the dispatcher includes the deletion of IAAs too. It is just moving this code to the area class that changed here. 
Reason for this change is the Markets class change in the rotate_markets method. The area_agents were popped there without being deleted. By moving the delete_past_agents to the Area class, I wanted to enforce the agents to be destroyed before the markets, in order to not fall into this pitfall again.

hannesdiedrich: Thanks for the explanation!

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -148,7 +148,8 @@ def _populate_plots_stats_for_supply_demand_curve(self, area):
 
     def update_plot_stats(self, area):
         self._get_stats_from_market_data(self.plot_stats, area, False)
-        self._get_stats_from_market_data(self.plot_balancing_stats, area, True)
+        if ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET:
+            self._get_stats_from_market_data(self.plot_balancing_stats, area, True)
```
hannesdiedrich: Thumbsup!

PR #676, opened 2020-03-16
D3ASIM-2083: Added finish event for connected/registered strategies, …

spyrostz:
…that gets transmitted at the end of the simulation.


src/d3a/models/strategy/__init__.py
```diff
@@ -390,6 +390,9 @@ def assert_if_trade_offer_price_is_too_low(self, market_id, trade):
     def can_offer_be_posted(self, offer_energy, available_energy, market):
         return self.offers.can_offer_be_posted(offer_energy, available_energy, market)
 
+    def deactivate(self):
+        pass
```
faizan2590: Is it really needed?

spyrostz: Yes, because then the strategy.deactivate call in the Area class would fail

PR #678, opened 2020-03-17
Delete bid impl early return when called even before the any bid was …

faizan2590:
…posted

### DIFF
--- a/src/d3a/models/strategy/__init__.py
+++ b/src/d3a/models/strategy/__init__.py
@@ -432,6 +432,8 @@ class BidEnabledStrategy(BaseStrategy):
             deleted_bid_ids = [bid.id for bid in self.get_posted_bids(market)]
         else:
             deleted_bid_ids = [bid_id]
+        if len(deleted_bid_ids) == 0:
+            return
         for b_id in deleted_bid_ids:
             if b_id in market.bids.keys():
                 market.delete_bid(b_id)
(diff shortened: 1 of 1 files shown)


src/d3a/models/strategy/__init__.py
```diff
@@ -432,6 +432,8 @@ def remove_bid_from_pending(self, market_id, bid_id=None):
             deleted_bid_ids = [bid.id for bid in self.get_posted_bids(market)]
         else:
             deleted_bid_ids = [bid_id]
+        if len(deleted_bid_ids) == 0:
```
spyrostz: I do not understand this actually. Even if deleted_bid_ids is an empty list, the former code should deal with it properly and return an empty list. Moreover, this implementation does not return anything in this line, while the rest of the code returns a list.

faizan2590: It will fail [here](https://github.com/gridsingularity/d3a/pull/678/files#diff-5c5fbb99e0155bbce18f8ca72347fd58L438) since `self._bids[market.id]` will give a key error when no bids were placed in first place

spyrostz: I see the problem now. For code uniformity, I would replace the `self._bids[market.id]` with the `self.get_posted_bids(market)` like it's done 2 lines above. If you choose to stick to this implementation though, please `return []` so that the return values are uniform.

src/d3a/models/strategy/__init__.py
```diff
@@ -435,7 +435,7 @@ def remove_bid_from_pending(self, market_id, bid_id=None):
         for b_id in deleted_bid_ids:
             if b_id in market.bids.keys():
                 market.delete_bid(b_id)
-        self._bids[market.id] = [bid for bid in self._bids[market.id]
+        self._bids[market.id] = [bid for bid in self.get_posted_bids(market)
```
hannesdiedrich: Ah! yes! Thats right! Thanks!

PR #679, opened 2020-03-18
D3ASIM-2131: Substracting posted bid energy before deleting.

faizan2590:


src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -120,11 +122,14 @@ def _offer_impl(self, arguments, response_channel):
                 arguments['energy'] + \
                 self.state.pledged_sell_kWh[self.market.time_slot] + \
                 self.state.offered_sell_kWh[self.market.time_slot]
-            assert cumulative_allowed <= self.state.energy_to_sell_dict[self.market.time_slot], \
+            print(f"cumulative_allowed: {cumulative_allowed}")
```
hannesdiedrich: please remove

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -80,6 +80,8 @@ def _delete_offer_impl(self, arguments, response_channel):
                 self.offers.remove_offer_from_cache_and_market(self.market, to_delete_offer_id)
             self.state.offered_sell_kWh[self.market.time_slot] -= \
```
spyrostz: Are you sure that this is the correct formula ? I was under the impression that offered_sell_kWh was the total amount of energy of the posted offers, therefore the `-=` should be replaced with `=` here. Please correct me if I have not understood the storage state operation correctly.

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -185,9 +190,11 @@ def _delete_bid(self, payload):
     def _delete_bid_impl(self, arguments, response_channel):
         try:
             to_delete_bid_id = arguments["bid"] if "bid" in arguments else None
-            deleted_bids = self.remove_bid_from_pending(self.market.id, bid_id=to_delete_bid_id)
             self.state.offered_buy_kWh[self.market.time_slot] -= \
```
spyrostz: Same comment as for the offers applies here, for the bids.

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -80,6 +80,8 @@ def _delete_offer_impl(self, arguments, response_channel):
                 self.offers.remove_offer_from_cache_and_market(self.market, to_delete_offer_id)
             self.state.offered_sell_kWh[self.market.time_slot] -= \
                 self.offers.posted_offer_energy(self.market.id)
+            self.state.energy_to_sell_dict[self.market.time_slot] += \
```
spyrostz: This formula adds the total amount of posted offer energy in the energy_to_sell every time a new offer is added. This means that this variable will not be correctly calculated in case there are still pending offers. For instance, if there are 100 posted offers with 1kWh each, when you post a new offer of 1 kWh, the energy_to_sell_dict variable will increase by 101 kWh, which is wrong, because you would like to increase it by one. 
I would recommend to have a look at the `clamp_energy_to_sell` and `clamp_energy_to_buy` methods in the StorageState class, these methods are used for precisely the thing you are trying to achieve in the normal storage strategy.

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -203,8 +206,9 @@ def _delete_bid_impl(self, arguments, response_channel):
         try:
             to_delete_bid_id = arguments["bid"] if "bid" in arguments else None
             deleted_bids = self.remove_bid_from_pending(self.market.id, bid_id=to_delete_bid_id)
-            self.state.offered_buy_kWh[self.market.time_slot] -= \
+            self.state.offered_buy_kWh[self.market.time_slot] = \
                 self.posted_bid_energy(self.market.id)
+            self.state.energy_to_buy_dict([self.market.time_slot])
```
spyrostz: Isn't it `clamp_energy_to_buy` ?

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -254,6 +258,7 @@ def _bid_impl(self, arguments, bid_response_channel):
                 buyer_origin=arguments["buyer_origin"]
             )
             self.state.offered_buy_kWh[self.market.time_slot] += bid.energy
+            self.state.energy_to_buy_dict[self.market.time_slot] -= bid.energy
```
spyrostz: Same here, you should use the `get_posted_energy` and `clamp_energy_to_buy` methods.

PR #680, opened 2020-03-19
D3ASIM-2115: Added transaction_id consumption and redirection in the …

hannesdiedrich:
…external strategy.


src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -25,29 +25,32 @@ def event_activate(self):
...
+    def _list_bids_impl(self, arguments, response_channel):
         try:
             filtered_bids = [{"id": v.id, "price": v.price, "energy": v.energy}
                              for _, v in self.market.get_bids().items()
                              if v.buyer == self.device.name]
             self.redis.publish_json(
                 response_channel,
-                {"command": "list_bids", "status": "ready", "bid_list": filtered_bids})
+                {"command": "list_bids", "status": "ready", "bid_list": filtered_bids,
+                 "transaction_id": arguments["transaction_id"]})
```
spyrostz: If a message with no transaction id is passed, I think the redis thread will crash, In this case a KeyError will be raised, which will get caught by the `except` block. But, since the except block also uses `arguments["transaction_id"]` , it will also raise an exception which will not get caught, and the redis thread that receives the incoming messages will terminate.

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -25,29 +25,32 @@ def event_activate(self):
...
         except Exception as e:
             logging.error(f"Error when handling list bids on area {self.device.name}: "
                           f"Exception: {str(e)}")
             self.redis.publish_json(
                 response_channel,
                 {"command": "list_bids", "status": "error",
-                 "error_message": f"Error when listing bids on area {self.device.name}."})
+                 "error_message": f"Error when listing bids on area {self.device.name}.",
+                 "transaction_id": arguments["transaction_id"]})
```
spyrostz: In order to alleviate the error I mentioned in the former comment, it's better to replace this with `"transaction_id": arguments.get("transaction_id", None)`

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -79,11 +81,17 @@ def _dispatch_tick_frequency(self):
             (DISPATCH_EVENT_TICK_FREQUENCY_PERCENT / 100)
         )
 
+    @staticmethod
+    def _extract_trans_id(payload):
+        return json.loads(payload["data"])["transaction_id"]
```
spyrostz: Same as before, `transaction_id` might not be part of the message.

tests/test_strategy_external.py
```diff
@@ -128,7 +128,7 @@ def test_accept_offer(self):
         assert response_payload["trade"] == trade.to_JSON_string()
 
     def test_bid(self):
-        payload = {"data": json.dumps({"energy": 12, "price": 32})}
+        payload = {"data": json.dumps({"energy": 12, "price": 32, "transaction_id": "some_id"})}
```
spyrostz: General comment, but since this is my major concern, could you also validate with a test that messages without a transaction_id are correctly rejected ? Thanks !

hannesdiedrich: I am definitely in favour of testing the external strategies. But I just realise, that this unittest only tests the `ExternalStrategy` which we dont use any more. Right?
If this is true i will revert https://github.com/gridsingularity/d3a/pull/680/files#diff-4227e14dfec828d6e1d346e65d981d94R197 too
I dont know why i added it.
And I will add unittedts for the external strategies ,ok?

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -79,24 +84,27 @@ def _delete_offer_impl(self, arguments, response_channel):
             self.redis.publish_json(
                 response_channel,
                 {"command": "offer_delete", "status": "ready",
-                 "deleted_offers": deleted_offers})
+                 "deleted_offers": deleted_offers,
+                 "transaction_id": arguments.get("transaction_id", None)})
```
faizan2590: Shouldn't `transaction_id` must be a non-none value ?

hannesdiedrich: yes, this is precaution implemented after Spyros comment: https://github.com/gridsingularity/d3a/pull/680#discussion_r395068333
If it is indeed not provided in the arguments, then the value of it would be none, which then will not be matched in the client.

tests/test_strategy_external.py
spyrostz: That would be great yes. I also got confused and thought that we already had unit tests for these. If you do not want to take the time to do it all, just add the register unit test, and add TODO comments so that we deal with these at the first time we need to solve a ticket for each of the strategies.

hannesdiedrich: OK! Thanks! Adding the unittest for the register command.

PR #683, opened 2020-03-24
D3ASIM-1967: Constant grid fee interface implementation. Replaced har…

spyrostz:
…dcoded fee class names with variable class names, in order for the class to be selected dynamically according to the selected fee type. Added integration tests for constant grid fees.


src/d3a/models/market/__init__.py
```diff
@@ -78,6 +80,13 @@ def __init__(self, time_slot=None, bc=None, notification_listener=None, readonly
             if transfer_fees is not None else 0
         self.transfer_fee_const = transfer_fees.transfer_fee_const \
             if transfer_fees is not None else 0
+
+        self.fee_class = GridFees
+        self.grid_fee_value = self.transfer_fee_ratio
+        if self.transfer_fee_const > 0:
+            self.fee_class = ConstantGridFees
+            self.grid_fee_value = self.transfer_fee_const
```
faizan2590: I guess it could be in the else clause rather then assigning and then overwriting if `transfer_fee_const >0`

src/d3a/models/market/grid_fees/constant_grid_fees.py
```diff
@@ -0,0 +1,89 @@
+from d3a.models.market.grid_fees import BaseClassGridFees
+from d3a.models.market.market_structures import TradeBidInfo
+
+
+class ConstantGridFees(BaseClassGridFees):
+
+    @staticmethod
+    def update_incoming_bid_with_fee(source_bid, original_bid, tax_ratio=None):
```
faizan2590: Better rename `tax_ratio` to something like `tax_value` as it is targeting the constant grid fee here

hannesdiedrich: Tax_ratio is not used here. This function is mainly used in the percentage based grid fees. Both Classes have to implement this function because of the basclass.

src/d3a/models/market/__init__.py
hannesdiedrich: One additional proposal, how about adding the grid_fee_value as an construction parameter and only use it inside the fee_class?

src/d3a/models/market/grid_fees/constant_grid_fees.py
hannesdiedrich: But I agree with renaming tax_ratio throughout the class. It is indeed a grid_fee_rate, right?

src/d3a/models/market/__init__.py
spyrostz: In order to summarize both comments, I agree it would be better to convert the grid fees from a class with only static methods to an instance, and store/manage the grid fee values in the constructor of this class. Changing this, thanks

src/d3a/setup/grid_fees/constant_grid_fees.py
```diff
@@ -0,0 +1,59 @@
+"""
```
hannesdiedrich: i can not leave comment directly there, but `src/d3a/setup/grid_fees/non_compounded_grid_fees2.py`  is empty. Intentional?

integration_tests/grid_fee.feature
```diff
@@ -68,3 +68,55 @@ Feature: GridFee integration tests
...
+     And d3a is installed
+     And d3a uses an two-sided pay-as-clear market
+     And the minimum offer age is <min_offer_age>
+     And d3a dispatches events from top to bottom
+     When we run the simulation with setup file grid_fees.constant_grid_fees and parameters [24, 60, 60, 0, 1]
+     Then trades on the House 1 market clear with 30.0 cents/kWh and at grid_fee_rate with 0.0 cents/kWh
+     Then trades on the Neighborhood 1 market clear with 30.0 cents/kWh and at grid_fee_rate with 1.0 cents/kWh
+     Then trades on the Grid market clear with 29 cents/kWh and at grid_fee_rate with 2.0 cents/kWh
+     Then trades on the Neighborhood 2 market clear with 27 cents/kWh and at grid_fee_rate with 1.0 cents/kWh
+     Then trades on the House 2 market clear with 26.0 cents/kWh and at grid_fee_rate with 0.0 cents/kWh
```
hannesdiedrich: Its a pleasure to look at these. It will make so much things simpler! Thanks!

src/d3a/setup/grid_fees/constant_grid_fees.py
spyrostz: Because apparently it was empty in the master branch too! I will remove it, I blindly moved these 2 files in the new directory without looking at their content. Since it was empty in the first place, no reason to keep it.

src/d3a/models/market/grid_fees/constant_grid_fees.py
spyrostz: You are correct, this is a grid_fee_rate, and it should be renamed throughout the class hierarchy (in ABC, base_model and here). Will rename this, thanks

PR #684, opened 2020-03-24
D3ASIM-2158: Added dso_market_stats to RedisMarketExternalConnection.

hannesdiedrich:


src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -41,15 +42,43 @@ def market_stats_callback(self, payload):
...
+
+    def event_market_cycle(self):
+        if self.area.current_market is None:
+            return
+        market_event_channel = f"{self.channel_prefix}/market-events/market"
+        current_market_info = self.area.current_market.info
+        current_market_info['last_market_stats'] = \
+            self.area.stats.get_price_stats_current_market()
+        current_market_info["self_sufficiency"] = \
+            self.area.endpoint_stats["kpi"].get("self_sufficiency", None)
```
faizan2590: What about `self-consumption` ?

src/d3a/d3a_core/simulation.py
```diff
@@ -241,6 +241,7 @@ def _run_cli_execute_cycle(self, slot_resume, tick_resume):
 
     def _update_and_send_results(self, is_final=False):
         self.endpoint_buffer.update_stats(self.area, self.status, self.progress_info)
+        self.endpoint_buffer.send_results_to_child_areas(self.area)
```
spyrostz: Why not calling this method in the endpoint_buffer.update_area_aggregated_stats ? It serves the same purpose, and you can avoid the second recursion.

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: This is not requested int the ticket: https://gridsingularity.atlassian.net/browse/D3ASIM-2158

src/d3a/d3a_core/simulation.py
hannesdiedrich: Nice! Didnt see that. Will add it there. Thanks!

hannesdiedrich:
> One general comment, I think I am missing the on_finish event here.

True. Will add it.

PR #685, opened 2020-03-24
Bug/d3 asim 2131

faizan2590:


src/d3a/models/state.py
```diff
@@ -130,17 +130,16 @@ def free_storage(self, time_slot):
         """
         in_use = self._used_storage \
             - self.pledged_sell_kWh[time_slot] \
-            + self.pledged_buy_kWh[time_slot] \
```
spyrostz: Ok, so to summarise, free_storage property should not include the amount of offered energy for buy/sell , right ? I am of course concerned that this might affect many things, but let's give it a try that way.

src/d3a/models/strategy/__init__.py
```diff
@@ -104,6 +104,23 @@ def is_offer_posted(self, market_id, offer_id):
...
+    def get_sold_offer_ids_in_market(self, market_id):
+        sold_offer_ids = []
+        for sold_offer in self.sold.get(market_id, []):
+            sold_offer_ids.append(sold_offer.id)
+        return sold_offer_ids
+
+    def open_in_market(self, market_id):
+        open_offer = []
+        sold_offer_ids = self.get_sold_offer_ids_in_market(market_id)
+        for offer, _market in self.posted.items():
```
spyrostz: As far as I understand, the reason for introducing these methods is in order to be able to calculate the energy of the existing offers. This is exactly what the posted offers dict keeps, essentially when an offer is deleted or traded, it should be removed from the posted dict. That's why I do not think that you need any of these 3 functions (`get_sold_offer_ids_in_market `, `open_in_market `, `open_offer_energy `), you would achieve the same result by using the `posted_offer_energy` method.
cc'ing @hannesdiedrich who has recently worked in this class.

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -306,6 +311,9 @@ def event_tick(self):
             super().event_tick()
         else:
             self.state.tick(self.market_area, self.market.time_slot)
+            self.state.clamp_energy_to_sell_kWh([self.market.time_slot])
+            self.state.clamp_energy_to_buy_kWh([self.market.time_slot])
+            # print(f"tick_info_dict: {self._device_info_dict}")
```
spyrostz: Trivial, reminder to remove the print before merging

src/d3a/models/state.py
faizan2590: The reason I did it this way is `free_storage` & `used_storage` doesn't have the same pattern of accounting energy, meaning `free_storage` takes into account the `promised energy that hasn't been traded` whereas `used_storage` is not doing so

spyrostz: I see, then I agree that these 2 properties should complement each other. Thanks for finding this!

src/d3a/models/strategy/__init__.py
hannesdiedrich: Yes, I agree, `posted_offer_energy` should be sufficient.

faizan2590: `posted_in_market` reports all the offer that has been yet posted (it is kept in `posted` even after the trade OR being deleted) whereas for `offered_sell_kWh` of ESS needs the offered energy that are not traded yet.

faizan2590: Judging from the doc-string of this [class](https://github.com/gridsingularity/d3a/pull/685/files#diff-5c5fbb99e0155bbce18f8ca72347fd58R59), it is expected to distinguish `open_offer`  from the `posted_offer`

spyrostz: Unfortunately the link does not work in your case, but I think I found the culprit here:
https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/strategy/__init__.py#L150
Therefore indeed the posted offers do contain the sold offers. In this case this works for me, and these functions serve a purpose. Thanks!

hannesdiedrich: Hmm, ok, so self.posted keeps track of ALL offers, including the sold and active ones. 
Why do we have this again? This is probably heritage from before, right? Anyways. 
I agree with Spyros. Leave your implementation. Thanks!

faizan2590: There is indeed a room for potential refactoring as I see in case of `bids` inside `BidEnabledStrategy`, there being popped-out from `_bids` dict after being traded.
If you guys agree I would be very much in favour of a future enhancement story

spyrostz: I agree that the behavior of both should be synced, feel free to create the enhancement story. Thanks!

PR #687, opened 2020-03-26
D3ASIM-2170: Added compression before sending the simulation results …

spyrostz:
…via Redis.

### DIFF
--- a/src/d3a/d3a_core/redis_connections/redis_communication.py
+++ b/src/d3a/d3a_core/redis_connections/redis_communication.py
@@ -23,6 +23,7 @@ from redis.exceptions import ConnectionError
 from rq import get_current_job
 from rq.exceptions import NoSuchJobError
 from d3a_interface.results_validator import results_validator
+from zlib import compress
 
 log = getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -111,9 +112,18 @@ def publish_results(self, endpoint_buffer):
             return
         result_report = endpoint_buffer.generate_result_report()
         results_validator(result_report)
+
         results = json.dumps(result_report)
+        message_size = utf8len(results)
+        if message_size > 64000:
+            log.error(f"Do not publish message bigger than 64 MB, current message size "
+                      f"{message_size / 1000.0} MB.")
+            return
```
hannesdiedrich: So what happens here? Te data package that is too large, will not be sent, so there will be a whole in the results on the d3a-web or d3a-ui side, right?

spyrostz: Exactly. The idea is to not stop the simulation, but to send an error message to Kibana/Sentry, and simply the UI results will have gaps from the data that are missing.

PR #696, opened 2020-03-30
D3ASIM-1967: Added ignore_fees parameter to the ABC in order to requi…

spyrostz:
…re its usage. Added handling of this parameter in constant grid fess.


src/d3a/models/market/__init__.py
```diff
@@ -102,11 +103,15 @@ def __init__(self, time_slot=None, bc=None, notification_listener=None, readonly
...
         if not transfer_fees:
             transfer_fees = TransferFees(grid_fee_percentage=0.0, transfer_fee_const=0.0)
-        if transfer_fees.transfer_fee_const is not None and transfer_fees.transfer_fee_const > 0.0:
-            self.fee_class = ConstantGridFees(transfer_fees.transfer_fee_const)
+        if grid_fee_type == 1:
+            if not (transfer_fees.transfer_fee_const is not None
+                    and transfer_fees.transfer_fee_const > 0.0):
+                self.fee_class = ConstantGridFees(0.0)
+            else:
+                self.fee_class = ConstantGridFees(transfer_fees.transfer_fee_const)
```
hannesdiedrich: Can you please explain this? grid_fee_type = 1 IS the constant grid fee type.

hannesdiedrich: Ahhh, no  I see it, it is a double "not"

hannesdiedrich:
```suggestion
        if grid_fee_type == 1:
            if transfer_fees.transfer_fee_const is None or transfer_fees.transfer_fee_const <= 0.0:
                self.fee_class = ConstantGridFees(transfer_fees.transfer_fee_const)
            else:
                self.fee_class = ConstantGridFees(0.0)
```

tests/test_area.py
```diff
@@ -81,6 +82,7 @@ def test_markets_are_cycled_according_to_market_count(self):
         self.area._bc = False
         for i in range(2, 97):
             self.config.market_count = i
+            self.config.grid_fee_type = 1
```
hannesdiedrich:
```suggestion
            self.config.grid_fee_type = ConstSettings.IAASettings.GRID_FEE_TYPE
```

src/d3a/models/market/__init__.py
spyrostz: I will revert the double negation, however I am not merging the suggestion because the 2 blocks should be reversed. Pushing a separate change that fixes this.

src/d3a/models/market/__init__.py
```diff
@@ -102,11 +103,15 @@ def __init__(self, time_slot=None, bc=None, notification_listener=None, readonly
...
-    def _create_fee_handler(self, transfer_fees):
+    def _create_fee_handler(self, grid_fee_type, transfer_fees):
         if not transfer_fees:
             transfer_fees = TransferFees(grid_fee_percentage=0.0, transfer_fee_const=0.0)
-        if transfer_fees.transfer_fee_const is not None and transfer_fees.transfer_fee_const > 0.0:
-            self.fee_class = ConstantGridFees(transfer_fees.transfer_fee_const)
+        if grid_fee_type == 1:
+            if transfer_fees.transfer_fee_const is None or \
+                    transfer_fees.transfer_fee_const <= 0.0:
+                self.fee_class = ConstantGridFees(0.0)
```
faizan2590, line 112: Would be nice to repeat the same validation for grid_fee_percentage too

src/d3a/setup/grid_fees/constant_grid_fees.py
```diff
@@ -20,9 +20,12 @@
 from d3a.models.strategy.load_hours import LoadHoursStrategy
 from d3a.models.appliance.pv import PVAppliance
 from d3a.models.strategy.pv import PVStrategy
+from d3a_interface.constants_limits import ConstSettings
 
 
 def get_setup(config):
+    ConstSettings.IAASettings.GRID_FEE_TYPE = 1
```
faizan2590, line 27: Umm is it really needed even after setting fee type via `config.grid_fee_type` ?

src/d3a/models/market/__init__.py
```diff
@@ -60,6 +60,7 @@ def wrapper(self, *args, **kwargs):
 class Market:
 
     def __init__(self, time_slot=None, bc=None, notification_listener=None, readonly=False,
+                 grid_fee_type=ConstSettings.IAASettings.GRID_FEE_TYPE,
```
faizan2590, line 63: Would be less error prone IMO if we discourage passing default anywhere other than `SimulationConfig`

src/d3a/models/market/__init__.py
spyrostz: You're right, I'll do the same for the percentage fees. Thanks

src/d3a/models/market/__init__.py
spyrostz: I assume you are talking about all market classes, right ? Reason that I added the default value was due to the order of the arguments. Even timeslot is a kwargs, therefore in order to put an argument after that, it has to provide a default value. The proper solution to that would be to remove default values for all parameters here, and reorder them (for instance, bc should not be the 2nd parameter IMO). Nevertheless, I will not do it in the context of this story since this change  is quite error prone. Better to do in its own ticket.

src/d3a/setup/grid_fees/constant_grid_fees.py
spyrostz: Indeed it's not really necessary. Removing it.

src/d3a/models/market/__init__.py
faizan2590: Please create a ticket.

spyrostz: https://gridsingularity.atlassian.net/browse/D3ASIM-2194

PR #697, opened 2020-03-31
Feature/d3 asim 2175

hannesdiedrich:


integration_tests/integration_tests.feature
```diff
@@ -35,14 +35,12 @@ Feature: Run integration tests
...
 
-  # TODO: Once iaa_fee is included in billing, this test to be run non-zero iaa_fee
   Scenario Outline: Test Balanced Energy Bills
     Given we have a scenario named <scenario>
     And d3a is installed
-    When we run the simulation with setup file <scenario> and parameters [24, 60, 60, 0, 1]
+    When we run the simulation with setup file <scenario> and parameters [24, 60, 60, 1]
     Then the traded energy report the correct accumulated traded energy
     And the energy bills report the correct accumulated traded energy price
-    And the energy bills report the correct external traded energy and price
```
faizan2590: Umm, can you explain the reason to remove this step?

integration_tests/grid_fee.feature
```diff
@@ -109,7 +109,7 @@ Feature: GridFee integration tests
      And d3a uses an two-sided pay-as-clear market
      And the minimum offer age is <min_offer_age>
      And d3a dispatches events from top to bottom
-     When we run the simulation with setup file grid_fees.constant_grid_fees and parameters [24, 60, 60, 0, 1]
+     When we run the simulation with setup file grid_fees.constant_grid_fees and parameters [24, 60, 60, 1]
```
spyrostz: Well, no harm having it here, but since most of the tests use the percentage fees, why not having a default value of percentage fees (without changing this parameter list) and if there is a need for a constant fee in a test, only then to add an extra parameter ? Since you did it this way it's not important now, just that you could have avoided the hassle of changing all the integration tests with this approach.

integration_tests/steps/integration_tests.py
```diff
@@ -728,35 +709,6 @@ def test_accumulated_energy(context):
     assert isclose(net_energy, 0, abs_tol=1e-10)
 
 
-@then('the energy bills report the correct external traded energy and price')
```
spyrostz: I think it is wrong to remove it now. Keeping the TODO there is important to not forget to reactivate it once the bug is resolved.

src/d3a/models/area/__init__.py
```diff
@@ -43,14 +43,13 @@
 log = getLogger(__name__)
 
 
+# TODO: As this is only used in the unittests, please move to the tests package
```
spyrostz: I would say to delete this and replace the usages of this class with constants_limits.GlobalConfig class. If you can change the TODO to reflect that it would be great

integration_tests/grid_fee.feature
hannesdiedrich: Hmm i dont understand. There is no global grid_fee setting any more. The grid fees are set in the setup file. So i deleted this parameter. 
Or can you rephrase please?

integration_tests/integration_tests.feature
hannesdiedrich: Please have a look in to the step function , it is disabled anyway

integration_tests/steps/integration_tests.py
hannesdiedrich: IMO this test is totally outdated. There is no bug but another way of calculation. Shall I add another story to add  a real `the energy bills report the correct external traded energy and price` test?

src/d3a/models/area/__init__.py
hannesdiedrich: Ok, will do!

integration_tests/grid_fee.feature
spyrostz: I see, I misread the parameters here. So the last parameter is always the market_count, right ? Then all is good.

integration_tests/alternative_pricing.feature
```diff
@@ -9,7 +9,7 @@ Feature: Alternative Pricing
   Scenario Outline: Pricing for alternative pricing is right
     Given we have a scenario named <scenario>
     And d3a is installed
-    When we run the d3a simulation with <scenario> [24, 60, 60]
+    When we run the simulation with setup file <scenario> and parameters [24, 60, 60, 1]
```
spyrostz: Regarding the latter comment, I originally was confused by this, because you added a parameter, and I incorrectly thought that it was the grid fee type. It is the market count though, and I assume you added the parameter here in order to remove a step definition, right ?

integration_tests/steps/integration_tests.py
spyrostz: Why not adding a link to the 2-sided PAC bug ticket? That way we know that this needs to be rewritten when the bug is resolved.

integration_tests/alternative_pricing.feature
hannesdiedrich: Exactly, this was bugging me for a while now, that we have two step functions for running a simple simulations.

integration_tests/steps/integration_tests.py
hannesdiedrich: Can you please provide the link to the ticket ? I cant find it.
So I leave the implementation here but link the ticket, yes?

hannesdiedrich: (also disabling this step again, of course)

integration_tests/alternative_pricing.feature
spyrostz: Perfect, glad that this was removed.

integration_tests/steps/integration_tests.py
spyrostz: https://gridsingularity.atlassian.net/browse/D3ASIM-1887 , it is part of the TODO comment on the step implementation. I had a deeper look in the test, and I cannot understand why the test is outdated. I think it is perfectly valid still, and it essentially checks out if the trade_energy_price graph is consistent with the energy bills, as far as I understand. The only reason why this step should fail, is due to the behavior described in the bug. Why is it important to delete it actually ?

hannesdiedrich: Ok, now it makes sense, thanks for charing the ticket!
Will re add the step again. Thanks!

integration_tests/integration_tests.feature
hannesdiedrich: Readding it again because of the other discussion

PR #706, opened 2020-04-02
D3ASIM-2103: Fixed demanded energy calculation to work with load prof…

spyrostz:
…iles. Reset demanded energy value when recursing, and not reset it on every recursion step. Adapted accounted_markets variable to be stored per area.


src/d3a/models/strategy/load_hours.py
```diff
@@ -133,6 +133,12 @@ def event_market_cycle(self):
                 self.energy_requirement_Wh[market.time_slot] = 0.0
                 self.state.desired_energy_Wh[market.time_slot] = 0.0
         self.event_market_cycle_prices()
+        if self.area.current_market:
+            self.state.total_energy_demanded_wh = sum(
+                e for t, e in self.state.desired_energy_Wh.items()
+                if t <= self.area.current_market.time_slot)
```
hannesdiedrich, line 139: Hmm, unfortunately, I dont see any point in the code, where the self.state.desired_energy_Wh is filled with values. CAn you help me?

spyrostz: This file, line 360 https://github.com/gridsingularity/d3a/pull/706/files#diff-5f60b6cdf41a34bdba5921c18ebab298L360

hannesdiedrich: Ok, thanks.
Sorry, but why are there multiple entries in the desired_energy_Wh with a time stamp older than the current market?

src/d3a/d3a_core/sim_results/kpi.py
```diff
@@ -117,6 +117,7 @@ def _accumulate_energy_trace(self):
                     self._dissipate_infinite_consumption(trade)
 
     def update_area_kpi(self, area):
+        self.total_energy_demanded_wh = 0
         self._accumulate_total_energy_demanded(area)
```
faizan2590, line 121: Correct me if I'm wrong `update_area_kpi ` is triggered for every tick meaning that the same_demanded energy would be recalculated for every ticket. I think it could be moved to class constructor as I don't see the need for it to re-calculate an static value for every tick

src/d3a/models/strategy/load_hours.py
spyrostz: I think you misunderstood the intention here. The problem was that the `total_energy_demanded_Wh` was calculated always using the avg_power_W, which is obviously incorrect for the case of load profiles. In addition to that, the value was not calculated after each market slot, but only at the beginning of the simulation. The self consumption formula though is total_energy_consumed / total_energy_demand (the value we are talking about now). Total energy consumed is aggregated every market slot, while this value always had the final value (precalculated at the beginning of the simulation). This had the very annoying effect IMO that the self-sufficiency and self-consumption values did not display the correct values during the simulation run (because the nominator of the aforementioned division was referring to the values until this point in the simulation, and the denominator for the whole simulation). Therefore in order to sanely debug this I had to also recalculate the `total_energy_demanded_wh` value on every market slot, in order for the results to be correct on every market slot. 
Therefore, I converted the meaning of this value from 'total energy demanded for the whole simulation' to 'total energy demanded until this point in the simulation'.
Hope this clarified things a bit.

src/d3a/d3a_core/sim_results/kpi.py
spyrostz: It is updated every market slot actually, not on every tick. Moreover, this was one of the major issues that this PR solves. `_accumulate_total_energy_demanded` is recursive. Therefore, if it is called with the Grid area, it will set the `total_energy_demanded_wh` to 0 at the beginning of every recursion. This will obviously work for single devices, but as I mentioned above, doing this on a neighbourhood or a grid will set this value to the demand of the last load in the recursion.

faizan2590: Yes you are right it is indeed called per market_slot BUT if you keep `total_energy_demanded_wh` 
 as part of `event_activate_energy ` then `_accumulate_total_energy_demanded` could be moved to the constructor of this class to save the re-accumulating at every slot.

spyrostz: This will have the effect though, that the self-sufficiency and self-consumption will not have correct results during the simulation. I'll redirect you to the second comment chain with @hannesdiedrich , where I describe exactly the reason why it is important IMO to recalculate this value on every market slot (as long as we want the self sufficiency and self consumption values to make sense during the simulation too, and not only at the end of the simulation)

faizan2590: Thanks a lot. Convinced :-)

src/d3a/models/strategy/load_hours.py
hannesdiedrich: Ahhhh! Thanks a lot! I confused `total_energy_demanded_wh` with something like `energy_requirement_Wh` here. But we are in the KPIs.
Ok. Your explanation makes sense. Thanks a lot!

src/d3a/d3a_core/sim_results/kpi.py
```diff
@@ -56,19 +51,24 @@ def _accumulate_devices(self, area):
...
-        if trade.seller_origin in self.producer_list:
+        # Trade seller origin should be equal to the trade seller in order to
+        # not double count trades in higher hierarchies
+        if trade.seller_origin in self.producer_list and trade.seller_origin == trade.seller:
             self.total_energy_produced_wh += trade.offer.energy * 1000
 
     def _accumulate_self_consumption(self, trade):
-        if trade.seller_origin in self.producer_list and trade.buyer_origin in self.consumer_list:
+        # Trade buyer origin should be equal to the trade buyer in order to
+        # not double count trades in higher hierarchies
```
hannesdiedrich, line 68: wow! good catch!

integration_tests/steps/kpi.py
```diff
@@ -33,13 +33,16 @@ def test_export_of_kpi_result(context, kpi, expected_kpis):
     expected_kpis = ast.literal_eval(expected_kpis)
     for area, value in expected_kpis.items():
         if kpi == "self_sufficiency":
-            assert kpi_data[area]['self_sufficiency'] == float(value)
+            assert isclose(kpi_data[area]['self_sufficiency'], float(value), rel_tol=1e-04)
         elif kpi == "self_consumption":
+            print(value)
+            print(kpi_data[area]['self_consumption'])
+            print(area)
```
hannesdiedrich, line 40: please remove before pushing. Thanks

PR #708, opened 2020-04-02
D3ASIM-2186: Added totals_with_penalties to bills.

hannesdiedrich:


src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -208,6 +212,19 @@ def _accumulate_by_children(self, area, flattened, results):
...
         return results
 
+    def _aggregate_totals_with_penalties(self, area, results):
+        if not area.children:
+            return
+        totals_with_penalties = 0
+        for child in area.children:
+            self._aggregate_totals_with_penalties(child, results)
+            if "totals_with_penalties" in results[child.name]:
+                if isinstance(results[child.name]["totals_with_penalties"], dict):
```
faizan2590, line 222: Can you tell me why `results[child.name]["totals_with_penalties"]` is expected to be in two different data type ?

hannesdiedrich: The reason for this is that is that it `totals_with_penalties` in an non-device area would be also a number. but as all members of an area are subdicts (e.g. `{"grid": {"House 1": {}, "House 2":{}, "Accumulated Trades": {}, "External Tradds": {}}}` )i followed this convention and created a subdict for the `totals_with_penalties` too and put it in a subdict.
As we foresie some additional parameters that need to be accumulated like this, a subdict is helpful too. 
IN the devices `totals_with_penalties` is a normal entry.
Does it make sense?

faizan2590: Thanks for the explanation, makes sense now BUT better add an additional check to validate its not a dict in any case if it is a device

hannesdiedrich: Hmm , i dont get it. This IS the check whether it is the device or not, by checking if it is a dict or not.

faizan2590: This is an implicit way of checking it and will work until devices' `totals_with_penalties` are not accidentally populated with dict due to some logical failure.

So, better a unittest validating that `_store_area_penalties` method will never populate a dict type `totals_with_penalties` for devices

integration_tests/steps/kpi.py
```diff
@@ -85,3 +85,22 @@ def penalty_rate_respected(context, device_name):
     assert isclose(bills[str(device.uuid)]["penalty_cost"],
                    bills[str(device.uuid)]["penalty_energy"] * DEVICE_PENALTY_RATE / 100.0,
                    rel_tol=0.0003 * DEVICE_PENALTY_RATE)
+
+
+@then('totals_with_penalties are correctly populated for all areas in the bills')
+def totals_with_penalties_bills_test(context):
+    bills = context.simulation.endpoint_buffer.market_bills.bills_results
+    expected_grid_totals = sum(bills[house]["totals_with_penalties"]["costs"]
```
faizan2590, line 93: Name `expected` is a bit deceiving as it is extracting the calculated_value

integration_tests/steps/kpi.py
```diff
@@ -85,3 +85,22 @@ def penalty_rate_respected(context, device_name):
...
+                               for house in ["House 1", "House 2"])
+    assert isclose(bills["Grid"]["totals_with_penalties"]["costs"], expected_grid_totals)
+
+    for house_name in ["House 1", "House 2"]:
+        house_obj = \
+            list(filter(lambda x: x.name == house_name, context.simulation.area.children))[0]
+        expected_house_totals = \
```
faizan2590, line 100: Same as [above](https://github.com/gridsingularity/d3a/pull/708/files?file-filters%5B%5D=.py#r403124573)

integration_tests/steps/kpi.py
hannesdiedrich: Actually no, it IS the expected value (the sum of all `totals_with_penalties` of all children.)
The actual value is `bills["Grid"]["totals_with_penalties"]["costs"]` (which i need to check)

integration_tests/steps/kpi.py
hannesdiedrich: Same as above

PR #710, opened 2020-04-03
D3ASIM-2160: Converting kpi_ratio to percentage for UI.

faizan2590:


src/d3a/d3a_core/sim_results/kpi.py
```diff
@@ -159,10 +159,24 @@ def area_performance_indices(self, area):
...
+            self_sufficiency_percentage = \
+                self.performance_indices[area_name]["self_sufficiency"] * 100
+        else:
+            self_sufficiency_percentage = 0.0
+        if self.performance_indices[area_name]["self_consumption"] is not None:
+            self_consumption_percentage = \
+                self.performance_indices[area_name]["self_consumption"] * 100
+        else:
+            self_consumption_percentage = 0.0
+        return {"self_sufficiency": self_sufficiency_percentage,
```
spyrostz, line 173: No need to return the variables here and reassign, you already change the `self.performance_indices` member of the class.

faizan2590: Sorry, I couldn't follow you, I only have used `self.performance_indices`  for calculation and assigned the calculated value to a local variable called `self_consumption_percentage & self_sufficiency_percentage `

spyrostz: I see now why you need a return variable here, you need to only update the stats sent to redis and not the normal kpi stats. That's good , LGTM once more :P

PR #711, opened 2020-04-07
Feature/d3 asim 2189

hannesdiedrich:


src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -592,3 +591,41 @@ def _convert_output_format(self, price_energy, csv_output, redis_output):
                 } for timeslot, trades in trade_rates.items()
             ]
             redis_output[node.uuid] = deepcopy(csv_output[node.name])
+
+
+class BaselinePeakEnergyStats:
+    def __init__(self):
+        self.baseline_peak_percentage_result = {}
+        self.baseline_peak_percentage_result_redis = {}
```
faizan2590, line 599: Are there any expected difference of results to be sent to UI and backend reports ?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -53,6 +53,7 @@ def __init__(self, job_id, initial_params, area):
         self.device_statistics = DeviceStatistics()
         self.file_export_endpoints = FileExportEndpoints()
         self.kpi = KPI()
+        self.baseline_peak_stats = BaselinePeakEnergyStats()
```
faizan2590, line 56: Would be better if renamed  `baseline_peak_stats` to `area_throughput_stats`

src/d3a/models/area/__init__.py
```diff
@@ -103,6 +109,23 @@ def __init__(self, name: str = None, children: List["Area"] = None,
         self.redis_ext_conn = RedisMarketExternalConnection(self) \
             if external_connection_available is True else None
 
+    def validate_baseline_energy_settings(self):
```
faizan2590, line 112: Could be moved to `d3a-interface` to be used in `d3a-web` too

src/d3a/setup/power_flow/baseline_peak_energy.py
```diff
@@ -0,0 +1,69 @@
...
+                        baseline_energy_settings={"baseline_peak_energy_export_kWh": 0.3}
+
+                    ),
+                    ], baseline_energy_settings={"baseline_peak_energy_export_kWh": 0.3}
+                 ),
+            Area('Cell Tower', strategy=CellTowerLoadHoursStrategy(avg_power_W=100,
+                                                                   hrs_per_day=24,
+                                                                   hrs_of_day=list(range(0, 24)),
+                                                                   final_buying_rate=35),
+                 appliance=SwitchableAppliance()),
```
faizan2590, line 65: I guess it could be safely removed

src/d3a/d3a_core/sim_results/area_statistics.py
hannesdiedrich: Yes, just like in all other stats: area names are replaced by area uuids and numbers are rounded

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: ok! Will do!

src/d3a/models/area/__init__.py
hannesdiedrich: Yes, I was thinking about it but first needed your opinion about the whole calculation and added it here. Will move to d3a-interface and use it in the d3a-web part too.

src/d3a/setup/power_flow/baseline_peak_energy.py
hannesdiedrich: No, it is actually good to have this case in there, to test that the import in the House 1 branch is not the same as the export in the House 2 branch.

hannesdiedrich: I could also replace it wiht another load area, if you prefer.

faizan2590: That would be be great. Thanks!

src/d3a/d3a_core/sim_results/area_statistics.py
faizan2590: Thanks for reminding, in that case it looks good

src/d3a/models/area/__init__.py
```diff
@@ -68,13 +71,16 @@ def __init__(self, name: str = None, children: List["Area"] = None,
                  event_list=[],
                  grid_fee_percentage: float = None,
                  transfer_fee_const: float = None,
-                 external_connection_available=False):
+                 external_connection_available: bool = False,
+                 baseline_energy_settings=None):
```
faizan2590, line 75:
```suggestion
                 baseline_energy_settings: BaselineSettingsDefault  =None):
```

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -592,3 +591,41 @@ def _convert_output_format(self, price_energy, csv_output, redis_output):
...
+        self.results_to_redis(area)
+
+    def update_exported_imported_energy(self, area):
+        if area.baseline_energy_settings is not None:
+            area.stats.aggregate_exported_imported_energy(area)
+            baseline_import = area.baseline_energy_settings["baseline_peak_energy_import_kWh"]
+            baseline_export = area.baseline_energy_settings["baseline_peak_energy_export_kWh"]
+            if baseline_import > 0:
+                create_subdict_or_update(self.baseline_peak_percentage_result, area.name,
+                                         {"import": {k: v/baseline_import
```
faizan2590, line 612: Here the v is a time profile value showing the cumulative imported energy per slot whereas as baseline_import also needs to be a profile value

hannesdiedrich: No, v is a value of the area.stats.imported_energy profile, baseline_import is a number.

src/d3a/models/area/__init__.py
hannesdiedrich: Let me check whether with this change the user is able to set only one member of BaselineSettingsDefault. If so, I will change this. Thanks

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -592,3 +591,41 @@ def _convert_output_format(self, price_energy, csv_output, redis_output):
...
+    def results_to_redis(self, area):
+        if area.name in self.baseline_peak_percentage_result:
+            rounded_result = {}
+            for key in self.baseline_peak_percentage_result[area.name].keys():
+                rounded_result[key] = \
+                    {k: round_floats_for_ui(v)
+                     for k, v in self.baseline_peak_percentage_result[area.name][key].items()}
+            self.baseline_peak_percentage_result_redis[area.uuid] = rounded_result
+        for child in area.children:
+            self.results_to_redis(child)
```
faizan2590, line 631: This additional recursion could potentially be  saved if `self.baseline_peak_percentage_result_redis` is populated right after the calculation of `self.baseline_peak_percentage_result`

src/d3a/d3a_core/sim_results/area_statistics.py
faizan2590: Yes you also said the same thing and I think `baseline_import` also needs to be a profile since its a comparison of results with and without grid fee

src/d3a/d3a_core/sim_results/area_statistics.py
hannesdiedrich: you are right! Will move this into `update_exported_imported_energy`

src/d3a/d3a_core/sim_results/area_statistics.py
hannesdiedrich: every value `v` of the `area.stats.imported_energy` is divided by the `baseline_import`

src/d3a/models/area/stats.py
```diff
@@ -30,6 +32,24 @@ def __init__(self, area_markets):
         self.aggregated_stats = {}
         self.market_bills = {}
         self.rate_stats_market = {}
+        self.exported_energy = {}
+        self.imported_energy = {}
+
+    def aggregate_exported_imported_energy(self, area):
```
spyrostz, line 38: I would strongly suggest to decouple it to its own class, that contains only `exported_energy` and `imported_energy` as members. That way the separation is going to be clearer, and you will not have to construct the AreaStats class twice just to use this function.

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -592,3 +591,41 @@ def _convert_output_format(self, price_energy, csv_output, redis_output):
...
+        if area.baseline_energy_settings is not None:
+            area.stats.aggregate_exported_imported_energy(area)
+            baseline_import = area.baseline_energy_settings["baseline_peak_energy_import_kWh"]
+            baseline_export = area.baseline_energy_settings["baseline_peak_energy_export_kWh"]
+            if baseline_import > 0:
+                create_subdict_or_update(self.baseline_peak_percentage_result, area.name,
+                                         {"import": {k: v/baseline_import
+                                          for k, v in area.stats.imported_energy.items()}})
+            if baseline_export > 0:
+                create_subdict_or_update(self.baseline_peak_percentage_result, area.name,
```
spyrostz, line 615: I would suggest to refactor this into a small function that updates both `self.baseline_peak_percentage_result` and `self.baseline_peak_percentage_result_redis` in one go, in order to avoid the second recursion. You can use the area.uuid parameter, in order to avoid the area name-> uuid conversion.

src/d3a/models/area/stats.py
hannesdiedrich: @spyrostz Why do I have to construct the AreaStats class twice? 
I dont see it

src/d3a/d3a_core/sim_results/area_statistics.py
hannesdiedrich: yes, good Idea! will do!

src/d3a/models/area/stats.py
spyrostz: Ah now I see it better, you do not really reconstruct the AreaStats object, but you simply store the `imported_energy` `exported_energy` parameters and the `aggregate_exported_imported_energy` here. If it is only used by the `BaselinePeakEnergyStats` class, why not store these parameters and methods there ? It is a bit misleading that these exist in a different class / module. 
Of course, this is under the assumption that these stats are not going to be visible via the external connection. If they indeed need to be exposed via the external connection it is ok to have them here. But I think that we are not going to publish them if I understand correctly.

src/d3a/d3a_core/area_serializer.py
```diff
@@ -141,7 +143,9 @@ def optional(attr):
         area = Area(name, children, uuid, optional('strategy'), optional('appliance'), config,
                     optional('budget_keeper'), grid_fee_percentage=grid_fee_percentage,
                     transfer_fee_const=grid_fee_constant,
-                    external_connection_available=external_connection_available)
+                    external_connection_available=external_connection_available,
+                    baseline_peak_energy_import_kWh=baseline_peak_energy_import_kWh,
```
spyrostz, line 147: I assume that due to the GraphQL issue we discussed yesterday, you decided to have individual constructor arguments here too, right ?

src/d3a/d3a_core/sim_results/area_statistics.py
```diff
@@ -592,3 +591,59 @@ def _convert_output_format(self, price_energy, csv_output, redis_output):
                 } for timeslot, trades in trade_rates.items()
             ]
             redis_output[node.uuid] = deepcopy(csv_output[node.name])
+
+
+class AreaThroughputStats:
```
spyrostz, line 596: Better to add this class to its own module / file. The `area_statistics` module is already big without this new class, and it would be great to start reducing its size.

src/d3a/d3a_core/area_serializer.py
hannesdiedrich: `due to the GraphQL issue we discussed yesterday, you decided to have individual constructor arguments in the constructor of the Area class too, right `  Exactly. I could think of an object that can hold several parameters that is not a JSON string. 
I had it all implemented here, but reversed it due to our discussion.

src/d3a/d3a_core/sim_results/area_statistics.py
hannesdiedrich: Agreed! Doing so.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -99,7 +103,8 @@ def generate_json_report(self):
             "progress_info": self.simulation_progress,
             "device_statistics": self.device_statistics.device_stats_time_str,
             "energy_trade_profile": self.file_export_endpoints.traded_energy_profile,
-            "kpi": self.kpi.performance_indices
+            "kpi": self.kpi.performance_indices,
+            "baseline_peak_energy": self.area_throughput_stats.results
```
faizan2590, line 107: Please rename the same here too. Thanks
```suggestion
            "area_throughput_stats": self.area_throughput_stats.results
```

src/d3a/d3a_core/sim_results/area_throughput_stats.py
```diff
@@ -0,0 +1,75 @@
...
+            create_subdict_or_update(self.results_redis, area.uuid, out_dict)
+
+    def update_results(self, area):
+        baseline_import = area.baseline_peak_energy_import_kWh
+        baseline_export = area.baseline_peak_energy_export_kWh
+        if (baseline_import is not None and baseline_import > 0) or \
+                (baseline_export is not None and baseline_export > 0):
+            self.aggregate_exported_imported_energy(area)
+            if baseline_import is not None and baseline_import > 0:
+                self._calc_results(area, baseline_import, self.imported_energy, "import")
```
faizan2590, line 51: I think you can save the `baseline_import > 0` check since its a numerator and a zero value wouldn't crash it

hannesdiedrich: Yes, it would. Devision by 0 is not defined.

integration_tests/steps/power_tests.py
```diff
@@ -26,3 +26,20 @@ def test_export_of_power_flow_result(context):
...
+
+
+@then('BaselinePeakEnergyStats are correctly calculated')
+def test_baseline_peak_energy_stats(context):
+    area_throughput_stats = \
+        context.simulation.endpoint_buffer.area_throughput_stats.results
+
+    expected_results = {'Neighborhood 1': {'import': {'peak_energy_kWh': 0.4,
+                                                      'peak_percentage': 100.0}},
+                        'House 1': {'import': {'peak_energy_kWh': 0.4,
```
faizan2590, line 38: To maintain the uniformity JSON structure, wouldn't it be better to always have the same JSON structure, for instance in this case `["House 1"]["export"]["peak_energy_kWh"] = 0` rather than silently having no key at all?
@spyrostz Do you have any opinion ?

hannesdiedrich: I know it is not a lot of data, but the less data we send, the better.
e.g. here: https://gridsingularity.atlassian.net/browse/D3ASIM-2227

spyrostz: I would disagree on that, because these 2 have different meanings. No value means that the user has not configured it. ["House 1"]["export"]["peak_energy_kWh"] = 0 means that there was no export at all during the simulation, 100% self consumption

faizan2590: discussed offline, no need for any further changes.

PR #712, opened 2020-04-07
D3ASIM-2185: Added penalty_energy and penalty_cost to Accumulated Tra…

hannesdiedrich:
…des bills.

### DIFF
--- a/integration_tests/steps/kpi.py
+++ b/integration_tests/steps/kpi.py
@@ -85,6 +85,10 @@ def penalty_rate_respected(context, device_name):
     assert isclose(bills[str(device.uuid)]["penalty_cost"],
                    bills[str(device.uuid)]["penalty_energy"] * DEVICE_PENALTY_RATE / 100.0,
                    rel_tol=0.0003 * DEVICE_PENALTY_RATE)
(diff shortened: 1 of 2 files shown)


integration_tests/steps/kpi.py
```diff
@@ -85,6 +85,10 @@ def penalty_rate_respected(context, device_name):
     assert isclose(bills[str(device.uuid)]["penalty_cost"],
                    bills[str(device.uuid)]["penalty_energy"] * DEVICE_PENALTY_RATE / 100.0,
                    rel_tol=0.0003 * DEVICE_PENALTY_RATE)
+    assert isclose(bills[str(house1.uuid)]["Accumulated Trades"]["penalty_cost"],
```
faizan2590, line 88: Just that it caught my attention now, having a two word dictionary key (Accumulated Trades) doesn't seems like a good approach. No need to do anything for now as its been there for a while

spyrostz: @faizan2590 this key is used by the UI as a row header, that's why it is a string here. It would be great if it remained as is, in order to avoid crashing the UI

PR #713, opened 2020-04-08
Feature/d3 asim 2138: Subclassing ResettableCommunicator into CommonResettableCommunicator

faizan2590:
... to enable external PV/ESS/Load strategies and market stats to share common redis connection.


src/d3a/models/area/__init__.py
```diff
@@ -148,6 +148,8 @@ def activate(self, bc=None):
         self.log.debug('Activating area')
         self.active = True
         self.dispatcher.broadcast_activate()
+        if self.redis_ext_conn is not None:
+            self.redis_ext_conn.sub_to_area_event()
```
hannesdiedrich, line 152: Just for understanding: You moved the subscribe into the activate method because there is no point listening to the channels before that, right?

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -137,6 +152,10 @@ def market_area(self):
     def device(self):
         return self.owner
 
+    @property
```
hannesdiedrich, line 155: Can you please tell me the reason why it is better to declare it as a property here? Thanks

src/d3a/d3a_core/redis_connections/redis_area_market_communicator.py
```diff
@@ -75,6 +92,26 @@ def publish_json(self, channel, data):
...
+        self.channel_callback_dict = {}
+
+    def sub_to_channel(self, channel, callback):
+        self.pubsub.subscribe(**{channel: callback})
+
+    def sub_to_multiple_channels(self, channel_callback_dict):
+        self.pubsub.subscribe(**channel_callback_dict)
+
+    def sub_to_response(self, channel, callback):
+        super().sub_to_response(channel, callback)
```
hannesdiedrich, line 107: I guess you can safely remove this. Right?

src/d3a/models/area/__init__.py
hannesdiedrich: Explained offline. Thanks

src/d3a/models/strategy/external_strategies/__init__.py
spyrostz: I think the property makes sense here, in order to conveniently access the communicator class without having to always do `self.owner.config.external_redis_communicator.publish()` for instance. An alternative would be to simply store the reference to the object of course (which would save some CPU cycles for looking up over these area members, but the savings will be minuscule), but to my eyes this solution also works. 
If the concern is why to not make it a function, it is to save the explicit () that you will have to add otherwise (essentially one would have to do self.redis().publish(), contrary to self.redis.publish() which is needed with this implementation)

src/d3a/d3a_core/simulation.py
```diff
@@ -274,6 +274,7 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
         slot_count = int(config.sim_duration / config.slot_length)
 
         self._update_and_send_results()
+        self.simulation_config.external_redis_communicator.start_communication()
```
spyrostz, line 277: Please move this before the update_and_send_results, otherwise the first results transmission will fail.

src/d3a/models/strategy/external_strategies/__init__.py
faizan2590: Well the main reason was in setup file I have to explicitly pass `config` as a parameter to all areas else they are considering the default `GlobalConfig` from here `src/d3a/models/area/__init__.py:253` which doesn't have `external_redis_communicator` BUT once all the areas have been activated  `config` obj is respectively passed to all area and `GlobalConfig` not being used later

src/d3a/d3a_core/simulation.py
faizan2590: Thanks doing it

PR #714, opened 2020-04-09
Feature/d3 asim 2213: Adding transformer limits as part of throughput stats.

faizan2590:


src/d3a/d3a_core/area_serializer.py
```diff
@@ -134,6 +134,8 @@ def optional(attr):
         external_connection_available = description.get('allow_external_connection', False)
         baseline_peak_energy_import_kWh = description.get('baseline_peak_energy_import_kWh', None)
         baseline_peak_energy_export_kWh = description.get('baseline_peak_energy_export_kWh', None)
+        import_capacity_kVA = description.get('import_capacity_kVA', None)
+        export_capacity_kVA = description.get('export_capacity_kVA', None)
```
hannesdiedrich, line 138: Hmm, @spyrostz didnt we discuss in length, to not increase the length of the parameter list of the area?
I reverted some code because of this decision. I think it was done after we had the discussion with Sarah and we split the two Stories (I never should have proposed that, sorry.)

integration_tests/steps/power_tests.py
```diff
@@ -34,12 +34,20 @@ def test_baseline_peak_energy_stats(context):
         context.simulation.endpoint_buffer.area_throughput_stats.results
 
     expected_results = {'Neighborhood 1': {'import': {'peak_energy_kWh': 0.4,
-                                                      'peak_percentage': 100.0}},
+                                                      'peak_percentage': 100.0,
+                                                      'capacity_kwh': 2.0,
```
spyrostz, line 38: It is better to stick to the common naming and uppercase the 'W' on 'kWh'. Therefore 'kwh' -> 'kWh'

src/d3a/d3a_core/area_serializer.py
hannesdiedrich: I know I saw you coment regarding hte performance on d3a-web, to move the transformer calculation into the d3a.
But still. Next time we have to discuss in more detail, what needs to be done.

src/d3a/d3a_core/area_serializer.py
```diff
@@ -145,7 +147,9 @@ def optional(attr):
                     transfer_fee_const=grid_fee_constant,
                     external_connection_available=external_connection_available,
                     baseline_peak_energy_import_kWh=baseline_peak_energy_import_kWh,
-                    baseline_peak_energy_export_kWh=baseline_peak_energy_export_kWh)
+                    baseline_peak_energy_export_kWh=baseline_peak_energy_export_kWh,
```
spyrostz, line 150: No need to do it now, but there could be a namedtuple that groups these 4 parameters (baseline_peak_energy_import/export, import/export_capacity) and use only the tuple as a constructor argument of the Area class. That way the number of constructor arguments of the area class will be decreased.

src/d3a/d3a_core/sim_results/area_throughput_stats.py
```diff
@@ -30,16 +30,25 @@ def update(self, area):
         self.update_results(area)
 
     def _calc_results(self, area, baseline_value, energy_profile, direction_key):
-        if len(energy_profile.keys()) > 0:
```
hannesdiedrich, line 33: This check is important. `max()` will fail if you parse an empty list.

hannesdiedrich: I see, there is a default value. Nice!

hannesdiedrich: Ignore this comment then.

src/d3a/d3a_core/area_serializer.py
spyrostz: I do not think there was any other way other than passing more parameters here, if we want to do this calculation on d3a. Moreover, which parameters are you talking about ? If this is about area class, there is a much better workaround in order to decrease the parameter / argument count (stated here https://github.com/gridsingularity/d3a/pull/714#discussion_r407919687). let's discuss it in this comment. 
However, what did the code that you reverted do ?

PR #715, opened 2020-04-14
D3ASIM-2227: Omitted saving bid/offer/trade ids in the csv result files.

hannesdiedrich:


src/d3a/setup/jira/d3asim_1139.py
```diff
@@ -9,7 +9,7 @@
 def get_setup(config):
     market_maker_rate = 3
     ConstSettings.IAASettings.MARKET_TYPE = 1
-    config.update_config_parameters(market_maker_rate=market_maker_rate, grid_fee_percentage=0)
+    config.update_config_parameters(market_maker_rate=market_maker_rate)
```
faizan2590, line 12: Can you explain the background of removing `grid_fee_percentage ` from here?

hannesdiedrich: grid_fee_percentage is only an area parameter now, not a global setting.

PR #717, opened 2020-04-14
Bug/d3 asim 2224: Adjusting energy bills table

faizan2590:


src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -67,11 +67,12 @@ def __init__(self, is_spot_market=True):
     @classmethod
     def _store_bought_trade(cls, result_dict, trade):
         # Division by 100 to convert cents to Euros
+        fee_price = trade.fee_price / 100. if trade.fee_price is not None else 0.
         result_dict['bought'] += trade.offer.energy
-        result_dict['spent'] += trade.offer.price / 100.
+        result_dict['spent'] += trade.offer.price / 100. - fee_price
```
hannesdiedrich, line 72: Can you please explain why you subtract the fees here and add them later again?

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -245,17 +246,25 @@ def _generate_external_and_total_bills(self, area, results, flattened):
...
             earned = external.pop("earned")
             bought = external.pop("bought")
             sold = external.pop("sold")
             external.update(**{"spent": earned, "earned": spent, "bought": sold,
-                               "sold": bought, "market_fee": market_fee})
+                               "sold": bought, "market_fee": 0.})
             external["total_energy"] = external["bought"] - external["sold"]
             external["total_cost"] = external["spent"] - external["earned"]
             results[area.name].update({"External Trades": external})
+            results[area.name]["Accumulated Trades"]["bought"] += external["bought"]
```
spyrostz, line 260: To summarize our discussion on slack, it is better if the Accumulated Trades are not recalculated here again. All that is needed is to take into account the External Trades row when calculating the Accumulated Trades row. In order to achieve that the calculation order of these 2 rows needs to be swapped, essentially to calculate first the External Trades row and after that the Accumulated Trades row (it involves just moving the external trade calculation code above the accumulated trades).

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -245,17 +246,25 @@ def _generate_external_and_total_bills(self, area, results, flattened):
...
+            results[area.name]["Accumulated Trades"]["sold"] += external["sold"]
+            results[area.name]["Accumulated Trades"]["spent"] += external["spent"]
+            results[area.name]["Accumulated Trades"]["earned"] += \
```
spyrostz, line 263: As a followup to my other comment, the Accumulated Trades grid fees need to also be a sum of the grid fees of its children (including External Trades). Reason I am not adding it there, is because this code does not appear in the PR and cannot leave a comment. I am referring to line 242 of this file.

src/d3a/d3a_core/sim_results/stats.py
faizan2590: If you look into the attached pdf of ticket "bought" --> "Trade Cost" , it excluded the grid fee. We can discuss offline too if you need further explanation. Thanks!

hannesdiedrich: Ah, i see, so you are omitting the fees fro the device bills, but the accumulated Trades include them. Ok then. Thanks!

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -226,36 +227,40 @@ def _aggregate_totals_with_penalties(self, area, results):
...
+            external["total_cost"] = external["spent"] - external["earned"]
+            results[area.name].update({"External Trades": external})
+
         all_child_results = [v for v in results[area.name].values()]
         results[area.name].update({"Accumulated Trades": {
             'bought': sum(v['bought'] for v in all_child_results),
             'sold': sum(v['sold'] for v in all_child_results),
             'spent': sum(v['spent'] for v in all_child_results),
-            'earned': sum(v['earned'] for v in all_child_results),
+            'earned': sum(v['earned'] for v in all_child_results) + self.market_fees[area.name],
```
spyrostz, line 252: I think this is not correct here. I think the earned value of the Accumulated Trades should not include the market fees (with an emphasis on 'I think', I might have misunderstood what needs to be shown here)

faizan2590: Resolved after offline discussion, no need for change

PR #718, opened 2020-04-15
Feature/d3 asim 2204

spyrostz:


src/d3a/models/market/two_sided_pay_as_bid.py
```diff
@@ -187,7 +187,7 @@ def accept_bid(self, bid: Bid, energy: float = None,
 
         if already_tracked is False:
             self._update_stats_after_trade(trade, bid, bid.buyer, already_tracked)
-            log.info(f"[TRADE][BID] [{self.time_slot_str}] {trade}")
+            log.info(f"[TRADE][BID] [{self.name}] [{self.time_slot_str}] {trade}")
```
hannesdiedrich, line 190: Thanks! I always forgot to add this!

integration_tests/steps/two_sided_market.py
```diff
@@ -199,3 +200,63 @@ def energy_origin(context, producer, consumer):
                 assert trade.seller_origin == producer
 
     assert trade_count > 0
+
+
+@then('trades are matched only on the Grid market')
+def trades_matched_on_grid(context):
+    # Assert that all grid trades contain Offer objects, bid trades are not tracked
+    assert all(type(trade.offer) == Offer
```
faizan2590, line 208: Im wondering if there would be any consequences of using `isinstance` here

spyrostz: There shouldn't be any actually, I'll give it a try to replace them. Thanks

integration_tests/steps/two_sided_market.py
```diff
@@ -199,3 +200,63 @@ def energy_origin(context, producer, consumer):
...
+               for market in context.simulation.area.past_markets
```
faizan2590, line 209: I think you need to explicitly set `KEEP_PAST_MARKETS=True` in the setup file being used else it will only be validated for only one last past markets.

spyrostz: I was under the impression that KEEP_PAST_MARKETS is True by default for the integration tests, however I will explicitly declare it in the setup file in order to cover our base.

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_agent.py
```diff
@@ -25,10 +25,16 @@ class TwoSidedPayAsBidAgent(OneSidedAgent):
...
+        self.engines = [
+            TwoSidedPayAsBidEngine('High -> Low', higher_market, lower_market, min_offer_age,
+                                   min_bid_age, self),
+            TwoSidedPayAsBidEngine('Low -> High', lower_market, higher_market, min_offer_age,
+                                   min_bid_age, self),
+        ]
+        super().__init__(owner=owner,
                          higher_market=higher_market, lower_market=lower_market,
-                         min_offer_age=min_offer_age)
+                         min_offer_age=min_offer_age, do_create_engine=False)
```
hannesdiedrich, line 37: I understand your motivation here, but is it maybe possible instead of setting this boolean variable rather send the engine class that the engines should be build of?

hannesdiedrich: Or is it something one just does not do, sending a class to the super.__init__() ?

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_agent.py
```diff
@@ -16,18 +16,17 @@
...
 class TwoSidedPayAsClearAgent(TwoSidedPayAsBidAgent):
 
     def __init__(self, *, owner, higher_market, lower_market,
-                 min_offer_age=ConstSettings.IAASettings.MIN_OFFER_AGE):
+                 min_offer_age=ConstSettings.IAASettings.MIN_OFFER_AGE,
+                 min_bid_age=ConstSettings.IAASettings.MIN_BID_AGE):
         super().__init__(owner=owner,
                          higher_market=higher_market, lower_market=lower_market,
-                         min_offer_age=min_offer_age,
-                         engine_type=TwoSidedPayAsClearEngine)
```
hannesdiedrich, line 30: Ah we did it here.

hannesdiedrich: So you totally deprecated the  2PAC engine because it is the same as 2PAB right?

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_agent.py
spyrostz: The problem is that now the engine classes do not accept the same arguments, therefore there is no way to have a constructor that accepts the same arguments (unless we use the 'creative' and confusing way that the Area/MarketEvents use, by using * arguments). That obliges us to not construct the engine in the base class, but in the respective agent class constructor. The alternative to this boolean flag is to have an optional engine_objects argument, that would pass the engines to the base classes, and if it is empty only then the base class should create its engine. I find it pretty much the same as the approach with the boolean variable, however if you prefer it that way let me know and I'll implement it .

integration_tests/steps/two_sided_market.py
```diff
@@ -199,3 +200,63 @@ def energy_origin(context, producer, consumer):
...
+               for trade in market.trades)
+
+    # Assert that all House 2 trades contain Offer objects
+    house2 = next(c for c in context.simulation.area.children if c.name == "House 2")
+    assert all(type(trade.offer) == Offer
+               for market in house2.past_markets
+               for trade in market.trades)
+
+
+@then('trades are matched only on the House 1 market')
```
faizan2590, line 225: Some potential enhancement if area name and `Offer/Bid` is parameterised

src/d3a/models/strategy/area_agents/two_sided_pay_as_clear_agent.py
spyrostz: Exactly. The only overridden method was the __repr__ method, therefore the only difference was when printing strings/logs (which we only do in the trace logs for the agents). Therefore definitely not worth it to maintain an extra class here.

integration_tests/steps/two_sided_market.py
spyrostz: I thought of that too, but I thought that it would be way too confusing to understand the test afterwards. If you prefer it that way I can do it quickly, let me know

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_agent.py
hannesdiedrich: OK, yes i get it. No then leave it.
But AFAIS, this boolean you dont need this boolean flag really, as you construct the engines in every agent explicitly anyway. Or did i miss something?

spyrostz: The problem is that the agents inherit from one another. Therefore if you construct the two sided market engines in the two sided agent, you need to avoid constructing the one-sided engine from the base class (which is the one-sided agent). That's why sucha  variable is actually required.

integration_tests/steps/two_sided_market.py
hannesdiedrich: Please leave it like is. Its not that much code and i think it is easier to understand how it is now.

src/d3a/models/area/event_dispatcher.py
```diff
@@ -126,19 +126,46 @@ def event_listener(self, event_type: Union[MarketEvent, AreaEvent], **kwargs):
             self.area.strategy.event_on_disabled_area()
 
     @staticmethod
-    def select_agent_class(is_spot_market):
+    def create_agent_object(owner, higher_market, lower_market, is_spot_market):
         if is_spot_market:
             if ConstSettings.IAASettings.MARKET_TYPE == 1:
                 if ConstSettings.IAASettings.AlternativePricing.PRICING_SCHEME != 0:
-                    return OneSidedAlternativePricingAgent
+                    return OneSidedAlternativePricingAgent(
```
faizan2590, line 133:
```suggestion
                    agent_class = OneSidedAlternativePricingAgent
```

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_agent.py
hannesdiedrich: ah, yes, sorry. You are right of course.

src/d3a/models/area/event_dispatcher.py
```diff
@@ -126,19 +126,46 @@ def event_listener(self, event_type: Union[MarketEvent, AreaEvent], **kwargs):
...
+                        owner=owner,
+                        higher_market=higher_market,
+                        lower_market=lower_market,
+                        min_offer_age=ConstSettings.IAASettings.MIN_OFFER_AGE
+                    )
                 else:
-                    return OneSidedAgent
+                    return OneSidedAgent(
```
faizan2590, line 140:
```suggestion
                    agent_class = OneSidedAgent
```

src/d3a/models/area/event_dispatcher.py
```diff
@@ -126,19 +126,46 @@ def event_listener(self, event_type: Union[MarketEvent, AreaEvent], **kwargs):
...
+                        owner=owner,
+                        higher_market=higher_market,
+                        lower_market=lower_market,
+                        min_offer_age=ConstSettings.IAASettings.MIN_OFFER_AGE
+                    )
```
faizan2590, line 145:
```suggestion
                return agent_class(owner=owner, higher_market=higher_market,
                                   lower_market=lower_market,
                                   min_offer_age=ConstSettings.IAASettings.MIN_OFFER_AGE)

```

spyrostz: I see, so you want to avoid duplication in assignment of constructor arguments here, good idea. I'll do it, thanks

src/d3a/setup/two_sided_market/one_cep_one_load_immediate_match.py
```diff
@@ -0,0 +1,58 @@
...
+            Area(
+                'House 1',
+                [
+                    Area('H1 General Load', strategy=LoadHoursStrategy(
+                        avg_power_W=200,
+                        hrs_per_day=24,
+                        hrs_of_day=list(range(24)),
+                        initial_buying_rate=35,
+                        final_buying_rate=35,
+                        update_interval=duration(minutes=14)
```
faizan2590, line 42: Very minor but it could be squashed into less number of lines.

spyrostz: Let me try to change it, this suggestion is contrary to the rest of the setup files, but I'll do it.

hannesdiedrich: IMO: Please leave it as is. I personally prefer to scroll through a list where each line has one setting than multiple.

spyrostz: Since I also tend to share @hannesdiedrich 's opinion, I left it as is.

PR #724, opened 2020-04-17
D3ASIM-2262: Fixed tracking of trades for storage devices in self suf…

spyrostz:
…ficiency/consumption calculation.

### DIFF
--- a/integration_tests/steps/kpi.py
+++ b/integration_tests/steps/kpi.py
@@ -30,6 +30,7 @@ def test_export_of_kpi_result(context, kpi, expected_kpis):
                                           "aggregated_results", "kpi.json"))
     with open(sim_data_csv[0], "r") as sf:
         kpi_data = json.load(sf)
+    print(kpi_data)
     expected_kpis = ast.literal_eval(expected_kpis)
     for area, value in expected_kpis.items():
         if kpi == "self_sufficiency":
--- a/integration_tests/steps/kpi.py
+++ b/integration_tests/steps/kpi.py
@@ -30,7 +30,6 @@ def test_export_of_kpi_result(context, kpi, expected_kpis):
                                           "aggregated_results", "kpi.json"))
     with open(sim_data_csv[0], "r") as sf:
         kpi_data = json.load(sf)
-    print(kpi_data)
     expected_kpis = ast.literal_eval(expected_kpis)
     for area, value in expected_kpis.items():
         if kpi == "self_sufficiency":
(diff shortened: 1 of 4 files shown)


integration_tests/steps/kpi.py
```diff
@@ -30,6 +30,7 @@ def test_export_of_kpi_result(context, kpi, expected_kpis):
                                           "aggregated_results", "kpi.json"))
     with open(sim_data_csv[0], "r") as sf:
         kpi_data = json.load(sf)
+    print(kpi_data)
```
faizan2590, line 33: In case not to forget before merging

spyrostz: Oops, thanks

src/d3a/d3a_core/sim_results/kpi.py
```diff
@@ -76,10 +76,12 @@ def _accumulate_self_consumption_buffer(self, trade):
             self.self_consumption_buffer_wh += trade.offer.energy * 1000
 
     def _dissipate_self_consumption_buffer(self, trade):
-        if trade.seller_origin in self.ess_list and trade.seller_origin == trade.seller:
```
faizan2590, line 79: I recall this check (... `and trade.seller_origin == trade.seller`) was to take into account the intermediate hierarchies. Right?

src/d3a/d3a_core/sim_results/kpi.py
```diff
@@ -76,10 +76,12 @@ def _accumulate_self_consumption_buffer(self, trade):
...
+        if trade.seller_origin in self.ess_list:
             # self_consumption_buffer needs to be exhausted to total_self_consumption
             # if sold to internal consumer
-            if trade.buyer_origin in self.consumer_list and self.self_consumption_buffer_wh > 0:
+            if trade.buyer_origin in self.consumer_list and \
+                    trade.buyer_origin == trade.buyer and \
```
hannesdiedrich, line 83: So there was a double counting and you added a check to only track the trade that happens in the market where the device is connected to?

src/d3a/d3a_core/sim_results/kpi.py
spyrostz: Exactly, now this is dealt by checking the buyer_origin in each individual if clause.

src/d3a/d3a_core/sim_results/kpi.py
spyrostz: As I mentioned in the reply to Faizan, the tracking is now done in the buyer side, there was a check to avoid double counting the same offer in the seller side before, but the tracking needed to be done on the buyer side for it to be successful.

PR #726, opened 2020-04-17
Add storage loss controls

faizan2590:


faizan2590:
> It would be nice to have a ticket that tracks the DoD of this.
> From what i see here, it can be usefull, but I can not validate whether the job is done.
> I guess that @chrisd12 is in contact with you about it!
> So, leaving a LGTM here

@hannesdiedrich If you look into my commit, it is tracked with `D3ASIM-1951` ticket, where @chrisd12 needed help to integrate his commit where he wrote an small loss function for storage. I agree though I should have renamed the branch as per ticket number to avoid this confusion

PR #727, opened 2020-04-20
D3ASIM-2079: Returning None values for non-valid slots as part of red…

faizan2590:
…is market_stats response channel.

### DIFF
--- a/src/d3a/models/area/stats.py
+++ b/src/d3a/models/area/stats.py
@@ -20,6 +20,14 @@ from statistics import mean, median
 from d3a_interface.constants_limits import DATE_TIME_FORMAT
 from d3a.constants import TIME_ZONE
 from d3a import limit_float_precision
+from copy import deepcopy
+
+default_trade_stats_dict = {
+    "min_trade_rate": None,
+    "max_trade_rate": None,
+    "avg_trade_rate": None,
+    "median_trade_rate": None,
+    "total_traded_energy_kWh": None}
 
 
 class AreaStats:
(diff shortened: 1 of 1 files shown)


src/d3a/models/area/stats.py
```diff
@@ -140,19 +140,18 @@ def current_market(self):
         return past_markets[-1] if len(past_markets) > 0 else None
 
     def get_market_stats(self, market_slot_list):
-        if self.current_market is None:
-            return {"INFO": "No market stats available yet"}
+        default_stats_dict = {"min_trade_rate": None,
+                              "max_trade_rate": None,
+                              "avg_trade_rate": None,
+                              "median_trade_rate": None,
+                              "total_traded_energy_kWh": None}
```
hannesdiedrich, line 147: Can you please make this a global variable and reuse it in the `min_max_avg_median_rate_current_market`?
Thanks!

src/d3a/models/area/stats.py
```diff
@@ -118,11 +126,7 @@ def get_price_stats_current_market(self):
                 if self.current_market.time_slot in self.rate_stats_market else None
 
     def min_max_avg_median_rate_current_market(self):
-        out_dict = {"min_trade_rate": None,
-                    "max_trade_rate": None,
-                    "avg_trade_rate": None,
-                    "median_trade_rate": None,
-                    "total_traded_energy_kWh": None}
+        out_dict = deepcopy(default_trade_stats_dict)
```
spyrostz, line 129: Performance impact is small, but please use a shallow copy here (essentially only `copy` and not `deepcopy`) because there are no references in the dict, and even a shallow copy will work fine.

PR #731, opened 2020-04-21
D3ASIM-2280: Incorporating unsold ess offer/bid in energy_to_sell/ene…

faizan2590:
…rgy_to_buy.

### DIFF
--- a/src/d3a/models/strategy/external_strategies/storage.py
+++ b/src/d3a/models/strategy/external_strategies/storage.py
@@ -286,10 +286,11 @@ class StorageExternalMixin(ExternalMixin):
 
     @property
     def _device_info_dict(self):
-        energy_to_sell = self.state.energy_to_sell_dict[self.market.time_slot] - \
-                         self.state.offered_sell_kWh[self.market.time_slot]
+        energy_to_sell = \
+            self.state.energy_to_sell_dict[self.market.time_slot] + \
+            self.state.offered_sell_kWh[self.market.time_slot]
         energy_to_buy = \
-            self.state.energy_to_buy_dict[self.market.time_slot] - \
+            self.state.energy_to_buy_dict[self.market.time_slot] + \
             self.state.offered_buy_kWh[self.market.time_slot]
 
         return {
--- a/src/d3a/models/strategy/external_strategies/storage.py
+++ b/src/d3a/models/strategy/external_strategies/storage.py
@@ -44,6 +44,19 @@ class StorageExternalMixin(ExternalMixin):
             f'{self.channel_prefix}/device_info': self._device_info,
         })
 
+    def _is_ess_buy_limit_surpassed(self, energy):
+        return energy > self.state.energy_to_buy_dict[self.market.time_slot]
+
+    def _is_ess_sell_limit_surpassed(self, energy):
+        return energy > self.state.energy_to_sell_dict[self.market.time_slot]
+
+    def _publish_error_message_to_response_channel(self, response_channel, transaction_id):
+        self.redis.publish_json(
+            response_channel,
+            {"command": "bid",
+             "error": "ESS energy limit for this market_slot has been surpassed.",
+             "transaction_id": transaction_id})
+
     def _list_offers(self, payload):
         self._get_transaction_id(payload)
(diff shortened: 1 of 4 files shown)


src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -129,6 +129,15 @@ def _offer(self, payload):
             assert set(arguments.keys()) == {'price', 'energy', 'transaction_id'}
             arguments['seller'] = self.device.name
             arguments['seller_origin'] = self.device.name
+            if arguments['energy'] > self.state.energy_to_sell_dict[self.market.time_slot]:
+                "ESS energy limit for this market_slot has been surpassed."
```
hannesdiedrich, line 133: This is probably a remnant. I wonder why the checks dont fail.
Please remove this line. Thanks!

spyrostz: This is not an illegal statement as far as I know. Python allows single variables as statements, even without assignment, that's why it does not fail. I agree to remove the line though

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -129,6 +129,15 @@ def _offer(self, payload):
             assert set(arguments.keys()) == {'price', 'energy', 'transaction_id'}
             arguments['seller'] = self.device.name
             arguments['seller_origin'] = self.device.name
+            if arguments['energy'] > self.state.energy_to_sell_dict[self.market.time_slot]:
```
spyrostz, line 132: I think it would be great to also add this check in the `_offer_impl` method too, as a second layer of protection. Reason for that, is in case a new offer is posted in the tick_event, before the incoming message is handled.

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -246,6 +253,14 @@ def _bid(self, payload):
             assert set(arguments.keys()) == {'price', 'energy', 'transaction_id'}
             arguments['buyer'] = self.device.name
             arguments['buyer_origin'] = self.device.name
+            if arguments["energy"] > self.state.energy_to_buy_dict[self.market.time_slot]:
```
spyrostz, line 256: Same here, the same check should apply to the `_bid_impl` IMO

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -44,6 +44,19 @@ def event_activate(self):
             f'{self.channel_prefix}/device_info': self._device_info,
         })
 
+    def _is_ess_buy_limit_surpassed(self, energy):
+        return energy > self.state.energy_to_buy_dict[self.market.time_slot]
+
+    def _is_ess_sell_limit_surpassed(self, energy):
+        return energy > self.state.energy_to_sell_dict[self.market.time_slot]
+
+    def _publish_error_message_to_response_channel(self, response_channel, transaction_id):
```
hannesdiedrich, line 53: This function is only for error messages concerning the ESS limit.
Please either rename the function or transform it into an even more generic function that takes the command, error and transaction_id as input.
I guess that the first will be sufficient though.

spyrostz, line 53: I would strongly suggest to remove this function and to revert the assert method. Reason for that, this function is used for both bids and offers, but the command field of the response is always 'bid'. Moreover, the status: 'error' field is not reported. Both these issues will be fixed if you revert to the former implementation.

spyrostz: @hannesdiedrich I would prefer if the old way was retained where an assertion is raised, and caught afterwards. That way we are sure that in the impl method there is only one publish_json for an error condition, which makes the code more uniform.

hannesdiedrich: You are right, agreed!

faizan2590: Sorry, now im confused, before my recent commit, I was using an if clause and publishing message to response channel, I wasn't using `assert` for the failure of `try except block`.

faizan2590: Discussed offline all the state validation will be moved to `_impl` method and keeping generic error message

PR #733, opened 2020-04-22
D3ASIM-1849: Added option to export energy trade plots for each market…

hannesdiedrich:
… slot.


src/d3a/d3a_core/export.py
```diff
@@ -662,3 +664,74 @@ def _plot_avg_trade_graph(self, stats, area_name, key, label):
...
+
+    def _plot_energy_profile_hr(self, area: Area, subdir: str):
+        """
+        Plots history of energy trades plotted for each market_slot
+        """
+        barmode = "relative"
+        xtitle = 'Time'
+        ytitle = 'Energy [kWh]'
+        market_name = area.name
+        title = f'HR Energy Trade Profile of {market_name}'
```
faizan2590, line 686:
```suggestion
        title = f'Hourly Energy Trade Profile of {market_name}'
```

hannesdiedrich: Why hourly? It is per market slot. 
I named it HR for High resolution. But I am open for other names as well.

faizan2590: Please `High resolution` as you proposed and at least avoid confusing short forms for the title as its purpose is to aid in understanding and not lead to more confusion

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -91,9 +91,9 @@ def tick(self, *, area):
 
         for bid_id, bid in self.markets.source.get_bids().items():
             if bid.id not in self.bid_age:
-                self.bid_age[bid.id] = area.current_tick
+                self.bid_age[bid.id] = area.current_tick_in_slot
```
spyrostz, line 94: This is not the current tick in the slot actually. IIRC this is the current tick for the whole simulation, therefore it never resets to 0.

src/d3a/d3a_core/export.py
```diff
@@ -135,6 +135,8 @@ def export(self, export_plots=True, power_flow=None):
             self.plot_ess_energy_trace(self.area, self.plot_dir)
             if ConstSettings.GeneralSettings.EXPORT_DEVICE_PLOTS:
                 self.plot_device_stats(self.area, [])
+            if ConstSettings.GeneralSettings.EXPORT_ENERGY_TRADE_PROFILE_HR:
+                self.plot_energy_trade_profile_hr(self.area, self.plot_dir)
             if ConstSettings.IAASettings.MARKET_TYPE == 3 and \
                     ConstSettings.GeneralSettings.SUPPLY_DEMAND_PLOTS:
```
spyrostz, line 141: Since you are in the neighbourhood, how about renaming this to `EXPORT_SUPPLY_DEMAND_PLOTS` ?

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
hannesdiedrich: yes, this is true. This due to autorefactoring of Pycharm probably.
I wonder why this does not lead to a failing test.
BUT: I found out, that the "clock" of a market can only run for one slot_length, this is why, the market clock has to be updated with the current_tick_in_slot.

src/d3a/models/area/stats.py
```diff
@@ -50,6 +52,21 @@ def update_area_market_stats(self):
                 if "bills" in self.aggregated_stats else None
             self.rate_stats_market[self.current_market.time_slot] = \
                 self.min_max_avg_median_rate_current_market()
+            # TODO: This accumulation of trade data could potentially also used for the
```
spyrostz, line 55: No need to do anything, but I think that there is a need to somehow manage the data that we accumulate over time, instead of just saving the data in each individual class that is part of the endpoint buffer. I mean that we could reuse this class in the future, instead of saving copies of our results in the MarketEnergyBills, DeviceStatistics ... and other classes in the endpoint_buffer.

hannesdiedrich: This is actually what i wanted to say in the comment. :-) Shall I rephrase it?

tests/test_inter_area_agent.py
```diff
@@ -349,7 +350,7 @@ def test_iaa_forwards_offers_according_to_percentage(iaa_fee):
                                 higher_market=higher_market,
                                 lower_market=lower_market)
     iaa.event_tick()
-    iaa.owner.current_tick = 14
+    iaa.owner.current_tic = 14
```
faizan2590, line 353: typo

src/d3a/d3a_core/export.py
```diff
@@ -662,3 +665,74 @@ def _plot_avg_trade_graph(self, stats, area_name, key, label):
...
+            seller_dict[data["seller"]]["timestamp"].append(time)
+            seller_dict[data["seller"]]["energy"].append(data["energy"] * ENERGY_SELLER_SIGN_PLOTS)
+            buyer_dict[data["buyer"]]["timestamp"].append(time)
+            buyer_dict[data["buyer"]]["energy"].append(data["energy"] * ENERGY_BUYER_SIGN_PLOTS)
+
+        # 2. Create bar plot objects and collect them in a list
+        for agent, data in seller_dict.items():
+            data_obj = go.Bar(x=data["timestamp"],
+                              y=data["energy"],
+                              width=GlobalConfig.tick_length.seconds * 1000,
```
faizan2590, line 728: Would be nice to replace the constant factor with any variable

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -204,10 +205,8 @@ def _calculate_devices_sold_bought_energy_past_markets(self, area, past_markets)
         out_dict = {"sold_energy": {}, "bought_energy": {}}
         for market in past_markets:
             for trade in market.trades:
-                trade_seller = trade.seller[4:] if trade.seller.startswith("IAA ") \
-                    else trade.seller
-                trade_buyer = trade.buyer[4:] if trade.buyer.startswith("IAA ") \
-                    else trade.buyer
+                trade_seller = area_name_from_area_or_iaa_name(trade.seller)
+                trade_buyer = area_name_from_area_or_iaa_name(trade.buyer)
```
faizan2590, line 209: 👍

src/d3a/d3a_core/export.py
hannesdiedrich: Yeah. After long search i found out, that if the xaxis is in datetime format, the widths of the bars can be set by providing a number in milliseconds. Shall I leave a comment about this in the code? I would not want to create a variable in constants.py. Or what are you suggesting?

faizan2590: Yes please leave a comment about this in the code.

src/d3a/models/area/stats.py
spyrostz: Ah great then! It is fine as it is, no need to rephrase it.

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
spyrostz: The latter indeed is correct, there is no need to send the tick number for the whole simulation, tick in slot should suffice.

hannesdiedrich:
> Reading the description of story, did you double check with @sarahhambridge if scroll bar is no more needed?

I asked Sarah and she is ok with it to have individual plots for now in the backend.

PR #734, opened 2020-04-22
D3ASIM-2267: Decoupled the transformer and baseline peak results to b…

spyrostz:
…e generated independently.


src/d3a/d3a_core/sim_results/area_throughput_stats.py
```diff
@@ -36,30 +36,48 @@ def _calc_results(self, area, baseline_value, energy_profile, direction_key):
...
+                area_results["export"].update(
+                    self._calc_results(area, baseline_export, self.exported_energy, "export")
+                )
+
+        import_capacity = area.import_capacity_kWh
+        export_capacity = area.export_capacity_kWh
+        if import_capacity is not None and import_capacity > 0:
+            area_results["import"].update(self._calc_transformer_results(area, "import"))
+        if export_capacity is not None and export_capacity > 0:
+            area_results["export"].update(self._calc_transformer_results(area, "export"))
```
faizan2590, line 76: From the description `send capacity_kWh whenever import_capacity_kVA and export_capacity_kVA have values`. Does that mean `capacity_kWh` has to be send when `import_capacity_kVA and export_capacity_kVA` have values ?

spyrostz: They are both supposed to be independent of each other, however thanks a lot for the comment, because due to the comment I had another look at the bug ticket where Sarah mentions to `always send peak_energy_kWh values`. I adapted that, and then I noticed strange values in the percentage (like 200% in cases where the calculated peak and the predefined peak are the same, where essentially the percentage had to be 100%). As a followup I discovered a bug where the imported_energy and exported_energy were not calculated per-area, but each area replaced the previous area's values. I also added a fix for this here, and (hopefully) this feature is ready to go now.

src/d3a/d3a_core/sim_results/area_throughput_stats.py
```diff
@@ -83,20 +88,26 @@ def update_results(self, area):
...
+                add_or_create_key(self.imported_energy[area.uuid], current_market.time_slot_str,
                                   trade.offer.energy)
-        if current_market.time_slot_str not in self.imported_energy:
-            self.imported_energy[current_market.time_slot_str] = 0.
-        if current_market.time_slot_str not in self.exported_energy:
-            self.exported_energy[current_market.time_slot_str] = 0.
+        if current_market.time_slot_str not in self.imported_energy[area.uuid]:
+            self.imported_energy[area.uuid][current_market.time_slot_str] = 0.
+        if current_market.time_slot_str not in self.exported_energy[area.uuid]:
+            self.exported_energy[area.uuid][current_market.time_slot_str] = 0.
```
hannesdiedrich, line 113: Ah yes, this is a remnant of moving this out of the AreaStats into AreaThroughputStats. Thanks for fixing!

src/d3a/d3a_core/sim_results/area_throughput_stats.py
```diff
@@ -34,17 +34,18 @@ def _calc_results(self, area, baseline_value, energy_profile, direction_key):
...
             baseline_peak_energy_kWh = round_floats_for_ui(area.baseline_peak_energy_export_kWh)
         return {
-            "peak_energy_kWh": peak_energy_kWh,
             "peak_percentage": peak_percentage,
             "baseline_peak_energy_kWh": baseline_peak_energy_kWh,
          }
 
+    def _calc_peak_energy_results(self, energy_profile):
+        return {"peak_energy_kWh": round_floats_for_ui(max(energy_profile.values(), default=0.0))}
+
```
hannesdiedrich, line 48: This is just refactoring, right?

spyrostz: No, this is actually a requirement from the bug ticket. Sarah wants the `peak_energy_kWh` value to always be reported for every area.

hannesdiedrich: Ah yes, makes sense! This is the value the user has to see to set a Baseline to see the percentage in the next run. Thanks!

PR #739, opened 2020-04-24
Bug/d3 asim 2179

spyrostz:


src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -98,13 +85,36 @@ def update_sold_bought_energy(self, area: Area):
...
+                    self.traded_energy[area.uuid] = {"sold_energy": {}, "bought_energy": {}}
+                self._calculate_devices_sold_bought_energy(self.traded_energy[area.uuid],
+                                                           area.current_market)
+                self.traded_energy[area.name] = self.traded_energy[area.uuid]
+                if area.name not in self.balancing_traded_energy:
+                    self.balancing_traded_energy[area.name] = {"sold_energy": {},
+                                                               "bought_energy": {}}
+                if len(area.past_balancing_markets) > 0:
+                    self._calculate_devices_sold_bought_energy(
+                        self.balancing_traded_energy[area.name], area.current_balancing_market)
```
hannesdiedrich, line 115: I think this whole mechanism could be refactored and simplified in the frame of 2212.
But thanks for the starting point!

src/d3a/d3a_core/simulation.py
```diff
@@ -279,9 +280,10 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
         self.simulation_config.external_redis_communicator.start_communication()
         self._update_and_send_results()
         for slot_no in range(slot_resume, slot_count):
+
             self._update_progress_info(slot_no, slot_count)
 
-            log.info(
+            log.warning(
```
hannesdiedrich, line 286: Probably a remnant of testing. Pleas revert to INFO. Thanks

src/d3a/d3a_core/simulation.py
```diff
@@ -356,13 +358,18 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
-                self.export.export()
+                self.export.export(self.should_export_plots)
 
         if self.use_repl:
             self._start_repl()
 
+    @property
+    def should_export_plots(self):
+        return not self.redis_connection.is_enabled()
+
```
hannesdiedrich, line 372: In line 338 the CLI flag is also incorporated: 
`if self.export_on_finish or self.redis_connection.is_enabled():`
How about adding it here too?

src/d3a/models/area/event_dispatcher.py
```diff
@@ -215,8 +215,13 @@ def _delete_past_agents(self, area_agent_member):
                     if hasattr(agent, "engines"):
                         for engine in agent.engines:
                             del engine.forwarded_offers
+                            del engine.offer_age
+                            del engine.trade_residual
+                            del engine.ignored_offers
```
hannesdiedrich, line 220: Out of interest: How much memory did this change save?

src/d3a/models/area/markets.py
```diff
@@ -87,14 +85,14 @@ def _delete_past_markets(self, past_markets, timeframe):
...
-                past_markets[pm].traded_energy = {}
-                past_markets[pm].accumulated_actual_energy_agg = {}
+                del past_markets[pm].offers
+                del past_markets[pm].trades
+                del past_markets[pm].offer_history
+                del past_markets[pm].notification_listeners
+                del past_markets[pm].bids
+                del past_markets[pm].bid_history
+                del past_markets[pm].traded_energy
+                del past_markets[pm].accumulated_actual_energy_agg
```
hannesdiedrich, line 95: Really? Does it make a difference or is it "just" a cleaner way?

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -225,6 +224,6 @@ def event_bid_split(self, *, market_id, original_bid, accepted_bid, residual_bid
                              f"{short_offer_bid_log_str(local_residual_bid)}")
 
     def _add_to_forward_bids(self, source_bid, target_bid):
-        bid_info = BidInfo(deepcopy(source_bid), deepcopy(target_bid))
+        bid_info = BidInfo(source_bid, target_bid)
```
hannesdiedrich, line 227: So I guess that the original objects of source_bid and target_bid stayed in memory and could not be deleted because the link was lost to them? Or is this something different?

src/d3a/models/strategy/load_hours.py
```diff
@@ -134,11 +134,23 @@ def event_market_cycle(self):
...
+        if ConstSettings.GeneralSettings.KEEP_PAST_MARKETS is True or \
+                self.area.current_market is None:
+            return
+        to_delete = []
+        for k in self.energy_requirement_Wh.keys():
+            if k < self.area.current_market.time_slot:
+                to_delete.append(k)
+        for k in to_delete:
+            del self.energy_requirement_Wh[k]
+            del self.state.desired_energy_Wh[k]
```
hannesdiedrich, line 153: Nice!

src/d3a/models/strategy/predefined_load.py
```diff
@@ -77,20 +77,20 @@ def __init__(self, daily_load_profile,
...
         for each slot.
         :return: None
         """
-        self.load_profile = read_arbitrary_profile(
+        load_profile = read_arbitrary_profile(
             InputProfileTypes.POWER,
             self.daily_load_profile)
-        self._update_energy_requirement()
+        self._update_energy_requirement(load_profile)
+        del self.daily_load_profile
```
hannesdiedrich, line 91: I guess that this makes a different when you have a large number of predefined loads, right?

src/d3a/d3a_core/sim_results/file_export_endpoints.py
spyrostz: I agree, it is definitely worth a refactoring. Adding it in the description of 2212.

src/d3a/d3a_core/simulation.py
spyrostz: Will do, thanks!

src/d3a/d3a_core/simulation.py
spyrostz: `self.export_on_finish` is responsible for controlling whether the csv files are exported too. In the case of redis connection, we still want the csv files to be produced, we do not want the plots to be exported though.

src/d3a/models/area/event_dispatcher.py
spyrostz: Cannot say exactly, because this was mostly an extra security layer. The real change that actually forced the IAAs to be deleted was this: https://github.com/gridsingularity/d3a/pull/739/files#diff-d67f7f839ba3365d818a4d178e4b487fL79 . I can run a (not so quick) test though if you are interested, ping me if you would like to. Thanks!

src/d3a/models/area/markets.py
spyrostz: Should not make a difference. I was trying to make it consistent with the IAAs actually. It also is clearer to my eyes the intent of these calls (essentially a `del` indicates that the argument does not plan to be reused, contrary to `{}` which indicates that the argument will be reused - to my eyes at least)

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
spyrostz: Indeed. I saw a huge amount of memory (about 50 MBs) of type `Bid` after using this tool https://pypi.org/project/memory-analyzer/ . Once I threw out the deepcopies, this was decreased by a lot.

src/d3a/models/strategy/predefined_load.py
spyrostz: Indeed. I am not sure if it is worth it actually, because this memory allocation is visible from the start of the simulation, despite that does not hurt to get rid of the profile since it is not needed.

src/d3a/d3a_core/simulation.py
hannesdiedrich: Yes, you are right. Ignore my comment then.

src/d3a/models/area/markets.py
hannesdiedrich: Agreed! Thanks!

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
hannesdiedrich: Awesome!

src/d3a/d3a_core/sim_results/file_export_endpoints.py
hannesdiedrich: Thanks!

PR #740, opened 2020-04-28
D3ASIM-2302: Added cumulative energy bills. Fixed issue with market f…

spyrostz:
…ee calculation. Added integration tests for the cumulative bills.

### DIFF
--- a/integration_tests/integration_tests.feature
+++ b/integration_tests/integration_tests.feature
@@ -50,6 +50,16 @@ Feature: Run integration tests
       |      two_sided_market.default_2a       |
       |    two_sided_pay_as_clear.default_2a   |
 
+  Scenario Outline: Test Cumulative Bills
+    Given we have a scenario named <scenario>
+    And d3a is installed
+    When we run the simulation with setup file <scenario> and parameters [24, 60, 60, 1]
+    Then the cumulative energy bills for each area are the sum of its children
+  Examples: Settings
+      |                scenario                |
+      |               default_2a               |
+      |      two_sided_market.default_2a       |
+      |    two_sided_pay_as_clear.default_2a   |
 
   Scenario Outline: Simulation subscribes on the appropriate channels
      Given d3a is installed
@@ -137,6 +147,16 @@ Feature: Run integration tests
      And the past markets are not kept in memory
      And we run the simulation with setup file default_2a and parameters [24, 60, 60, 1]
      Then the energy bills are identical no matter if the past markets are kept
+
+  Scenario: Cumulative bills are the same with and without keeping the past markets
+     Given we have a scenario named default_2a
+     And d3a is installed
+     And the past markets are kept in memory
+     When we run the simulation with setup file default_2a and parameters [24, 60, 60, 1]
+     And the reported cumulative bills are saved
+     And the past markets are not kept in memory
+     And we run the simulation with setup file default_2a and parameters [24, 60, 60, 1]
+     Then the cumulative bills are identical no matter if the past markets are kept
 
   Scenario Outline: Price energy day results are the same with and without keeping the past markets
      Given we have a scenario named <scenario>
--- a/integration_tests/integration_tests.feature
+++ b/integration_tests/integration_tests.feature
@@ -60,6 +60,7 @@ Feature: Run integration tests
       |               default_2a               |
       |      two_sided_market.default_2a       |
       |    two_sided_pay_as_clear.default_2a   |
+      |          grid_fees.default_2a          |
 
   Scenario Outline: Simulation subscribes on the appropriate channels
      Given d3a is installed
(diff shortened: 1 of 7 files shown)


src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -136,6 +147,65 @@ def _store_area_penalties(self, results_dict, area):
...
+            }
+            for uuid, results in self.cumulative_bills_results.items()
+        }
+
+    def update_cumulative_bills(self, area):
+        for child in area.children:
+            self.update_cumulative_bills(child)
+
+        if area.uuid not in self.cumulative_bills_results or \
+                ConstSettings.GeneralSettings.KEEP_PAST_MARKETS is True:
```
faizan2590, line 168: I think this check could be skipped if only last_past_market data is accumulated.

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -82,6 +84,7 @@ def _store_sold_trade(cls, result_dict, trade):
         result_dict['earned'] += (trade.offer.price - fee_price) / 100.
         result_dict['total_energy'] -= trade.offer.energy
         result_dict['total_cost'] -= (trade.offer.price - fee_price) / 100.
+        result_dict['market_fee'] += fee_price / 100.0
```
hannesdiedrich, line 87: This change has to be decided in D3ASIM-2303

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -112,18 +115,26 @@ def _get_child_data(self, area):
...
                 for market in self._get_past_markets_from_area(area.parent, "past_markets"))
         elif isinstance(area.strategy, PVStrategy):
-            penalty_energy = sum(
+            return sum(
                 area.strategy.state.available_energy_kWh.get(market.time_slot, 0)
                 for market in self._get_past_markets_from_area(area.parent, "past_markets"))
         else:
+            return None
+
+    def _store_area_penalties(self, results_dict, area):
```
hannesdiedrich, line 133: This will deprecated in the frame of D3ASIM-2303.
OR, you could also remove all penalties and totals_with_penalties from the "normal" bills code.

src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -136,6 +147,65 @@ def _store_area_penalties(self, results_dict, area):
...
+            spent_total = sum(trade.offer.price
+                              for trade in trades
+                              if trade.buyer == area.name) / 100.0
+            earned = sum(trade.offer.price - trade.fee_price
+                         for trade in trades
+                         if trade.seller == area.name) / 100.0
+            penalty_energy = self._calculate_device_penalties(area)
+            if penalty_energy is None:
+                penalty_energy = 0.0
+            penalties = penalty_energy * DEVICE_PENALTY_RATE / 100.0
```
hannesdiedrich, line 202: please rename to `penalty_cost` or similar. Thanks!

integration_tests/steps/integration_tests.py
```diff
@@ -662,6 +666,7 @@ def run_sim(context, scenario, total_duration, slot_length, tick_length, market_
         )
         context.simulation.run()
     except Exception as er:
+        root_logger.critical(f"Error reported when running the simulation: {er}")
```
hannesdiedrich, line 669: Nice!

integration_tests/integration_tests.feature
```diff
@@ -50,6 +50,16 @@ Feature: Run integration tests
...
       |    two_sided_pay_as_clear.default_2a   |
 
+  Scenario Outline: Test Cumulative Bills
+    Given we have a scenario named <scenario>
+    And d3a is installed
+    When we run the simulation with setup file <scenario> and parameters [24, 60, 60, 1]
+    Then the cumulative energy bills for each area are the sum of its children
+  Examples: Settings
+      |                scenario                |
+      |               default_2a               |
```
hannesdiedrich, line 60: As I intended in the D3ASIM-2303 we should include a setup file that has non zero grid fees. Thanks!

src/d3a/d3a_core/sim_results/stats.py
spyrostz: Of course it can work even without it, if we access only the last past market. However, this defeats the purpose of always recalculating it when the KEEP_PAST_MARKETS is set to true.

src/d3a/d3a_core/sim_results/stats.py
spyrostz: Will remove it from here.

src/d3a/d3a_core/sim_results/stats.py
spyrostz: As discussed, will do the latter. Thanks

src/d3a/d3a_core/sim_results/stats.py
spyrostz: Fixing it thanks

integration_tests/integration_tests.feature
spyrostz: Actually, after watching it, I did not include the grid fees in it (was not required). They were included implicitly though through the spent_total, therefore adding these here.

PR #743, opened 2020-04-29
D3ASIM-2302: Refactored cumulative bills to their own class.

spyrostz:


src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -188,6 +147,51 @@ def update_cumulative_bills(self, area):
...
+    @classmethod
+    def _store_sold_trade(cls, result_dict, trade):
+        # Division by 100 to convert cents to Euros
+        fee_price = trade.fee_price if trade.fee_price is not None else 0.
+        result_dict['sold'] += trade.offer.energy
+        result_dict['earned'] += (trade.offer.price - fee_price) / 100.
+        result_dict['total_energy'] -= trade.offer.energy
+        result_dict['total_cost'] -= (trade.offer.price - fee_price) / 100.
+
+    def _default_area_dict(self, area):
```
faizan2590, line 177: I think it could also be a class method

spyrostz: Indeed no need for self here, converting this to a class method.

PR #750, opened 2020-05-04
D3ASIM-2286: Adapted precision on price energy day endpoint to 3 deci…

spyrostz:
…mal points.


spyrostz:
> LGTM but would be nice to use pre-existed utility function `round_floats_for_ui`

True, I will adapt it.

PR #754, opened 2020-05-06
D3ASIM-2325: Enhancing start, pause, resume, rerun, abort support of …

faizan2590:
…siulation via UI.

### DIFF
--- a/src/d3a/d3a_core/d3a_jobs.py
+++ b/src/d3a/d3a_core/d3a_jobs.py
@@ -40,7 +40,7 @@ def decompress_and_decode_queued_strings(queued_string):
 
 @job('d3a')
 def start(scenario, settings, events):
-    logging.getLogger().setLevel(logging.ERROR)
+    logging.getLogger().setLevel(logging.INFO)
 
     scenario = decompress_and_decode_queued_strings(scenario)
 
--- a/src/d3a/d3a_core/d3a_jobs.py
+++ b/src/d3a/d3a_core/d3a_jobs.py
@@ -40,7 +40,7 @@ def decompress_and_decode_queued_strings(queued_string):
 
 @job('d3a')
 def start(scenario, settings, events):
-    logging.getLogger().setLevel(logging.INFO)
+    logging.getLogger().setLevel(logging.ERROR)
 
     scenario = decompress_and_decode_queued_strings(scenario)
 
--- a/src/d3a/d3a_core/redis_connections/redis_communication.py
+++ b/src/d3a/d3a_core/redis_connections/redis_communication.py
@@ -69,16 +69,62 @@ class RedisSimulationCommunication:
     def _reset_callback(self, _):
         self._simulation.reset()
 
-    def _stop_callback(self, _):
+    def _stop_callback(self, payload):
+        response = json.loads(payload["data"])
         self._simulation.stop()
+        stop_response_channel = f'{self._simulation_id}/response/stop'
+        if self._simulation.is_stopped:
+            self.publish_json(
+                stop_response_channel,
+                {"command": "pause", "status": "success",
+                 "simulation_id": str(self._simulation_id),
+                 "transaction_id": response["transaction_id"]})
+        else:
+            self.publish_json(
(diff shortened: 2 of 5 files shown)


src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -40,7 +40,7 @@ def decompress_and_decode_queued_strings(queued_string):
 
 @job('d3a')
 def start(scenario, settings, events):
-    logging.getLogger().setLevel(logging.ERROR)
+    logging.getLogger().setLevel(logging.INFO)
```
spyrostz, line 43: Can you please revert this to its initial value before merging? Thanks!

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -69,16 +69,62 @@ def _subscribe_to_channels(self):
     def _reset_callback(self, _):
         self._simulation.reset()
 
-    def _stop_callback(self, _):
+    def _stop_callback(self, payload):
+        response = json.loads(payload["data"])
         self._simulation.stop()
+        stop_response_channel = f'{self._simulation_id}/response/stop'
+        if self._simulation.is_stopped:
```
spyrostz, line 76: The publish of the response in all 3 commands could be refactored and reused. The publish_json calls are almost identical, and even the if / else blocks can be abstracted in a method that will be reused by all 3 endpoints.

src/d3a/d3a_core/simulation.py
```diff
@@ -305,12 +305,16 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
             log.debug(f"Used {mbs_used} MBs.")
 
             for tick_no in range(tick_resume, config.ticks_per_slot):
+                tick_start = time.monotonic()
```
spyrostz, line 308: Why did you use monotonic clock here ? DST is not enabled in our services IIRC, therefore the monotonic clock will not be of use. Better to revert to the standard `time` method that we use elsewhere.

src/d3a/d3a_core/simulation.py
```diff
@@ -305,12 +305,16 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
+                while self.paused:
```
spyrostz, line 309: You need to add some kind of sleep in the while loop, because the current implementation would run the CPU in 100% needlessly. Alternatively, look at the existing implementation which does not exhibit this behavior.

src/d3a/d3a_core/simulation.py
faizan2590: Umm, tick_start was already using monotonic_clock so I continued the legacy and reused it to hold the simulation for certain time before stopping it in case the used the user didn't resume it.

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -66,19 +65,49 @@ def _subscribe_to_channels(self):
         self.pubsub.subscribe(**self._sub_callback_dict)
         self.pubsub.run_in_thread(sleep_time=0.5, daemon=True)
 
-    def _reset_callback(self, _):
-        self._simulation.reset()
-
-    def _stop_callback(self, _):
+    def _generate_redis_response(self, response, simulation_id, flag, command_type):
```
spyrostz, line 68: I think that `flag` would be more appropriately named as `is_successful`

src/d3a/d3a_core/simulation.py
spyrostz: I see now that this was there right from the start. I would actually suggest to change this to `time.time()`. I think that there will be no problem with it, and then timing will be consistent across d3a.

hannesdiedrich: @faizan2590 I dont see Spyros' suggestion adapted. Was there an offline discussion about this?

faizan2590: I need to incorporate it indeed but was waiting for your review to do it in one go

PR #757, opened 2020-05-11
Feature/d3 asim 1849

hannesdiedrich:


src/d3a/models/area/stats.py
```diff
@@ -60,13 +60,21 @@ def update_area_market_stats(self):
...
-        return dict((trade.time.timestamp(), {
-            "energy": trade.offer.energy,
-            "seller": area_name_from_area_or_iaa_name(trade.seller),
-            "buyer": area_name_from_area_or_iaa_name(trade.buyer)})
-                    for trade in self.current_market.trades)
+        oudict = {}
+        for trade in self.current_market.trades:
+            trade_time = trade.time.timestamp()
+            if trade.time.timestamp() in oudict:
+                oudict[trade_time]["energy"].append(trade.offer.energy)
```
spyrostz, line 70: I understand that this works as it is, still I cannot shake the feeling that not grouping the energy, seller and buyer in each trade and only distinguishing them via the index is not that pretty. How about instead of using the index, to create a list for every case, which would contain dicts with the following format:
`[{energy: 0.1, seller: "a", buyer: "b"}, {energy: 0.1, seller: "a", buyer: "b"}]`
That way it will be probably easier to parse in the export.py module too.

hannesdiedrich: Thanks a lot! This is a good idea. I was thinking about a better way too as I didnt like the list idea either. Will change this.

PR #758, opened 2020-05-13
Feature/d3 asim 2349

hannesdiedrich:


src/d3a/models/area/__init__.py
```diff
@@ -115,6 +121,35 @@ def __init__(self, name: str = None, children: List["Area"] = None,
...
+
+        if key_in_dict_and_not_none(kwargs, 'import_capacity_kVA') or \
+                key_in_dict_and_not_none(kwargs, 'export_capacity_kVA'):
+            import_capacity_kVA = kwargs["import_capacity_kVA"] \
+                if key_in_dict_and_not_none(kwargs, 'import_capacity_kVA') else None
+            export_capacity_kVA = kwargs["export_capacity_kVA"] \
+                if key_in_dict_and_not_none(kwargs, 'export_capacity_kVA') else None
+            validate_area(import_capacity_kVA=import_capacity_kVA,
+                          export_capacity_kVA=export_capacity_kVA)
+            self._convert_area_throughput_kva_to_kwh(import_capacity_kVA, export_capacity_kVA)
```
spyrostz, line 151: The problem that I see with this parameter is that it is not saved to the market, but only in the Area object. This will have the effect that the event will update this parameter immediately and not from the next market slot. I am not sure if this is in the context of this story though, if it is not I would leave a comment to deal with it in the context of https://gridsingularity.atlassian.net/browse/D3ASIM-2332

hannesdiedrich: You are right. it will be updated immediately, which is not wanted.
I guess we only need to trigger this function at the beginning of the market slot, indeed in your mentioned story. Thanks!

src/d3a/models/strategy/load_hours.py
```diff
@@ -152,18 +152,42 @@ def _delete_past_state(self):
...
+        if key_in_dict_and_not_none(kwargs, 'fit_to_limit'):
+            self.bid_update.fit_to_limit = kwargs['fit_to_limit']
+        if key_in_dict_and_not_none(kwargs, 'update_interval'):
+            if isinstance(kwargs['update_interval'], int):
+                update_interval = duration(minutes=kwargs['update_interval'])
+            else:
+                update_interval = kwargs['update_interval']
+            self.bid_update.update_interval = update_interval
+        if key_in_dict_and_not_none(kwargs, 'use_market_maker_rate'):
+            self.use_market_maker_rate = ['use_market_maker_rate']
```
faizan2590, line 176:
```suggestion
            self.use_market_maker_rate = kwargs['use_market_maker_rate']
```

src/d3a/models/strategy/load_hours.py
```diff
@@ -152,18 +152,42 @@ def _delete_past_state(self):
...
+        self._validate_rates()
+
+    def area_reconfigure_event(self, **kwargs):
+        if key_in_dict_and_not_none(kwargs, 'hrs_per_day') or \
+                key_in_dict_and_not_none(kwargs, 'hrs_of_day'):
+            self.assign_hours_of_per_day(kwargs['hrs_of_day'], kwargs['hrs_per_day'])
             self.hrs_per_day = {day: self._initial_hrs_per_day
                                 for day in range(self.area.config.sim_duration.days + 1)}
+        if key_in_dict_and_not_none(kwargs, 'avg_power_W'):
+            self.avg_power_W = ['avg_power_W']
```
faizan2590, line 187: Same as [here](https://github.com/gridsingularity/d3a/pull/758/files#r424964819)

src/d3a/models/strategy/load_hours.py
hannesdiedrich: Nice Catch! Thanks!

hannesdiedrich:
> LGTM, once Spyros comments are addressed

@faizan2590 Please note the following comment here https://gridsingularity.atlassian.net/browse/D3ASIM-2332?focusedCommentId=18423
We plan to do it in the other story.
Will address your comments though

PR #761, opened 2020-05-15
Feature/d3 asim 2332

spyrostz:


src/d3a/d3a_core/simulation.py
```diff
@@ -297,6 +300,8 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
                 sleep(5)
                 break
 
+            self.live_events.handle_all_events(self.area)
```
faizan2590, line 303: Should the event only be applied only at the start of the slot, how about moving it to every tick?

spyrostz: The events should only be applied at the start of the market slot, in order to not mess up with the operation of the current market slot, and are only applied on the next market slot. Have a look at the story description too for more info https://gridsingularity.atlassian.net/browse/D3ASIM-2332.

PR #762, opened 2020-05-18
Feature/d3 asim 2331

spyrostz:


src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -66,15 +68,20 @@ def _subscribe_to_channels(self):
         self.pubsub.subscribe(**self._sub_callback_dict)
         self.pubsub.run_in_thread(sleep_time=0.5, daemon=True)
 
-    def _generate_redis_response(self, response, simulation_id, is_successful, command_type):
-        stop_response_channel = f'{simulation_id}/response/stop'
```
faizan2590, line 70: Thanks a lot for the fix!

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -66,15 +68,20 @@ def _subscribe_to_channels(self):
...
+    def _generate_redis_response(self, response, simulation_id, is_successful, command_type,
+                                 response_params=None):
+        if response_params is None:
+            response_params = {}
```
faizan2590, line 74: How about assigning this default as part of method argument?

spyrostz: Reason I did that is in order to circumvent the default value binding in Python methods, by using a sentinel value (None in this case). You can find a more detailed explanation here, http://effbot.org/zone/default-values.htm I highly recommend you to read this.

setup.py
```diff
@@ -12,7 +12,7 @@
         # TODO: Workaround for https://github.com/ethereum/py-solc/issues/64
         REQUIREMENTS.extend(
             ['d3a-interface @ '
-             f'git+https://github.com/gridsingularity/d3a-interface.git',
+             f'git+https://github.com/gridsingularity/d3a-interface.git@feature/D3ASIM-2331',
```
hannesdiedrich, line 15: please revert before pushing

src/d3a/d3a_core/simulation.py
```diff
@@ -71,6 +72,8 @@ def __init__(self):
         self.eta = duration(seconds=0)
         self.elapsed_time = duration(seconds=0)
         self.percentage_completed = 0
+        self.current_slot_str = ""
```
hannesdiedrich, line 75: This is not used ATM. Were do you plan to use it in the future?

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -126,6 +133,21 @@ def _slowdown_callback(self, message):
             return
         self._simulation.slowdown = slowdown
 
+    def _live_event_callback(self, message):
+        data = json.loads(message["data"])
+        try:
+            self._live_events.add_event(data)
```
hannesdiedrich, line 139: AFAIS, there is no `self._live_events` yet. This is the placeholder for the deserialisation of the events, right?

spyrostz: This PR is based on branch `feature/D3ASIM-2332` , where the live events are actually implemented in d3a and such an object exists.

src/d3a/d3a_core/simulation.py
spyrostz: I thought it would actually be useful in the future, in order to replace the 'current_market' timestamp that we currently have in the DB. However, you are right, let's introduce this when it is needed. Removing this.

PR #763, opened 2020-05-19
D3ASIM-1848: Plotly slider fix in case only value needs to be plotted.

faizan2590:


src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -152,6 +152,10 @@ def plot_slider_graph(cls, fig, stats_plot_dir, area_name, market_slot_data_mapp
             for k in range(market_slot_data_mapping[i].start,
                            market_slot_data_mapping[i].end):
                 step["args"][0]["visible"][k] = True  # Toggle i'th trace to "visible"
+            if market_slot_data_mapping[i].start == market_slot_data_mapping[i].end and \
+                    market_slot_data_mapping[i].start > 0 and \
+                    market_slot_data_mapping[i].end < len(fig.data):
```
hannesdiedrich, line 157: Sorry, I have trouble understanding the if clause
it is basically: `a==b and a>0 and b<x`
What are you trying to achieve?

faizan2590: if you have a look at the for loop above this code diff, it basically toggles the data point to be displayed per slider, the issue with that is in case when we have only one data point, 
lets say at the point where start=46 & end=46, in such a case code inside for loop wont be triggered being the same start and end value.

spyrostz, line 157: Maybe this will make it a bit more understandable:
```suggestion
            if market_slot_data_mapping[i].start == market_slot_data_mapping[i].end and \
                    0 < market_slot_data_mapping[i].start < len(fig.data):
```

spyrostz: To understand it a bit better, the problem that you want to tackle here is that the element is not rendered at all ? In this case, why not including the end in the for loop by doing this `for k in range(market_slot_data_mapping[i].start,  market_slot_data_mapping[i].end + 1):` ?

PR #765, opened 2020-05-20
Feature/d3 asim 2329

spyrostz:


src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -194,7 +194,7 @@ def event_activate_price(self):
 
     def _area_reconfigure_prices(self, final_buying_rate):
         if not self.connected:
-            super()._area_reconfigure_prices(final_buying_rate)
+            super()._area_reconfigure_prices(final_buying_rate=final_buying_rate)
```
hannesdiedrich, line 197: This is just cosmetics right?

spyrostz: Unfortunately no, the reason for that is that `_area_reconfigure_prices` uses only `**kwargs` as an argument, therefore when it is called the keyword argument needs to be provided, not a positional argument.

PR #769, opened 2020-05-25
D3ASIM-2395: Updating simulation result status=timeout in case of sim…

faizan2590:
…ulation pause timeout.


src/d3a/d3a_core/simulation.py
```diff
@@ -94,6 +94,7 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig =
         self.export_path = export_path
 
         self.sim_status = "initializing"
+        self.is_time_out = False
```
spyrostz, line 97: I think it would be better to name it `is_timed_out`

PR #772, opened 2020-05-27
D3ASIM-2355: Added some fixes for LiveEvents.

hannesdiedrich:


src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -72,10 +72,10 @@ def _generate_redis_response(self, response, simulation_id, is_successful, comma
                                  response_params=None):
         if response_params is None:
             response_params = {}
-        stop_response_channel = f'{simulation_id}/response/{command_type}'
```
spyrostz, line 75: Thank you!

src/d3a/models/appliance/inter_area.py
```diff
@@ -52,6 +55,8 @@ def event_tick(self):
         )
 
     def slot_energy(self, market):
+        if not hasattr(market, "trades"):
```
spyrostz, line 58: This was really needed at the end ?

hannesdiedrich: Ah no, didnt try it. I thought we decided to keep it anyway. Shall I remove it?

spyrostz: I am not sure if it is even worth your time. If you can quickly try it out then give it a go, otherwise leave it, I am fine with either.

hannesdiedrich: I just tried it again. We indeed need it.

spyrostz: Thanks for checking it out!

PR #779, opened 2020-06-04
D3ASIM-2406: Added additional sending of results when paused.

hannesdiedrich:


hannesdiedrich:
> It is ok for now, but I still think that either this while loop or the _handle_paused should be removed (since they essentially do the same thing). LGTM nevertheless, let's deal with it next time we refactor simulation.py

Sorry @spyrostz I have to dig deeper here: I dont see that they are doing the same. The while loop is actually handling the pause. `_handle_paused` is dealing with the functionality that allows the user to resume via the CLI. It could be renamed but IMO we still need both. 
Right?

spyrostz:
> > It is ok for now, but I still think that either this while loop or the _handle_paused should be removed (since they essentially do the same thing). LGTM nevertheless, let's deal with it next time we refactor simulation.py
> 
> Sorry @spyrostz I have to dig deeper here: I dont see that they are doing the same. The while loop is actually handling the pause. `_handle_paused` is dealing with the functionality that allows the user to resume via the CLI. It could be renamed but IMO we still need both.
> Right?

The main thing they do is exactly the same: they both wait when in paused state in order to resume when the corresponding resume event arrives. Whether the event is coming from the cli or from Redis should be irrelevant to this part of the code, what matters is that the `paused` member is set to False from somewhere. Therefore I disagree, we do not need 2 while loops here, one should suffice.

src/d3a/d3a_core/simulation.py
```diff
@@ -109,6 +109,7 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig =
         self.live_events = LiveEvents(self.simulation_config)
         self.redis_connection = RedisSimulationCommunication(self, redis_job_id, self.live_events)
         self._started_from_cli = redis_job_id is None
+        print(self._started_from_cli)
```
faizan2590: Reminder to remove before merging

src/d3a/d3a_core/simulation.py
```diff
@@ -444,20 +438,36 @@ def _handle_input(self, console, sleep: float = 0):
...
-            self.redis_connection.publish_intermediate_results(self.endpoint_buffer)
-            while self.paused:
+    def _handle_paused(self, console, tick_start):
+        if console is not None:
+            self._handle_input(console)
+            if self.pause_after and self.time_since_start >= self.pause_after:
+                self.paused = True
+                self.pause_after = None
+
+        started_pause = True
```
spyrostz, line 448: Why do you need the started_pause here? You can add the logic that needs to happen once outside of the while loop, by having an additional check `if self.paused: ` , like the former implementation did. That way you are able to trigger things that need to happen only once before the loop starts .

hannesdiedrich: True! Will change this.

hannesdiedrich: https://github.com/gridsingularity/d3a/pull/779/commits/371dfdb5f89f0997df4df3aca50e248411256316

src/d3a/d3a_core/simulation.py
```diff
@@ -444,20 +438,33 @@ def _handle_input(self, console, sleep: float = 0):
...
-            self.paused = True
-            self.pause_after = None
+    def _handle_paused(self, console, tick_start):
+        if console is not None:
+            self._handle_input(console)
+            if self.pause_after and self.time_since_start >= self.pause_after:
+                self.paused = True
+                self.pause_after = None
+
+        paused = False
```
spyrostz, line 448: Better name it to `paused_flag`, it looks very confusing to me

src/d3a/d3a_core/simulation.py
```diff
@@ -444,20 +438,33 @@ def _handle_input(self, console, sleep: float = 0):
...
         if self.paused:
+            if console:
+                log.critical("Simulation paused. Press 'p' to resume or resume from API.")
+            else:
+                self._update_and_send_results()
             start = time.monotonic()
```
spyrostz, line 454: Since we are here, how about removing the monotonic clock in favor of the normal clock (time.time) ? Thanks!

PR #782, opened 2020-06-08
Feature/d3 asim 2348

faizan2590:


src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -575,21 +531,36 @@ def _device_plot_layout(barmode, title, xaxis_caption, yaxis_caption, y2axis_cap
...
-                domain=[0.33, 0.66],
+                domain=[0.6, 0.8],
                 autorange=False
             ),
             yaxis3=dict(
-                title='Energy rate [€/kWh]',
-                domain=[0.0, 0.33],
+                title='Demand/Traded [kWh]',
+                side='left',
+                # range=y2axis_range,
```
spyrostz, line 549: Please remove before merging.

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -55,35 +55,51 @@ def _calc_min_max_from_sim_dict(subdict: Dict, key: str):
 
     @classmethod
     def _device_price_stats(cls, area: Area, subdict: Dict):
-        key_name = "trade_price_eur"
+        sold_key_name = "sold_trade_price_eur"
```
spyrostz, line 58: Are you certain this is not breaking the simulation results in the UI ? IIRC, this class is also used to generate results for the Load/PV in the UI, therefore it would be great if you could validate that the UI does not fail if you include these changes.

integration_tests/steps/jira_issues_tests.py
```diff
@@ -255,13 +255,16 @@ def device_statistics(context):
     counter = 0
     for house in ["House 1", "House 2"]:
         for device in output_dict[house]:
+            print(f"device: {device}")
```
spyrostz, line 258: Did not see this before, please make sure that you remove the prints before merging

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -40,7 +40,7 @@ def decompress_and_decode_queued_strings(queued_string):
 
 @job('d3a')
 def start(scenario, settings, events):
-    logging.getLogger().setLevel(logging.ERROR)
+    logging.getLogger().setLevel(logging.INFO)
```
spyrostz, line 43: Please revert to ERROR, because otherwise the d3a staging log will get spammed with info messages, and I am not sure if Kibana can support that many.

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -62,28 +61,45 @@ def _device_price_stats(cls, area: Area, subdict: Dict):
...
-        traded_energy = 0
-        for t in market.trades:
-            if t.seller == area.name:
-                traded_energy -= t.offer.energy
-            if t.buyer == area.name:
-                traded_energy += t.offer.energy
-
-        _create_or_append_dict(subdict, key_name,
-                               {market.time_slot: traded_energy})
+        from d3a.models.strategy.infinite_bus import InfiniteBusStrategy
```
spyrostz, line 75: It can be moved to the top of the module.

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -62,28 +61,45 @@ def _device_price_stats(cls, area: Area, subdict: Dict):
...
+        if type(area.strategy) == InfiniteBusStrategy:
```
spyrostz, line 76: I think it is better if you create 2 functions, one for the if block and one for the else block. Then it will be clear which function is for calculating the statistics of the Infinite Bus, and which function is for the rest of the devices.

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -62,28 +61,45 @@ def _device_price_stats(cls, area: Area, subdict: Dict):
...
+            sold_key_name = "sold_trade_energy_kWh"
+            bought_key_name = "bought_trade_energy_kWh"
+            sold_traded_energy = 0
+            bought_traded_energy = 0
+            for t in market.trades:
```
spyrostz, line 81: I have a general comment for the entirety of this class, but let's start it from here: can we not retrieve the sold, bought and traded energy directly from the strategy ? Reason I am asking is that this method iterates over all market trades, and is way more slower than just reading one value from the strategy class. There is a potential for a big performance improvement here IMO, and also for much easier refactoring of the code.

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -157,10 +173,4 @@ def _gather_device_statistics(cls, area: Area, subdict: Dict, flat_result_dict:
                                    {market.time_slot: area.strategy.energy_per_slot_kWh})
             cls._calc_min_max_from_sim_dict(subdict, "production_kWh")
 
-        elif isinstance(area.strategy, CommercialStrategy):
```
spyrostz, line 160: Do we not need this anymore ? I think that at least in the d3a plots, this block is still needed.

faizan2590: Judging from the comments of @sarahhambridge in ticket, where she mentioned `energy buffer is not needed`. So, I have omitted it

src/d3a/d3a_core/d3a_jobs.py
hannesdiedrich: Still toto

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -62,29 +62,54 @@ def _device_price_stats(cls, area: Area, subdict: Dict):
...
     def _device_energy_stats(cls, area: Area, subdict: Dict):
-        key_name = "trade_energy_kWh"
         market = list(area.parent.past_markets)[-1]
+        if type(area.strategy) == InfiniteBusStrategy:
+            cls.calculate_stats_for_infinite_bus(area, market, subdict)
+        else:
+            cls.calculate_stats_for_non_infinite_bus(area, market, subdict)
+
+    @classmethod
+    def calculate_stats_for_non_infinite_bus(cls, area, market, subdict):
```
spyrostz, line 82: Minor, but `calculate_stats_for_device` is a more appropriate name IMO

src/d3a/d3a_core/sim_results/device_statistics.py
faizan2590: Discussed offline, will be dealt later in this ticket: https://gridsingularity.atlassian.net/browse/D3ASIM-2459

PR #788, opened 2020-06-11
D3ASIM-2410: Publishing simulation pulses on a redis channel.

faizan2590:


src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -35,12 +37,21 @@
...
 ZIP_RESULTS_KEY = "d3a-zip-results-key/"
+HEARTBEAT_CHANNEL = "d3a-heartbeat"
+PULSE_RATE = 5  # in secs
 
 
 def utf8len(s):
     return len(s.encode('utf-8')) / 1000.0
 
 
+class RepeatingTimer(Timer):
```
spyrostz, line 48: I would suggest to move the RepeatTimer class from d3a-web to d3a-interface, and reuse it in both places. That way we will not have 2 clones of the same class across our repos

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -35,12 +37,21 @@
 RESULTS_CHANNEL = "d3a-results"
 ZIP_RESULTS_CHANNEL = "d3a-zip-results"
 ZIP_RESULTS_KEY = "d3a-zip-results-key/"
+HEARTBEAT_CHANNEL = "d3a-heartbeat"
```
spyrostz, line 40: This is not going to work with multiple simulations. The simulation_id needs to be part of the channel name

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -63,6 +74,8 @@ def __init__(self, simulation, simulation_id, live_events):
             log.error("Redis is not operational, will not use it for communication.")
             del self.pubsub
             return
+        self.heartbeat = RepeatingTimer(PULSE_RATE, self.heartbeat_tick)
```
spyrostz, line 77: I think it would be much better to name this `HEARTBEAT_INTERVAL`, it is more understandable to my eyes. Keep it that way if you disagree though

src/d3a/d3a_core/redis_connections/redis_communication.py
faizan2590: `simulation_id` is sent as part of published message.

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -35,12 +37,21 @@
 RESULTS_CHANNEL = "d3a-results"
 ZIP_RESULTS_CHANNEL = "d3a-zip-results"
 ZIP_RESULTS_KEY = "d3a-zip-results-key/"
+HEARTBEAT_CHANNEL = "d3a-heartbeat"
+PULSE_RATE = 5  # in secs
```
spyrostz, line 41: It would be even better if this was part of the constants.py. Or, if this value is used from the d3a-web too, in d3a-interface.

src/d3a/d3a_core/redis_connections/redis_communication.py
spyrostz: Understood. Then all is good, ignore my comment.

src/d3a/d3a_core/simulation.py
```diff
@@ -262,6 +262,8 @@ def _update_and_send_results(self, is_final=False):
             filename = self.export.export_to_zip_file()
             self.redis_connection.write_zip_results(filename)
             self.export.delete_exported_files()
+            if hasattr(self.redis_connection, 'heartbeat'):
+                self.redis_connection.heartbeat.cancel()
```
hannesdiedrich, line 266: So if the d3a is sending the results, the `RepeatingTimer` gets stopped?
`_update_and_send_results` is called on every market slot. Is this really what you wanted? I thought is should send a heartbeat tick every PULSE_RATE  seconds no matter what.

faizan2590: Have a look at L257 at this point of commit, it is when simulation simulation has finished and now `RepeatingTimer ` should stop too

spyrostz: This is actually the place where the simulation has actually finished and is preparing the zip files. Therefore I think it is a correct place to have it there, although this will probably be refactored shortly after we skip generating the archive with the csv files in the d3a.

PR #790, opened 2020-06-15
D3ASIM-2479: Omitted assert for parameters in the area_reconfigure_ev…

hannesdiedrich:
…ent.

### DIFF
--- a/src/d3a/models/strategy/load_hours.py
+++ b/src/d3a/models/strategy/load_hours.py
@@ -153,7 +153,6 @@ class LoadHoursStrategy(BidEnabledStrategy):
             del self.state.desired_energy_Wh[k]
 
     def _area_reconfigure_prices(self, **kwargs):
-        assert all(k in self.parameters for k in kwargs.keys())
         if key_in_dict_and_not_none(kwargs, 'initial_buying_rate'):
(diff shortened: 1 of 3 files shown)


src/d3a/models/strategy/load_hours.py
```diff
@@ -153,7 +153,6 @@ def _delete_past_state(self):
             del self.state.desired_energy_Wh[k]
 
     def _area_reconfigure_prices(self, **kwargs):
-        assert all(k in self.parameters for k in kwargs.keys())
```
spyrostz, line 156: Why is this needed here ? The parameters should have exactly the same names as the incoming arguments from the live events

hannesdiedrich: Unfortunately, the update_event sents parameters like `uuid` and `name`, too which are not part of the `parameters` tuple. I agree that this was a quick and dirty fix and actually wanted to trigger the discussion here.
Why do we really need this assert here? We check only for the needed in the if clauses later anyway.

spyrostz: You are right, better to not have this discussion here, and I understand the reason for removal. The assert is just a sanity check that validates that all the parameters passed are accepted by the strategy, but if parameters like uuid are passing through the update live event then feel free to remove this.

PR #791, opened 2020-06-15
Bug/d3 asim 2467

hannesdiedrich:


src/d3a/d3a_core/simulation.py
```diff
@@ -346,6 +347,8 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
             if self.export_on_finish and not self.redis_connection.is_enabled():
                 self.export.data_to_csv(self.area, True if slot_no == 0 else False)
 
+            self.live_events.handle_all_events(self.area)
```
spyrostz, line 350: Correct me if I am wrong, but I think that this is not correct. The live events will be applied after the current market slot has been completed. However, the current_market_slot_time is used as the time that the event will start taking effect. This is wrong, because the event will start taking effect after the current market slot has been completed.

hannesdiedrich: I tried it out and moving this line from the beginning of the market slot to the end didnt seem to make a difference. Maybe I didnt get into the case where it would make a difference. 
I will revert this though because you are right.

PR #792, opened 2020-06-15
D3ASIM-2476: Adding devices in bills results added later via event.

faizan2590:


src/d3a/d3a_core/sim_results/stats.py
```diff
@@ -241,6 +241,12 @@ def _get_child_data(self, area):
                         for child in area.children}
             return self.bills_results[area.name]
 
+    def _get_child_instance(self, name, area):
```
spyrostz, line 244: A better explanation here would be great to have, because I fail to see what this change actually fixes. The energy bills for the child areas should be accumulated by the `_accumulate_by_children` method IIRC

faizan2590: Child area being added later via event wasn't part bills area dict being populated [here](https://github.com/gridsingularity/d3a/pull/792/files#diff-672c71151977db649f4d45c99e348f73R240). As far as I ha e understood `_accumulate_by_children ` uses `flattened` as a reference which was being populated via `bills` where nested area's bills being added via event were not part of it

spyrostz: I understand that. Despite that though, the reason I am insisting on this is because the child areas should be taken into account throughout this calculation even for live events, because of [this line](https://github.com/gridsingularity/d3a/pull/792/files#diff-672c71151977db649f4d45c99e348f73R284) . Essentially this reads the children member of the area, which is updated whenever a child is added via a live event. Therefore areas created via live events will be calculated, and then taken into account in the flattened and accumulated versions of the bills. By calculating the bills specifically for the children in the loop here, I suspect that the results for the child areas are double-counted

PR #794, opened 2020-06-16
D3ASIM-2487: Added bulk-live-event handling.

hannesdiedrich:


src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -149,6 +151,22 @@ def _live_event_callback(self, message):
             {"activation_time": self._simulation.progress_info.current_slot_str}
         )
 
+    def _bulk_live_event_callback(self, message):
+        data = json.loads(message["data"])
+        try:
+            for event in data["bulk_event_list"]:
```
spyrostz, line 157: There is a small cornercase here, but I am not sure if we should deal with it. Let's assume that 5 events are registered, and the third one encounters an error when applied. This will be handled correctly as far as I recall, however the problem is that the first 2 events will be applied, while the remaining 3 will not. If we want to consider this as a single operation, then we should revert the first 2 events too. 
Having said that, it is not that critical. Feel free to skip it if you think it is an overkill

hannesdiedrich: You are right and i tried something: 
https://github.com/gridsingularity/d3a/pull/794/commits/9296825d4595d3d92fa6ede68f6712f60cbcdf70

PR #798, opened 2020-06-22
D3ASIM-2401: Added support for aggregator commands and events.

spyrostz:


src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -28,6 +28,7 @@
 from d3a.models.config import SimulationConfig
 from d3a.d3a_core.util import available_simulation_scenarios, update_advanced_settings
 from d3a.d3a_core.simulation import run_simulation
+from d3a.d3a_core.singletons import aggregator
```
hannesdiedrich, line 31: Do you need to import the singleton here? If yes, why?

src/d3a/d3a_core/redis_connections/aggregator_connection.py
```diff
@@ -0,0 +1,84 @@
...
+
+        self.batch_events[aggregator_uuid].append(event)
+
+    def receive_batch_commands_callback(self, payload):
+        batch_command_message = json.loads(payload["data"])
+        transaction_id = batch_command_message["transaction_id"]
+        with self.lock:
+            self.pending_batch_commands[transaction_id] = batch_command_message
+
+    def approve_batch_commands(self):
```
hannesdiedrich, line 43: I am having trouble understanding why this additional approval step is needed. In the live events we also "only" have a buffer that is consumed every market slot. Can you help me please? thanks!

spyrostz: The extra step is required here exactly for the same reason as the live events. Tick event takes time. Therefore we need to buffer any incoming commands that arrive during the tick to be processed at the next tick. So, at the start of each tick, we move these commands from the 'incoming' buffer to the 'processing' buffer. 
Only difference with the live events, is that this buffer is per-tick, not per-market cycle.

hannesdiedrich: OK, understood. What I meant is that in the liveEvents, we have inly one buffer. The `event_buffer` . It gets filled in the frame of one slot and the digested at the beginning of the next slot. Why do we need two buffers (incoming and processing) here?

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -166,17 +193,25 @@ def _reset_event_tick_counter(self):
         self._last_dispatched_tick = 0
 
     def _dispatch_event_tick_to_external_agent(self):
+        if not self.connected and not self.is_aggregator_controlled:
```
hannesdiedrich, line 196: This will disable the publishing of the tick event to "normal" external clients, right?

spyrostz: No actually, it will allow it. This clause early-returns if neither an external client nor an aggregator is connected, therefore if either of the 2 is connected, the tick event will be published.

src/d3a/d3a_core/redis_connections/aggregator_connection.py
spyrostz: Answered on Slack already, the difference here is that in our case the first buffer does not need to be consumed in one go, since it is consumed as long as the tick event propagates. This adds an extra danger of messages arriving while existing messages are being processed. We do not have this problem in live-events, because we consume all the events in one go, which is protected by a lock.

src/d3a/d3a_core/redis_connections/aggregator_connection.py
```diff
@@ -19,34 +22,49 @@ def set_aggregator_device_mapping(self, aggregator_device):
...
+        self._add_batch_event(device_uuid, event, self.batch_market_cycle_events)
+
+    def add_batch_tick_event(self, device_uuid, event):
+        self._add_batch_event(device_uuid, event, self.batch_tick_events)
+
+    def add_batch_trade_event(self, device_uuid, event):
+        self._add_batch_event(device_uuid, event, self.batch_trade_events)
+
+    def add_batch_finished_event(self, device_uuid, event):
+        self._add_batch_event(device_uuid, event, self.batch_finished_events)
```
hannesdiedrich, line 50: Is the distinction between the event types done for better debugging or does it have also a performance or stabilisation purpose?

hannesdiedrich: Dont get me wrong, I like it :-)

spyrostz: I wanted to avoid the caller code to pass dict key-value pairs actually. It was way more convenient to just call the method name rather than create a dict with e.g. "event": "market"

src/d3a/d3a_core/simulation.py
```diff
@@ -286,6 +288,7 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
         tick_lengths_s = config.tick_length.total_seconds()
         slot_count = int(config.sim_duration / config.slot_length)
 
+        self.simulation_config.external_redis_communicator.sub_to_aggregator(aggregator)
```
faizan2590, line 291: Can you point me from where is argument `aggregator ` coming from ?

spyrostz: https://github.com/gridsingularity/d3a/pull/798/files/82a65e57e35ecb71dd06e43c888007f2ee81346c#diff-fa25e2850120fece94bc9fbdfbd1266eR50

PR #802, opened 2020-06-23
Bug/d3 asim 2418

faizan2590:


integration_tests/steps/kpi.py
```diff
@@ -33,7 +33,7 @@ def test_export_of_kpi_result(context, kpi, expected_kpis):
     expected_kpis = ast.literal_eval(expected_kpis)
     for area, value in expected_kpis.items():
         if kpi == "self_sufficiency":
-            assert isclose(kpi_data[area]['self_sufficiency'], float(value), rel_tol=1e-04)
+            assert isclose(kpi_data[area]['self_sufficiency'], float(value), rel_tol=1e-01)
```
hannesdiedrich, line 36: I always get a headache seeing `rel_tol` reduced. Can you please explain why this was needed?

src/d3a/setup/two_sided_market/infinite_bus.py
```diff
@@ -18,38 +18,25 @@
...
-                                                      initial_selling_rate=24.1),
-                         appliance=PVAppliance()),
-                ], transfer_fee_const=1
-            ),
-            Area('Infinite Bus', strategy=InfiniteBusStrategy(energy_buy_rate=24,
-                                                              energy_sell_rate=24),
-                 appliance=SimpleAppliance()
-                 ),
-
+            Area('Market Maker', strategy=MarketMakerStrategy(energy_rate=30),
```
hannesdiedrich, line 34: So, this is not testing the infinite bus any more, so either revert the changes or rename the file. Also, please remove lines that are commented. Thanks

integration_tests/kpi.feature
```diff
@@ -35,27 +35,33 @@ Feature: KPI integration tests
      Then self_sufficiency of {'Grid': 1.0} are correctly reported
      And self_consumption of {'Grid': 1.0} are correctly reported
 
-  Scenario Outline: D3ASIM-2103, self sufficiency reports correct values on levels higher than houses
-     Given we have a scenario named <scenario_file>
+  Scenario: D3ASIM-2103, self sufficiency reports correct values on levels higher than houses for SSM
```
hannesdiedrich, line 38: SSM? What is it?

faizan2590: single sided market

integration_tests/steps/kpi.py
faizan2590: compared values were something like `0.83333333334`, so I had to reduce the precision level for expected and actual value comparision

src/d3a/setup/two_sided_market/infinite_bus.py
faizan2590: Thanks for pointing it out, have reverted to its original state.

integration_tests/steps/kpi.py
spyrostz: Yes, but this would succeed if this was indeed the issue, the number you pasted here is satisfying the -04 tolerance

faizan2590: Here are the actual values:
actual: `0.6666666666666666`
expected: `0.666`

hannesdiedrich: yeah, but wouldnt `expected = 0.666666` with `rel_tol=1e-04` work?

hannesdiedrich: There is also an `abs_tol` parameter in `isclose`. That might be more obvious when counting the number after the comma.

integration_tests/kpi.feature
hannesdiedrich: As I see the acronym in the code the first time i suggest to write "1-sided market" or similar instead of it.

faizan2590: There is always the first time for everything, so I think its okay to keep it

PR #809, opened 2020-06-30
Feature/d3 asim 2537: Enabling aggregator to handled/tested in an standalone scenario

faizan2590:


src/d3a/d3a_core/cli.py
```diff
@@ -98,11 +98,14 @@ def main(log_level):
 @click.option('--enable-bc', is_flag=True, default=False, help="Run simulation on Blockchain")
 @click.option('--compare-alt-pricing', is_flag=True, default=False,
               help="Compare alternative pricing schemes")
+@click.option('--enable-external-connection', is_flag=True, default=False,
+              help="External Agents interaction to simulation during runtime")
```
hannesdiedrich, line 102: having trouble understanding the need for this flag. We lived without it for now. I vote to omit it.

src/d3a/d3a_core/redis_connections/aggregator_connection.py
```diff
@@ -49,6 +52,91 @@ def add_batch_trade_event(self, device_uuid, event):
...
+                "device_uuid": message['device_uuid'],
+                "transaction_id": message['transaction_id']}
+            self.redis_db.publish(
+                "crud_aggregator_response", json.dumps(success_response_message)
+            )
+
+    def _create_aggregator(self, message):
+        if message['transaction_id'] not in self.selected_aggregators:
+            with self.lock:
+                self.selected_aggregators.append(message['transaction_id'])
```
hannesdiedrich, line 103: why do you need `selected_aggregators`? Isnt `aggregator_device_mapping.keys()` sufficient?

src/d3a/d3a_core/cli.py
faizan2590: I added it since it opens redis communication that runs in thread and slightly slowsdown the simulation which could be avoided for cases when external communication is not needed.

src/d3a/d3a_core/redis_connections/aggregator_connection.py
faizan2590: You are right, will remove it!

src/d3a/d3a_core/cli.py
hannesdiedrich: Yes, ok, understood. Although speed is not an issue locally. 
Leave it here though.

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -88,7 +82,8 @@ def start(scenario, settings, events, aggregator_device_mapping):
             "max_panel_power_W": settings.get('max_panel_power_W',
                                               ConstSettings.PVSettings.MAX_PANEL_OUTPUT_W),
             "grid_fee_type": settings.get('grid_fee_type', GlobalConfig.grid_fee_type),
-            "external_connection_enabled": True
+            "external_connection_enabled": True,
```
hannesdiedrich, line 85: We should think of setting the default value here to False and control this via the d3a-web for the reason you mentioned above in cli.py

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -88,7 +82,8 @@ def start(scenario, settings, events, aggregator_device_mapping):
...
+            "aggregator_device_mapping": aggregator_device_mapping
```
hannesdiedrich, line 86: This will not work, `aggregator_device_mapping ` is not defined, because you deleted line 65-68.

hannesdiedrich: sorry, forget it

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -423,6 +444,19 @@ def _offer_aggregator(self, arguments):
...
+        for bid in existing_bids:
+            assert bid.buyer == self.owner.name
+            if bid.id in self.market.bids.keys():
+                bid = self.market.bids[bid.id]
+            self.market.delete_bid(bid.id)
+
+            self.remove_bid_from_pending(self.market.id, bid.id)
+        self.post_bid(self.market, arguments["price"], arguments["energy"],
+                      buyer_origin=bid.buyer_origin)
+
```
hannesdiedrich, line 459: This is basically a new feature. Right? The "normal" device client does not have one of those. When was this decided?

faizan2590: This is after colin notice a bug where energy requirement was more than zero so client posted a bid/ offer. However since agrgegator is batch_processing that info in the same time old offer/bid got traded. SO, I propose him to actually add a new method called `update_offer/update` which would only check if the old offer/bid is not yet traded and if still persist then update the price

hannesdiedrich: I see, thanks for explaining. 
But again, this favours the aggregator before the single-device api. Me, as a single user must first list all bids, then delete the old one and then post a new one.  
IMO the case you are mentioning, is not a bug but something we just have to catch and report back to the aggregator. 
So, either we add it to the "normal" device too or we just say now: The aggregator as it is now will be thrown away anyway so need for that.

faizan2590: That case is not visible in the device since the device client request is instantly processed (bid/delete_bid/delete_offer/offer) and updates the `energy_requirement/available_energy` where as in case of aggregator case it is being cached and then batch processing is done every tick, so there could be a case where old bid is in cache in the mean time client posted the new one and in between the old one got accepted. However, still I would be in favour supporting device client with update_offer/update_bid as it looks more cleaner to me than client deleting the old bid and posting bid with updated price

src/d3a/d3a_core/d3a_jobs.py
faizan2590: Yes, I agree, will use `False` as default

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -42,7 +41,7 @@ def decompress_and_decode_queued_strings(queued_string):
 
 @job('d3a')
 def start(scenario, settings, events, aggregator_device_mapping):
-    logging.getLogger().setLevel(logging.ERROR)
+    logging.getLogger().setLevel(logging.INFO)
```
hannesdiedrich, line 44: please revert before merging

src/d3a/d3a_core/redis_connections/aggregator_connection.py
```diff
@@ -49,6 +51,87 @@ def add_batch_trade_event(self, device_uuid, event):
     def add_batch_finished_event(self, device_uuid, event):
         self._add_batch_event(device_uuid, event, self.batch_finished_events)
 
+    def crud_aggregator_callback(self, payload):
```
hannesdiedrich, line 54: Also here, please remove all "crud" occurrences.  create, read, update, and delete does not apply here, because the aggregator is not some storage. IMO it is confusing and should be renamed. Thanks

src/d3a/d3a_core/simulation.py
```diff
@@ -47,7 +47,6 @@
 from d3a.d3a_core.exceptions import D3AException
 from d3a.models.area.event_deserializer import deserialize_events_to_areas
 from d3a.d3a_core.live_events import LiveEvents
-from d3a.d3a_core.singletons import aggregator
```
hannesdiedrich, line 50: Please remove the de declaration of `aggregator` in the src/d3a/d3a_core/singletons.py (or remove the whole file)

hannesdiedrich: sorry, you did already.

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -396,33 +396,99 @@ def _list_offers_aggregator(self, arguments):
...
+            for bid in existing_bids:
+                assert bid.buyer == self.owner.name
+                if bid.id in self.market.bids.keys():
+                    bid = self.market.bids[bid.id]
+                self.market.delete_bid(bid.id)
+
+                self.remove_bid_from_pending(self.market.id, bid.id)
+            if len(existing_bids) > 0:
+                updated_bid = self.post_bid(self.market, bid_rate * existing_bid_energy,
+                                            existing_bid_energy, buyer_origin=self.device.name)
```
hannesdiedrich, line 479: Ok, i just realise that you also combining multiple old bids into one new one. I definitely not agree with this. But as Colin wants it like this. Leave it.

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -318,18 +329,32 @@ def trigger_aggregator_commands(self, command):
...
+            elif command["type"] == "list_market_stats":
+                market_data = self.device.parent.stats.get_market_stats(
+                    command['data']['market_slots']
+                )
+                return {
+                    "command": "list_market_stats", "status": "ready",
+                    "market_data": market_data,
+                    "transaction_id": command.get("transaction_id", None),
+                    "area_uuid": self.device.uuid
+                }
```
hannesdiedrich, line 357: Please refactor this into its own function, just like the other commands. Thanks!

src/d3a/models/strategy/external_strategies/storage.py
faizan2590: I agree with your reservations but for now his api-script is also placing one offer/bid at a time, so its okay for now.

PR #812, opened 2020-07-02
D3ASIM-2560: Re-initiated area._markets area.stats and area.dispatche…

hannesdiedrich:
…r when DeleteAreaEvent.

### DIFF
--- a/src/d3a/d3a_core/live_events.py
+++ b/src/d3a/d3a_core/live_events.py
@@ -4,7 +4,6 @@ import traceback
 from d3a.d3a_core.area_serializer import area_from_dict
 from d3a.d3a_core.exceptions import D3AException
 from d3a.models.area.event_dispatcher import DispatcherFactory
-from d3a.models.area.stats import AreaStats
 from d3a.models.area.markets import AreaMarkets
 
 
@@ -67,7 +66,6 @@ class DeleteAreaEvent:
         area.children = [c for c in area.children if c.uuid != self.area_uuid]
         if len(area.children) == 0:
             area._markets = AreaMarkets(area.log)
-            area.stats = AreaStats(area._markets)
             area.dispatcher = DispatcherFactory(area)()
         return True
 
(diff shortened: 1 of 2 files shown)


src/d3a/d3a_core/live_events.py
```diff
@@ -62,6 +65,10 @@ def apply(self, area):
             return False
 
         area.children = [c for c in area.children if c.uuid != self.area_uuid]
+        if len(area.children) == 0:
+            area._markets = AreaMarkets(area.log)
+            area.stats = AreaStats(area._markets)
```
faizan2590, line 70: As far as I see `stats` contain some critical infos historical market stats that are being loose here via initialising it

hannesdiedrich: @faizan2590 You are right, i removed the re-initiation of AreaStats here:
https://github.com/gridsingularity/d3a/pull/812/commits/70b9b1ffe53a035af250400c7a51ffb15b9272a7

src/d3a/d3a_core/live_events.py
```diff
@@ -62,6 +64,9 @@ def apply(self, area):
             return False
 
         area.children = [c for c in area.children if c.uuid != self.area_uuid]
+        if len(area.children) == 0:
+            area._markets = AreaMarkets(area.log)
```
faizan2590, line 68: Are you sure it wouldn't mess up with existing markets?
As far as I see `AreaMarkets ` contains current market too that are needed for ledgering the results

hannesdiedrich: You are right in general. However, in the UI case we dont keep the past markets, `AreaMarkets._market_rotation` will take care of deleting them anyway. This will "only" be a problem if you simulate with multiple future markets. But then again, this here, is only the case, when you delete the last child of an area. After deletion, there is no point of having markets anyway. So, IMO this reinitiating of the  `AreaMarkets` is unproblematic for this special case. Do you agree?

faizan2590: Thanks for the explanation. Still in case of multiple future markets and  only one of the child being removed would distort the result. 

Nevertheless, you may merge this PR after dropping a comment in code to think of a better solution in case of multiple future spot markets.

hannesdiedrich: Thanks! Will leave a comment here and merge.

PR #821, opened 2020-07-09
D3ASIM-2606: Renamed area parameter 'transfer_fee_const' into 'grid_f…

hannesdiedrich:
…ee_constant'.


src/d3a/models/area/__init__.py
```diff
@@ -59,10 +59,6 @@
 
 class Area:
 
-    reconfig_parameters = ('grid_fee_percentage', 'transfer_fee_const',
-                           'baseline_peak_energy_import_kWh', 'baseline_peak_energy_export_kWh',
-                           'import_capacity_kVA', 'export_capacity_kVA')
```
faizan2590, line 64: Why are you completely getting rid of it?

hannesdiedrich: It is not used any more. This is a remnant of my first implementation of liveEvents.

PR #824, opened 2020-07-13
D3ASIM-2599: Creating an aggregator_device mapping while selecting if…

faizan2590:
… not already exist.

### DIFF
--- a/src/d3a/d3a_core/redis_connections/aggregator_connection.py
+++ b/src/d3a/d3a_core/redis_connections/aggregator_connection.py
@@ -62,15 +62,16 @@ class AggregatorHandler:
 
     def _select_aggregator(self, message):
         if message['aggregator_uuid'] not in self.aggregator_device_mapping:
-            msg = f"{message['aggregator_uuid']} aggregator not found."
-            error_response_message = {
-                "status": "error", "aggregator_uuid": message['aggregator_uuid'],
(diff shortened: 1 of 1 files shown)


src/d3a/d3a_core/redis_connections/aggregator_connection.py
```diff
@@ -62,15 +62,16 @@ def aggregator_callback(self, payload):
 
     def _select_aggregator(self, message):
         if message['aggregator_uuid'] not in self.aggregator_device_mapping:
-            msg = f"{message['aggregator_uuid']} aggregator not found."
-            error_response_message = {
-                "status": "error", "aggregator_uuid": message['aggregator_uuid'],
+            self.aggregator_device_mapping[message['aggregator_uuid']] = []
```
hannesdiedrich, line 65: this will overwrite the list every time. So only one device can be connected to an aggregator.

faizan2590: This case would only exist when there was no aggregator in first place, so no way it would overwrite since the aggregator_device never existed with the targeted aggregator_uuid.

hannesdiedrich: True. Sorry. Understood.

PR #834, opened 2020-07-28
Slowdown Factor in Cli to flot

psurender:
D3ASIM2634- Slowdown Factor In Cli to float

### DIFF
--- a/src/d3a/d3a_core/cli.py
+++ b/src/d3a/d3a_core/cli.py
@@ -82,7 +82,8 @@ _setup_modules = available_simulation_scenarios
               help="Settings file path")
 @click.option('--slowdown', type=float, default=0,
               help="Slowdown factor [0 - 10,000]. "
-                   "Where 0 means: no slowdown, ticks are simulated as fast as possible;")
+                   "Where 0 means: no slowdown, ticks are simulated as fast as possible;"
+                   "and 100: ticks are simulated in realtime")
 @click.option('--seed', help="Manually specify random seed")
 @click.option('--paused', is_flag=True, default=False, show_default=True,
               help="Start simulation in paused state")
--- a/src/d3a/d3a_core/cli.py
+++ b/src/d3a/d3a_core/cli.py
@@ -80,9 +80,10 @@ _setup_modules = available_simulation_scenarios
                   ', '.join(_setup_modules)))
 @click.option('-g', '--settings-file', default=None,
               help="Settings file path")
-@click.option('--slowdown', type=int, default=0,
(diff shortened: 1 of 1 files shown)


src/d3a/d3a_core/cli.py
```diff
@@ -80,9 +80,10 @@ def main(log_level):
                   ', '.join(_setup_modules)))
 @click.option('-g', '--settings-file', default=None,
               help="Settings file path")
-@click.option('--slowdown', type=int, default=0,
+@click.option('--slowdown', type=float, default=0,
               help="Slowdown factor [0 - 10,000]. "
-                   "Where 0 means: no slowdown, ticks are simulated as fast as possible; "
+                   "Where 0 means: no slowdown, ticks are simulated as fast as possible;"
+                   "You can give float value also which makes simulation slower"
```
faizan2590, line 86: This can be skipped as the user would already know via `d3a run --help` that `slowdown` can now be configured as float value

psurender: Faizen you mean to skip which line ? in the above change?....user can enter int value and also float value.

psurender: Changed the comments to more accurate... "You can give float value also which makes simulation slower"

src/d3a/d3a_core/cli.py
```diff
@@ -80,9 +80,10 @@ def main(log_level):
                   ', '.join(_setup_modules)))
 @click.option('-g', '--settings-file', default=None,
               help="Settings file path")
-@click.option('--slowdown', type=int, default=0,
+@click.option('--slowdown', type=float, default=0,
               help="Slowdown factor [0 - 10,000]. "
-                   "Where 0 means: no slowdown, ticks are simulated as fast as possible; "
+                   "Where 0 means: no slowdown, ticks are simulated as fast as possible;"
+                   "You can give also float value to precisely control the simulation speed"
                    "and 100: ticks are simulated in realtime")
```
hannesdiedrich, line 87: IMO you dont need the added sentence, but if you want, please only add it to the end.
```suggestion
                   "Where 0 means: no slowdown, ticks are simulated as fast as possible;"
                   "and 100: ticks are simulated in realtime."
                   "You can give also float value to precisely control the simulation speed")
```

psurender: Changes are done

hannesdiedrich: @psurender you deleted one line too much. PLease re-add and 100: "ticks are simulated in realtime"

psurender: ok

PR #835, opened 2020-07-29
D3ASIM-2459: Used basestrategy trade list in for device statistic.

hannesdiedrich:


hannesdiedrich:
I made a e2e test for the device plot. It works.

PR #839, opened 2020-08-04
Feature/d3 asim 2212

hannesdiedrich:


src/d3a/d3a_core/util.py
```diff
@@ -455,6 +455,12 @@ def create_subdict_or_update(indict, key, subdict):
     return indict
 
 
+def _create_or_update_subdict(indict, key, subdict):
```
faizan2590, line 458:
```suggestion
def create_or_update_subdict(indict, key, subdict):
```
Minor but its no more a private method and seems to be better qualified for `d3a-interface`

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -67,7 +70,7 @@ def generate_result_report(self):
             "job_id": self.job_id,
```
faizan2590, line 70: Since we are already in result enhancement domain, good time to deal with `D3ASIM-2288` after this ticket to reduce some more memory issues via replacing pendulum timestamp to UNIX.

src/d3a/d3a_core/sim_results/bills.py
```diff
@@ -297,9 +262,6 @@ def _accumulate_market_fees(self, area, past_market_types):
             self._accumulate_market_fees(child, past_market_types)
 
     def _update_market_fees(self, area, market_type):
-        if ConstSettings.GeneralSettings.KEEP_PAST_MARKETS:
-            # If all the past markets remain in memory, reinitialize the market fees
-            self.market_fees = {}
         self._accumulate_market_fees(area, market_type)
 
     def update(self, area):
```
faizan2590, line 267: Good time to think of solving some more memory issues via trying to getting rid of duplicate results dict for backend and redis like [here](https://github.com/gridsingularity/d3a/pull/839/files?file-filters%5B%5D=.py#diff-dbad1988984f5b467d3f69255b611e42R273-R274) in L273-L274 at this point of commits.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -45,19 +45,22 @@ def __init__(self, job_id, initial_params, area, export_plots=True):
...
             "percentage_completed": 0
         }
+        self.should_export_plots = should_export_plots
+
         self.market_unmatched_loads = MarketUnmatchedLoads(area)
-        self.price_energy_day = MarketPriceEnergyDay()
+        self.price_energy_day = MarketPriceEnergyDay(should_export_plots)
         self.market_bills = MarketEnergyBills()
         self.cumulative_bills = CumulativeBills()
         self.balancing_bills = MarketEnergyBills(is_spot_market=False)
```
faizan2590, line 54: Try to skip initialising of results related to balancing markets when simulation is executed without balancing market.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -139,61 +121,36 @@ def update_stats(self, area, simulation_status, progress_info):
...
-                                                market_type)
-        self.accumulated_trades, self.current_trades = \
-            export_cumulative_grid_trades(area, self.accumulated_trades,
-                                          market_type, all_devices=True)
-        self.accumulated_balancing_trades, self.current_balancing_trades = \
-            export_cumulative_grid_trades(area, self.accumulated_balancing_trades,
-                                          balancing_market_type)
+            # TODO: (D3ASIM-2670) Trades are already stored in bids_offers_trades:
+            #  accumulation of area.stats.market_trades endpoint should be deprecated
+            self.last_energy_trades_high_resolution[area.uuid] = area.stats.market_trades
```
spyrostz, line 143: `area.stats.market_trades` object is storing a dict of market time slots -> trades that happened on this market slot. Therefore in this case, the last_energy_trades_high_resolution is always storing the trades for the whole simulation. This is a potential memory issue, because this can cause the memory to grow in size indefinitely. 
If I have understood something wrong in the PR please ping me, because there are probably other uses of market_trades that I do not see in the PR.

src/d3a/d3a_core/export.py
```diff
@@ -86,13 +88,35 @@ def __init__(self, root_area: Area, path: str, subdir: str, endpoint_buffer):
             _log.error("Could not open directory for csv exports: %s" % str(ex))
             return
 
-    def export_json_data(self, directory: dir, area: Area):
+    def generate_json_report(self):
```
spyrostz, line 91: I would very much prefer this method to stay in the endpoint_buffer class. Reason for that is that this method only depends on the endpoint_buffer, therefore you can replace all the `self.endpoint_buffer` to just `self`. In general, this is usually a very good indication that a method is a better fit to another class.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -139,61 +121,36 @@ def update_stats(self, area, simulation_status, progress_info):
         self.generate_result_report()
 
         self.bids_offers_trades.clear()
+
+        self.trade_profile.update(area)
```
spyrostz, line 125: Correct me if I am wrong, but this method is called no matter if the csv files / plots are exported or not. This method calculates both global and last slot data irregardless if global data are needed or not. If this is avoided, the corresponding memory leak will be resolved, and we will only keep global data when we need them (simulation is running locally and needs to generate plots)

hannesdiedrich: Actually not really. If you look into the function, you will see that it calls `_populate_area_children_data` and this will only calculates global slot data, if self.should_export_plots is False.

src/d3a/d3a_core/sim_results/bills.py
```diff
@@ -104,8 +72,7 @@ def update_cumulative_bills(self, area):
         for child in area.children:
             self.update_cumulative_bills(child)
 
-        if area.uuid not in self.cumulative_bills_results or \
-                ConstSettings.GeneralSettings.KEEP_PAST_MARKETS is True:
```
spyrostz, line 108: If it is 100% deprecated, please remove it from d3a-interface too, thanks

hannesdiedrich: @spyrostz I could not deprecate it. 100% unfortunately, because a lot of integrationtest depend on setting it to True.
I suggest to open a separate story for that. Do you agree?

spyrostz: Please do, thanks a lot, this is anyways outside the scope of this story.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
spyrostz: Awesome. Thanks a lot for explaining!

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: I agree! I started dealing with this, but realised, that it is more complex and created the D3ASIM-2670 story that deals with deprecating `area.stats.market_trades`.

src/d3a/d3a_core/sim_results/bills.py
hannesdiedrich: True! Let me try that.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: I am definitely a fan of D3ASIM-2288, but it needs to be discussed further, in the frame of the discussion of the simulation results service.

src/d3a/d3a_core/sim_results/bills.py
hannesdiedrich: https://gridsingularity.atlassian.net/browse/D3ASIM-2678

src/d3a/d3a_core/util.py
hannesdiedrich: Agreed, I will add it do d3a-interface, as I needed the same function in https://github.com/gridsingularity/d3a-web/pull/912 , too, where I will reuse it, too.

hannesdiedrich:
> You can safely remove `export_to_zip_file` and `delete_exported_files ` [here](https://github.com/gridsingularity/d3a/pull/839/files?file-filters%5B%5D=.py#diff-9e95bed0f62db1b1f967c65f0823d49dR129-R138)

True! Will do! Yeah!

src/d3a/d3a_core/sim_results/bills.py
hannesdiedrich: @faizan2590 A gave my best, but it is not possible without changing the whole (very complicated) structure of the bills creation.
The problem is, that the bills structure is so different for areas and devices. The code would bemore complicated than the actual saving of memory. Regarding the memory: Even for large grids, the bills dont take up a lot of space, so i suggest to go with the easy solution and to only populate the bills_redis_results if `should_export_plots is False`. Does it make sense?

hannesdiedrich: Ok, i thought about it some more. the cumulative_bills are created with uuid keys, too. 
To be consistent, I would exchange all "name" handling in the MarketBills into uuids and only exchange them into area names just before exporting to local json. OK?

hannesdiedrich: OK, I gave it another try, but it is too complicated. The problem is, that the root keys in the bills dict are uuids, but all subdictkeys are area names. This is not possible to handle with the current bills implementation. Will go with the way proposed in my last comment @faizan2590

hannesdiedrich: Actually, I just reverted my changes as CumulativeBills is depending on MarketEnergyBills.bills_redis_results.

src/d3a/d3a_core/util.py
```diff
@@ -455,6 +455,12 @@ def create_subdict_or_update(indict, key, subdict):
     return indict
 
 
+# def _create_or_update_subdict(indict, key, subdict):
+#     if key not in indict.keys():
+#         indict[key] = {}
+#     indict[key].update(subdict)
+
+
```
faizan2590, line 463: better delete these lines now

src/d3a/d3a_core/sim_results/bills.py
```diff
@@ -335,7 +298,8 @@ def _accumulate_by_children(self, area, flattened, results):
                 )
         return results
 
-    def _write_acculumated_stats(self, area, results, all_child_results, key_name):
+    @staticmethod
+    def _write_acculumated_stats(area, results, all_child_results, key_name):
```
spyrostz, line 302: Very slight typo: `accumulated`

src/d3a/d3a_core/simulation.py
```diff
@@ -24,6 +24,9 @@
 import dill
 import click
 import platform
+import os
```
spyrostz, line 27: Thanks a lot for that! Definitely not part of that story, but we can think about activating this https://www.python.org/dev/peps/pep-0008/#imports in flake, in order to support this:

Imports should be grouped in the following order:

    Standard library imports.
    Related third party imports.
    Local application/library specific imports.

No need to open a story yet, because my plan is to have a general discussion about code styling (essentially make our rules stricter), which will involve d3a-web too. We can discuss it then.

PR #840, opened 2020-08-04
Updating vagrantfile for d3a installation on windows.

faizan2590:


vagrant/Vagrantfile
```diff
@@ -2,9 +2,7 @@ Vagrant.configure("2") do |config|
   config.vm.box = "ubuntu/bionic64"
   config.vm.provider "virtualbox" do |v|
     v.memory = 4096
-    v.gui = true
```
hannesdiedrich, line 5: I guess that this disables the GUI for ubuntu, right?
It may be helpful though. E.g. to look at the plots. If you set up the VM with a shared directory, where you can open the plots also on the host OS (Windows), than its fine by me.

faizan2590: The reason that I had to get rid of it is due vm getting timeout since gui was not getting loaded in first place, so took the suggestion from [here](https://stackoverflow.com/questions/23293071/timed-out-while-waiting-for-the-machine-to-boot-when-vagrant-up)

hannesdiedrich: Wait, but in the forum link you posted, the guy ADDED the line to see that there was a BIOS issue, AFAIU.
Why are you disabling the GUI?

faizan2590: Sorry, yes you are right but still we were still not able to make it work. 
Nevertheless, closing this PR in favour of https://github.com/gridsingularity/d3a/pull/842 where folder sharing between guest and host machine was enabled and I don't think that anyone would really be needing the GUI, since all he/she needs vagrant for is to make him run simulation via his windows machine.

PR #841, opened 2020-08-05
Feature/d3 asim 1890

faizan2590:


src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -41,7 +41,7 @@ def decompress_and_decode_queued_strings(queued_string):
 
 @job('d3a')
 def start(scenario, settings, events, aggregator_device_mapping):
-    logging.getLogger().setLevel(logging.ERROR)
+    logging.getLogger().setLevel(logging.INFO)
```
spyrostz, line 44: Hmm, I am not sure that we want to print info logs by default in production (we probably don't want to)

src/d3a/d3a_core/launcher.py
```diff
@@ -56,13 +60,26 @@ def run(self):
...
+        redis_host = urlparse(REDIS_URL).netloc
+        info = None
+        while not info:
+            try:
+                info = StrictRedis(redis_host).info()
+                logging.debug("REDIS IS ALIVE")
+            except ConnectionError:
+                logging.error("Redis not yet available - sleeping")
+                time.sleep(5)
+
```
spyrostz, line 82: Since this method is used on both d3a-web and d3a, I think it better belongs to d3a-interface

src/d3a/d3a_core/d3a_jobs.py
faizan2590: Sorry, remnants of testing will set the log level back to ERROR

src/d3a/d3a_core/launcher.py
faizan2590: You are right, will do it.

setup.py
```diff
@@ -12,7 +12,7 @@
         # TODO: Workaround for https://github.com/ethereum/py-solc/issues/64
         REQUIREMENTS.extend(
             ['d3a-interface @ '
-             f'git+https://github.com/gridsingularity/d3a-interface.git',
+             f'git+https://github.com/gridsingularity/d3a-interface.git@feature/D3ASIM-1890',
```
hannesdiedrich, line 15: Please remove before merging

src/d3a/d3a_core/launcher.py
```diff
@@ -56,6 +55,7 @@ def run(self):
             self.job_array = [j for j in self.job_array if j.poll() is None]
 
     def is_crowded(self):
+        check_redis_health(redis_db=StrictRedis.from_url(REDIS_URL, retry_on_timeout=True))
```
hannesdiedrich, line 58: AFAIU, the call of the function does not have an effect (the return value "reconnect_needed" is not used. Do I miss something?

faizan2590: Since the same method is used in `d3a-web` too where it is needed to know if reconnect_needed or not whereas here it would simply work as wait for redis to be available again to receive incoming jobs and no resubscription is needed in `d3a` so I think  it is safe to simply ignore `reconnect_needed` here

hannesdiedrich: I see. ok.

PR #845, opened 2020-08-17
D3ASIM-2670: Recreating energy_trades_high_resolution from bids_offer…

faizan2590:
…s_trades.

### DIFF
--- a/src/d3a/d3a_core/export.py
+++ b/src/d3a/d3a_core/export.py
@@ -741,13 +741,13 @@ class ExportAndPlot:
         zero_point_dict = {"timestamp": [market_slot - GlobalConfig.tick_length],
                            "energy": [0.0]}
         # 1. accumulate data by buyer and seller:
-        for stock_time, stocks in market_trades.items():
-            for stock in stocks:
-                if stock['tag'] == "trade":
-                    trade_time = stock_time
-                    seller = stock["seller_origin"]
-                    buyer = stock["buyer_origin"]
-                    energy = stock["energy"]
+        for market_slot_time, market_slot_trades in market_trades.items():
+            for trade in market_slot_trades:
+                if trade['tag'] == "trade":
+                    trade_time = market_slot_time
+                    seller = trade["seller_origin"]
+                    buyer = trade["buyer_origin"]
+                    energy = trade["energy"]
                     if seller not in seller_dict:
                         seller_dict[seller] = deepcopy(zero_point_dict)
                     if buyer not in buyer_dict:
--- a/src/d3a/d3a_core/export.py
+++ b/src/d3a/d3a_core/export.py
@@ -26,14 +26,12 @@ import operator
 from slugify import slugify
 from sortedcontainers import SortedDict
 from collections import namedtuple
-from pendulum import from_timestamp
 from copy import deepcopy
-from d3a.constants import TIME_ZONE
 from d3a.models.market.market_structures import Trade, BalancingTrade, Bid, Offer, BalancingOffer
 from d3a.models.area import Area
 from d3a_interface.constants_limits import ConstSettings, GlobalConfig, DATE_TIME_FORMAT
 from d3a_interface.utils import mkdir_from_str
-from d3a.d3a_core.util import constsettings_to_dict, generate_market_slot_list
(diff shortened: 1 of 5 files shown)


src/d3a/d3a_core/sim_results/area_market_stock_stats.py
```diff
@@ -36,33 +35,28 @@ def update(self, area):
             self.state[area.name][last_past_market.time_slot] = {}
 
         for bid in last_past_market.bid_history:
-            tool_tip = f"{bid.buyer_origin} " \
```
hannesdiedrich, line 39: The changes you did to this file are not needed for this story, but to sync the information on the bids_offers_trades graph on the UI and the backend, right?

src/d3a/d3a_core/export.py
```diff
@@ -725,19 +741,21 @@ def add_plotly_graph_dataset(market_trades, market_slot):
...
-            if seller not in seller_dict:
-                seller_dict[seller] = deepcopy(zero_point_dict)
-            if buyer not in buyer_dict:
-                buyer_dict[buyer] = deepcopy(zero_point_dict)
-            seller_dict[seller]["timestamp"].append(trade_time)
-            seller_dict[seller]["energy"].append(energy * ENERGY_SELLER_SIGN_PLOTS)
-            buyer_dict[buyer]["timestamp"].append(trade_time)
-            buyer_dict[buyer]["energy"].append(energy * ENERGY_BUYER_SIGN_PLOTS)
+        for stock_time, stocks in market_trades.items():
+            for stock in stocks:
```
hannesdiedrich, line 745: I dont think stock is the right naming here. I suggest to rename it to market_slots and to loop over the trades. so:
```suggestion
        for market_slot_time, market_slot_trades in market_trades.items():
            for trade in market_slot_trades:
```

src/d3a/d3a_core/sim_results/area_market_stock_stats.py
hannesdiedrich: Also the changes in export.py Right?

faizan2590: It is needed as part of this story too as info needed for `high_resoulution_energy_trades` were part of tool_tip which is essentially an string as well as not consistent for offer/bids and trades, so quite hard and also not good approach to extract info from an string when such info could easily be saved in dict and tool_tips are only needed for backend plots and could be generated on fly while exporting backend plots.

hannesdiedrich: yep, makes sense, Thanks!

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -159,10 +159,8 @@ def update_area_aggregated_stats(self, area):
 
     def _update_bids_offer_trades(self, area):
         if area.current_market is not None:
-            self.bids_offers_trades[area.uuid] = area.current_market.get_bids_offers_trades()
-            # TODO: (D3ASIM-2670) Trades are already stored in bids_offers_trades:
-            #  accumulation of area.stats.market_trades endpoint should be deprecated
-            self.last_energy_trades_high_resolution[area.uuid] = area.stats.market_trades
+            self.bids_offers_trades[area.uuid] = \
+                {area.current_market.time_slot_str: area.current_market.get_bids_offers_trades()}
```
spyrostz, line 163: I agree in general with this implementation, but I am not sure you really need the time_slot_str of the current market here. The current market time is already transmitted in a separate field in the simulation results, therefore you can populate it from there. This would save you the effort in stripping away this time slot in the d3a-web PR (essentially instead of sending it and removing it, you could not send it at all here and add it via the results.current_market)

In addition to that, is this format absolutely required for the UI implementation of `energy_trades_high_resolution` ? If not I would prefer to not send it at all, since these are either way reported per requested market. cc'ing @hannesdiedrich  to pitch in

hannesdiedrich: I agree. AFAIS, you can omit calling `normalized_bid_offer_trade_data` on the d3a-web side then.

PR #852, opened 2020-08-20
D3ASIM-2678: Remoce KEEP_MARKET_STATS from code.

hannesdiedrich:


src/d3a/d3a_core/sim_results/export_unmatched_loads.py
```diff
@@ -54,16 +54,16 @@ def __init__(self, area):
...
         if hasattr(self.area, "past_markets") and len(list(self.area.past_markets)) > 0:
             self.latest_time_slot = list(self.area.past_markets)[-1].time_slot
         else:
-            if ConstSettings.GeneralSettings.KEEP_PAST_MARKETS is False:
+            if all_past_markets is False:
                 self.latest_time_slot = self.hour_list[0].add(
                     seconds=self.area.config.tick_length.seconds * self.area.current_tick)
             else:
-                self.latest_time_slot = self.hour_list[0]
+                self.latest_time_slot = self.hour_list[-1]
```
spyrostz, line 66: I do not think this is right. This needs to be set to the initial market slot (hour_list[0]), if we want to keep the past markets, in order to always force to recalculate the unmatched loads.

tox.ini
```diff
@@ -84,11 +84,11 @@ commands =
 	python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
     pip-sync requirements/base.txt requirements/tests.txt requirements/pandapower.txt
     pip install -e .
-    flake8
```
spyrostz, line 87: Don't forget to revert these changes before merging

spyrostz (dismissed):
One comment, LGTM otherwise

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
hannesdiedrich: This is actually for the case, then there is no `self.area.past_markets` yet (in the first market slot).
This is why it works with both values. But really, the latest_time_slot should be the -1th element of the  self. hour_list.
If you have a look here:
https://github.com/gridsingularity/d3a/blob/39a08adc41c4c93e5c81f0d8bf917c61338c895b/src/d3a/d3a_core/sim_results/export_unmatched_loads.py#L202
It is used to stop accumulating when the current market slot is reached. Speaking of that, I see a potential improvement for this handling. Let me add it and ping you then.

hannesdiedrich: @spyrostz I added the change. Tests are passing locally.
But you are right. Exactly because if there is no current market yet, the 0th element of hour_list should be chosen.

PR #854, opened 2020-08-24
D3ASIM-2421: Fixed tooltips in bids offers trade history local graph.…

spyrostz:
… Added only residual offers and bids to history tracking.


faizan2590 (approved):
LGTM

PR #855, opened 2020-08-26
Feature/d3 asim 2689

faizan2590:


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -121,10 +144,40 @@ def generate_json_report(self):
...
+        for child in area.children:
+            self._populate_core_stats(child)
+
     def update_stats(self, area, simulation_status, progress_info):
         self.status = simulation_status
         if area.current_market is not None:
             self.current_market = area.current_market.time_slot_str
+            self.current_market_unix = area.current_market.time_slot.timestamp()
+        self._populate_core_stats(area)
+        # print(f"flattened_area_core_stats_dict: {self.flattened_area_core_stats_dict}")
```
spyrostz, line 180: Make sure to remove it before it is merged.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -66,6 +70,24 @@ def __init__(self, job_id, initial_params, area, should_export_plots):
                 ConstSettings.GeneralSettings.EXPORT_ENERGY_TRADE_PROFILE_HR:
             self.area_market_stocks_stats = OfferBidTradeGraphStats()
 
+    @staticmethod
+    def _structure_results_from_area_object(target_area):
+        area_dict = dict()
+        area_dict['name'] = target_area.name
+        area_dict['uuid'] = target_area.uuid
+        area_dict['strategy'] = str(target_area.strategy.__class__.__name__)
+        area_dict['appliance'] = str(target_area.strategy.__class__.__name__)
```
spyrostz, line 79: There are a couple of points that would be nice to have here. First, appliance is not needed, and probably won't in the future, since we will get rid of the appliance mechanism, therefore I would recommend to remove it. Second, if the strategy is None (which is the case for area objects) the clause `target_area.strategy.__class__` will probably not fail, but will return the class type of the `None` object. Not that good for using it in the API which is transmitted across our services. What I would recommend to not use the term strategy, but `type` instead. That way, if strategy is None, you can assign a type `Area`.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -96,7 +118,8 @@ def generate_result_report(self):
             "last_energy_trades_high_resolution": convert_pendulum_to_str_in_dict(
                 self.last_energy_trades_high_resolution, {}),
             "bids_offers_trades": self.bids_offers_trades,
-            "results_area_uuids": list(self.result_area_uuids)
+            "results_area_uuids": list(self.result_area_uuids),
+            "flattened_area_core_stats_dict": self.flattened_area_core_stats_dict
```
spyrostz, line 122: Please do not send it now. At the moment we will continue using the old way.

spyrostz (commented):

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -66,6 +70,23 @@ def __init__(self, job_id, initial_params, area, should_export_plots):
                 ConstSettings.GeneralSettings.EXPORT_ENERGY_TRADE_PROFILE_HR:
             self.area_market_stocks_stats = OfferBidTradeGraphStats()
 
+    @staticmethod
+    def _structure_results_from_area_object(target_area):
+        area_dict = dict()
+        area_dict['name'] = target_area.name
+        area_dict['uuid'] = target_area.uuid
+        area_dict['type'] = str(target_area.strategy.__class__.__name__)
```
spyrostz, line 78:
```suggestion
        area_dict['type'] = str(target_area.strategy.__class__.__name__) if target_area.strategy is not None else "Area"
```
Maybe also save the "Area" string as a global constant, so that it can be reused from other places.

spyrostz (dismissed):
one comment, LGTM once resolved.

spyrostz (approved):
LGTM

PR #856, opened 2020-08-26
D3ASIM-2181: Disable the redis queue print -Show D3A Logging Thru Logger

psurender:
Please Review


src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -138,7 +138,7 @@ def main():
                                          retry_on_timeout=True)):
         Worker(
             ['d3a'],
-            name='simulation.{}.{:%s}'.format(getpid(), now())
+            name='simulation.{}.{:%s}'.format(getpid(), now()),log_job_description=False
```
faizan2590, line 141:
```suggestion
            name='simulation.{}.{:%s}'.format(getpid(), now()), log_job_description=False
```
Please follow flake8 in order of travis build to pass.

Also curious to know if you used git cli to commit or a git gui tool because git cli always pre-catch it while committing.

faizan2590 (commented):
Have left one comment, will LGTM once addressed.

src/d3a/d3a_core/d3a_jobs.py
psurender: while commiting its not identified this actually for d3a-web its identifying, ok will remove this space

spyrostz: Actually it is vice-versa: d3a should include `flake8` checks, while d3a-web does not. Having said that, what I usually do to avoid such problems is to run `tox -e travis` locally before I open a PR to make sure that my changes pass all required syntactic, unit and integration tests.

spyrostz (commented):
A general comment, this PR indeed disables the logs from the Redis queue. What I do not see here though is where are the corresponding logs from d3a printed. The idea behind this ticket AFAIK is to not print the scenario in compressed binary format (which is sent by the redis queue) but to print it once it has been uncompressed and in string format. If you can point me to the corresponding log then all is good and I will LGTM, otherwise such a log needs to be added.

src/d3a/d3a_core/d3a_jobs.py
psurender: Spyros ,Sure , I will do that

psurender:
> needs

Yes Spyros, me and Faizan was discussed on this and I have analysed on this, but I dont see option to convert certain set of logs  , I have used decompress option with format of utf-8, but still conversion is not happening, This will take some time to identify.

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -41,9 +41,13 @@ def decompress_and_decode_queued_strings(queued_string):
 
 @job('d3a')
 def start(scenario, settings, events, aggregator_device_mapping):
-    logging.getLogger().setLevel(logging.ERROR)
+    logging.getLogger().setLevel(logging.INFO)
```
spyrostz, line 44: Please revert this to error

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -41,9 +41,13 @@ def decompress_and_decode_queued_strings(queued_string):
...
 
     scenario = decompress_and_decode_queued_strings(scenario)
+    logging.getLogger().info(f"Scenario details are : {0}: {scenario}")
+    logging.getLogger().info(f"Settings details are : {0}: {settings}")
+    logging.getLogger().info(f"Events details are : {0}: {events}")
+    logging.getLogger().info(f"Aggregator device mapping details are : {0}: {aggregator_device_mapping}")
```
spyrostz, line 50: Change these logs to error accordingly

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -44,6 +44,10 @@ def start(scenario, settings, events, aggregator_device_mapping):
     logging.getLogger().setLevel(logging.ERROR)
 
     scenario = decompress_and_decode_queued_strings(scenario)
+    logging.getLogger().error(f"Scenario details are : {0}: {scenario}")
+    logging.getLogger().error(f"Settings details are : {0}: {settings}")
+    logging.getLogger().error(f"Events details are : {0}: {events}")
+    logging.getLogger().error(f"Aggregator device mapping details are : {0}: {aggregator_device_mapping}")
```
spyrostz, line 50: Travis is complaining about this line, that it is too long. I think that there is something wrong with the git precommit hooks in your local machine, otherwise you would not be able to commit this change.

As for the actual fix, if you split this line in 2 lines the problem will resolve.

spyrostz (commented):
LGTM once Travis passes

spyrostz (approved):
LGTM

faizan2590 (approved):
LGTM

spyrostz (approved):
LGTM

PR #858, opened 2020-09-01
D3ASIM-1826: Removed the restriction of the duplicate area names,as w…

psurender:
…e are using UUID

### DIFF
--- a/src/d3a/d3a_core/area_serializer.py
+++ b/src/d3a/d3a_core/area_serializer.py
@@ -160,3 +160,13 @@ def area_from_dict(description, config=None):
 def area_from_string(string, config=None):
     """Recover area from its json string representation"""
     return area_from_dict(json.loads(string), config)
+
+
+def are_all_areas_unique(area, list_of_areas=set()):
+    assert area.name not in list_of_areas
+    list_of_areas.add(area.name)
+
+    for child in area.children:
+        list_of_areas = are_all_areas_unique(child, list_of_areas)
+
+    return list_of_areas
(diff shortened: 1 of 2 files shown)


src/d3a/d3a_core/area_serializer.py
```diff
@@ -163,7 +163,6 @@ def area_from_string(string, config=None):
 
 
 def are_all_areas_unique(area, list_of_areas=set()):
```
spyrostz, line 165: I think it is much better to remove the call of this function rather than remove the assertion from inside it. That way the recursion that this function includes will be avoided.

psurender: Yes Spyros ,I have removed completely from both files, I thought function call will be utilised in other way to us in future to add areas like dummy in backend,

spyrostz (approved):
LGTM

src/d3a/d3a_core/area_serializer.py
```diff
@@ -160,13 +160,3 @@ def optional(attr):
 def area_from_string(string, config=None):
     """Recover area from its json string representation"""
     return area_from_dict(json.loads(string), config)
-
```
spyrostz, line 163: Better to keep this function and not to delete it. That way we might reuse it if we need it in the future.

psurender: Sure Spyros, I have kept the method as it is and now Integration tests are passing.

spyrostz (approved):
LGTM

faizan2590 (approved):
LGTM

PR #859, opened 2020-09-01
D3ASIM-2758: Do not prepopulate the area uuid keys, do it in the recu…

spyrostz:
…rsive method instead, in order to take into account incoming live events.


faizan2590 (approved):
Nice catch, didn't think about the case of device being added during live simulation.
LGTM!

PR #860, opened 2020-09-02
D3ASIM-2741: Simulation folders are changed to Area-UUID(Part Of D3ASIM-1826)

psurender:


psurender:
Need definition of the duplicates in the grid setup , research team has to come up with outline definition of duplicates at parent and children level

PR #862, opened 2020-09-03
D3ASIM-2703: Calculating device statistics using raw results data dict.

faizan2590:


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -187,7 +230,8 @@ def update_stats(self, area, simulation_status, progress_info):
 
         self.market_unmatched_loads.update_unmatched_loads(area)
 
-        self.device_statistics.update(area)
+        self.device_statistics.update(area, self.area_result_dict,
+                                      self.flattened_area_core_stats_dict)
```
hannesdiedrich, line 234: I thought the goal of this story is to get rid of parsing the area object.
Do you really need it here? IMO it should look like:
```suggestion
        self.device_statistics.update(self.area_result_dict,
                                      self.flattened_area_core_stats_dict)
```
Right?

faizan2590: Yes you are right, already got rid of using the area object inside the `update` method, will now also change this method signature too.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -158,8 +164,44 @@ def _populate_core_stats(self, area):
...
+                    for t in area.strategy.trades[area.parent.current_market]:
+                        core_stats_dict['trades'].append(t.serializable_dict())
+
+            elif type(area.strategy) == InfiniteBusStrategy:
+                core_stats_dict['production_kWh'] = area.strategy.energy_per_slot_kWh
+                if area.parent.current_market is not None:
+                    for t in area.strategy.trades[area.parent.current_market]:
+                        core_stats_dict['trades'].append(t.serializable_dict())
+
             self.flattened_area_core_stats_dict[area.uuid] = {}
```
hannesdiedrich, line 203: Can you tell me why you decided to move this into the endpoint buffer? AFAIR, this is device statistics code and i dont see it being useful for other statistics. 
Also, Is there a reason why you are writing the `production_kWh`, `load_profile_kWh` etc directly to the `self.flattened_area_core_stats_dict[area.uuid][self.current_market]` instead of having another subdict called "device_statitics" or so?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -145,8 +151,8 @@ def generate_json_report(self):
     def _populate_core_stats(self, area):
         if area.uuid not in self.flattened_area_core_stats_dict:
             self.flattened_area_core_stats_dict[area.uuid] = {}
-        if self.current_market_unix is not None and \
-                self.current_market_unix not in self.flattened_area_core_stats_dict[area.uuid]:
+        if self.current_market != "" and \
+                self.current_market not in self.flattened_area_core_stats_dict[area.uuid]:
```
hannesdiedrich, line 155: So you have decided to move from unix time to time string again. Any particular reason for that?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
faizan2590: Yes, that is the whole essence of `flattened_area_core_stats_dict ` to have all kinds of raw data being needed for results at one place, for example `load_profile_kWh ` is not just needed for `device_statistics` but also for `unmatched_loads`, so rather than having the same stats/data being duplicated at multiple places is to have in one common pleace

src/d3a/d3a_core/sim_results/endpoint_buffer.py
faizan2590: Yes, wanted to start with following unix_time_stamp but realised would be needing a lot of reconversion in between, so postponed and to be dealt carefully as part of `1739`

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: ok, makes sense. Thanks!

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: ok, I see now. The d3a work for `1739` is done btw. Have a look here:
https://github.com/gridsingularity/d3a/blob/02068b2ca1c125626e8765afd144c9883c825e64/src/d3a/d3a_core/sim_results/endpoint_buffer.py#L117
Where the `convert_pendulum_to_str_in_dict` is used. When we tackle https://gridsingularity.atlassian.net/browse/D3ASIM-2288, we will enable the `unix_time=True` and convert it in d3a-web.
So, yes, you are right, no need to use this unix_time here.

faizan2590: Yes, but my idea is to scratch the datetime format and string formatted datetime completely from d3a and Im sure it will make the simulation optimized at least in terms of memory usage

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -27,155 +20,193 @@ def _calc_min_max_from_sim_dict(subdict: Dict, key: str):
...
-            cls._soc_stats(area, subdict)
-
-        elif isinstance(area.strategy, LoadHoursStrategy):
-            cls._load_profile_stats(area, subdict)
-
-        elif type(area.strategy) == FinitePowerPlant:
-            market = list(area.parent.past_markets)[-1]
-            create_or_update_subdict(subdict, "production_kWh",
-                                     {market.time_slot: area.strategy.energy_per_slot_kWh})
+    def _gather_device_statistics(cls, area: Dict, subdict: Dict,
```
hannesdiedrich, line 184: AFAIS, You pnly need the area_uuid here and in all sub_functions. So you can only parse this.

faizan2590: Sorry, I didn't get it, I don't just need `area_uuid ` but also the area type here too, to distinguish between the type of data needs to be documented

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -150,7 +150,7 @@ def _load_profile_stats(cls, area: Dict, subdict: Dict, core_stats={},
 
         cls._calc_min_max_from_sim_dict(subdict, key_name)
 
-    def update(self, area, area_result_dict={}, core_stats={}, current_market_time_slot=None):
+    def update(self, area_result_dict={}, core_stats={}, current_market_time_slot=None):
```
hannesdiedrich, line 153: As mentioned offline, please rename all subfunctions too, Thanks

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -145,22 +152,56 @@ def generate_json_report(self):
...
+            if area.parent.current_market is not None:
+                for t in area.strategy.trades[area.parent.current_market]:
+                    core_stats_dict['trades'].append(t.serializable_dict())
+
+        elif type(area.strategy) in [InfiniteBusStrategy, MarketMakerStrategy]:
+            if area.parent.current_market is not None:
+                for t in area.strategy.trades[area.parent.current_market]:
+                    core_stats_dict['trades'].append(t.serializable_dict())
+
+        self.flattened_area_core_stats_dict[area.uuid] = {}
```
hannesdiedrich, line 203: You can omit this now, AFAIU.

src/d3a/d3a_core/sim_results/device_statistics.py
hannesdiedrich: yes, we discussed this offline. You are right, but `area` is not the right variable name IMO. It should be `area_result_dict` (or similar) here and in all sub functions.

hannesdiedrich: `area_dict` would be sufficient actually.

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -27,155 +20,188 @@ def _calc_min_max_from_sim_dict(subdict: Dict, key: str):
...
+            trade_stats_daily[time] += value
 
         for time_str, value in trade_stats_daily.items():
             min_trade_stats_daily[time_str] = limit_float_precision(min(value)) \
                 if len(value) > 0 else FILL_VALUE
             max_trade_stats_daily[time_str] = limit_float_precision(max(value)) \
                 if len(value) > 0 else FILL_VALUE
 
-        min_trade_stats = dict((time, min_trade_stats_daily[time.format(TIME_FORMAT)])
+        min_trade_stats = dict((time, min_trade_stats_daily[time])
```
hannesdiedrich, line 35: great! This will save some resources actually!

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -27,155 +20,193 @@ def _calc_min_max_from_sim_dict(subdict: Dict, key: str):
...
-    def _soc_stats(cls, area: Area, subdict: Dict):
+    def _soc_stats(cls, area: Dict, subdict: Dict, core_stats={}):
         key_name = "soc_history_%"
-        market = list(area.parent.past_markets)[-1]
-        create_or_update_subdict(subdict, key_name,
-                                 {market.time_slot:
-                                  area.strategy.state.charge_history[market.time_slot]})
+        if core_stats[area['uuid']] == {}:
+            return
+        # market = list(area.parent.past_markets)[-1]
```
hannesdiedrich: please remove

integration_tests/steps/integration_tests.py
```diff
@@ -1059,6 +1059,8 @@ def identical_cumulative_bills(context):
...
     device_stats_dict = context.simulation.endpoint_buffer.device_statistics.device_stats_dict
     load_profile = device_stats_dict['House 1']['H1 DefinedLoad']['load_profile_kWh']
-    for time_slot, value in load_profile.items():
-        if time_slot.add(days=1) < today(tz=TIME_ZONE).add(days=2):
-            assert value == load_profile[time_slot.add(days=1)]
+    load_profile_ts = {int(from_format(time_slot, DATE_TIME_FORMAT).timestamp()): value
+                       for time_slot, value in load_profile.items()}
+    start = int(list(load_profile_ts.keys())[0])
+    end = int(list(load_profile_ts.keys())[23])
+    assert all([load_profile_ts[i] == load_profile_ts[i + 86400] for i in range(start, end, 3600)])
```
hannesdiedrich, line 1066: This ps probably not needed any more, right?
It may be my opinion, but the old implementation is more human readable.

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -27,155 +20,188 @@ def _calc_min_max_from_sim_dict(subdict: Dict, key: str):
...
-                                 {market.time_slot: bought_traded_energy})
+        create_or_update_subdict(
+            subdict, bought_key_name,
+            {current_market_time_slot: bought_traded_energy})
         cls._calc_min_max_from_sim_dict(subdict, bought_key_name)
 
     @classmethod
-    def _pv_production_stats(cls, area: Area, subdict: Dict):
+    def _pv_production_stats(cls, area: Dict, subdict: Dict, core_stats={},
+                             current_market_time_slot=None):
```
hannesdiedrich, line 118: Very minor, but to keep the variable names descriptive but short, I would say that `current_market_slot` should be sufficient.

hannesdiedrich (commented):
Some more refactoring comments, in general LGTM

integration_tests/steps/integration_tests.py
faizan2590: In old implementation `time_slot ` was in and now it is in string, so now how had to parse it in any format anyway for integration test, so timestamp seems better to me with less hustle

src/d3a/d3a_core/sim_results/device_statistics.py
faizan2590: Ok sir!

integration_tests/steps/integration_tests.py
hannesdiedrich: Ok, this makes sense. Thanks!

hannesdiedrich (dismissed):
LGTM

integration_tests/steps/integration_tests.py
```diff
@@ -1059,6 +1059,8 @@ def identical_cumulative_bills(context):
 def identical_profiles(context):
     device_stats_dict = context.simulation.endpoint_buffer.device_statistics.device_stats_dict
     load_profile = device_stats_dict['House 1']['H1 DefinedLoad']['load_profile_kWh']
-    for time_slot, value in load_profile.items():
-        if time_slot.add(days=1) < today(tz=TIME_ZONE).add(days=2):
-            assert value == load_profile[time_slot.add(days=1)]
+    load_profile_ts = {int(from_format(time_slot, DATE_TIME_FORMAT).timestamp()): value
+                       for time_slot, value in load_profile.items()}
+    start = int(list(load_profile_ts.keys())[0])
+    end = int(list(load_profile_ts.keys())[23])
```
spyrostz, line 1065: Doesn't `-1` work (for the test to be more flexible than having exactly 24 hours)?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -145,22 +152,55 @@ def generate_json_report(self):
...
+        if hasattr(area.current_market, 'bid_history'):
+            for bid in area.current_market.bid_history:
+                core_stats_dict['bids'].append(bid.serializable_dict())
+        if hasattr(area.current_market, 'trades'):
+            for trade in area.current_market.trades:
+                core_stats_dict['trades'].append(trade.serializable_dict())
+
+        if isinstance(area.strategy, PVStrategy) or \
+                isinstance(area.strategy, PVUserProfileStrategy) or \
+                isinstance(area.strategy, PVPredefinedStrategy):
```
spyrostz, line 170: No need for these 2, they inherit from the PVStrategy therefore isinstance should detect them.

spyrostz (dismissed):
One comment, LGTM otherwise. Very interested to see any performance improvements that these changes might have.

integration_tests/steps/integration_tests.py
spyrostz: Discussed offline, please ignore my comment, thanks!

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #867, opened 2020-09-10
D3ASIM-2778: Only track energy origin in storage for one trade type d…

hannesdiedrich:
…epending on the market type.


faizan2590 (approved):
LGTM

PR #868, opened 2020-09-10
Feature/d3 asim 2812

faizan2590:


src/d3a/d3a_core/export.py
```diff
@@ -132,6 +133,22 @@ def export(self, export_plots=True, power_flow=None):
...
+        if not subdirectory.exists():
+            subdirectory.mkdir(exist_ok=True, parents=True)
+        json_file = os.path.join(self.directory, "area_tree_summary.json")
+        with open(json_file, 'w') as outfile:
+            json.dump(data, outfile, indent=2)
+
+    def raw_data_to_json(self, time_slot, data: Dict):
+        subdirectory = pathlib.Path(self.directory, "raw_data")
+        if not subdirectory.exists():
+            subdirectory.mkdir(exist_ok=True, parents=True)
```
hannesdiedrich, line 147: you could create this subdirectory before launching the simulation and also deleting it at the end (after exporting of course) Thanks!

src/d3a/d3a_core/export.py
```diff
@@ -132,6 +133,22 @@ def export(self, export_plots=True, power_flow=None):
     def data_to_csv(self, area, is_first):
         self._export_area_with_children(area, self.directory, is_first)
 
+    def area_tree_summary_to_json(self, data: Dict):
+        subdirectory = pathlib.Path(self.directory, "raw_data")
+        if not subdirectory.exists():
+            subdirectory.mkdir(exist_ok=True, parents=True)
+        json_file = os.path.join(self.directory, "area_tree_summary.json")
+        with open(json_file, 'w') as outfile:
+            json.dump(data, outfile, indent=2)
```
hannesdiedrich, line 142: I am still not sure whether JSON is the right format for buffering the results here. I thought about pickle but this will lead to issues at some point (from my experience and apparently also others: https://www.benfrederickson.com/dont-pickle-your-data/#:~:text=Just%20use%20JSON&text=Pickle%20on%20the%20other%20hand,data%20they%20can%20write%20out. ) 
Maybe @spyrostz  will have a better idea or will just say: JSON is good! Lets wait for that.

src/d3a/d3a_core/simulation.py
```diff
@@ -256,6 +256,11 @@ def _run_cli_execute_cycle(self, slot_resume, tick_resume):
 
     def _update_and_send_results(self, is_final=False):
         self.endpoint_buffer.update_stats(self.area, self.status, self.progress_info)
+        if self.export_on_finish and not self.redis_connection.is_enabled() and is_final:
```
hannesdiedrich, line 259:
```suggestion
        if self.export_on_finish and self.should_export_plots and is_final:
```

src/d3a/d3a_core/simulation.py
```diff
@@ -361,6 +366,11 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
             self._update_and_send_results()
             if self.export_on_finish and not self.redis_connection.is_enabled():
                 self.export.data_to_csv(self.area, True if slot_no == 0 else False)
+                if self.area.current_market is not None:
```
hannesdiedrich, line 369: I would actually put this check into `raw_data_to_json` do also have this check on every place. Thanks!

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
```diff
@@ -167,36 +156,29 @@ def _get_hover_info(cls, indict, slot_time):
...
         for child_name, child_ul_list in indict.items():
             for time in child_ul_list:
-                if time.hour == slot_time.hour and \
-                   time.day == slot_time.day and \
-                   time.month == slot_time.month and \
-                   time.year == slot_time.year:
+                if int(time[11:13]) == int(slot_time[11:13]) and \
+                   int(time[8:10]) == int(slot_time[8:10]) and \
+                   int(time[5:7]) == int(slot_time[5:7]) and \
+                   int(time[:4]) == int(slot_time[:4]):
```
hannesdiedrich, line 162: I think you could parse a pendulum object to `update_unmatched_loads` and then you can save this here.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -226,7 +226,9 @@ def update_stats(self, area, simulation_status, progress_info):
 
         self.cumulative_bills.update_cumulative_bills(area)
 
-        self.market_unmatched_loads.update_unmatched_loads(area)
+        self.market_unmatched_loads.update_unmatched_loads(
+            self.area_result_dict, self.flattened_area_core_stats_dict, self.current_market
```
hannesdiedrich, line 230: Here i like to mention, what i wanted to mention already in 2703: 
Lets stick to the naming convention we implicitly have and rename this to `self. current_market_time_slot_str`, because i would expect a market object here judging from the current naming.
so, i would rename like this:
        self.current_market --> self.current_market_time_slot_str
        self.current_market_unix --> self.current_market_time_slot_unix
        self.current_market_datetime --> self.current_market_time_slot

Would you also like this?

src/d3a/d3a_core/export.py
faizan2590: As far as I see `.exists()` already check if the subdirectory needs to be created or not, so why to replace with another check if it has to be created at start or not, considering the second point of deleting at the end, I would be skeptical to delete it at the end either as the integration tests uses it for testing the past markets stats after the simulation successfully finishes.

src/d3a/d3a_core/export.py
faizan2590: Will wait for Spyros feedback on that!

src/d3a/d3a_core/export.py
hannesdiedrich: Please think about performance. IMO we should not check every market slot whether the "raw_data" folder exists. We should just create it at the beginning and save 3 (or if you count area_tree_summary_to_json, too, 6) extra commands.
As for the deletion of the data: IMO there should be a global constant that (or you reuse D3A_TEST_RUN) that decides of keeping the rwa_data results. By default it should not keep the. Only for tests and debugging.

src/d3a/d3a_core/simulation.py
faizan2590: okay they both have same `not self.redis_connection.is_enabled()` and `self.should_export_plots`, if you prefer the later I would like to rename `should_export_plots ` to something generic as the target here is not to export plots but simply raw market stats.

hannesdiedrich: I am definitely in favour of renaming  `should_export_plots` ! I never liked this naming.
How about simply `export_results` ?

hannesdiedrich: Can you please rename it everywhere, also in endpoints_buffer please?

src/d3a/d3a_core/util.py
```diff
@@ -508,3 +508,10 @@ def child_buys_from_area(trade, area_name, child_names):
 def if_not_in_list_append(target_list, obj):
     if obj not in target_list:
         target_list.append(obj)
+
+
+def area_name_uuid_map(tree_dict, name_uuid_map={}):
+    name_uuid_map[tree_dict['name']] = tree_dict['uuid']
+    for child in tree_dict['children']:
+        area_name_uuid_map(child, name_uuid_map)
+    return name_uuid_map
```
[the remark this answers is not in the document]
faizan2590: I think you meant the one in `d3a-api-client` [here](https://github.com/gridsingularity/d3a-api-client/blob/master/d3a_api_client/utils.py#L136) . I definitely agree but this is something I would keep a note to my self and could easily take care of it when similar PRs are merged and settled in master

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
faizan2590: Yes, I could parse it to pendulum object but again I have to re-convert to string to save it in results string, so I would actually prefer to save some back and forth datetime conversion

src/d3a/d3a_core/util.py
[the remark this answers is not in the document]
hannesdiedrich: Ah, now, that was wishful thinking on my side. I meant the one on d3a-web. https://github.com/gridsingularity/d3a-web/blob/ebcbd0df8e11eccd61bc45146ee66a14a6eeec82/simulation/utils.py#L43
I know it is always a hassle to move stuff to d3a-interface, but i would be in favour of it. Thanks!

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
hannesdiedrich: I dont see your point the pendulum object for the current_time_slot is already there, you only have to parse it to `update_unmatched_loads`. There is only the `.format(DATE_TIME_FORMAT)` that has to be done here. But in your implementation, you need to transform 8 strings into integer which is not very effective either. And it is not very elegant IMO, too. Sorry to say.

src/d3a/d3a_core/simulation.py
faizan2590: Again think from the perspective of performance here too, this function is only used in two places, in this place it is always needed to check `if self.area.current_market` since it loops from the start of the simulation where current market is None and in other place it is guaranteed that there would be `current_market ` so no need to have this check in second place, so indeed I  can put this check inside `raw_data_to_json ` and have an early return if `current_market is not None` but wouldnt that degrades the *performance* that you first have to jump inside the function and then do the same check, if the method `raw_data_to_json ` would have been used in a lot of places then I would definitely be in favour of it but here im skeptical to do so

hannesdiedrich: I am just realising, that you call `self.export.raw_data_to_json` twice ? why is this?
Addotionally, you need to check for `if self.area.current_market` every time you run `raw_data_to_json` anyway because you want to access a current_market member in there. So I have to be relentless here, and ask you to put the check into the function. Thanks!

src/d3a/d3a_core/sim_results/endpoint_buffer.py
faizan2590: Okay, doing it now, just to get your memory recalled, you were skeptical to be so verbose back then (https://github.com/gridsingularity/d3a/pull/862#discussion_r484862864) but thats minor and will adapt this comment and be verbose

hannesdiedrich: @faizan2590 OMG, yes. Thanks for reminding me. But see above, i mean the suffix of the name. to be consistent with the market slot naming inside of the d3a code. If you like, please remove the "current" then. so :
self.current_market --> self.market_time_slot_str
self.current_market_unix --> self.market_time_slot_unix
self.current_market_datetime --> self.market_time_slot

src/d3a/d3a_core/sim_results/export_unmatched_loads.py
hannesdiedrich: Thanks for the offline explanation, you are right, in the future the results service will not get the pendulum object but a string from the d3a. Does make any sense to convert it to pendulum object. before just to have pretty code here.
However, i wonder if you could get rid of the `int()` as you comparing characters which each other and don't use `>` or `<`

faizan2590: `However, i wonder if you could get rid of the int() as you comparing characters which each other and don't use > or <`

Yup, just checked it, no need for an explicit conversion to `int` here, `str` conversion also works fine here

integration_tests/jira_issues_tests.feature
```diff
@@ -119,7 +118,7 @@ Feature: Jira Issues Tests
   Scenario: D3ASIM-1690: No unmatched load
     Given we have a scenario named two_sided_pay_as_clear/jira_d3asim_1690
     And d3a is installed
-    When we run the simulation with setup file two_sided_pay_as_clear.jira_d3asim_1690 and parameters [24, 60, 60, 1]
+    When we run the d3a simulation on console with two_sided_pay_as_clear.jira_d3asim_1690 for 24 hrs (60, 60)
```
spyrostz, line 121: I am a bit confused by the PR. Can you point me where is this step implementation function? Thanks

integration_tests/steps/integration_tests.py
```diff
@@ -678,21 +686,21 @@ def run_sim(context, scenario, total_duration, slot_length, tick_length, market_
...
+    assert len(context.raw_sim_data.keys()) == 24
     if scenario == "default":
-        street1 = list(filter(lambda x: x.name == "Street 1", context.simulation.area.children))[0]
-        house1 = list(filter(lambda x: x.name == "S1 House 1", street1.children))[0]
-        permanent_load = list(filter(lambda x: x.name == "S1 H1 Load", house1.children))[0]
-        energy_profile = [ki for ki in permanent_load.strategy.state.desired_energy_Wh.values()]
-        assert all([permanent_load.strategy.energy == ei for ei in energy_profile])
+        assert "Street 1" in context.name_uuid_map.keys()
+        assert "S1 House 1" in context.name_uuid_map.keys()
+        assert "S1 H1 Load" in context.name_uuid_map.keys()
```
spyrostz, line 701:
```suggestion
        assert {"Street 1", "S1 House 1", "S1 H1 Load"}.is_subset(set(context.name_uuid_map.keys()))
```
I would also give this a try

integration_tests/steps/load_tests.py
```diff
@@ -79,16 +79,19 @@ def check_user_pv_dict_profile(context):
...
+        unmatched_data, _ = unmatched.get_current_market_results(
+            context.area_tree_summary_data, core_stats, time_slot
+        )
+        count += get_number_of_unmatched_loads(unmatched_data)
     # There are two loads with the same final_buying_rate profile that should report unmatched
     # energy demand for the first 6 hours of the day:
-    assert get_number_of_unmatched_loads(unmatched) == int(number_of_loads * 6. * 60 /
-                                                           house.config.slot_length.minutes)
+    number_of_loads = 2
+    assert count == number_of_loads * 6 * 60 / 30
```
spyrostz, line 94: I assume that you cannot use the GlobalConfig object because d3a is run via the console, right ?

integration_tests/steps/integration_tests.py
```diff
@@ -1064,3 +1072,20 @@ def identical_profiles(context):
     start = int(list(load_profile_ts.keys())[0])
     end = int(list(load_profile_ts.keys())[23])
     assert all([load_profile_ts[i] == load_profile_ts[i + 86400] for i in range(start, end, 3600)])
+
+
+def get_simulation_raw_results(context):
+    area_tree_summary = glob.glob(os.path.join(context.export_path, "*", "area_tree_summary.json"))
+    print(f"area_tree_summary: {area_tree_summary}")
```
spyrostz, line 1079: Please don't forget to remove it before merging

integration_tests/steps/integration_tests.py
```diff
@@ -1064,3 +1072,20 @@ def identical_profiles(context):
...
+    with open(area_tree_summary[0], "r") as sf:
+        context.area_tree_summary_data = json.load(sf)
+    from d3a.d3a_core.util import area_name_uuid_map
```
spyrostz, line 1082: I assume that you have tested this, and it is required to be imported lazily. I am asking because it would be great if we could avoid lazy import here.

src/d3a/d3a_core/export.py
spyrostz: Better to avoid pickle altogether here actually. In addition to the points that the blog that Hannes added correctly mentions, I have also encountered another issue, which is that it does not work across different Python versions (Python 2 pickle cannot be read by Python 3 pickle, and this behavior exists even across 3.6 and 3.8 versions). 

TLDR: JSON is great for this purpose

spyrostz (approved):
Took a while, LGTM as soon as comments are addressed

integration_tests/steps/load_tests.py
faizan2590: Yes that was the reason but since now Im using the Simulation class to launch the simulation, reverting these two lines now

integration_tests/steps/integration_tests.py
faizan2590: no, just double checked, moving it to top

integration_tests/jira_issues_tests.feature
faizan2590: https://github.com/gridsingularity/d3a/blob/feature/D3ASIM-2812/integration_tests/steps/integration_tests.py#L286

src/d3a/d3a_core/simulation.py
faizan2590: Adapted as per offline discussion, `raw_data_to_json `  is now being called only once now and `self.area.current_market` stays outside of the function next to other checks if raw data needs to be exported to disk or not

src/d3a/d3a_core/export.py
faizan2590: Created raw data dir at the time of instantiating class `ExportAndPlot` as well as created an story to deal with the same pattern being found in old places like for CSV output: ExportAndPlot https://gridsingularity.atlassian.net/browse/D3ASIM-2834

src/d3a/d3a_core/export.py
```diff
@@ -143,8 +148,6 @@ def area_tree_summary_to_json(self, data: Dict):
 
     def raw_data_to_json(self, time_slot, data: Dict):
         subdirectory = pathlib.Path(self.directory, "raw_data")
```
hannesdiedrich, line 150: I suggest to add this as a self member and also reuse it in the area_tree_summary_to_json (where you can also omit the check of existence) Thanks!

integration_tests/steps/integration_tests.py
```diff
@@ -660,9 +661,10 @@ def run_sim(context, scenario, total_duration, slot_length, tick_length, market_
     paused = False
     pause_after = duration()
     repl = False
-    no_export = True
-    export_path = None
+    no_export = False
```
hannesdiedrich, line 664: Hmm, this will delay the integrationtests that dont need exported results (and adds up significantly to the runtime of all tests). I suggest to also parse this as an step argument.

src/d3a/d3a_core/export.py
```diff
@@ -82,6 +82,11 @@ def __init__(self, root_area: Area, path: str, subdir: str, file_stats_endpoint,
             self.directory = pathlib.Path(self.rootdir, subdir)
             self.zip_filename = pathlib.Path(self.rootdir, subdir + "_results")
             mkdir_from_str(str(self.directory))
+            import d3a.constants
```
hannesdiedrich, line 85: I think you can move this to the top and also do it like:
`from d3a.constants import D3A_TEST_RUN`

spyrostz, line 85: I think you can change it to a normal import

src/d3a/d3a_core/simulation.py
```diff
@@ -256,7 +255,9 @@ def _run_cli_execute_cycle(self, slot_resume, tick_resume):
 
     def _update_and_send_results(self, is_final=False):
         self.endpoint_buffer.update_stats(self.area, self.status, self.progress_info)
-        if self.export_on_finish and self.should_export_results and is_final:
+        import d3a.constants
+        if self.export_on_finish and self.should_export_results and \
+                self.area.current_market is not None and d3a.constants.D3A_TEST_RUN:
```
hannesdiedrich, line 260: Here, too: Change to `from d3a.constants import D3A_TEST_RUN` and move to top.

src/d3a/d3a_core/simulation.py
```diff
@@ -366,11 +367,11 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
-                if self.area.current_market is not None:
-                    self.export.raw_data_to_json(
-                        self.area.current_market.time_slot_str,
-                        self.endpoint_buffer.flattened_area_core_stats_dict
-                    )
+                # if self.area.current_market is not None:
+                #     self.export.raw_data_to_json(
+                #         self.area.current_market.time_slot_str,
+                #         self.endpoint_buffer.flattened_area_core_stats_dict
+                #     )
```
hannesdiedrich, line 374: Please remove

hannesdiedrich (approved):
Left some comments. After addressed, LGTM

integration_tests/steps/integration_tests.py
spyrostz: I would rather add it in the context actually, and use an extra 'given' step in order to set it. That way we will avoid to change lots of step definitions in the feature files.

src/d3a/d3a_core/export.py
spyrostz: He is actually obliged to import the module and not the variable, because if the variable is imported at the beginning of the file, the value of the variable will remain the same even if its value is modified in the `before_all` function. By importing the module reference we can access the current value of the variable after we change it.

spyrostz (approved):
Some comments, once resolved LGTM

src/d3a/d3a_core/simulation.py
faizan2590: https://github.com/gridsingularity/d3a/pull/868#discussion_r490020878

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #871, opened 2020-09-15
Bug/d3 asim 2825

spyrostz:


tests/test_strategy_pvpredefined.py
```diff
@@ -20,7 +20,9 @@
 import uuid
 import pathlib
 import os
-from pendulum import DateTime, duration, today
+from pendulum import DateTime, duration, today, instance
+from datetime import date, datetime
```
hannesdiedrich, line 24:
```suggestion
from pendulum import DateTime, duration, today, instance, date, datetime
```

hannesdiedrich (approved):
One minor comment, LGTM

tests/test_strategy_pvpredefined.py
spyrostz: I agree about date and I will move it now, however this will not work with the datetime import. Reason for that is that the datetime.combine is used in the test, and as far as I know this function does not exist in pendulum. We can replace this function of course, but I would very much use the same function that the d3a_interface uses (datetime.combine)

hannesdiedrich: ok, convinced, ignore my comment then.

faizan2590 (approved):
LGTM

PR #875, opened 2020-09-22
Feature/d3 asim 2705: Calculating cumulative grid trades from the raw offer/bid/trade data

faizan2590:


src/d3a/d3a_core/sim_results/cumulative_grid_trades.py
```diff
@@ -15,16 +15,40 @@
...
+from d3a.d3a_core.util import area_name_from_area_or_iaa_name, add_or_create_key, \
+    make_iaa_name_from_dict, subtract_or_create_key, round_floats_for_ui
 
 
-def export_cumulative_grid_trades(area, accumulated_trades_redis, past_market_types):
-    accumulated_trades = \
-        accumulate_grid_trades_all_devices(area, accumulated_trades_redis, past_market_types)
-    return accumulated_trades, generate_cumulative_grid_trades_for_all_areas(accumulated_trades,
-                                                                             area, {})
+def _is_cell_tower_type(area):
```
hannesdiedrich, line 23: you can now probably delete the corresponding `_is_cell_tower` method in the area_statistics.py, right?

hannesdiedrich: And of course for the other types too, if they are not needed somewhere else.

hannesdiedrich (approved):
One minor comment, LGTM

src/d3a/d3a_core/sim_results/cumulative_grid_trades.py
```diff
@@ -15,16 +15,40 @@
...
+    return area['type'] == "CellTowerLoadHoursStrategy"
+
+
+def _is_load_node_type(area):
```
spyrostz, line 27: I see that these functions have been renamed from `_is_load_node` to `is_load_node_type`, but the old functions `is_load_node` are still there. Therefore, either the old function definitions need to be deleted if they are not needed anymore in the area_statistics module, or these functions need to be moved to a separate module (maybe the __init__) in order to be reused by all the modules that might need them

src/d3a/d3a_core/sim_results/cumulative_grid_trades.py
```diff
@@ -34,12 +58,297 @@ def __init__(self):
...
+        return results
+
+    @classmethod
+    def _area_trade_from_parent(cls, area, parent, flattened_area_core_stats_dict,
+                                accumulated_trades):
+        area_IAA_name = make_iaa_name_from_dict(area)
+        parent_trades = flattened_area_core_stats_dict.get(parent['uuid'], {}).get('trades', [])
+
+        for trade in parent_trades:
+            trade_price = trade['energy'] * trade['energy_rate']
```
spyrostz, line 278: This is a strong point in favor of including the trade price in the trade JSON objects.

spyrostz (approved):
1 comment and 1 point for future discussion, LGTM otherwise.

src/d3a/d3a_core/sim_results/cumulative_grid_trades.py
faizan2590: Okay, will add now trade price in the trade JSON objects, the reason I was skeptical in first is to avoid sending redundant data via communication channels (in our case now it is redis) to save some data cache which could easily be calculated by result service using the already existing infos which is energy and rate here

src/d3a/d3a_core/sim_results/cumulative_grid_trades.py
faizan2590: Yes, moved all `*_type` method to __init__.py module, though keeping the old implementation `is_load_node ` as still being used in some places for instance for kpis and will completely deprecate once all the subtasks of this parent story has been completed

src/d3a/d3a_core/sim_results/cumulative_grid_trades.py
faizan2590: https://github.com/gridsingularity/d3a/pull/875#discussion_r493424610

src/d3a/d3a_core/sim_results/cumulative_grid_trades.py
hannesdiedrich: I agree with Faizan. Is it really worth to send and store a redundant information in the DB to avoid a multiplication operation?

spyrostz: It depends how frequently do you want to show it. I agree that 'just' for calculating cumulative grid trades it is not worth it. However, if we ever need to display it in a graph (for instance in the bid/offer/trade graph) or if we want to feed the trade price in more stats than that then it is definitely worth to precalculate (assuming that DB space is not the issue there) instead of having to have a separate step that would perform this multiplication. 
Having said that, since we only have this ATM I agree to not add the price yet. However I predict such a need will appear soon

spyrostz (approved):
LGTM

src/d3a/d3a_core/sim_results/cumulative_grid_trades.py
hannesdiedrich: ok no, if you put it like this, we can add the price indeed.

hannesdiedrich (approved):
LGTM

src/d3a/d3a_core/sim_results/cumulative_grid_trades.py
faizan2590: So, I will keep the pre-calculated `trade_price ` as part of trade JSON object and will see later if it makes sense to be pre-calculated or some DB space needs to be saved and calculated on the fly.

PR #877, opened 2020-09-24
Feature/d3 asim 2827

hannesdiedrich:


spyrostz (dismissed):
LGTM

src/d3a/models/area/__init__.py
```diff
@@ -158,6 +158,15 @@ def _set_grid_fees(self, transfer_fee_const, grid_fee_percentage):
...
 
+    def get_grid_fee(self):
+        grid_fee_type = self.config.grid_fee_type \
+            if self.config is not None \
+            else ConstSettings.IAASettings.GRID_FEE_TYPE
+        if grid_fee_type == 1:
+            return self.grid_fee_constant
+        elif grid_fee_type == 2:
+            return self.grid_fee_percentage
+
```
faizan2590, line 169:
```suggestion
       return self.grid_fee_constant if grid_fee_type == 1 else self.grid_fee_percentage
```

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -104,7 +101,10 @@ def event_market_cycle(self):
             return
         market_event_channel = f"{self.channel_prefix}/market-events/market"
         current_market_info = self.area.current_market.info
-        current_market_info['last_market_stats'] = \
+        current_market_info["current_market_fee"] = \
+            self.area.current_market.fee_class.grid_fee_rate
+        current_market_info["next_market_fee"] = str(self.area.get_grid_fee())
```
faizan2590, line 106: For the next market, you are checking the grid fee type and then returning it whereas in `current_market_fee ` you are only returning the constant grid fee. Better to have a uniform implementation for both `current_market_fee  & next_market_fee ` .

faizan2590 (commented):
Had to leave a comment for clarification, would give another look once addressed. Thanks!

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: @faizan2590 not really, for the `current_market_fee` i return the grid_fee that is set, independent of the type. Have a look in the current_market.fee_class.

faizan2590: Ah yes got it, again why didnt you followed the same approach for getting the `next_market_fee ` too ?

faizan2590 (approved):
LGTM

src/d3a/models/area/redis_external_market_connection.py
faizan2590: No need for the further change: https://gridsingularity.atlassian.net/browse/D3ASIM-2827?focusedCommentId=20174

PR #878, opened 2020-09-24
Feature/d3 asim 2768

faizan2590:


src/d3a/d3a_core/sim_results/market_price_energy_day.py
```diff
@@ -0,0 +1,111 @@
...
+        price_lists[area_dict['uuid']][current_market_slot].extend(trade_rates)
+
+        if area_dict['uuid'] not in fee_lists:
+            fee_lists[area_dict['uuid']] = OrderedDict()
+        if current_market_slot not in fee_lists[area_dict['uuid']].keys():
+            fee_lists[area_dict['uuid']][current_market_slot] = []
+        fee_rates = [
+            # Convert from cents to euro
+            t['fee_price'] / 100.0 / t['energy']
+            for t in core_stats.get(area_dict['uuid'], {}).get('trades', [])
```
hannesdiedrich, line 50: A potential improvement with regard of computational effort would be to use one for loop only to populate `fee_rates` and `trade_rates`. What do you think?

src/d3a/d3a_core/sim_results/market_price_energy_day.py
```diff
@@ -0,0 +1,111 @@
...
+            ]
+        for node_name, fee_rates in fees.items():
+            redis_output[node_name]["grid-fee-rate"] = [
+                {
+                    "time": timeslot,
+                    "av_price": round_floats_for_ui(mean(fee_rate) if len(fee_rate) > 0 else 0),
+                    "min_price": round_floats_for_ui(min(fee_rate) if len(fee_rate) > 0 else 0),
+                    "max_price": round_floats_for_ui(max(fee_rate) if len(fee_rate) > 0 else 0),
+                } for timeslot, fee_rate in fee_rates.items()
+            ]
```
hannesdiedrich, line 111: I dont know whether this is really a valid concern, but what if the `price_energy` and `fees` are not in sync. How about adding the fees also to the price_energy dict and avoid that?

hannesdiedrich (commented):
In general it LGTM, but i left some comments.

src/d3a/d3a_core/sim_results/market_price_energy_day.py
```diff
@@ -0,0 +1,111 @@
...
+                }
+            redis_output[node_name]["price-energy-day"] = [
+                {
+                    "time": timeslot,
+                    "av_price": round_floats_for_ui(mean(trades) if len(trades) > 0 else 0),
+                    "min_price": round_floats_for_ui(min(trades) if len(trades) > 0 else 0),
+                    "max_price": round_floats_for_ui(max(trades) if len(trades) > 0 else 0),
+                } for timeslot, trades in trade_rates.items()
+            ]
+        for node_name, fee_rates in fees.items():
```
spyrostz, line 103: Judging from the ticket description, grid fees has to be one line, therefore there has to be exactly one value per market slot. The min/max/avg values of the fees for one market slot are not needed according to the ticket, only the sum of the fees for one market slot. All these assuming that there was no offline discussion that is not reflected in the ticket

src/d3a/d3a_core/sim_results/market_price_energy_day.py
```diff
@@ -0,0 +1,111 @@
...
+                    "load-unit": "kWh",
+                    "price-energy-day": [],
+                    "grid-fee-rate": []
+                }
+            redis_output[node_name]["price-energy-day"] = [
+                {
+                    "time": timeslot,
+                    "av_price": round_floats_for_ui(mean(trades) if len(trades) > 0 else 0),
+                    "min_price": round_floats_for_ui(min(trades) if len(trades) > 0 else 0),
+                    "max_price": round_floats_for_ui(max(trades) if len(trades) > 0 else 0),
```
spyrostz, line 100: Would be far easier and more elegaant for the UI to add the grid fees as an additional value here.

spyrostz (commented):
There is one requirement missing from the ticket, which is to not report grid fees in the case of percentage grid fees. The PR needs to handle this case too

src/d3a/d3a_core/sim_results/market_price_energy_day.py
faizan2590: I confirmed from Sarah  and its neither  `min/max/avg values of the fees for one market slot` nor the `sum of the fees for one market slot` but simply `the  grid fee set in the market`

src/d3a/d3a_core/sim_results/market_price_energy_day.py
faizan2590: Outdated, as fee being set at the market only needs to be plotted.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -160,6 +161,9 @@ def _populate_core_stats(self, area):
         if hasattr(area.current_market, 'trades'):
             for trade in area.current_market.trades:
                 core_stats_dict['trades'].append(trade.serializable_dict())
+        if getattr(area, 'strategy', None) is None and area.current_market is not None:
+            core_stats_dict['grid_fee_constant'] = area.current_market.fee_class.grid_fee_rate \
+                if isinstance(area.current_market.fee_class, ConstantGridFees) else None
```
spyrostz, line 166: This works but looks very counter-intuitive to me to use introspection in order to find out the area fee type. I would rather expose an extra property method in the Area class (have a look at Area._set_grid_fees, you can either add an extra argument in order to save the fee type, or deduce the fee type by None-checking self.grid_fee_constant and self.grid_fee_percentage)

src/d3a/d3a_core/sim_results/market_price_energy_day.py
```diff
@@ -72,40 +59,33 @@ def calculate_csv_output(self, area_dict, price_energy_redis_output):
...
-                    "min_price": round_floats_for_ui(min(fee_rate) if len(fee_rate) > 0 else 0),
-                    "max_price": round_floats_for_ui(max(fee_rate) if len(fee_rate) > 0 else 0),
-                } for timeslot, fee_rate in fee_rates.items()
-            ]
+
+            area_core_stats = core_stats.get(node_uuid, {})
+            fee = area_core_stats['grid_fee_constant'] / 100 \
+                if key_in_dict_and_not_none(area_core_stats, 'grid_fee_constant') else None
+
+            redis_output[node_uuid]["price-energy-day"][0].update({"fee_price": fee})
```
spyrostz, line 91: Just for my understanding, since we always calculate one market slot at a time, I guess that there is always going to be one element in this list, right ?

spyrostz (dismissed):
One last cosmetic comment, LGTM once resolved.

hannesdiedrich (approved):
LGTM once Spyros' comments area addressed.

src/d3a/d3a_core/sim_results/market_price_energy_day.py
faizan2590: Yes!

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #879, opened 2020-09-25
Feature/d3 asim 2765

hannesdiedrich:


src/d3a/models/area/stats.py
```diff
@@ -50,16 +50,26 @@ def __init__(self, area_markets):
     def update_aggregated_stats(self, area_stats):
         self.aggregated_stats = area_stats
 
+    def _extract_from_bills(self, trade_key):
+        if self.current_market is None:
+            return None
+        return {key: self.aggregated_stats["bills"][trade_key][key]
+                for key in ["earned", "spent", "bought", "sold"]} \
+            if "bills" in self.aggregated_stats \
+               and trade_key in self.aggregated_stats["bills"] else None
```
faizan2590, line 59:
```suggestion
               and trade_key in self.aggregated_stats["bills"] else {}
```
looks better to return an empty dictionary than `None` to have a uniform data type pattern  irrespective of if value is there or not.

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -175,18 +177,23 @@ def event_market_cycle(self):
...
+            if self.is_aggregator_controlled:
+                current_market_info.update(default_market_info)
             current_market_info['device_info'] = self._device_info_dict
             current_market_info["event"] = "market"
             current_market_info["area_uuid"] = self.device.uuid
             current_market_info['device_bill'] = self.device.stats.aggregated_stats["bills"] \
                 if "bills" in self.device.stats.aggregated_stats else None
-            current_market_info['last_market_stats'] = \
-                self.market_area.stats.get_price_stats_current_market()
+            current_market_info["last_market_maker_rate"] = \
```
faizan2590, line 187:
```suggestion
            current_market_info["current_market_maker_rate"] = \
```
in order to have uniform name as the method being called here

src/d3a/d3a_core/global_objects.py
```diff
@@ -0,0 +1,39 @@
...
+            self._create_grid_tree_dict(area, self.area_tree_dict)
+
+    def _create_grid_tree_dict(self, area, outdict):
+        outdict[area.name] = {}
+        if area.children:
+            outdict[area.name].update(
+                {"last_market_slot": area.current_market.time_slot_str,
+                 "last_market_bill": area.stats.get_current_market_stats_for_grid_tree(),
+                 "last_market_stats": area.stats.get_price_stats_current_market(),
+                 "last_market_fee": str(area.current_market.fee_class.grid_fee_rate),
```
faizan2590, line 36: Would be better to prefix all with `current_` rather than `last_`.
Thanks!

hannesdiedrich: @faizan2590 This was a specific request from Colin and I agree. From a clients perspective, it is really the last market slot. If you look into the implementation of the current market, it is really the last past market (`past_markets[-1]`)

faizan2590: Okay, Im indifferent to naming being followed here, but would press on having a uniform naming convention, at least for future to not being confused like `get_current_market_stats_for_grid_tree ` to be renamed to `get_last_market_stats_for_grid_tree ` and `get_current_market_maker_rate `  to `get_last_market_maker_rate ` .

hannesdiedrich: I would agree in general, but in the code, we use `current_market` as a name for the last_past_market. So either we rename it in the whole code (which is a huge refactoring story) or we keep it as is. I would go with the latter.
Maybe @spyrostz has a better idea?

faizan2590: Okay, but you could at least rename `get_current_market_stats_for_grid_tree `, the one you added in this PR.

hannesdiedrich: :-) Sorry Faizan, but to be consistent, I dont agree. The `get_current_market_stats_for_grid_tree` is indeed using the current_market.

faizan2590: > :-) Sorry Faizan, but to be consistent, I dont agree. The get_current_market_stats_for_grid_tree is indeed using the current_market.

You are adding the value of `current_market ` being returned via `get_current_market_stats_for_grid_tree` to `last_market_bill `.
Since this is just added in this PR, you can simply rename `get_current_market_stats_for_grid_tree` to `get_last_market_stats_for_grid_tree`

hannesdiedrich: OK, i give up, really not worth this long discussion here. Renaming it.

src/d3a/d3a_core/redis_connections/aggregator_connection.py
```diff
@@ -39,7 +48,15 @@ def _add_batch_event(self, device_uuid, event, batch_event_dict):
 
         batch_event_dict[aggregator_uuid].append(event)
 
-    def add_batch_market_event(self, device_uuid, event):
+    def _add_grid_stats_to_market_event(self, device_uuid, global_objects):
+        market_info = default_market_info
```
spyrostz, line 52: Unfortunately I think a deepcopy is necessary here, in order to generate a new instance of the default market info dict. The way it is implemented ATM only the reference is copied, and will buffer incrementally the value the more times this method is called.

hannesdiedrich: hach, yes. You just shared this this morning. Will do so, Thanks

src/d3a/models/area/stats.py
```diff
@@ -50,16 +50,26 @@ def __init__(self, area_markets):
...
-                 for key in ["earned", "spent", "bought", "sold"]} \
-                if "bills" in self.aggregated_stats \
-                   and "Accumulated Trades" in self.aggregated_stats["bills"] else None
+                {key: self._extract_from_bills(key)
+                 for key in ["Accumulated Trades"]}
             self.rate_stats_market[self.current_market.time_slot] = \
                 self.min_max_avg_median_rate_current_market()
 
+    def get_current_market_stats_for_grid_tree(self):
+        return {key.lower().replace(" ", "_"): self._extract_from_bills(key)
```
spyrostz, line 70: I assume this is the area name, right? If this is the case I would very much prefer to use slugify for that

spyrostz (dismissed):
Some comments, LGTM though

src/d3a/models/area/stats.py
hannesdiedrich: Now, actually, this is just converting e.g. "Accumulated Trades" into "accumulated_trades" as of a request from Colin.

spyrostz: I see, does this apply for area names too though ? Because in this case I could foresee that Colin does not want such a behavior

spyrostz (approved):
LGTM

src/d3a/models/area/stats.py
hannesdiedrich: No, this does not apply for area names. If you see one line below, it is only for ["Accumulated Trades", "External Trades"] for now.

spyrostz: My sight is getting worse apparently. Thanks a lot for bearing with me!

PR #880, opened 2020-09-29
D3ASIM-2821: Calculating area_throughput stats from raw result data.

faizan2590:


src/d3a/models/area/stats.py
```diff
@@ -172,20 +169,3 @@ def get_market_stats(self, market_slot_list, dso=False):
                     self._get_market_area_throughput(time_slot)
 
         return out_dict
-
-    def aggregate_exported_imported_energy(self, area):
```
hannesdiedrich, line 176: Please have a look at `_get_market_area_throughput` . The DSO stats need the exported and imported energy. This is why i put this function here into the stats.py.

hannesdiedrich (commented):
One important comment for now, will wait until addressed and resume reviewing then.

src/d3a/models/area/stats.py
faizan2590: Thanks a lot for pointing, got overlooked by me, will come-up  with a solution.

src/d3a/setup/power_flow/baseline_peak_energy.py
```diff
@@ -40,10 +40,6 @@ def get_setup(config):
                     grid_fee_percentage=0, grid_fee_constant=0,
                     baseline_peak_energy_import_kWh=0.4
                 ),
-                Area(
-                    'House 1 2', [],
-                    import_capacity_kVA=2.0, export_capacity_kVA=2.0
-                ),
```
hannesdiedrich, line 46: Can you please explain why you deleted this one here? Thanks

integration_tests/steps/power_tests.py
```diff
@@ -44,8 +44,6 @@ def test_baseline_peak_energy_stats(context):
                                                'peak_percentage': 50.0,
                                                'baseline_peak_energy_kWh': 0.4},
                                     'export': {'peak_energy_kWh': 0.0}},
-                        'House 1 2': {'import': {'peak_energy_kWh': 0.0, 'capacity_kWh': 2.0},
-                                      'export': {'peak_energy_kWh': 0.0, 'capacity_kWh': 2.0}},
```
hannesdiedrich, line 48: continue from https://github.com/gridsingularity/d3a/pull/880/files#r500194689 : 
In order to have something like this here?

src/d3a/d3a_core/sim_results/area_throughput_stats.py
```diff
@@ -22,66 +22,56 @@ class AreaThroughputStats:
...
-            "baseline_peak_energy_kWh": baseline_peak_energy_kWh,
-         }
+    def update_results(self, area_dict, core_stats, current_market_time_slot_str):
+        area_throughput = core_stats.get(area_dict['uuid'], {}).get('area_throughput', {})
+        imported_peak = round_floats_for_ui(area_throughput.get('imported_energy_kWh', 0.))
+        exported_peak = round_floats_for_ui(area_throughput.get('exported_energy_kWh', 0.))
+        area_results = {
+            "import": {'peak_energy_kWh': imported_peak},
+            "export": {'peak_energy_kWh': exported_peak},
+        }
```
hannesdiedrich, line 38: Minor but you could merge these lines.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -160,7 +161,18 @@ def _populate_core_stats(self, area):
         if hasattr(area.current_market, 'trades'):
             for trade in area.current_market.trades:
                 core_stats_dict['trades'].append(trade.serializable_dict())
-        if getattr(area, 'strategy', None) is None and area.current_market is not None:
+        if area.strategy is None and area.current_market is not None:
+            area_stats = {
```
hannesdiedrich, line 165: Here too, you dont need to assign a separate variable if only needed once.

src/d3a/models/area/__init__.py
```diff
@@ -252,8 +252,6 @@ def _cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fa
 
         # area_market_stats have to updated when cycling market of each area:
         self.stats.update_area_market_stats()
-        self.stats.aggregate_exported_imported_energy(self)
-
```
hannesdiedrich, line 256: I am not sure whether this is even worth mentioning. But i would prefer this to stay here and not to trigger it in the endpointbuffer as these are area related results. Maybe Spyros has a different opinion.

hannesdiedrich (commented):
Left some minor comments and questions. Apart from that, LGTM.
Will look into it later

src/d3a/d3a_core/sim_results/area_throughput_stats.py
faizan2590: The reason I did it this way is due to one line getting longer than 100 and flake8 started complaining.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
faizan2590: Okay, will do so, if you prefer that way.

src/d3a/models/area/__init__.py
faizan2590: The reason I moved it to `SimulationEndpointBuffer` class is that markets are not cycled at the end of simulation, so the last market after the end of simulation was missing with imported/exported energy stats.

src/d3a/setup/power_flow/baseline_peak_energy.py
faizan2590: Re-added

src/d3a/models/area/__init__.py
spyrostz: I would actually be of a third opinion, why is there a need to be called from either place ? Could this not be part of the `update_area_market_stats` method ? In general it will work as it is, but it is kind of strange to call the area stats update from the endpoint buffer, and it is also strange to call 2 methods from the AreaStats in the same place, which, judging by their name, look similar.

spyrostz (commented):
One unresolved comment, apart from that LGTM though!

src/d3a/models/area/stats.py
```diff
@@ -183,18 +184,22 @@ def get_market_stats(self, market_slot_list, dso=False):
...
+
+        child_names = [area_name_from_area_or_iaa_name(c.name) for c in self._area.children]
+        if getattr(self.current_market, 'trades', None) is not None:
+            for trade in self.current_market.trades:
+                if child_buys_from_area(trade, self._area.name, child_names):
+                    add_or_create_key(self.exported_energy, self.current_market.time_slot,
+                                      trade.offer.energy)
+                if area_sells_to_child(trade, self._area.name, child_names):
+                    add_or_create_key(self.imported_energy, self.current_market.time_slot,
+                                      trade.offer.energy)
```
hannesdiedrich, line 202: I think there is no member self.current_market here. You probably meant self.area.current_market.

src/d3a/models/area/stats.py
```diff
@@ -183,18 +184,22 @@ def get_market_stats(self, market_slot_list, dso=False):
 
         return out_dict
 
-    def aggregate_exported_imported_energy(self, area):
-        if self.current_market is None:
+    def _aggregate_exported_imported_energy(self):
+        if self._area.current_market is None:
```
hannesdiedrich, line 188:
```suggestion
        if self.area.current_market is None:
```

hannesdiedrich (approved):
Once addressed for the last two comments, LGTM

src/d3a/models/area/stats.py
faizan2590: There is `src/d3a/models/area/stats.py:166`

src/d3a/models/area/stats.py
faizan2590: Its only used internally in this class, so I specially defined it as private inside via `_` as prefix.

src/d3a/models/area/stats.py
hannesdiedrich: aahh, but then you dont need to parse the area object (you can get rid of self._area). Perfect!

src/d3a/models/area/stats.py
hannesdiedrich: Ah yes, Sorry, I overlooked the `_`
https://github.com/gridsingularity/d3a/pull/880/files/f3c7edb6a47cb97786138bb2e762dc1eff7d4316#r501052676

src/d3a/models/area/stats.py
faizan2590: I cant get rid of `self._area` as I need it to accumulate `child_names `

spyrostz (approved):
LGTM

PR #881, opened 2020-09-29
D3ASIM-2820: Adapted energy trade profile to use only core stats and …

spyrostz:
…not the Area objects.

Deliberately omitted the balancing market trade profiles, since the balancing market results will differ significantly from the spot market results.

### DIFF
--- a/src/d3a/d3a_core/sim_results/device_statistics.py
+++ b/src/d3a/d3a_core/sim_results/device_statistics.py
@@ -114,8 +114,10 @@ class DeviceStatistics:
         cls._calc_min_max_from_sim_dict(subdict, bought_key_name)
 
     @classmethod
-    def _pv_production_stats(cls, area_dict: Dict, subdict: Dict, core_stats={},
+    def _pv_production_stats(cls, area_dict: Dict, subdict: Dict, core_stats=None,
                              current_market_slot=None):
+        if core_stats is None:
+            core_stats = {}
         key_name = "pv_production_kWh"
         if core_stats[area_dict['uuid']] == {}:
             return
@@ -127,7 +129,9 @@ class DeviceStatistics:
         cls._calc_min_max_from_sim_dict(subdict, key_name)
 
     @classmethod
-    def _soc_stats(cls, area_dict: Dict, subdict: Dict, core_stats={}, current_market_slot=None):
+    def _soc_stats(cls, area_dict: Dict, subdict: Dict, core_stats=None, current_market_slot=None):
+        if core_stats is None:
+            core_stats = {}
         key_name = "soc_history_%"
         if core_stats[area_dict['uuid']] == {}:
             return
@@ -138,8 +142,10 @@ class DeviceStatistics:
         cls._calc_min_max_from_sim_dict(subdict, key_name)
 
     @classmethod
-    def _load_profile_stats(cls, area_dict: Dict, subdict: Dict, core_stats={},
+    def _load_profile_stats(cls, area_dict: Dict, subdict: Dict, core_stats=None,
                             current_market_slot=None):
+        if core_stats is None:
+            core_stats = {}
         key_name = "load_profile_kWh"
         if core_stats[area_dict['uuid']] == {}:
             return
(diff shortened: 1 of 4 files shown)


src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -150,7 +156,11 @@ def _load_profile_stats(cls, area_dict: Dict, subdict: Dict, core_stats={},
 
         cls._calc_min_max_from_sim_dict(subdict, key_name)
 
-    def update(self, area_result_dict={}, core_stats={}, current_market_slot=None):
+    def update(self, area_result_dict=None, core_stats=None, current_market_slot=None):
+        if area_result_dict is None:
+            area_result_dict = {}
+        if core_stats is None:
+            core_stats = {}
```
faizan2590, line 163: Can you tell me the need for this change?

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -114,8 +114,10 @@ def calculate_stats_for_infinite_bus(cls, area_dict, subdict, core_stats, curren
         cls._calc_min_max_from_sim_dict(subdict, bought_key_name)
 
     @classmethod
-    def _pv_production_stats(cls, area_dict: Dict, subdict: Dict, core_stats={},
+    def _pv_production_stats(cls, area_dict: Dict, subdict: Dict, core_stats=None,
                              current_market_slot=None):
+        if core_stats is None:
+            core_stats = {}
```
hannesdiedrich, line 120: Is this a good practice as it is bad ti have an empty dict as default value or what is the rationale behind that?

hannesdiedrich (approved):
One minor comment, LGTM.

src/d3a/d3a_core/sim_results/energy_trade_profile.py
```diff
@@ -120,3 +126,22 @@ def add_sold_bought_lists(res_dict):
...
+            for sold_bought in ("sold_energy", "bought_energy"):
+                if sold_bought not in profile[k]:
+                    continue
+                outdict[k][sold_bought] = {}
+                for dev in profile[k][sold_bought].keys():
+                    outdict[k][sold_bought][dev] = {}
+                    for target in profile[k][sold_bought][dev].keys():
+                        outdict[k][sold_bought][dev][target] = {}
+                        for timestamp_str in profile[k][sold_bought][dev][target].keys():
+                            timestamp_object = str_to_pendulum_datetime(timestamp_str)
```
faizan2590, line 144: Correct me if Im wrong, from the description of method name, I thought it will convert timestamp strings to timestamps  whereas it actually converted to datetime object.

src/d3a/d3a_core/sim_results/device_statistics.py
spyrostz: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

src/d3a/d3a_core/sim_results/device_statistics.py
spyrostz: Answered to @faizan2590 above too, https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

src/d3a/d3a_core/sim_results/energy_trade_profile.py
spyrostz: Well, datetime and timestamp is the same, isn't it? I could change it to `convert_timestamp_strings_to_timestamp_objects` or `convert_timestamp_strings_to_datetimes` but this is not easier to read IMO.

faizan2590: > Well, datetime and timestamp is the same, isn't it? 

 Okay, Im not sure actually, the default method of pendulum `timestamp()` returns a UNIX_timestamp, so I related the same and expected here the `unix timestamp` too

faizan2590 (approved):
One minor comment, once settled, LGTM

src/d3a/d3a_core/sim_results/energy_trade_profile.py
spyrostz: Ahhhh I see now, that the source of the confusion was the UNIX timestamp. In this case I agree with you, renaming this to use datetime. Thanks!

faizan2590 (approved):
LGTM

PR #882, opened 2020-09-29
Feature/d3 asim 2796

hannesdiedrich:


src/d3a/models/area/__init__.py
```diff
@@ -268,6 +268,10 @@ def _cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fa
             self.redis_ext_conn.event_market_cycle()
 
     def tick(self):
+        if self.redis_ext_conn is not None and not self.strategy:
+            if self.redis_ext_conn.is_aggregator_controlled:
+                self.redis_ext_conn.aggregator.consume_all_area_commands(
```
spyrostz, line 273: I would remove the load/pv/storage strategy code that calls consume_all_area_commands and call this method only from here. This has the advantage that there is only one place where the aggregator is consuming the commands, instead of calling the same function in the tick method of all the strategies.

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -40,64 +48,104 @@ def _market_stats_channel(self):
...
+        sub_channel_dict = {
             f"{self.channel_prefix}/market_stats": self.market_stats_callback,
             f"{self.channel_prefix}/dso_market_stats": self.dso_market_stats_callback,
-            f"{self.channel_prefix}/grid_fees": self.set_grid_fees_callback
-        })
+            f"{self.channel_prefix}/grid_fees": self.set_grid_fees_callback,
+            f"{self.channel_prefix}/register_participant": self._register,
+            f"{self.channel_prefix}/unregister_participant": self._unregister}
+        if self.area.config.external_redis_communicator.is_enabled:
+            self.aggregator = self.area.config.external_redis_communicator.aggregator
```
spyrostz, line 77: In order not to forget and to summarize our discussion about it, could you please add a TODO comment in the SimulationConfig class that we should find a more appropriate place for the external_redis_communicator object than the config class.

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -40,64 +48,104 @@ def _market_stats_channel(self):
...
+        self.redis_com.sub_to_multiple_channels(sub_channel_dict)
 
     def market_stats_callback(self, payload):
         market_stats_response_channel = f"{self.channel_prefix}/response/market_stats"
-        payload_data = json.loads(payload["data"])
+        payload_data = json.loads(payload["data"]) \
+            if isinstance(payload["data"], str) else payload["data"]
```
spyrostz, line 83: I am curious why this is needed, would be great if you took some time to explain this. I also stumbled upon a similar change in a d3a-web PR, could it be that these 2 are related ?

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -40,64 +48,104 @@ def _market_stats_channel(self):
...
         ret_val = {"status": "ready",
+                   "area_uuid": self.area.uuid,
                    "command": "market_stats",
                    "market_stats":
                        self.area.stats.get_market_stats(payload_data["market_slots"]),
                    "transaction_id": payload_data.get("transaction_id", None)}
-        self.redis_db.publish_json(market_stats_response_channel, ret_val)
+        if self.is_aggregator_controlled:
+            ret_val.pop("transaction_id")
```
spyrostz, line 91: This would not be needed IMO if the transaction_id was popped and not `get` from the payload_data 3 lines above.

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -40,64 +48,104 @@ def _market_stats_channel(self):
...
-                "command": "grid_fees", "status": "error",
+            ret_val = {
+                "status": "error",
                 "error_message": "GridFee parameter conflicting with GlobalConfigFeeType",
-                "transaction_id": payload_data.get("transaction_id", None)}
-             )
+                **base_dict}
+
+        if self.is_aggregator_controlled:
+            ret_val.pop("transaction_id")
```
spyrostz, line 126: Same as above

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: Ah, yes, how about in the new added global_objects module in the frame of D3ASIM-2765:
https://github.com/gridsingularity/d3a/pull/879/files#diff-2756a77886549ce43c1ecfd27073b805 ?
Shall I add a story for that? I fear that A TODO will not change much soon ;-)

spyrostz (dismissed):
Some comments, LGTM once resolved/answered

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: IIRC, this is needed for the batch_commands. Here the individual commands are already parsed dicts, but if this method is called directly via the endpoint, the payload["data"] is a string.

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: So you are suggest to rather add it to the retval for the case when the market is not aggregator_controlled and not pop it here?

spyrostz: Ah I confused the logic here, so this is the inverse than what I thought, transaction_id is always populated and deleted if we have an aggregator and transaction_ids do not make sense for individual commands. 
I think it would be cleaner though, if the transaction_id was not populated at all in the aggregator case and was only populated in individual commands. This is a matter of personal preference though and YMMV, therefore feel free to ignore this comment and its follow-up. It is fine as it is even now

src/d3a/models/area/redis_external_market_connection.py
spyrostz: Discussed above, you can ignore it

src/d3a/models/area/redis_external_market_connection.py
spyrostz: Hmm, I assume you mean this method `trigger_aggregator_commands` when you mention `but if this method is called directly via the endpoint,` . How about doing a json.loads from the trigger_aggregator_commands method, so that the `market_stats_callback` accepts only dicts ?

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: Oh no, you convinced me actually. It would be more clearer your way. Will change this, thanks!

src/d3a/models/area/redis_external_market_connection.py
spyrostz: Please go ahead to add a story, you are right that TODOs can be long-lived :) thanks!

hannesdiedrich: FYI: https://gridsingularity.atlassian.net/browse/D3ASIM-2868

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: I will actually have a look again why this is different for the devices and markets. Will keep you posted

spyrostz: Thanks for checking it out!

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -40,64 +48,104 @@ def _market_stats_channel(self):
     def _grid_fees_channel(self):
         return f"{self.channel_prefix}/grid_fees"
 
-    def sub_to_area_event(self):
-        self.redis_db = self.area.config.external_redis_communicator
-        self.redis_db.sub_to_multiple_channels({
+    @staticmethod
+    def _get_transaction_id(payload):
+        data = json.loads(payload["data"])
+        if "transaction_id" in data and data["transaction_id"] is not None:
```
faizan2590, line 54: Nice to use predefined utility function `key_in_dict_and_not_none` here.

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: I found out that it is actually the other way around. for the individual commands payload["data"] is a string, but for the batch commands, it is already deserialised to a dict. In the devices, we have different functions for the aggregator and for the direct commands. Here, i propose to keep the if/then implementation and not implement an extra function. Do you agree?

spyrostz: Agreed, in this case there is no need to change the implementation

src/d3a/models/area/__init__.py
```diff
@@ -115,6 +115,7 @@ def __init__(self, name: str = None, children: List["Area"] = None,
         self._markets = AreaMarkets(self.log)
         self.stats = AreaStats(self._markets)
         log.debug(f"External connection {external_connection_available} for area {self.name}")
+        # print(self.name, external_connection_available)
```
faizan2590, line 118: Please remove before merging

faizan2590 (approved):
LGTM

PR #883, opened 2020-10-01
Store D3A tx via Substrate Smart Contract

lana-shanghai:
Added code to connect to a local running Substrate node in SubstrateMarketInterface, helper functions, and added a trigger to the one sided market.


src/d3a/models/market/blockchain_interface.py
```diff
@@ -59,11 +65,88 @@ def bc_listener(self):
         pass
 
 
+class SubstrateBlockchainInterface():
+    def __init__(self):
+        self.contracts = {'main': main_address}
+        self.substrate = SubstrateInterface(
+            url="ws://127.0.0.1:9944",
```
faizan2590, line 72: Better to introduce and read from an environment variable than having a hard coded url here.

src/d3a/models/market/blockchain_interface.py
```diff
@@ -59,11 +65,88 @@ def bc_listener(self):
...
+                'value': 1 * 10**18,
+                'gas_limit': 1000000000000,
+                'data': data
+            }
+        )
+        return call
+
+    def storage_contract_call(self, contract_address, function, path):
+        trade_id = random.randint(10, 16)
+        print('TRADE ID {}'.format(trade_id))
```
faizan2590, line 93: Prints are indeed handy while developing code but please replace it with log messages before merging this PR. Thanks!

src/d3a/models/market/blockchain_interface.py
```diff
@@ -59,11 +65,88 @@ def bc_listener(self):
...
+        call = self.substrate.compose_call(
+            call_module='Contracts',
+            call_function='call',
+            call_params={
+                'dest': contract_address,
+                'value': 1 * 10**18,
+                'gas_limit': 1000000000000,
+                'data': data
+            }
+        )
```
faizan2590, line 108: Please abstract it into a method, I see a code repetition [here](https://github.com/gridsingularity/d3a/pull/883/files#diff-797fb81c9cd9bd008da4c6892e371a50R79-R88)

src/d3a/models/market/blockchain_interface.py
```diff
@@ -59,11 +65,88 @@ def bc_listener(self):
...
+        return trade_id, call
+
+    def create_new_offer(self, energy, price, seller):
```
faizan2590, line 111:
```suggestion
    def create_new_offer_id(self, energy, price, seller):
```
Since its only returning a uuid here.

src/d3a/models/market/blockchain_interface.py
lana-shanghai: Good point, will do

src/d3a/models/market/blockchain_interface.py
lana-shanghai: Not sure where exactly to place it. Is it in the root, such as .env file?

faizan2590: You can take [this](https://github.com/gridsingularity/d3a/blob/master/src/d3a/d3a_core/redis_connections/redis_communication.py#L35) as an example, so you define this url  default value at the top of module and eventually in production could be overwritten by respective url. Does it sounds like a good approach to you?

lana-shanghai: Yes, sounds good, now it imports a default url from utils

src/d3a/models/market/blockchain_interface.py
lana-shanghai: Changed to logging

src/d3a/models/market/blockchain_interface.py
lana-shanghai: Done

src/d3a/models/market/blockchain_interface.py
lana-shanghai: I didn't touch it for now because it is a snippet from an older interface and it is used in a lot of places. We will change all the interface functions in the future to match d3a / substrate

src/d3a/models/market/one_sided.py
```diff
@@ -25,23 +25,20 @@
...
-    def __init__(self, time_slot=None, bc=None, notification_listener=None,
+    def __init__(self, time_slot=None, bc='substrate', notification_listener=None,
                  readonly=False, grid_fee_type=ConstSettings.IAASettings.GRID_FEE_TYPE,
                  transfer_fees=None, name=None, in_sim_duration=True):
         super().__init__(time_slot, bc, notification_listener, readonly, grid_fee_type,
                          transfer_fees, name)
-        self.bc_interface = MarketBlockchainInterface(bc) \
-            if bc is not None \
-            else NonBlockchainInterface()
+        self.bc_interface = SubstrateBlockchainInterface()
```
faizan2590, line 41: One nice to have refactoring could be is to initialise `bc_interface` from the `Market ` class constructor, it would indeed work even now since `TwoSidedPayAsBid` inherits `OneSidedMarket` but would be clear that this is something not dependant on market type.

src/d3a/blockchain/contracts/ink/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+
+    #[ink(event)]
+    struct Approval {
+        #[ink(topic)]
+        owner: Option<AccountId>,
+        #[ink(topic)]
+        spender: Option<AccountId>,
+        #[ink(topic)]
+        value: Balance,
+    }
```
faizan2590, line 39: Assuming the target of this contract to only track trade, remnants of token transfer could safely be removed.

src/d3a/blockchain/contracts/ink/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+
+        fn balance_of_or_zero(&self, owner: &AccountId) -> Balance {
+            let balance = self.balances.get(owner).unwrap_or(&0);
+            *balance
+        }
+
+        fn allowance_of_or_zero(&self, owner: &AccountId, spender: &AccountId) -> Balance {
+            let allowance = self.allowances.get(&(*owner, *spender)).unwrap_or(&0);
+            *allowance
+        }
```
faizan2590, line 135: Same comment(https://github.com/gridsingularity/d3a/pull/883/files#r498653888) goes here too

src/d3a/blockchain/contracts/ink/lib.rs
```diff
@@ -0,0 +1,203 @@
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+use ink_lang as ink;
+
+#[ink::contract(version = "0.1.0")]
+mod erc20 {
```
faizan2590, line 7: Could rename to something as simple as `Trade` to get an idea right from the start, what this contract is trying to achieve.

lana-shanghai: Good point, I will delete the erc20 code snippets and rename to trade.wasm for now

src/d3a/models/market/metadata.json
```diff
@@ -0,0 +1,1119 @@
+{
```
faizan2590, line 1: This could safely be removed as it could be generated based on need from the provided ink contract.

src/d3a/models/market/blockchain_interface.py
```diff
@@ -59,11 +72,81 @@ def bc_listener(self):
...
+            self.contracts['main'],
+            "trade",
+            str(path)
+        )
+        keypair = Keypair.create_from_mnemonic(mnemonic)
+        extrinsic = self.substrate.create_signed_extrinsic(call=call, keypair=keypair)
+
+        try:
+            result = self.substrate.submit_extrinsic(extrinsic, wait_for_inclusion=True)
+            print("Extrinsic '{}' sent and included in block '{}'".format(
```
spyrostz, line 135: Please do not forget to replace prints with logging.debug before merging, thanks!

lana-shanghai: Ah yes, this is a code block from the interface, will fix

src/d3a/blockchain/contracts/ink/lib.rs
lana-shanghai: For now, we decided that we will leave the ERC20 standard plus the trade structures as extension

src/d3a/blockchain/contracts/ink/lib.rs
lana-shanghai: And the same regarding this

src/d3a/blockchain/__init__.py
```diff
@@ -16,71 +16,20 @@
...
-        self.contracts[contract.address] = contract
-        if id_:
-            self.contracts[id_] = contract
-        if listeners:
-            self.listeners[contract.address].extend(listeners)
-        return contract
+    def __init__(self):
+        self.substrate = SubstrateInterface(
+            url=default_url,
+            address_type=42,
```
spyrostz, line 33: better to use a module variable (like default_url) for the address type too, because 42 is specifically for template nodes.

src/d3a/models/market/metadata.json
lana-shanghai: This is not the case, because the names of the functions convert to hexes via the metadata file

src/d3a/blockchain/contracts/ink/lib.rs
```diff
@@ -0,0 +1,203 @@
...
+    #[ink(storage)]
+    struct Erc20 {
+        /// The total supply.
+        total_supply: storage::Value<Balance>,
+        /// The balance of each user.
+        balances: storage::HashMap<AccountId, Balance>,
+        /// Approval spender on behalf of the message's sender.
+        allowances: storage::HashMap<(AccountId, AccountId), Balance>,
+        /// The balance of each user.
+        trades: storage::HashMap<Hash, (AccountId, AccountId, Balance, i8)>,
```
spyrostz, line 19: Can this not be of type `storage::HashMap<Hash, Trade>,` ? Is the tuple needed here?

lana-shanghai: Depends on which data we want to see in the response from the contract call! What type would the Trade be then?

src/d3a/blockchain/__init__.py
lana-shanghai: Sure

src/d3a/blockchain/contracts/ink/lib.rs
spyrostz: If I understand the contract correctly though, this is the type of the storage, therefore it is the type that the Trade objects are going to be stored. The response from the contract call is defined by the get_trades contract function, which has a return value of a tuple indeed, therefore theoretically we could store the trade as a Trade object, and convert it to a tuple on the get_trades method. That way inside the smart contract, we will be aware that Trades are dealt with using their own type.

In addition to that, I am intrigued to see if Ink provides a standardized way to handle structs as contract return values. Not really important though, we can live without this functionality

src/d3a/models/market/blockchain_interface.py
```diff
@@ -59,11 +72,81 @@ def bc_listener(self):
...
+        )
+
+    def compose_call(self, data, contract_address):
+        call = self.substrate.compose_call(
+            call_module='Contracts',
+            call_function='call',
+            call_params={
+                'dest': contract_address,
+                'value': 1 * 10**18,
+                'gas_limit': 1000000000000,
```
spyrostz, line 91: Maybe you can add variables for these hardcoded parameters.

src/d3a/models/market/blockchain_interface.py
```diff
@@ -19,8 +19,21 @@
...
 import platform
+import random
+import logging
 
+logging.basicConfig(filename="trades.log",
+                    format='%(asctime)s %(message)s',
+                    filemode='w')
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
```
spyrostz, line 36: Base logger should be already configured, you can just `import logging` and use `logging.info` etc.

spyrostz (approved):
Some comments, once dealt with LGTM

src/d3a/models/market/one_sided.py
lana-shanghai: Will put this in backlog ticket

spyrostz (approved):
LGTM

PR #884, opened 2020-10-01
Bug/d3 asim 2852

hannesdiedrich:


src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -97,8 +102,14 @@ def set_grid_fees_callback(self, payload):
         grid_fees_response_channel = f"{self.channel_prefix}/response/grid_fees"
         payload_data = payload["data"] \
             if isinstance(payload["data"], dict) else json.loads(payload["data"])
-        validate_area(grid_fee_percentage=payload_data.get("fee_percent", None),
-                      grid_fee_constant=payload_data.get("fee_const", None))
+        try:
+            print("validate")
```
faizan2590, line 106: Please remove before merging.

src/d3a/models/strategy/load_hours.py
```diff
@@ -160,27 +175,44 @@ def _delete_past_state(self):
...
                 update_interval = kwargs['update_interval']
-            self.bid_update.update_interval = update_interval
+        else:
+            update_interval = self.bid_update.update_interval
+
         if key_in_dict_and_not_none(kwargs, 'use_market_maker_rate'):
             self.use_market_maker_rate = kwargs['use_market_maker_rate']
 
-        self._validate_rates()
+        if self._validate_rates(initial_rate, final_rate, energy_rate_change_per_update,
```
faizan2590, line 209: I don't like this approach of using validator method for logic testing, but can't quickly think of any alternate and better way either :-(

faizan2590 (dismissed):
LGTM

src/d3a/models/area/__init__.py
```diff
@@ -127,27 +128,39 @@ def area_reconfigure_event(self, **kwargs):
                 if key_in_dict_and_not_none(kwargs, 'grid_fee_constant') else 0
             grid_fee_percentage = kwargs["grid_fee_percentage"] \
                 if key_in_dict_and_not_none(kwargs, 'grid_fee_percentage') else 0
-
-            validate_area(grid_fee_percentage=grid_fee_percentage,
-                          grid_fee_constant=grid_fee_constant)
-            self._set_grid_fees(grid_fee_constant, grid_fee_percentage)
+            try:
```
spyrostz, line 131: I think you can try/except once for the whole method. That way the code will become more DRY

src/d3a/models/area/__init__.py
```diff
@@ -127,27 +128,39 @@ def area_reconfigure_event(self, **kwargs):
...
                 key_in_dict_and_not_none(kwargs, 'export_capacity_kVA'):
             import_capacity_kVA = kwargs["import_capacity_kVA"] \
                 if key_in_dict_and_not_none(kwargs, 'import_capacity_kVA') else None
             export_capacity_kVA = kwargs["export_capacity_kVA"] \
                 if key_in_dict_and_not_none(kwargs, 'export_capacity_kVA') else None
-            validate_area(import_capacity_kVA=import_capacity_kVA,
-                          export_capacity_kVA=export_capacity_kVA)
-            self._convert_area_throughput_kva_to_kwh(import_capacity_kVA, export_capacity_kVA)
+            try:
+                validate_area(import_capacity_kVA=import_capacity_kVA,
```
spyrostz, line 159: Optional, but IIRC we discussed about being able to not apply the whole event if the validation fails, this is not followed here since the event is only partially applied if, e.g. the validation fails the `baseline_peak_energy_export_kWh` check. Had we settled on requirements about that ?

src/d3a/models/strategy/load_hours.py
```diff
@@ -111,15 +116,25 @@ def _init_price_update(self, fit_to_limit, energy_rate_increase_per_update, upda
...
-            validate_load_device_price(
-                initial_buying_rate=self.bid_update.initial_rate[time_slot],
-                energy_rate_increase_per_update=rate_change,
-                final_buying_rate=self.bid_update.final_rate[time_slot],
-                fit_to_limit=self.bid_update.fit_to_limit)
+    @staticmethod
+    def _validate_rates(initial_rate, final_rate, energy_rate_change_per_update,
+                        fit_to_limit, init=False):
+        try:
+            for time_slot in generate_market_slot_list():
```
spyrostz, line 123: I would very much prefer to use `enumerate`, or even validate each parameter separately, instead of generating the market slot list and try to access each element. The disadvantage of this method is that it is obviously not applicable for Canary Network, since the market slot list is going to be infinite. Furthermore, this way entries in the initial/final rate that are not market slots are not evaluated (not useful for now, but will be if we decide to change the slot_length via a live event).

src/d3a/models/strategy/load_hours.py
spyrostz: To be frank, I think I prefer this than the reactive approach before. This way all event parameters are either discarded or applied. There is no way that, for instance, only initial_rate and final_rate to be applied , and for energy_rate_change_per_update and fit_to_limit to not be applied.

src/d3a/models/area/__init__.py
hannesdiedrich: So you mean to follow the same approach as in the device `area_reconfigure_event`: Do validate all the inputs first and only if there are no exceptions, we update? 
I cant remember discussing this, but i totally agree and will change it.

src/d3a/models/strategy/storage.py
```diff
@@ -127,61 +131,108 @@ def __init__(self, initial_soc: float = StorageSettings.MIN_ALLOWED_SOC,
...
+                                        energy_rate_increase_per_update=bid_rate_change,
+                                        energy_rate_decrease_per_update=offer_rate_change,
+                                        fit_to_limit=bid_fit_to_limit)
+            return True
+        except D3AException as e:
+            if init:
+                raise e
+            else:
+                log.error(str(e))
+                return False
```
spyrostz, line 224: I do not like that much the mixture of boolean return value and exception raise logic. Why not just raising an exception and dealing with it in the calling code ? If I see correctly, you only need boolean return values in the area_reconfigure_event methods. Consequently you can catch the exceptions there and either leave the parameters unchanged in case an exception is raised, or change the full parameter list if no exception is raised. You can easily do that using a try/except/else block.

src/d3a/models/area/__init__.py
spyrostz: Exactly! Thanks!

src/d3a/models/strategy/load_hours.py
hannesdiedrich: I am not sure what you mean here. I understand that this approach is not applicable for the CN, however, the user will probably also provide profiles in the CN case. And yes, `generate_market_slot_list` will not be usable here. I think this needs some more discussion actually, but only if we decide to use template strategies in the CN. Or didn't i get it right?

src/d3a/models/strategy/storage.py
hannesdiedrich: True! You are right. The motivation here was to put all work into one function, bu ti agree with you and will change it accordingly.

src/d3a/models/area/__init__.py
hannesdiedrich: ok!

spyrostz (approved):
LGTM

src/d3a/models/area/__init__.py
```diff
@@ -122,32 +123,55 @@ def area_reconfigure_event(self, **kwargs):
...
-                key_in_dict_and_not_none(kwargs, 'export_capacity_kVA'):
-            import_capacity_kVA = kwargs["import_capacity_kVA"] \
-                if key_in_dict_and_not_none(kwargs, 'import_capacity_kVA') else None
-            export_capacity_kVA = kwargs["export_capacity_kVA"] \
-                if key_in_dict_and_not_none(kwargs, 'export_capacity_kVA') else None
-            validate_area(import_capacity_kVA=import_capacity_kVA,
+
+        grid_fee_constant = kwargs["grid_fee_constant"] \
+            if key_in_dict_and_not_none(kwargs, 'grid_fee_constant') \
+            else self.grid_fee_constant
```
faizan2590, line 129: Assigning None as part of else would be more safe and would look clear.

faizan2590: sorry, ignore my comment, just realised, its part of `area_reconfigure_event `.

faizan2590 (approved):
LGTM

PR #889, opened 2020-10-09
D3ASIM-2892: Added base implementations for the area, strategy and si…

spyrostz:
…mulation get_state/load_state functionalities.


src/d3a/models/state.py
```diff
@@ -115,6 +128,37 @@ def __init__(self,
...
+            "charge_history_kWh": convert_pendulum_to_str_in_dict(self.charge_history_kWh),
+            "offered_history": convert_pendulum_to_str_in_dict(self.offered_history),
+            "used_history": convert_pendulum_to_str_in_dict(self.used_history),
+            "energy_to_buy_dict": convert_pendulum_to_str_in_dict(self.energy_to_buy_dict),
+            "energy_to_sell_dict": convert_pendulum_to_str_in_dict(self.energy_to_sell_dict),
+            "used_storage": self._used_storage,
+            "battery_energy_per_slot": self._battery_energy_per_slot,
+        }
+
+    def load_state(self, state_dict):
```
hannesdiedrich, line 147: Just FYI: I was confused at first. I thought that this is the state of the load. If you can think of another naming loading the state, please change. If not, its fine really.

spyrostz: Maybe `restore_state` ?

hannesdiedrich (approved):
Straight forward. Thanks! LGTM

src/d3a/models/state.py
hannesdiedrich: yes, great!

src/d3a/models/state.py
```diff
@@ -46,13 +47,25 @@ def __init__(self):
         self.available_energy_kWh = \
             {slot: 0. for slot in generate_market_slot_list()}
 
+    def get_state(self):
+        return {"available_energy_kWh": convert_pendulum_to_str_in_dict(self.available_energy_kWh)}
+
+    def load_state(self, state_dict):
+        self.available_energy_kWh = state_dict["available_energy_kWh"]
+
```
faizan2590, line 55: One potential optimisation that I see in terms of memory is rather than saving and retrieving the whole profile is to rather purge the value of market that has already been surpassed as they are no more needed and simply been there now there as a slug and increasing memory requirement.

faizan2590 (dismissed):
Left one comment, once addressed, LGTM

src/d3a/models/state.py
spyrostz: We do it already in the strategies IIRC:
https://github.com/gridsingularity/d3a/blob/119abaaa8776afb9d68568573991a6d807b0c084/src/d3a/models/strategy/load_hours.py#L156

faizan2590: Ah yes, ignore my comment.

PR #892, opened 2020-10-12
Feature/d3 asim 2204

faizan2590:


src/d3a/models/area/__init__.py
```diff
@@ -277,10 +277,14 @@ def tick(self):
                     market.match_offers_bids()
 
         self.events.update_events(self.now)
+
+    def update_area_current_tick(self):
```
hannesdiedrich, line 281: Can you please explain why you are calling the `update_area_current_tick` from the simulation.py instead of from the area tick method?

src/d3a/setup/d3a-settings.json
```diff
@@ -21,6 +21,7 @@
       "EXPORT_ENERGY_TRADE_PROFILE_HR": false,
       "EXPORT_OFFER_BID_TRADE_HR": false,
       "RUN_REAL_TIME": false,
+      "KEEP_PAST_MARKETS": false,
```
hannesdiedrich, line 24: I think we can now safely deprecate this setting (i though i did already)

src/d3a/models/area/__init__.py
faizan2590: Since the ticks are dispatched from bottom to top, current_tick was getting updated more than once per area for an intermediate area having multiple children.

hannesdiedrich (commented):
One important question that probably explains how the fix works.

src/d3a/models/area/__init__.py
hannesdiedrich: Ahhh, I understand now. Nice catch!

src/d3a/setup/d3a-settings.json
hannesdiedrich: :-) https://github.com/gridsingularity/d3a-interface/pull/118
Only one LGTM needed.

faizan2590: I would rather be in favour of completely deprecating it once simulation results has completely be moved to a separate service

integration_tests/grid_fee.feature
```diff
@@ -83,7 +83,7 @@ Feature: GridFee integration tests
   Examples: Minimum Offer Age
      | min_offer_age | slot_length | tick_length |
      |       0       |     60      |     60      |
-     |       5       |     60      |     15      |
+     |       5       |     60      |     10      |
```
spyrostz, line 86: I assume that this is not necessary and only a remnant of testing, right ?

spyrostz (approved):
One question, LGTM though

integration_tests/grid_fee.feature
faizan2590: Resolved after offline discussion,

> It kind of is necessary, as the more you increase the delay in offer/bid propagation while keeping tick_length constant, it would be delay propagation to each and every corner, the reason this test was passing before is due to multiple increment of current_tick, can also see the correlation one line below your comment where when  tick_length  is decreased even further to 5 when min_offer_age was increased as high as 9.

PR #894, opened 2020-10-12
D3ASIM-2892: Renamed function that populated core stats in the tests …

spyrostz:
…too.


hannesdiedrich (dismissed):
LGTM

faizan2590 (approved):
LGTM

lana-shanghai (approved):

src/d3a/d3a_core/simulation.py
```diff
@@ -135,7 +135,6 @@ def current_state(self):
             "paused": self.paused,
             "slowdown": self.slowdown,
             "seed": self.initial_params["seed"],
-            "pause_after": self.pause_after,
```
spyrostz, line 138: In order to prevent any comments here, the pause_after parameter is a constructor parameter that never gets modified, that's why it should not be a part of the state (if we restart the simulation with exactly the same parameters the pause_after will have exactly the same value). Apart from that, main incentive that I removed this was because it was a duration type and crashed the JSON serialization.

hannesdiedrich (approved):
LGTM

PR #898, opened 2020-10-13
Feature/d3 asim 2842

lana-shanghai:

PR #906, opened 2020-10-14
D3ASIM-2765: Added next_market_fee to grid_stats_tree, fixed order of…

hannesdiedrich:
… cration of grid_stats_tree.


src/d3a/models/area/__init__.py
```diff
@@ -209,7 +209,7 @@ def get_grid_fee(self):
         grid_fee_type = self.config.grid_fee_type \
             if self.config is not None \
             else ConstSettings.IAASettings.GRID_FEE_TYPE
-        return self.get_path_to_root_fees() if grid_fee_type == 1 else 0
+        return self.grid_fee_constant if grid_fee_type == 1 else self.grid_fee_percentage
```
spyrostz, line 212: Maybe @lana-shanghai can help more here, but I think this is deliberate. IIRC Sarah mandated to display 0 if the grid fee type is percentage.

spyrostz (dismissed):
One comment, LGTM once resolved

src/d3a/models/area/__init__.py
hannesdiedrich: @spyrostz get_grid_fees is only used for the external endpoints (so for the api clients) They dont want to have the gridfee of the whole tree included.

faizan2590, line 212:
```suggestion
        return self.grid_fee_constant if grid_fee_type == 1 else 0.
```

spyrostz: @hannesdiedrich you are right, I misread the function name's intent

hannesdiedrich: @faizan2590 Can you explain please?

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -155,7 +155,7 @@ def event_market_cycle(self):
         if self.area.current_market is None:
             return
         market_event_channel = f"{self.channel_prefix}/market-events/market"
-        current_market_info = self.area.current_market.info
+        current_market_info = self.area.next_market.info
```
faizan2590, line 158: Can you please explain the intuition behind it?

hannesdiedrich: yes, in the `market_info` of the external devices the `next_market` is called. 
Here, it was the current market. If you rund a device aggregator and a market aggregator in parallel, you see the difference of one slot. The `next_market` makes more sense here actually, as the client user is informed, that a new market has been started. Does it make sense?

src/d3a/models/area/__init__.py
faizan2590: It was decided to only return constant grid fee and ignore the percentage based fee: https://github.com/gridsingularity/d3a/pull/898#discussion_r503948590

src/d3a/models/area/redis_external_market_connection.py
faizan2590: Okay, please rename the variable `current_market_info ` to `market_info `

src/d3a/models/area/__init__.py
hannesdiedrich: hmm, ok, but then the whole function is useless actually, then we can just return the `self.grid_fee_constant` without getting the grid_fee type.
The user can not set up the percentage pased grid fee anyway. So I propose to deal with this in a separate "deprecate percentage grid fee story" What do you think. Also tagging @spyrostz

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: will do

hannesdiedrich: hmm, you are right. the statistics are targeted to the current market, not the next market. So we should rather use the current_market for the devices instead of changing this here in the market api, right?

src/d3a/models/area/__init__.py
spyrostz: I would actually prefer to not touch the percentage grid fee, since first of all it is exposed in the UI as far as I recall

faizan2590: > So I propose to deal with this in a separate "deprecate percentage grid fee story" 

Yes please make a separate ticket for deprecating it. Thanks!

hannesdiedrich: https://github.com/gridsingularity/d3a/pull/906#discussion_r506587199

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -155,20 +159,20 @@ def event_market_cycle(self):
         if self.area.current_market is None:
             return
         market_event_channel = f"{self.channel_prefix}/market-events/market"
-        current_market_info = self.area.next_market.info
-        current_market_info["current_market_fee"] = \
+        market_info = self.market.info
```
faizan2590, line 162:
```suggestion
        market_info = self.area.next_market.info
```
I guess

faizan2590 (approved):
Left one more comment, once addressed; LGTM.

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: @faizan2590 now, please have a closer look, i created a `market` property, the same as in the external strategies.

faizan2590: Ah yes, then ignore my comment but please don't be implicit in variable naming specially for markets, I would rather prefer to be explicit and name it `next_market `

hannesdiedrich: I used the naming from the strategies and found it good, as we will change this in in the frame of https://gridsingularity.atlassian.net/browse/D3ASIM-2955

spyrostz (approved):
LGTM

PR #909, opened 2020-10-16
Feature/d3 asim 2859

faizan2590:


src/d3a/models/strategy/market_maker_strategy.py
```diff
@@ -41,3 +42,9 @@ def event_market_cycle(self):
 
     def event_activate(self):
         pass
+
+    def area_reconfigure_event(self, *args, **kwargs):
+        super().area_reconfigure_event(*args, **kwargs)
+        if key_in_dict_and_not_none(kwargs, 'grid_connected') and \
+                type(kwargs['grid_connected']) == bool:
```
hannesdiedrich, line 49: Just out of curiosity: What other type it could be?

src/d3a/d3a_core/live_events.py
```diff
@@ -46,6 +49,20 @@ def __init__(self, area_uuid, area_params):
...
+                del self.area_params['energy_rate_profile_uuid']
+                self.area_params = {'strategy_object': strategy_obj(**self.area_params)}
+            elif self.area_params['type'] == "InfiniteBus":
+                strategy_obj = InfiniteBusStrategy
+                del self.area_params['energy_rate_profile_uuid']
+                del self.area_params['buying_rate_profile_uuid']
+                self.area_params = {'strategy_object': strategy_obj(**self.area_params)}
+
+            del self.area_params['type']
+            del self.area_params['number_of_clones']
```
hannesdiedrich, line 65: Can you explain why this is needed? This reminds me actually, to remove these already in the d3a-web.
But even now: if there is no "number_of_clones" member, this will fail, so i suggest to remove this line, if you can live without it (and maybe create a small PR that deals with this in the d3a-web) Thanks

hannesdiedrich (approved):
Two comments, LGTM once addressed

src/d3a/models/strategy/market_maker_strategy.py
faizan2590: `grid_connected ` must always be boolean.

src/d3a/models/area/__init__.py
```diff
@@ -134,10 +134,14 @@ def restore_state(self, saved_state):
         self.stats.restore_state(saved_state)
 
     def area_reconfigure_event(self, **kwargs):
-        if self.strategy is not None:
+        if key_in_dict_and_not_none(kwargs, 'strategy_object'):
+            self.strategy = kwargs['strategy_object']
+            self.strategy.area = self.parent
+            self.strategy.owner = self
```
spyrostz, line 140: Aren't these already done by the `activate` method https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/area/__init__.py#L232 ?

src/d3a/d3a_core/live_events.py
spyrostz: Agree with @hannesdiedrich , and in order to circumvent the KeyError here, you can use the pop method: 
`self.area_params.pop('number_of_clones', None)`

src/d3a/models/strategy/market_maker_strategy.py
spyrostz: That reminds me, at some point we should think about using the scenario validator for live events (or do we use it even now ? I cannot recall). Should be dealt on its own story though

src/d3a/d3a_core/live_events.py
```diff
@@ -46,6 +49,20 @@ def __init__(self, area_uuid, area_params):
     def apply(self, area):
         if area.uuid != self.area_uuid:
             return False
+        if key_in_dict_and_not_none(self.area_params, 'type'):
+            if self.area_params['type'] == "MarketMaker":
+                strategy_obj = MarketMakerStrategy
+                del self.area_params['type']
```
spyrostz, line 55: This line can be deleted. type is also removed from the dict couple lines below

spyrostz (commented):
Some comments, LGTM once resolved

src/d3a/models/strategy/market_maker_strategy.py
hannesdiedrich: You are right, we dont validate the live events with the scenario validator. shall i add a story?

src/d3a/models/area/__init__.py
faizan2590: Yes, tested it and you are right.

src/d3a/d3a_core/live_events.py
faizan2590: Actually I have deleted from there and have to keep it here else the new strategy initialisation would complain as `type` parameter is unknown to it and ATM strategy doesn't accept arbitrary args and kwargs either.

src/d3a/models/strategy/market_maker_strategy.py
faizan2590: Done: https://gridsingularity.atlassian.net/browse/D3ASIM-2961

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -41,7 +41,7 @@ def decompress_and_decode_queued_strings(queued_string):
 
 @job('d3a')
 def start(scenario, settings, events, aggregator_device_mapping):
-    logging.getLogger().setLevel(logging.ERROR)
+    logging.getLogger().setLevel(logging.INFO)
```
hannesdiedrich, line 44: please change before merging

src/d3a/d3a_core/live_events.py
hannesdiedrich: Why are you handling `type` differently from e.g. `energy_rate_profile_uuid ` (meaning the difference between `del` and `pop`?

src/d3a/models/area/__init__.py
```diff
@@ -134,7 +134,15 @@ def restore_state(self, saved_state):
         self.stats.restore_state(saved_state)
 
     def area_reconfigure_event(self, **kwargs):
-        if self.strategy is not None:
+        if key_in_dict_and_not_none(kwargs, 'strategy_object'):
+            self.strategy = kwargs['strategy_object']
+            self.activate()
+            # self.strategy.area = self.parent
+            # self.strategy.owner = self
```
hannesdiedrich, line 141: Please delete

src/d3a/models/strategy/market_maker_strategy.py
hannesdiedrich: Thanks for adding the ticket, still, the check for boolean can be removed.

src/d3a/d3a_core/live_events.py
hannesdiedrich: Could you please add another story to deprecate `number_of_clones` in d3a_web (at least the sending of this parameter to the d3a, which was never needed). Thanks!

hannesdiedrich (dismissed):
Some more comments, once addressed, LGTM

src/d3a/d3a_core/live_events.py
spyrostz: Also, if you are using pop, there could be only one place where all these parameters get deleted (exactly after `if key_in_dict_and_not_none(self.area_params, 'type'):`). There could even be a small function called, for instance, sanitize_market_maker_dict that could remove all these parameters from the dict regardless of the type value. Then you could also keep the type as a variable, and use this for the if/else blocks.

src/d3a/d3a_core/live_events.py
```diff
@@ -46,6 +49,22 @@ def __init__(self, area_uuid, area_params):
...
+            if area.strategy is None:
+                return False
+            if self.area_params['type'] == "MarketMaker":
+                strategy_obj = MarketMakerStrategy
+                del self.area_params['type']
+                self.area_params.pop('number_of_clones', None)
+                self.area_params.pop('energy_rate_profile_uuid', None)
+                self.area_params = {'strategy_object': strategy_obj(**self.area_params)}
+            elif self.area_params['type'] == "InfiniteBus":
+                strategy_obj = InfiniteBusStrategy
```
spyrostz, line 62: You can also construct these objects without assigning the class object first if I understand the code correctly.

src/d3a/models/area/__init__.py
```diff
@@ -134,7 +134,15 @@ def restore_state(self, saved_state):
...
+            self.strategy.event_activate()
+            self.strategy.event_market_cycle()
```
spyrostz, line 143: I think you can also move all these lines to the event too. I am mentioning this in order to leave the `area_reconfigure_event` completely unaffected from the live events.

spyrostz (approved):
Some cosmetic/code organization comments, LGTM however once these are discussed.

src/d3a/models/strategy/market_maker_strategy.py
faizan2590: I would rather be cautious and keep it here till the validators have been added.

src/d3a/d3a_core/live_events.py
faizan2590: https://gridsingularity.atlassian.net/browse/D3ASIM-2989
Done!

src/d3a/d3a_core/live_events.py
```diff
@@ -46,11 +49,31 @@ def __init__(self, area_uuid, area_params):
     def apply(self, area):
         if area.uuid != self.area_uuid:
             return False
-
-        area.area_reconfigure_event(**self.area_params)
+        self.sanitize_live_event_paramters_for_strategy_update()
+        if key_in_dict_and_not_none(self.area_params, 'type'):
+            if area.strategy is None:
```
spyrostz, line 54: you could have a clause `area_type = self.area_params.pop("type")`. That way you will avoid the explicit del's below

spyrostz (approved):
One small comment, LGTM though

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #914, opened 2020-10-19
D3ASIM-2948: Adapting price-energy-day graph as per market type.

faizan2590:


src/d3a/d3a_core/sim_results/market_price_energy_day.py
```diff
@@ -49,11 +50,18 @@ def gather_rates_one_market(cls, area_dict, core_stats, current_market_slot, pri
...
-            t['energy_rate'] / 100.0
-            for t in core_stats.get(area_dict['uuid'], {}).get('trades', [])
-        ]
+        if ConstSettings.IAASettings.MARKET_TYPE == 1:
+            trade_rates = [
+                # Convert from cents to euro
+                (t['energy_rate'] + (t['fee_price'] / t['energy'])) / 100.0
+                for t in core_stats.get(area_dict['uuid'], {}).get('trades', [])
+            ]
+        else:
```
hannesdiedrich, line 59: Hu, am an surprised here. Doesn't the energy_rate already include the grid fee (by our definition)?

hannesdiedrich (commented):
One important question. Will continue reviewing after answered.

src/d3a/d3a_core/sim_results/market_price_energy_day.py
spyrostz: And, if I may continue on the same train of thought as Hannes, is it really dealt differently across market types? This would result in the grid fees for 2-sided markets not working at all, because 2 sided markets use the 1 sided market code.

integration_tests/steps/jira_issues_tests.py
```diff
@@ -302,3 +302,28 @@ def area_external_trade(context):
     ext_trade = list(filter(lambda x: x['areaName'] == "External Trades",
                             cumulative_trade[house1.uuid]))[0]['bars'][0]['energy']
     assert isclose(ext_trade, -1 * 0.666, rel_tol=1e-05)
+
+
+@then('we test the min/max/avg trade and devices bill')
+def check_area_trade_and_bill(context):
+    from integration_tests.steps.integration_tests import get_simulation_raw_results
+    from d3a.d3a_core.sim_results.market_price_energy_day import MarketPriceEnergyDay
+    from d3a.d3a_core.sim_results.bills import CumulativeBills
```
spyrostz, line 311: Could these be imported at the beginning of the file ? Thanks

src/d3a/d3a_core/sim_results/market_price_energy_day.py
faizan2590: Thats not how it was implemented for example in case of offer fee price is subtracted from trade price `src/d3a/models/market/one_sided.py:127`, whereas in case of bid it is not `src/d3a/models/market/two_sided_pay_as_bid.py:140`, and later for results calculation is also arranged differently dependant on market type, like in bills(`src/d3a/d3a_core/sim_results/bills.py:91`, `src/d3a/d3a_core/sim_results/bills.py:139` etc) and same for other results enpoints too. Thats the reason I added an integration test to be sure that at least price graph and bills area same in all market types (`integration_tests/jira_issues_tests.feature:151`).

integration_tests/steps/jira_issues_tests.py
faizan2590: Yes, `MarketPriceEnergyDay ` and `CumulativeBills ` could be skipped from the lazy imports.

src/d3a/d3a_core/sim_results/market_price_energy_day.py
hannesdiedrich: A I see, i think the motivation for that was that in a two siede market, the buyer should pay the fee, right?

hannesdiedrich (dismissed):
LGTM

src/d3a/d3a_core/sim_results/market_price_energy_day.py
faizan2590: > A I see, i think the motivation for that was that in a two sieded market, the buyer should pay the fee, right?

Yes, thats what I recall since this feature has been there for over an year and the intuition behind it indeed would be to buyer should always pay the fee and fee being subtracted from supplier's revenue.

spyrostz: Thanks a lot for clarifying that. Makes sense now

spyrostz (approved):
LGTM

spyrostz (approved):
LGTM

PR #918, opened 2020-10-20
D3ASIM-2893: Added restore simulation while the simulation is launched.

spyrostz:


src/d3a/d3a_core/cli.py
```diff
@@ -19,8 +19,7 @@
 from logging import getLogger
 
 import click
-import dill
```
hannesdiedrich, line 22: Now that you removed all usages of `dill` you could remove it from the requirements.

src/d3a/d3a_core/simulation.py
```diff
@@ -448,15 +423,14 @@ def _handle_input(self, console, sleep: float = 0):
...
+                if cmd not in {'i', 'p', 'q', 'r', 'R', 's', '+', '-'}:
                     log.critical("Invalid command. Valid commands:\n"
                                  "  [i] info\n"
                                  "  [p] pause\n"
                                  "  [q] quit\n"
                                  "  [r] reset\n"
-                                 "  [S] stop\n"
+                                 "  [s] stop\n"
                                  "  [R] start REPL\n"
-                                 "  [s] save state\n"
```
hannesdiedrich, line 459: Thanks a lot for this refactoring! I guess that it is not worth it to reuse the new current_state into a pickle file for this option? WE never used it right? So I agree to remove it.

hannesdiedrich (approved):
Thanks for the additional refactoring! LGTM

src/d3a/d3a_core/simulation.py
spyrostz: No worries, we never used pickle for saving the state of the simulation (it used to be part of our legacy code, was not working right from the start IIRC), and it is guaranteed to fail (Area and Market classes are not pickleable)

src/d3a/d3a_core/cli.py
spyrostz: Will do so , thanks !

src/d3a/d3a_core/simulation.py
```diff
@@ -241,24 +222,18 @@ def deactivate_areas(self, area):
...
-                slot_resume, tick_resume = divmod(self.area.current_tick,
-                                                  self.simulation_config.ticks_per_slot)
-            else:
+            if initial_slot == 0:
                 self.run_start = DateTime.now(tz=TIME_ZONE)
                 self.paused_time = 0
-                slot_resume = tick_resume = 0
+                slot_resume = 0
+            else:
+                slot_resume = initial_slot
```
faizan2590, line 234: I think `slot_resume ` could be pulled out of this if else block, simply via always setting `slot_resume = initial_slot`

faizan2590 (approved):
Left one comment, other than that LGTM

src/d3a/d3a_core/simulation.py
spyrostz: Agreed, fixing this

faizan2590 (approved):
lgtm

PR #919, opened 2020-10-20
Feature/d3 asim 2908

lana-shanghai:


faizan2590 (dismissed):
LGTM

src/d3a/d3a_core/sim_results/bills.py
```diff
@@ -257,8 +258,16 @@ def _accumulate_market_fees(self, area_dict, area_core_stats):
     def _update_market_fees(self, area_dict, area_core_stats):
         self._accumulate_market_fees(area_dict, area_core_stats)
 
+    def _accumulate_grid_fee_charged(self, area_dict, area_core_stats):
+        area_stats = area_core_stats.get(area_dict['uuid'])
+        for trade in area_stats.get('trade', []):
+            self.cumulative_fee_charged_per_market += trade['fee_price'] / 100.
```
hannesdiedrich, line 264: AFAIK, this is already done in the market:
https://github.com/gridsingularity/d3a/blob/aaaabe1cc5e2fdf6dae05aa4a450aa21d2f2a99e/src/d3a/models/market/__init__.py#L152
So you can save some computational effort here and accumulate the market.market_fee members. Would be nice if you could test this. Thanks

hannesdiedrich (commented):
Left one comment, once addressed, will give a LGTM

src/d3a/d3a_core/sim_results/bills.py
lana-shanghai: @hannesdiedrich  It is not clear if the trade.fee_price is unit price per kWh/h, or if it's needed to be multiplied by the amount traded. From previous discussions, I was under the impression it is actually price, not the final fee to be paid. Also, who do you mean by market members? All the areas that participated in the sim from start to end?

hannesdiedrich: @lana-shanghai We have tried to stick to the following naming convention in the code:
1. "prices" have the unit [cents]
2. "rates" have the unit [cents/kWh]
I had a look and AFAIS, the self.market_fee should be indeed the money that was accumulated in the market by the earned fees (in cents). Consequently, you could loop over all the markets to achieve the same as you. 
BUUUT, you don't have access to the markets in the sim_results any more as we refactored this part of code exactly for not being dependent on the objects in the simulation in order to move this code to a different place. (all this only FYI)
TLDR; Your implementation is correct. I will approve your PR now.

setup.py
```diff
@@ -12,7 +12,7 @@
         # TODO: Workaround for https://github.com/ethereum/py-solc/issues/64
         REQUIREMENTS.extend(
             ['d3a-interface @ '
-             'git+https://github.com/gridsingularity/d3a-interface.git',
+             'git+https://github.com/gridsingularity/d3a-interface.git@feature/D3ASIM-2908',
```
hannesdiedrich, line 15: please remove this before merging!

hannesdiedrich (approved):
One last comment, once addressed, LGTM

src/d3a/d3a_core/sim_results/bills.py
```diff
@@ -257,8 +258,16 @@ def _accumulate_market_fees(self, area_dict, area_core_stats):
     def _update_market_fees(self, area_dict, area_core_stats):
         self._accumulate_market_fees(area_dict, area_core_stats)
 
+    def _accumulate_grid_fee_charged(self, area_dict, area_core_stats):
+        area_stats = area_core_stats.get(area_dict['uuid'])
+        for trade in area_stats.get('trade', []):
```
faizan2590, line 263:
```suggestion
        for trade in area_stats.get('trades', []):
```

faizan2590 (approved):
LGTM

spyrostz (approved):
LGTM

hannesdiedrich (approved):
LGTM

PR #922, opened 2020-10-20
D3ASIM-2964: Added check for price on the external strategies, bid/of…

spyrostz:
…fer price should always be positive.


tests/contracts/token_fixtures.py
```diff
@@ -27,7 +27,7 @@ def ganachecli_command():
     if "GANACHE_BINARY" in os.environ:
         return os.environ["GANACHE_BINARY"], 5
     else:
-        return 'ganache-cli', 2
+        return 'ganache-cli', 8
```
faizan2590, line 30: I assume thats a mistake, you are not testing this in ganache :-)

faizan2590 (approved):
LGTM

hannesdiedrich (approved):
LGTM

tests/contracts/token_fixtures.py
spyrostz: You are right, this should be reverted :)

PR #924, opened 2020-10-20
D3ASIM-2966: Added method that triggers recalculation of the descenda…

spyrostz:
…nts' strategies price recalculation in case there is an event in the area. At the moment only useful for the grid fees.


hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #926, opened 2020-10-21
D3ASIM-2948: Incorporating device trade_rate adjustment dependent on …

faizan2590:
…market type.

### DIFF
--- a/src/d3a/d3a_core/sim_results/device_statistics.py
+++ b/src/d3a/d3a_core/sim_results/device_statistics.py
@@ -3,6 +3,7 @@ from d3a import limit_float_precision
 from d3a_interface.utils import create_or_update_subdict
 from d3a.d3a_core.sim_results import is_load_node_type, is_cell_tower_type, is_pv_node_type, \
     is_prosumer_node_type
+from d3a_interface.constants_limits import ConstSettings
 FILL_VALUE = None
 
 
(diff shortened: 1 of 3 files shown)


src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -49,8 +50,19 @@ def _device_price_stats(cls, area_dict: Dict, subdict: Dict, core_stats, current
...
+                    trade_rate = t['energy_rate'] / 100.0
+                else:
+                    trade_rate = (t['energy_rate'] - t['fee_price'] / t['energy']) / 100.0
+                trade_price_list.append(trade_rate)
+            elif t['buyer'] == area_dict['name']:
+                if ConstSettings.IAASettings.MARKET_TYPE == 1:
+                    trade_rate = (t['energy_rate'] + t['fee_price'] / t['energy']) / 100.0
+                else:
+                    trade_rate = t['energy_rate'] / 100.0
+                trade_price_list.append(trade_rate)
```
hannesdiedrich, line 64: Minor, but There is some duplication in code (including src/d3a/d3a_core/sim_results/market_price_energy_day.py) i suggest to create two small utility functions called `trade_rate_without_fee` and `trade_rate_with_fee`. What do you think?

src/d3a/d3a_core/sim_results/device_statistics.py
```diff
@@ -49,8 +50,19 @@ def _device_price_stats(cls, area_dict: Dict, subdict: Dict, core_stats, current
         area_core_trades = core_stats[area_dict['uuid']].get('trades', [])
         trade_price_list = []
         for t in area_core_trades:
-            if t['seller'] == area_dict['name'] or t['buyer'] == area_dict['name']:
-                trade_price_list.append(t['energy_rate'] / 100.0)
+            if t['seller'] == area_dict['name']:
```
hannesdiedrich, line 53: But i also have a general comment here. Why is this only coming up now? 
I guess the one sided market wasn't looked at for a long time. 
I don't think that changing the results here (or also in https://github.com/gridsingularity/d3a/pull/914) really fixes the issue that we count the grid fees only for the bids.
Also, you missed one bid AFAIS; have a look in the video in the story. The difference between offer and trade rate in the offer/bid/trade graph is making me worried. 
I hate to say it but we might need another offline discussion about this.

PR #928, opened 2020-10-21
D3ASIM-2948: Fixed the one sided market trade price to take into acco…

spyrostz:
…unt grid fees.

### DIFF
--- a/src/d3a/models/strategy/area_agents/one_sided_engine.py
+++ b/src/d3a/models/strategy/area_agents/one_sided_engine.py
@@ -143,7 +143,7 @@ class IAAEngine:
                 f"offer: source_rate ({source_rate}) is not lower than target_rate ({target_rate})"
 
             try:
-                if ConstSettings.IAASettings.MARKET_TYPE == 1 and trade is not None:
+                if ConstSettings.IAASettings.MARKET_TYPE == 1:
                     # One sided market should subtract the fees
                     trade_offer_rate = trade.offer.energy_rate - \
                                        trade.fee_price / trade.offer.energy
--- a/src/d3a/models/strategy/area_agents/one_sided_engine.py
+++ b/src/d3a/models/strategy/area_agents/one_sided_engine.py
@@ -31,7 +31,7 @@ ResidualInfo = namedtuple('ResidualInfo', ('forwarded', 'age'))
 
 class IAAEngine:
     def __init__(self, name: str, market_1, market_2, min_offer_age: int,
-                 owner: "InterAreaAgent"):
+                 owner):
         self.name = name
         self.markets = Markets(market_1, market_2)
         self.min_offer_age = min_offer_age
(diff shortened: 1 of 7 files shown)


src/d3a/models/strategy/area_agents/one_sided_engine.py
```diff
@@ -143,7 +143,13 @@ def event_trade(self, *, trade):
...
 
             try:
-                trade_offer_rate = trade.offer.energy_rate
+                if ConstSettings.IAASettings.MARKET_TYPE == 1 and trade is not None:
+                    # One sided market should subtract the fees
+                    trade_offer_rate = trade.offer.energy_rate - \
+                                       trade.fee_price / trade.offer.energy
+                else:
+                    # trade_offer_rate not used in two sided markets, trade_bid_info used instead
+                    trade_offer_rate = None
```
hannesdiedrich, line 152: really? not `trade_offer_rate = trade.offer.energy_rate` ?

hannesdiedrich: Ah yes, in the accept_offer, the offer rate is used then.

hannesdiedrich (approved):
LGTM

src/d3a/models/strategy/area_agents/one_sided_engine.py
spyrostz: It is not used at all in the 2-sided market indeed. The clause you mention is already handled by the market internally https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/market/one_sided.py#L214

hannesdiedrich:
Thanks a lot for dealing with this so quickly. I wonder whether https://github.com/gridsingularity/d3a/pull/914 should be reverted now. (also maybe the market dependant behaviour in the bills?

src/d3a/models/strategy/area_agents/one_sided_engine.py
```diff
@@ -143,7 +143,13 @@ def event_trade(self, *, trade):
                 f"offer: source_rate ({source_rate}) is not lower than target_rate ({target_rate})"
 
             try:
-                trade_offer_rate = trade.offer.energy_rate
+                if ConstSettings.IAASettings.MARKET_TYPE == 1 and trade is not None:
```
faizan2590, line 146: I think the second check `..and trade is not None` could be avoided as at its been validated at R138 at this point of commit ?

spyrostz: Good catch. Fixing it, thanks!

spyrostz:
> Thanks a lot for dealing with this so quickly. I wonder whether #914 should be reverted now. (also maybe the market dependant behaviour in the bills?

Thank you, this actually exhibited in a test failure! Fixing this

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #930, opened 2020-10-21
Bug/d3 asim 2964 negative bids

spyrostz:


src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -48,15 +48,23 @@ def _forward_bid(self, bid):
...
-        forwarded_bid = self.markets.target.bid(
-            price=(self.markets.source.fee_class.update_forwarded_bid_with_fee(
-                bid.price / bid.energy, bid.original_bid_price / bid.energy)) * bid.energy,
-            energy=bid.energy,
-            buyer=self.owner.name,
-            seller=self.markets.target.name,
-            original_bid_price=bid.original_bid_price,
-            buyer_origin=bid.buyer_origin
-        )
+        if bid.price < 0.0:
```
faizan2590, line 51:
```suggestion
        if bid.price <= 0.0:
```
how about this?

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -244,6 +244,12 @@ def event_market_cycle_prices(self):
     def _update_bid_aggregator(self, arguments):
         assert set(arguments.keys()) == {'price', 'energy', 'type', 'transaction_id'}
         bid_rate = arguments["price"] / arguments["energy"]
+        if bid_rate < 0.0:
```
faizan2590, line 247: same https://github.com/gridsingularity/d3a/pull/930/files#r509555830

spyrostz: Hmm, here it actually makes sense to have a zero value, because the user might want to set the bid price to zero.

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
spyrostz: Great! Accepting it

src/d3a/models/strategy/commercial_producer.py
```diff
@@ -60,15 +62,28 @@ def event_market_cycle(self):
...
+        except MarketException:
+            logging.error(f"Offer posted with negative energy rate {energy_rate}."
+                          f"Posting offer with zero energy rate instead.")
+            offer = market.offer(
+                0.0,
+                self.energy_per_slot_kWh,
+                self.owner.name,
+                original_offer_price=0.0,
+                seller_origin=self.owner.name
+            )
```
faizan2590, line 84: I would rather not self assume and post offer with zero price in case the external agent has placed an offer with negative rate, the reason is that it may be he/she wants to sell energy at 10 cents but due to any of his code bug end up placing offer at -10cents, its better to only throw an exception back to researcher so he can adapt his script accordingly rather him wonder why his agent is selling at 0cents. Hope my reservations makes sense

spyrostz: Sure, makes sense. In this case I'll leave just the error log, the researcher can then act accordingly indeed. Keep in mind though that the same applies for the UI, that's why we cannot raise an exception here, it would crash the simulation run. An error log looks fine though

faizan2590 (dismissed):
lgtm

PR #931, opened 2020-10-22
D3ASIM-2925: Added energy forcast endpoint for external pv and load.

hannesdiedrich:


src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -34,16 +34,24 @@ class LoadExternalMixin(ExternalMixin):
...
             f'{self.channel_prefix}/unregister_participant': self._unregister,
             f'{self.channel_prefix}/bid': self._bid,
             f'{self.channel_prefix}/delete_bid': self._delete_bid,
             f'{self.channel_prefix}/list_bids': self._list_bids,
             f'{self.channel_prefix}/device_info': self._device_info,
-        })
+            }
+
+    @property
+    def channel_dict(self):
```
faizan2590, line 49: Any specific reason to wrap same private property to public?

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -345,6 +356,79 @@ def _list_bids_aggregator(self, arguments):
...
 
+class LoadForecastExternalStrategy(LoadExternalMixin, LoadForecastStrategy):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+    @property
+    def channel_dict(self):
+        return {**self._channel_dict,
+                f'{self.channel_prefix}/set_load_power_forecast': self._set_power_forecast}
```
faizan2590, line 367: Since all strategies have unique channel prefix, we can  generalise the suffix like here `set_load_power_forecast` to simply `power_forecast`

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -212,22 +221,25 @@ def _area_reconfigure_prices(self, **kwargs):
...
-                    self._offer_impl(req.arguments, req.response_channel)
-                elif req.request_type == "delete_offer":
-                    self._delete_offer_impl(req.arguments, req.response_channel)
-                elif req.request_type == "list_offers":
-                    self._list_offers_impl(req.arguments, req.response_channel)
-                elif req.request_type == "device_info":
-                    self._device_info_impl(req.arguments, req.response_channel)
-                else:
-                    assert False, f"Incorrect incoming request name: {req}"
+                self._incoming_commands_callback_selection(req)
```
faizan2590, line 242: Good to be abstracted into a method but can you help me understand the need to be called from `event_activate ` too?

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: Agreed!

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: Yes, in order to overwrite the `self._channel_dict` in the LoadForecastExternalStrategy.

src/d3a/models/strategy/external_strategies/pv.py
hannesdiedrich: I havent added it to the `event_activate`. Or can you point me to it please?

src/d3a/models/strategy/external_strategies/load.py
faizan2590: Why not re-signature the existing `_channel_dict ` to `channel_dict ` ?

src/d3a/models/strategy/predefined_load.py
```diff
@@ -126,3 +129,56 @@ def area_reconfigure_event(self, **kwargs):
         self._area_reconfigure_prices(**kwargs)
         if key_in_dict_and_not_none(kwargs, 'daily_load_profile'):
             self._event_activate_energy(kwargs['daily_load_profile'])
+
+
+class LoadForecastStrategy(DefinedLoadStrategy):
+    """
+        Strategy responsible for reading single forecast consumption data via hardware API
+    """
+    parameters = ('daily_load_profile', 'fit_to_limit', 'energy_rate_increase_per_update',
```
faizan2590, line 138:
```suggestion
    parameters = ('power_forecast_W', 'fit_to_limit', 'energy_rate_increase_per_update',
```

src/d3a/models/strategy/predefined_load.py
```diff
@@ -126,3 +129,56 @@ def area_reconfigure_event(self, **kwargs):
...
+    def event_activate_energy(self):
+        self.update_energy_forecast()
+
+    def update_energy_forecast(self):
+        self._initiate_hrs_per_day()
+
+        # sets energy forecast for next_market
+        energy_forecast_Wh = (self.area.config.slot_length / duration(hours=1)) * \
+            self.power_forecast_buffer_W
+        assert energy_forecast_Wh >= 0.0
```
faizan2590, line 180:
```suggestion
        assert energy_forecast_Wh >= 0.0, energy_forecast_Wh cannot be a negative value.
```
Minor cosmetic.

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: Because i need both. Please have a look at the LoadForecastExternalStrategy.
 I can not update a property (`channel_dict`), but i can change the 
`_channel_dict` which is not a property.

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -172,3 +172,47 @@ def area_reconfigure_event(self, **kwargs):
         if key_in_dict_and_not_none(kwargs, 'power_profile'):
             self._power_profile_W = kwargs['power_profile']
         self.read_config_event()
+
+
+class PVForecastStrategy(PVPredefinedStrategy):
+    """
+        Strategy responsible for reading single production forecast data via hardware API
+    """
+    parameters = ('power_profile', 'panel_count', 'initial_selling_rate', 'final_selling_rate',
```
faizan2590, line 181:
```suggestion
    parameters = ('power_forecast_W', 'panel_count', 'initial_selling_rate', 'final_selling_rate',
```

src/d3a/models/strategy/predefined_load.py
faizan2590: On second thought, why not raise an exception, so that simulation doesn't crash due to external clients' wrong value ?
```suggestion
        if energy_forecast_Wh <= 0.0:
            raise MarketException("energy_forecast_Wh cannot be a negative value.")
```

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -172,3 +172,47 @@ def area_reconfigure_event(self, **kwargs):
...
+        self.power_forecast_buffer_W = power_forecast_W
+
+    def event_activate_energy(self):
+        self.produced_energy_forecast_kWh()
+
+    def produced_energy_forecast_kWh(self):
+        # sets energy forecast for next_market
+        energy_forecast_kWh = (self.area.config.slot_length / duration(hours=1)) / 1000 * \
+            self.power_forecast_buffer_W
+        assert energy_forecast_kWh >= 0.0
```
faizan2590, line 214: [same](https://github.com/gridsingularity/d3a/pull/931/files#r511863864) applies here too.

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -172,3 +172,47 @@ def area_reconfigure_event(self, **kwargs):
...
+        slot_time = self.area.next_market.time_slot
+        print(self.area.next_market.time_slot_str, "write_to_market", self.power_forecast_buffer_W)
```
faizan2590, line 216: Minor, to be removed before merging.

src/d3a/models/strategy/predefined_load.py
hannesdiedrich: Its a matter of taste really. 
1. An AssertionError behaves the same as an Exception AFAIK
2. The user can not set a negative value, please have a look into the `_set_power_forecast_impl`, there it is already checked (and does not make the simulation crash because it is caught by the try/except.

TLDR; will raise an Exception as you suggested.

src/d3a/models/strategy/external_strategies/load.py
faizan2590: umm, both are property.

hannesdiedrich: oh ,yes, sorry, i had so many forth and backs here that i lost track. 
But here too it is a matter of taste. I either do as I did or I have to overwrite the event_activate method in the LoadForecastExternalStrategy. 
Maybe @spyrostz  is best to be ask here for the best practice.

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: @faizan2590 Sorry, I didn't read properly the first time.
I thought about this when implementing too, and then the `_set_power_forecast` could even live in the ExternalMixin.
However, in the `_set_power_forecast_impl` different Strategy attributes are set and this shows also that pv and load are not identical here. This is why I kept the difference between load and pv.
But if @spyrostz agrees, I will refactor it as you said.

hannesdiedrich: @faizan2590 Actually, no, i think yours is a goof idea and will try to implement it like this.

src/d3a/models/strategy/external_strategies/pv.py
faizan2590: Sorry yes you are right, my bad.

src/d3a/models/strategy/external_strategies/load.py
faizan2590: > I thought about this when implementing too, and then the _set_power_forecast could even live in the ExternalMixin.

As per MRO `_set_power_forecast ` would be overwritten by `LoadForecastExternalStrategy` and `PVForecastExternalStrategy` respectively, so I don't  think you should be worried about base implementation in `ExternalMixin `

hannesdiedrich: Og course this includes to remove the implementations in `LoadForecastExternalStrategy` and `PVForecastExternalStrategy`, currently trying it.

src/d3a/models/strategy/predefined_load.py
faizan2590: You can remove it at all from here as it is already covered in `_set_power_forecast_impl `.

hannesdiedrich: yeah, i wanted to keep it as extra layer of security (in case something goes wrong in the conversion between power and energy) But you are right, its not needed as there is just a factor applied.

src/d3a/models/strategy/external_strategies/load.py
spyrostz: I would actually remove the private property and keep only the public property, unless I am not seeing something.

src/d3a/models/strategy/predefined_load.py
```diff
@@ -126,3 +130,55 @@ def area_reconfigure_event(self, **kwargs):
         self._area_reconfigure_prices(**kwargs)
         if key_in_dict_and_not_none(kwargs, 'daily_load_profile'):
             self._event_activate_energy(kwargs['daily_load_profile'])
+
+
+class LoadForecastStrategy(DefinedLoadStrategy):
```
spyrostz, line 135: I am fairly certain that we will never use this strategy without an external API. I would recommend to merge it with the LoadForecastExternalStrategy

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -172,3 +172,45 @@ def area_reconfigure_event(self, **kwargs):
         if key_in_dict_and_not_none(kwargs, 'power_profile'):
             self._power_profile_W = kwargs['power_profile']
         self.read_config_event()
+
+
+class PVForecastStrategy(PVPredefinedStrategy):
```
spyrostz, line 177: Same as above, we'll never need to instantiate it without an external connection

src/d3a/models/strategy/pv.py
```diff
@@ -98,8 +98,7 @@ def area_reconfigure_event(self, **kwargs):
             self.panel_count = kwargs['panel_count']
         if key_in_dict_and_not_none(kwargs, 'max_panel_power_W'):
             self.max_panel_power_W = kwargs['max_panel_power_W']
-
-        self.produced_energy_forecast_kWh()
+            self.produced_energy_forecast_kWh()
```
spyrostz, line 101: Are you certain ? This will have the effect that the produced_energy_forecast will not be updated if e.g. the panel_count parameter is modified, which does not look right to my eyes

spyrostz (commented):
Added some comments, will have another look once dealt with

src/d3a/models/strategy/predefined_load.py
hannesdiedrich: ok, makes sense!

src/d3a/models/strategy/pv.py
hannesdiedrich: Ah yes, you are right, i didnt consider the panel count as an energy related parameter. But it certainly is.

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: ok guys, i gave it yet another look. I dont want to subscribe to `f'{self.channel_prefix}/set_power_forecast'` channel for the normal `LoadExternalMixin`. So i have to change the list of channels in the `LoadForecastExternalStrategy`. I dont see another way of doing it the way i did. I can not overwrite the `event_activate` because i need to call the event_activate of the LoadHours Strategy there.
But maybe there is another way, could you please tell me? Thanks

faizan2590: Updating channel dict inside the constructor of this new class could be an option like `self.channel_dict.update({f'{self.channel_prefix}/set_load_power_forecast': self._set_power_forecast})`

hannesdiedrich: No, this wont work, as it is a property. You can not overwrite a property.

faizan2590: This property returns dictionary to which you are updating, I haven't tried yet but that was from the top of my head to give a try

hannesdiedrich: I have tried it in the past. 
yes, indeed, the property returns a dict, but it is really a function. You can not call `.update()` on a function. 
(already pycharm is complaining about it. )

hannesdiedrich: what i could do here is to initiate a dictionary `channel_dict` in the constructor and follow the approach you suggested. (so removing the property function.

spyrostz: You can also use `super()`

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -377,3 +376,47 @@ def _reject_all_pending_requests(self):
...
+                 "transaction_id": transaction_id})
+        else:
+            self.pending_requests.append(
+                IncomingRequest("set_power_forecast", arguments,
+                                power_forecast_response_channel))
+
+    def _set_power_forecast_impl(self, arguments, response_channel):
+        try:
+            assert arguments["power_forecast"] >= 0.0
+            self.power_forecast_buffer_W = arguments["power_forecast"]
```
spyrostz, line 408: Might be an overkill but I think that in the long run there needs to be an extra validation here, that will check the power forecast against the nameplate characteristics of the load and PV (essentially the max PV/Load power)

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -345,6 +355,82 @@ def _list_bids_aggregator(self, arguments):
...
+        elif req.request_type == "delete_bid":
+            self._delete_bid_impl(req.arguments, req.response_channel)
+        elif req.request_type == "list_bids":
+            self._list_bids_impl(req.arguments, req.response_channel)
+        elif req.request_type == "device_info":
+            self._device_info_impl(req.arguments, req.response_channel)
+        elif req.request_type == "set_power_forecast":
+            self._set_power_forecast_impl(req.arguments, req.response_channel)
+        else:
+            assert False, f"Incorrect incoming request name: {req}"
```
spyrostz, line 413:
```suggestion
        if req.request_type == "set_power_forecast":
            self._set_power_forecast_impl(req.arguments, req.response_channel)
        else:
            super(). _incoming_commands_callback_selection(req)
```

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -345,6 +355,82 @@ def _list_bids_aggregator(self, arguments):
...
+
+    def event_market_cycle(self):
+        self.update_energy_forecast()
+        super().event_market_cycle()
+
+    def event_activate_energy(self):
+        self.update_energy_forecast()
+
+    def update_energy_forecast(self):
+        self._initiate_hrs_per_day()
```
spyrostz, line 423: Why do you need to call this again ? This should be called by the _update_energy_requirement (essentially the first time that the energy_requirement is populated), therefore there should be no need to be called from here too.

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -354,6 +364,71 @@ def _offer_aggregator(self, arguments):
...
+        elif req.request_type == "delete_offer":
+            self._delete_offer_impl(req.arguments, req.response_channel)
+        elif req.request_type == "list_offers":
+            self._list_offers_impl(req.arguments, req.response_channel)
+        elif req.request_type == "device_info":
+            self._device_info_impl(req.arguments, req.response_channel)
+        elif req.request_type == "set_power_forecast":
+            self._set_power_forecast_impl(req.arguments, req.response_channel)
+        else:
+            assert False, f"Incorrect incoming request name: {req}"
```
spyrostz, line 414: Same as with the load, this method can be refactored.

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -354,6 +364,71 @@ def _offer_aggregator(self, arguments):
...
+        self.produced_energy_forecast_kWh()
+        super().event_market_cycle()
+
+    def event_activate_energy(self):
+        self.produced_energy_forecast_kWh()
+
+    def produced_energy_forecast_kWh(self):
+        # sets energy forecast for next_market
+        energy_forecast_kWh = (self.area.config.slot_length / duration(hours=1)) / 1000 * \
+            self.power_forecast_buffer_W
```
spyrostz, line 426: This calculation exists in many places now. How about refactoring to two utility functions called `convert_energy_to_power` and `convert_power_to_energy` respectively ?

src/d3a/models/strategy/load_hours.py
```diff
@@ -321,7 +321,8 @@ def event_market_cycle_prices(self):
         if ConstSettings.IAASettings.MARKET_TYPE == 1:
             return
         for market in self.active_markets:
-            if self.energy_requirement_Wh[market.time_slot] > 0:
+            if market.time_slot in self.energy_requirement_Wh and \
```
spyrostz, line 324: Not sure why this was needed too. I would expect that the energy requirement should always be populated with the time_slot value (for the simulation case at least, where this value is populated at the beginning of the simulation)

spyrostz (commented):
Some comments, LGTM once dealt with

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: I agree in general, but then we need to add this as an input argument for the new Strategy which will also include a UI change. So i suggest to do it later. Shall I add a ticket?

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: Yes, thanks a lot! I only learned about the value of `super()` last week :-)

src/d3a/models/strategy/external_strategies/__init__.py
spyrostz: Please do so yes, but as I mentioned above this is for the long run, not urgent for the hackathon. Thanks!

src/d3a/models/strategy/load_hours.py
hannesdiedrich: Indeed. Remnant from debugging. Will remove it.

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: I was not set, because the `event_activate_energy` is overwritten here. But you are right, it should live in there. Moving it.

spyrostz (approved):
LGTM

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: Done https://gridsingularity.atlassian.net/browse/D3ASIM-3010

faizan2590 (approved):
LGTM

PR #932, opened 2020-10-22
D3ASIM-2946: Adding uniform area types for UI.

faizan2590:


src/d3a/d3a_core/sim_results/__init__.py
```diff
@@ -29,6 +33,23 @@ def is_buffer_node_type(area):
...
+        return "PV"
+    elif is_load_node_type(area):
+        return "Load"
+    elif is_cell_tower_type(area):
+        return "CellTower"
+    elif is_prosumer_node_type(area):
+        return "Storage"
+    elif is_bulk_power_producer(area) or is_buffer_node_type(area):
+        return "MarketMaker"
+    elif area['type'] == "FinitePowerPlant":
```
spyrostz, line 47: Even if it is just one value, I would prefer it if you created one more function that checks the finite power plant, in order to be more consistent.

spyrostz (approved):
One comment, once resolved LGTM

src/d3a/d3a_core/sim_results/__init__.py
```diff
@@ -29,6 +33,23 @@ def is_buffer_node_type(area):
     return area['type'] == "InfiniteBusStrategy"
 
 
+def get_unified_area_type(area):
+    if is_pv_node_type(area):
+        return "PV"
+    elif is_load_node_type(area):
+        return "Load"
+    elif is_cell_tower_type(area):
+        return "CellTower"
```
hannesdiedrich, line 42: I know, this is here because of completeness, but we should think about deprecating this special type at some point (as it is only a loadhoursStreategy and the UI isnt using this anyway.

hannesdiedrich (approved):
One general comment, LGTM

src/d3a/d3a_core/sim_results/__init__.py
faizan2590: Yes, made a ticket to be taken into account later: https://gridsingularity.atlassian.net/browse/D3ASIM-2993

src/d3a/d3a_core/sim_results/__init__.py
faizan2590: Okay, will do in second round, as it is cosmetics and doing it as part of this PR will further delay frontend.

PR #938, opened 2020-10-23
fixed bids offers trades accumulation in the endpoint buffer

hannesdiedrich:


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -312,7 +312,8 @@ def update_offer_bid_trade(self):
         if self.current_market_time_slot_str == "":
             return
         for area_uuid, area_result in self.flattened_area_core_stats_dict.items():
-            self.bids_offers_trades[area_uuid] = area_result
+            self.bids_offers_trades[area_uuid] = \
+                {k: area_result[k] for k in ('offers', 'bids', 'trades')}
```
faizan2590, line 316: One more thing would be nice to have is to sanitise `offers', 'bids', 'trades'` i.e there is no need to send the complete serialised version of items contained in these objects

faizan2590 (approved):
One enhancement comments, other than that LGTM

src/d3a/d3a_core/sim_results/endpoint_buffer.py
hannesdiedrich: @faizan2590 You mean by only sending the bids/offers/trades that belong to the area? it hink this is already done by design of `flattened_area_core_stats_dict` or am i misunderstanding?

hannesdiedrich: Discussed offline: Only sanitise the offer bid and trades dicts from data that is not needed to be stored in the DB when research team decided which data is valueable or not.

PR #943, opened 2020-10-26
D3ASIM-1742: wip, refactor TradeBidInfo

lana-shanghai:


src/d3a/models/market/market_structures.py
```diff
@@ -252,8 +270,20 @@ def trade_from_JSON_string(trade_string, current_time):
         trade_dict['residual'] = offer_or_bid_from_JSON_string(trade_dict['residual'],
                                                                current_time)
     trade_dict['time'] = parse(trade_dict['time'])
-    # if 'offer_bid_trade_info' in trade_dict:
-    #     trade_dict['offer_bid_trade_info'] = TradeBidInfo(*trade_dict['offer_bid_trade_info'])
+    if 'offer_bid_trade_info' in trade_dict and trade_dict['offer_bid_trade_info'] is not None:
```
faizan2590, line 273:
```suggestion
    if key_in_dict_and_not_none(trade_dict, 'offer_bid_trade_info'):
```
could import `key_in_dict_and_not_none` from `d3a_interface.utils`

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
```diff
@@ -132,9 +132,16 @@ def event_bid_traded(self, *, bid_trade):
...
+                    original_bid_rate=bid_trade.offer_bid_trade_info.original_bid_rate,
+                    propagated_bid_rate=bid_trade.offer_bid_trade_info.propagated_bid_rate,
+                    original_offer_rate=bid_trade.offer_bid_trade_info.original_offer_rate,
+                    propagated_offer_rate=bid_trade.offer_bid_trade_info.propagated_offer_rate,
+                    trade_rate=bid_trade.offer_bid_trade_info.trade_rate
+                )
                 updated_trade_offer_info = \
                     self.markets.source.fee_class.propagate_original_offer_info_on_bid_trade(
-                        [None, None, *bid_trade.offer_bid_trade_info]
+                        trade_offer_info
```
faizan2590, line 144:
```suggestion
                        bid_trade.offer_bid_trade_info
```
Sorry I know I was the one suggested earlier to create [here](https://github.com/gridsingularity/d3a/pull/943/files#diff-4c2698dbd37d0660ae1460c62b57278e468c8d9674f2ed19b95a59b5d251ff64R135-R141) `TradeBidOfferInfo` but now that I'm looking into it again realised `bid_trade` already has an `offer_bid_trade_info` and recreation of `TradeBidOfferInfo` could simply be avoided in this case .

lana-shanghai: It was failing with the list before, so it was a fix, not a suggestion I think

src/d3a/models/market/market_structures.py
```diff
@@ -252,8 +271,21 @@ def trade_from_JSON_string(trade_string, current_time):
         trade_dict['residual'] = offer_or_bid_from_JSON_string(trade_dict['residual'],
                                                                current_time)
     trade_dict['time'] = parse(trade_dict['time'])
-    # if 'offer_bid_trade_info' in trade_dict:
-    #     trade_dict['offer_bid_trade_info'] = TradeBidInfo(*trade_dict['offer_bid_trade_info'])
+    if key_in_dict_and_not_none(trade_dict, 'offer_bid_trade_info'):
+        if len(trade_dict['offer_bid_trade_info']) == 5:
```
faizan2590, line 275:
```suggestion
        if len(trade_dict['offer_bid_trade_info']) == TradeBidOfferInfo.len():
```
A bit overkill but better to check by actual number of elements in the namedtupple via adding the following implementation inside `class TradeBidOfferInfo`

`    
@classmethod
    def len(cls):
        return len(cls._fields)
`

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
faizan2590: Indeed it will fail with list but what I am proposing now is to replace `list` with `bid_trade.offer_bid_trade_info` which is already in the  `TradeBidOfferInfo ` format and a duplicate.

faizan2590 (approved):
Have left two comments related to enhancement, LGTM once addressed.

src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py
lana-shanghai: oh, I see now, OK

src/d3a/models/market/market_structures.py
```diff
@@ -252,8 +275,21 @@ def trade_from_JSON_string(trade_string, current_time):
...
+            if _is_bid(trade_string):
+                trade_dict['offer_bid_trade_info'].insert(2, None)
+                trade_dict['offer_bid_trade_info'].insert(3, None)
+                trade_dict['offer_bid_trade_info'] = TradeBidOfferInfo(
+                    *trade_dict['offer_bid_trade_info'])
+            elif _is_offer(trade_string):
+                trade_dict['offer_bid_trade_info'].insert(0, None)
+                trade_dict['offer_bid_trade_info'].insert(1, None)
+                trade_dict['offer_bid_trade_info'] = TradeBidOfferInfo(
+                    *trade_dict['offer_bid_trade_info'])
```
spyrostz, line 292: Interesting approach, but how about making it a bit more explicit ? In order to avoid this complex deserialization code, the serialization could be simplified and transmit a dict with key-value pairs of the TradeBidOfferInfo (instead of a list that is transmitted now), even including None values. That way the deserialization code will be as simple as `trade_dict['offer_bid_trade_info'] = TradeBidOfferInfo(**trade_dict['offer_bid_trade_info'])` (in theory at least)

spyrostz (approved):
One comment, LGTM once resolved

src/d3a/models/market/market_structures.py
lana-shanghai: Attempting to do the fix ass suggested

spyrostz (approved):
LGTM

src/d3a/models/market/market_structures.py
```diff
@@ -177,17 +177,40 @@ def offer_or_bid_from_JSON_string(offer_or_bid, current_time):
...
-                              ('original_bid_rate', 'propagated_bid_rate',
-                               'original_offer_rate', 'propagated_offer_rate',
-                               'trade_rate'))):
+def _is_offer(offer_or_bid):
+    offer_bid_dict = json.loads(offer_or_bid)
+    object_type = offer_bid_dict.pop("type")
+    if object_type == "Offer":
+        return True
+    else:
+        return False
```
hannesdiedrich, line 186: Minor, but this could be simplified as:
```suggestion
    return object_type == "Offer":
```
(also in _is_bid)

src/d3a/models/market/market_structures.py
```diff
@@ -177,17 +177,40 @@ def offer_or_bid_from_JSON_string(offer_or_bid, current_time):
...
+
+
+def _is_bid(offer_or_bid):
```
hannesdiedrich, line 189: But also: I am not seeing `_is_bid` and `is_offer` being used anywhere. Is this a preparation for the future?

lana-shanghai: Good point, there were used before refactoring, I'll clean that up, thanks

src/d3a/models/market/market_structures.py
lana-shanghai: Right, that's better

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #953, opened 2020-11-03
D3ASIM-2987: Added check for the config/collab_uuid when selecting an…

spyrostz:
… aggregator, because channel mapping is not done in this case.

### DIFF
--- a/src/d3a/d3a_core/redis_connections/aggregator_connection.py
+++ b/src/d3a/d3a_core/redis_connections/aggregator_connection.py
@@ -71,6 +71,9 @@ class AggregatorHandler:
 
     def aggregator_callback(self, payload):
         message = json.loads(payload["data"])
+        if d3a.constants.EXTERNAL_CONNECTION_WEB is True and \
+                message['config_uuid'] != d3a.constants.COLLABORATION_ID:
+            return
         if message["type"] == "CREATE":
             self._create_aggregator(message)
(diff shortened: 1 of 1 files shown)


hannesdiedrich (approved):
Ahhh! Thanks for adding this check!

src/d3a/d3a_core/redis_connections/aggregator_connection.py
```diff
@@ -71,6 +71,9 @@ def add_batch_finished_event(self, device_uuid, event):
 
     def aggregator_callback(self, payload):
         message = json.loads(payload["data"])
+        if d3a.constants.EXTERNAL_CONNECTION_WEB is True and \
```
faizan2590, line 74: Why not doing the same for local run?

spyrostz: Because local runs do not care about multiple simulations, there is only one simulation running at any time. We could of course change the API client to also populate the config_uuid for the case of Redis (essentially when we are selecting an aggregator via the CLI), if you would like to do that ping me to change it.

faizan2590: No no keep it as it is for now, no need to complicate thing for now until multiple collaboration local run is needed.

faizan2590 (approved):
LGTM

PR #955, opened 2020-11-04
D3ASIM-3009: Synced the simulation time with real-time in case the si…

spyrostz:
…mulation is a Canary network.


src/d3a/constants.py
```diff
@@ -44,3 +44,5 @@
 SIMULATION_PAUSE_TIMEOUT = 600
 
 D3A_TEST_RUN = False
+
+IS_CANARY_NETWORK = True
```
hannesdiedrich, line 48: IMO the default value should be `False` for local runs. Or am misunderstanding?

src/d3a/d3a_core/simulation.py
```diff
@@ -301,6 +311,30 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
         self.simulation_config.external_redis_communicator.sub_to_aggregator()
         self.simulation_config.external_redis_communicator.start_communication()
         self._update_and_send_results()
+
+        if d3a.constants.IS_CANARY_NETWORK is True:
+            slot_count = sys.maxsize
+            import datetime
```
hannesdiedrich, line 317: please move to top

src/d3a/d3a_core/simulation.py
```diff
@@ -301,6 +311,30 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
+
+            ConstSettings.GeneralSettings.RUN_REAL_TIME = True
+
+            today = datetime.date.today()
+            seconds_since_midnight = time.time() - time.mktime(today.timetuple())
```
hannesdiedrich, line 322: I guess that you wanted to use `mktime` here so you used datetime instead of pendulum, right?

hannesdiedrich (dismissed):
Some minor comments, LGTM in general

src/d3a/d3a_core/simulation.py
```diff
@@ -301,6 +311,30 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
         self.simulation_config.external_redis_communicator.sub_to_aggregator()
         self.simulation_config.external_redis_communicator.start_communication()
         self._update_and_send_results()
+
+        if d3a.constants.IS_CANARY_NETWORK is True:
+            slot_count = sys.maxsize
+            import datetime
+
+            ConstSettings.GeneralSettings.RUN_REAL_TIME = True
```
faizan2590, line 319: This flag could be removed, since canary network is expected to run in realtime.

src/d3a/constants.py
spyrostz: Remnant from testing, changing this, thanks

src/d3a/d3a_core/simulation.py
spyrostz: Fixing this

src/d3a/d3a_core/simulation.py
spyrostz: Exactly, I wanted to minimise usage of pendulum here. Should not make a difference if you prefer to use pendulum though

src/d3a/d3a_core/simulation.py
spyrostz: As discussed, removing this. Thanks

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #956, opened 2020-11-05
Feature/d3 asim 3004

hannesdiedrich:


src/d3a/constants.py
```diff
@@ -46,3 +46,6 @@
 D3A_TEST_RUN = False
 
 IS_CANARY_NETWORK = False
+CN_PROFILE_EXPANSION_DAYS = 7
+
+RUN_IN_REALTIME = IS_CANARY_NETWORK
```
faizan2590, line 51: if `RUN_IN_REALTIME ` is supposed to follow `IS_CANARY_NETWORK ` then why not simply having one of the flag?

src/d3a/d3a_core/util.py
```diff
@@ -293,11 +292,11 @@ def read_settings_from_file(settings_file):
...
         simulation_settings = {
             "sim_duration": IntervalType('H:M')(
-                settings["basic_settings"].get('sim_duration', timedelta(hours=24))),
+                settings["basic_settings"].get('sim_duration', duration(hours=24))),
             "slot_length": IntervalType('M:S')(
-                settings["basic_settings"].get('slot_length', timedelta(minutes=15))),
+                settings["basic_settings"].get('slot_length', duration(minutes=15))),
             "tick_length": IntervalType('M:S')(
-                settings["basic_settings"].get('tick_length', timedelta(seconds=15))),
+                settings["basic_settings"].get('tick_length', duration(seconds=15))),
```
faizan2590, line 299: Better to replace default `sim_duration `, `slot_length `, `tick_length ` with some constant_variable.

src/d3a/d3a_core/util.py
```diff
@@ -527,6 +537,25 @@ def convert_area_throughput_kVA_to_kWh(transfer_capacity_kWA, slot_length):
         if transfer_capacity_kWA is not None else 0.
 
 
+def find_timestamp_of_same_weekday_and_time(indict, time_slot, ignore_not_found=False):
+    if IS_CANARY_NETWORK:
+        start_time = list(indict.keys())[0]
```
faizan2590, line 542: Are you sure that `indict ` is an ordered dict else I think its not guaranteed to be the first index as `start_time ` ?

src/d3a/d3a_core/util.py
```diff
@@ -527,6 +537,25 @@ def convert_area_throughput_kVA_to_kWh(transfer_capacity_kWA, slot_length):
...
+        key = datetime(year=start_time.year, month=start_time.month, day=start_time.day,
```
faizan2590, line 543: A bit more verbose variable name would be easier to understand its purpose.

src/d3a/d3a_core/util.py
```diff
@@ -527,6 +537,25 @@ def convert_area_throughput_kVA_to_kWh(transfer_capacity_kWA, slot_length):
         if transfer_capacity_kWA is not None else 0.
 
 
+def find_timestamp_of_same_weekday_and_time(indict, time_slot, ignore_not_found=False):
```
faizan2590, line 540: Again, what do you mean by `ignore_not_found `?

src/d3a/constants.py
hannesdiedrich: In order to test the behaviour of the strategies, if `IS_CANARY_NETWORK=True` you want to disable the `RUN_IN_REALTIME` here manually and locally. We have to discuss how to rerun the strategy tests with `IS_CANARY_NETWORK=True` on travis.

src/d3a/d3a_core/util.py
hannesdiedrich: True! Will use the ones from d3a-interface

src/d3a/d3a_core/util.py
hannesdiedrich: No, you are right. I mean it is ordered by the creation method, but i can not be sure and thus will use an OrderedDict (as long as this can oder by pendulum time stamp).

src/d3a/d3a_core/util.py
hannesdiedrich: agreed

src/d3a/d3a_core/util.py
hannesdiedrich: This parameter is only used in the `read_user_profile._fill_gaps_in_profile` Where the gaps in the profiles are filled with the last available non-0 value. There, it is ok to not find the timestamp in the input profile.

src/d3a/constants.py
spyrostz: I just started the PR, but judging from this I guess that you essentially use IS_CANARY_NETWORK in order to control whether the strategy needs to recycle its profiles. I assume that when you say strategy tests you are talking about the unit tests, right ? In the unit tests though the simulation.py is not invoked, therefore the `RUN_IN_REALTIME` is not really needed. I think I have misunderstood something though.

src/d3a/d3a_core/simulation.py
```diff
@@ -313,7 +313,7 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
         self.simulation_config.external_redis_communicator.start_communication()
         self._update_and_send_results()
 
-        if d3a.constants.IS_CANARY_NETWORK is True:
+        if RUN_IN_REALTIME:
```
spyrostz, line 316: That's ok, but keep in mind that not accessing the constant with the full path will have the effect that you cannot override its value in tests. If that's okay then no problem at all

src/d3a/models/read_user_profile.py
```diff
@@ -150,8 +146,13 @@ def _calculate_energy_from_power_profile(profile_data_W: Dict[str, float],
     :param slot_length: slot length duration
     :return: a mapping from time to energy values in kWh
     """
-
     input_time_list = list(profile_data_W.keys())
+
+    # add one market slot in order to have another data point for integration
+    additional_time_stamp = input_time_list[-1] + slot_length
+    profile_data_W[additional_time_stamp] = profile_data_W[input_time_list[-1]]
+    input_time_list.append(additional_time_stamp)
```
spyrostz, line 154: Can you please explain where is this needed ? Should not hurt, just out of curiosity.

src/d3a/constants.py
hannesdiedrich: I meant the integrationtests that test the strategies (correct bidding, energy consumption, energy provision, following the provided profiles). As i added this rotation of profiles, it might be needed.

src/d3a/d3a_core/simulation.py
hannesdiedrich: ah ok, not, will use the `d3a.constants. RUN_IN_REALTIME ` then,

src/d3a/models/read_user_profile.py
hannesdiedrich: I was waiting for this question and i though i understood it, but I apparently not. I am currently investigating why i needed this. Will come back to you

hannesdiedrich: Ah yes, I was right after all (in my head ;-)) It took me a while now to understand again. This function is really noe easy to understand and i remember long discussions last year about it. 
In short: I is because of the `range()` in line 162.
In length: Due to the nature of `range()`, the last market slot will not be part of the `slot_time_list` (which should be renamed to `slot_time_list_seconds`, if you agree, will do here). Consequently, we need to add another time stamp at the end right from the start. 
If you abstract it into formula (which i did in my head), for calculating the energy you need the power and the time duration. Of course, the time duration should always be the slot length here. But the way the function is implemented, it is calculating the time duration on the fly by difference between the time stamps in order to also take profiles that don't follow the market_slot length. So you need an additional sampling point for that. 
Of course this function could be refactored now, as the `_fill_gaps_in_profile` is called before. But frankly i can not be bothered. Too much time was spent by me in this function already only for this ticket. 
Let me know if my lengthly description doesn't make sense.

src/d3a/d3a_core/simulation.py
hannesdiedrich: will be fixed when merging D3ASIM-3004_fix_cn_run.

src/d3a/d3a_core/util.py
```diff
@@ -332,16 +331,22 @@ def update_nested_settings(class_object, class_name, settings_dict):
...
-    if not hasattr(config, 'market_slot_list') or len(config.market_slot_list) == 0:
-        config.market_slot_list = generate_market_slot_list_from_config(
-            config.sim_duration, config.start_date, config.market_count, config.slot_length
-        )
-    return config.market_slot_list
+def generate_market_slot_list(start_date=None, time_span=None):
+    if start_date is None:
+        start_date = GlobalConfig.start_date
+    if time_span is None:
+        time_span = GlobalConfig.sim_duration
```
faizan2590, line 338:
```suggestion
def generate_market_slot_list(start_date=GlobalConfig.start_date, time_span=GlobalConfig.sim_duration):
```

src/d3a/d3a_core/util.py
```diff
@@ -332,16 +331,22 @@ def update_nested_settings(class_object, class_name, settings_dict):
...
+    sim_duration_plus_future_markets = time_span + GlobalConfig.slot_length * \
+        (GlobalConfig.market_count - 1)
```
faizan2590, line 340: Can you help me understand the reason for one subtraction offset here?

src/d3a/d3a_core/util.py
```diff
@@ -332,16 +331,22 @@ def update_nested_settings(class_object, class_name, settings_dict):
...
+    market_slot_list = \
+        generate_market_slot_list_from_config(sim_duration=sim_duration_plus_future_markets,
+                                              start_date=start_date,
+                                              market_count=GlobalConfig.market_count,
+                                              slot_length=GlobalConfig.slot_length)
+
+    if not hasattr(GlobalConfig, 'market_slot_list') or len(GlobalConfig.market_slot_list) == 0:
```
faizan2590, line 347:
```suggestion
    if not getattr(GlobalConfig, 'market_slot_list', []):
```

src/d3a/d3a_core/util.py
```diff
@@ -459,6 +464,11 @@ def create_subdict_or_update(indict, key, subdict):
     return indict
 
 
+def write_default_to_dict(indict, key, default_value):
+    if key not in indict:
+        indict[key] = default_value
```
faizan2590, line 469: could be a candidate to be moved to d3a-interface

src/d3a/d3a_core/util.py
hannesdiedrich: yes, 
if `GlobalConfig.market_count=1`, the last market slot should be: `23:45`
if `GlobalConfig.market_count=2`, the last market slot should be: `00:00` (next day)
ans so on. If you add another slot to `time_span` there will be an offset of one market slot that is not wanted here.

src/d3a/models/read_user_profile.py
```diff
@@ -53,26 +55,20 @@ def _str_to_datetime(time_str, time_format) -> DateTime:
...
+    Returns dictionary with default values for all market slots.
+    :param val: Default value
     """
     if val is None:
         val = 0
+    if IS_CANARY_NETWORK:
+        market_slot_list = \
+            generate_market_slot_list(start_date=today(tz=TIME_ZONE),
+                                      time_span=duration(
+                                          days=CN_PROFILE_EXPANSION_DAYS))
```
faizan2590, line 67: Im wondering wouldn't it be good to implement via skipping datetime/pendulum object to make d3a as light weight as possible for CN?

hannesdiedrich: You mean instead of using datetime/pendulum to use unix-timestamp? 
This will be done in the future probably. But for now, as the profiles are rotated and we are not going into the BC soon, pendulum is too handy to be refactored. Just think about to reimplement something like `weekday()` in unix timestamps.
It is really not that heavy, even if we have 1000 communities (from the memory points of view). For the CPU it really is not big of a deal as we run in real time. 
I hope this answers you question.

src/d3a/models/read_user_profile.py
```diff
@@ -196,8 +197,14 @@ def _fill_gaps_in_profile(input_profile: Dict = None) -> Dict:
...
     for time in out_profile.keys():
-        if time in input_profile.keys():
-            current_val = input_profile[time]
+        if IS_CANARY_NETWORK:
+            temp_val = find_timestamp_of_same_weekday_and_time(input_profile, time,
+                                                               ignore_not_found=True)
+            if temp_val is not None:
+                current_val = temp_val
+        else:
+            if time in input_profile.keys():
```
faizan2590, line 206:
```suggestion
            if time in input_profile:
```

src/d3a/d3a_core/util.py
faizan2590, line 540:
```suggestion
def find_object_of_same_weekday_and_time(indict, time_slot, ignore_not_found=False):
```
Initially I thought this method would return a `timestamp` but later looking into the implementation, realised it either is returning a value or profile having an offset of weekday_and_time

src/d3a/models/strategy/pv.py
```diff
@@ -245,8 +250,13 @@ def _delete_past_state(self):
         for k in to_delete:
             self.state.available_energy_kWh.pop(k, None)
             self.energy_production_forecast_kWh.pop(k, None)
+            del self.offer_update.initial_rate[k]
```
faizan2590, line 253:
```suggestion
            del self.offer_update.initial_rate[market_time_slot]
```
...and following the same convention inside `_delete_past_state `

faizan2590 (dismissed):
Some comments, ping  me to approve once addressed

src/d3a/models/strategy/pv.py
hannesdiedrich: Sorry, i copied this from the other `_delete_past_state` and there a `k` . 
If you like, will refactor this.

faizan2590: Always better to have self-explanatory as well as precise variable name than one letter variable.
..and I didn't mean to point at you, just saw it while reviewing and see a potential enhancement.

src/d3a/constants.py
spyrostz: ATM this should be dealt with. Variables RUN_IN_REALTIME and IS_CANARY_NETWORK should be completely independent, therefore we should be able to run the tests with IS_CANARY_NETWORK=True without having to wait for ages.

src/d3a/models/read_user_profile.py
spyrostz: No that makes sense, thanks for explaining

PR #957, opened 2020-11-08
Feature/d3 asim 3007: Re-adjusting area throughput calculations.

faizan2590:


src/d3a/d3a_core/sim_results/area_throughput_stats.py
```diff
@@ -36,10 +36,14 @@ def update_results(self, area_dict, core_stats, current_market_time_slot_str):
...
+        import_peak_energy_net_kWh = net_peak if net_peak > 0 else 0.
+        export_peak_energy_net_kWh = abs(net_peak) if net_peak < 0 else 0.
         area_results = {
-            "import": {'peak_energy_kWh': imported_peak},
-            "export": {'peak_energy_kWh': exported_peak},
+            "import": {'peak_energy_trade_kWh': imported_peak,
+                       "peak_energy_net_kWh": import_peak_energy_net_kWh},
+            "export": {'peak_energy_trade_kWh': exported_peak,
+                       "peak_energy_net_kWh": export_peak_energy_net_kWh},
             "net_energy_flow": {'peak_energy_kWh': net_peak}
```
hannesdiedrich, line 47: Will the frontend use this too? If yes, please call `round_floats_for_ui` too, Thanks

hannesdiedrich (dismissed):
One comment. LGTM in general. 
I wonder why we need to fill up our DB with dicts that have the same value for each key. IMO  a single value should be sufficient. But unfortunately, this is what is requested. 
Maybe we can reused this later if the peak is a moving average over the peak and not he max/min of the whole simulation.

spyrostz (approved):
LGTM, but I agree with @hannesdiedrich 's concerns about duplicate data. We need to at some point discuss with frontend to delete or refactor some of the endpoints that use duplicate data. Not for now though

src/d3a/d3a_core/sim_results/area_throughput_stats.py
faizan2590: I will wait for Sarah to drop a comment in the ticket.

hannesdiedrich (approved):
LGTM

PR #958, opened 2020-11-09
D3ASIM-3004: Some more fixes towards the running CN, most notable is …

spyrostz:
…the usage of next_market in the endpoint buffer when sending the results to the d3a-web. That way the UI is synced to real time.


faizan2590 (approved):
LGTM

PR #961, opened 2020-11-09
D3ASIM-3005: Sanitzation parameter for strategy update via live event.

faizan2590:


src/d3a/d3a_core/live_events.py
```diff
@@ -50,14 +50,15 @@ def __init__(self, area_uuid, area_params):
     def apply(self, area):
         if area.uuid != self.area_uuid:
             return False
-        self.sanitize_live_event_paramters_for_strategy_update()
         area_type = self.area_params.pop("type", None)
         if area_type is not None:
+            self.sanitize_live_event_paramters_for_strategy_update()
```
spyrostz, line 55: This implies that the number_of_clones and the energy_rate parameters do not get removed if the area is a market and not a strategy. Is this really intended ? Moreover, the case that the "type" == "Area" is not covered (if the logic is the same, the `sanitize_live_event_paramters_for_strategy_update` call should be avoided too).

src/d3a/d3a_core/live_events.py
```diff
@@ -50,14 +50,15 @@ def __init__(self, area_uuid, area_params):
...
             if area.strategy is None:
                 return False
             if area_type == "MarketMaker":
                 area.strategy = MarketMakerStrategy(**self.area_params)
             elif area_type == "InfiniteBus":
+                self.area_params['energy_sell_rate'] = self.area_params.pop('energy_rate', None)
```
spyrostz, line 61: You can also do the conversion on the d3a-web side in order to be more consistent with the other parameters. Leave it here for now, just leave a TODO comment to move it to d3a-web later.

spyrostz (approved):
2 comments, LGTM once addressed

src/d3a/models/strategy/market_maker_strategy.py
```diff
@@ -30,8 +31,9 @@ def __init__(self, energy_rate_profile=None, energy_rate=None, grid_connected=Tr
         if energy_rate_profile is not None:
             energy_rate = read_and_convert_identity_profile_to_float(energy_rate_profile)
         else:
-            energy_rate = ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE \
-                if energy_rate is None else read_and_convert_identity_profile_to_float(energy_rate)
+            energy_rate = read_arbitrary_profile(
+                InputProfileTypes.IDENTITY, ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE
+            ) if energy_rate is None else read_and_convert_identity_profile_to_float(energy_rate)
```
hannesdiedrich, line 36: Hmm, here I am not sure, how about using `read_and_convert_identity_profile_to_float` for both cases?
```suggestion
            energy_rate = read_and_convert_identity_profile_to_float(ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE
            if energy_rate is None else energy_rate)
```

src/d3a/d3a_core/live_events.py
hannesdiedrich: I agree with the latter.

hannesdiedrich (commented):
2 comments, looking into it once addressed.

src/d3a/models/strategy/market_maker_strategy.py
```diff
@@ -30,9 +30,10 @@ def __init__(self, energy_rate_profile=None, energy_rate=None, grid_connected=Tr
         if energy_rate_profile is not None:
             energy_rate = read_and_convert_identity_profile_to_float(energy_rate_profile)
         else:
-            energy_rate = ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE \
+            energy_rate = read_and_convert_identity_profile_to_float(
+                ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE) \
                 if energy_rate is None else read_and_convert_identity_profile_to_float(energy_rate)
```
hannesdiedrich, line 35: Just out of curiosity: doesn't it work to have an if statement in the brackets of `read_and_convert_identity_profile_to_float` ?

src/d3a/models/strategy/market_maker_strategy.py
```diff
@@ -30,9 +30,10 @@ def __init__(self, energy_rate_profile=None, energy_rate=None, grid_connected=Tr
...
-        GlobalConfig.market_maker_rate = energy_rate
+            GlobalConfig.market_maker_rate = energy_rate
```
hannesdiedrich, line 36: Why do you only update the Globalconfig if `energy_rate_profile is None` ?

src/d3a/d3a_core/live_events.py
```diff
@@ -50,14 +50,16 @@ def __init__(self, area_uuid, area_params):
     def apply(self, area):
         if area.uuid != self.area_uuid:
             return False
-        self.sanitize_live_event_paramters_for_strategy_update()
         area_type = self.area_params.pop("type", None)
+        self.sanitize_live_event_paramters()
```
spyrostz, line 54: Since you renamed it, slight typo: `paramters` -> `parameters`

src/d3a/d3a_core/live_events.py
```diff
@@ -50,14 +50,16 @@ def __init__(self, area_uuid, area_params):
...
         if area_type is not None:
             if area.strategy is None:
                 return False
             if area_type == "MarketMaker":
                 area.strategy = MarketMakerStrategy(**self.area_params)
             elif area_type == "InfiniteBus":
+                # After hack to move this parameter casting at the web side
```
hannesdiedrich, line 61: Please prepend 'TODO:' In order to be better visible in Pycharm Thanks!

src/d3a/models/strategy/market_maker_strategy.py
faizan2590: Thanks, tried it, works both way, adapted as per your suggestion

src/d3a/models/strategy/market_maker_strategy.py
faizan2590: Nice catch, was remnant of testing, reverted now to out of if else block

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #963, opened 2020-11-10
D3ASIM-3046: Calling area_reconfigure_event if area_type == Area .

faizan2590:


src/d3a/d3a_core/live_events.py
```diff
@@ -53,6 +53,9 @@ def apply(self, area):
         self.sanitize_live_event_paramters_for_strategy_update()
         area_type = self.area_params.pop("type", None)
         if area_type is not None:
+            if area_type == "Area":
+                area.area_reconfigure_event(**self.area_params)
+                return True
```
spyrostz, line 58:
```suggestion
        if area_type is not None and area_type != "Area":
```

spyrostz (dismissed):
LGTM

src/d3a/d3a_core/live_events.py
```diff
@@ -53,6 +53,9 @@ def apply(self, area):
         self.sanitize_live_event_paramters_for_strategy_update()
         area_type = self.area_params.pop("type", None)
         if area_type is not None:
```
hannesdiedrich: I know it is a quickfix, but please refactor it so that `area.area_reconfigure_event(**self.area_params)` is always called (apart from the case when if area_type == "MarketMaker": or elif area_type == "InfiniteBus":

hannesdiedrich: The area_type should never be None, right?

hannesdiedrich (commented):
2 comments

hannesdiedrich (approved):
LGTM

PR #964, opened 2020-11-11
Fixed the boolean that controls whether the market is in the simulati…

spyrostz:
…on duration in case of the CN where we do not have a duration at all.


faizan2590 (approved):
lgtm

hannesdiedrich (approved):
LGTM

PR #974, opened 2020-11-11
fix_d3a_worker_name

hannesdiedrich:


faizan2590 (approved):

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -159,7 +159,7 @@ def main():
                                          retry_on_timeout=True)):
         Worker(
             [get_simulation_queue_name()],
-            name='simulation.{}.{:%s}'.format(getpid(), now()), log_job_description=False
+            name=f'simulation.{getpid()}.{now().timestamp()}', log_job_description=False
```
spyrostz, line 162: In order to make it a bit more inventive I would recommend to use time() in order to get milliseconds too. That way we can be almost certain that no 2 workers will start concurrently.

hannesdiedrich: `.timestamp()` uses milliseconds too, this was the whole point. :-)

spyrostz: all good, thanks for explaining!

spyrostz (approved):
LGTM

PR #981, opened 2020-11-19
Bug/d3 asim 3059

hannesdiedrich:
if add_days is one week (6), only add one day (happens when start_time is last time_slot on the week and time_slot already in new week)


src/d3a/d3a_core/util.py
```diff
@@ -543,9 +543,10 @@ def convert_area_throughput_kVA_to_kWh(transfer_capacity_kWA, slot_length):
 def find_object_of_same_weekday_and_time(indict, time_slot, ignore_not_found=False):
     if d3a.constants.IS_CANARY_NETWORK:
         start_time = list(indict.keys())[0]
+        add_days = abs(time_slot.weekday() - start_time.weekday())
```
spyrostz, line 546: Please bear with me for a moment in order to better understand it. What is done here is that the requested time_slot weekday is subtracted from the start time weekday. This depends on which weekday is '0' though. What I mean, is that if the start_time is Sunday (weekday == 6) and the time_slot is Friday (weekday == 4), then the add_days is going to be 2 (abs(4 -6) == 2). I am under the impression though that this is wrong, because in reality, from a profile that starts from Sunday, the add_days would need to be 5 (Sunday - Friday timedelta).
Assuming that I am correct above, the most straightforward way to fix this is to do this:
add_days = time_slot.weekday() - start_time.weekday()
if add_days < 0:
   add_days += 7
What do you think ?

spyrostz (commented):
One comment, LGTM once answered or dealt with

src/d3a/d3a_core/util.py
hannesdiedrich: You are totally right. This was already wring when i implemented the first time.
I agree with you proposal and implemented already. It also fixes the issue here. Thanks a lot for that!

spyrostz (approved):
LGTM

faizan2590 (approved):
LGTM

PR #984, opened 2020-11-25
D3ASIM-3073: Deprecated appliance.

hannesdiedrich:


src/d3a/models/base.py
```diff
@@ -25,12 +25,10 @@
 
 
 class AreaBehaviorBase:
-    """Base class used by area behaviour defining classes (`BaseStrategy`, `BaseAppliance`)"""
+    """Base class used by area behaviour defining classes `BaseStrategy`"""
     def __init__(self):
-        # `area` is the area we trade in
```
spyrostz, line 30: I think that these comments are useful, please leave these here. Thanks!

src/d3a/models/state.py
```diff
@@ -27,20 +27,6 @@
 
 StorageSettings = ConstSettings.StorageSettings
 
-# Complex device models should be split in three classes each:
```
spyrostz, line 30: This comment is important to keep here IMO. We already violate this a lot by not assigning enough things to the state.

spyrostz (approved):
Some small comments, LGTM once dealt with.

src/d3a/models/base.py
hannesdiedrich: Agreed! I thought they are just wrong, but no, they are right!. Will re-add them

src/d3a/models/state.py
hannesdiedrich: Ok, keeping in mind our offline discussion i will re-add and slightly change this comment.

spyrostz (dismissed):
Still LGTM

PR #989, opened 2020-11-26
D3ASIM-2993: Deprecated Celltower strategy.

hannesdiedrich:


src/d3a/d3a_core/sim_results/__init__.py
```diff
@@ -1,65 +0,0 @@
-from d3a.d3a_core.util import area_name_from_area_or_iaa_name
```
spyrostz, line 1: Are you really sure that this can be removed ? Was it ported to d3a_interface and forgotten to be deleted from here ?

spyrostz (dismissed):
One comment that needs clarification, LGTM otherwise

src/d3a/d3a_core/sim_results/__init__.py
hannesdiedrich: yes, i double checked, It was ported to the d3a-interface and not removed here.

PR #991, opened 2020-11-26
D3ASIM-2247: Removed 'seller' from Bid class.

hannesdiedrich:


tests/test_market.py
```diff
@@ -168,32 +168,29 @@ def test_market_offer(market, offer):
 
 
 def test_market_bid(market: TwoSidedPayAsBid):
-    bid = market.bid(1, 2, 'bidder', 'seller', 'bidder')
+    bid = market.bid(1, 2, 'bidder', 'bidder')
```
lana-shanghai, line 171: I wonder why is 'bidder' repeated 2 times here?

src/d3a/models/strategy/external_strategy.py
```diff
@@ -171,7 +171,7 @@ def _offer(self, payload):
             assert set(arguments.keys()) == {'price', 'energy'}
             arguments['seller'] = self.area.name
             arguments['seller_origin'] = self.area.name
-        except Exception as e:
+        except Exception:
```
spyrostz, line 174: How about instead of ignoring the exception to add a log that displays the error ? Maybe also the `arguments` would be useful to be printed here.

spyrostz (approved):
LGTM

tests/test_market.py
hannesdiedrich: If you look at the two sided market function `bid` you see that there are 4 positional arguments:
`price: float, energy: float, buyer: str, buyer_origin` hence, the last `bidder` is the buyer_origin.
It is ok that these are equal as this is the initial bit (which was not forwarded yet).

src/d3a/models/strategy/external_strategy.py
hannesdiedrich: yeah, flake complained here. I wanted to be consistent with the `_bid` method. But ok, i will add it to the end of the error log (in both places).

faizan2590 (approved):
LGTM

PR #994, opened 2020-11-27
Feature/d3 asim 2894 pv

spyrostz:


src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -170,10 +170,8 @@ def _specific_row(self, slot, market):
...
         elif isinstance(self.area.strategy, PVStrategy):
-            produced = self.area.strategy.state.available_energy_kWh.get(slot, 0)
-            return [produced,
-                    round(produced -
-                          self.area.strategy.energy_production_forecast_kWh.get(slot, 0), 4),
-                    self.area.strategy.energy_production_forecast_kWh.get(slot, 0)
-                    ]
+            produced = self.area.strategy.state.get_available_energy_kWh(slot, 0.0)
+            forecasted = self.area.strategy.state.get_energy_production_forecast_kWh(slot, 0.0)
+            curtailed = round(produced - forecasted, 4)
```
hannesdiedrich, line 175: minor, but how about renaming this to `residue`?

hannesdiedrich (approved):
One minor comment,
LGTM

src/d3a/models/state.py
```diff
@@ -44,14 +44,55 @@
...
+    def get_energy_production_forecast_kWh(self, time_slot, default_value=None):
+        assert self._energy_production_forecast_kWh[time_slot] >= -FLOATING_POINT_TOLERANCE
+        if default_value is not None:
+            return self._energy_production_forecast_kWh.get(time_slot, default_value)
+        return self._energy_production_forecast_kWh[time_slot]
+
+    def get_available_energy_kWh(self, time_slot, default_value=None):
+        assert self._available_energy_kWh[time_slot] >= -FLOATING_POINT_TOLERANCE
+        if default_value is not None:
+            return self._available_energy_kWh.get(time_slot, default_value)
```
faizan2590, line 59: Here `.get` is used to access value from `_available_energy_kWh ` dictionary for a `time_slot ` and if not found, a  `default_value ` would return BUT two lines above via assert would already fail for cases when the key doesn't exist.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
spyrostz: Well, because specifically for the renewable power plants/devices, the operation of limiting the generating power is called curtailment, and the energy is called curtailed_energy. Residual energy is mostly encountered as a term for loads (and for a very specific circumstance, it refers to the remaining load in a grid after renewable devices have been introduced), and not for other devices. I know that it is not a big deal, but it will be misleading if someone looks at the code later. 

We use this term only in the trade object IIRC, in which IMO it is also not 100% correct. However, since I have not seen a commonly used term in energy markets for that, I am fine with keeping it there.

src/d3a/models/state.py
spyrostz: Agreed! I'll disable the assert in case the caller expects a default_value. Thanks for catching this!

src/d3a/d3a_core/sim_results/file_export_endpoints.py
hannesdiedrich: Ah, ok. Thanks for explaining in detail. I am still getting used to the terms used in electrical grids and renewable power plants. So, of course: Keep it like it is.

faizan2590 (approved):
LGTM

PR #996, opened 2020-11-30
D3ASIM-3080: Fixed Electrolyzer integrationtest; Disabled Alternative…

hannesdiedrich:
… pricing integrationtest; enabled raising of exceptions when running simulation in integrationtests; fixed energyorigin in storage.

### DIFF
--- a/integration_tests/jira_issues_tests.feature
+++ b/integration_tests/jira_issues_tests.feature
@@ -52,6 +52,7 @@ Feature: Jira Issues Tests
      When we run the simulation with setup file jira.d3asim_871 and parameters [24, 60, 60, 1]
      Then there should be no unmatched loads
 
+  @disabled
   Scenario: D3ASIM-874, alternative pricing can buy energy from IAA if there is not enough self-consumption
      Given we have a scenario named jira/d3asim_869
      And d3a is installed
@@ -91,6 +92,7 @@ Feature: Jira Issues Tests
   Scenario: D3ASIM-1535, User should not be able to add area to leaf area
     Given we have a scenario named jira/d3asim_1535
     And d3a is installed
+    And raising exceptions when running the simulation is disabled
     When we run the simulation with setup file jira.d3asim_1535 and parameters [24, 60, 60, 1]
     Then an AreaException is raised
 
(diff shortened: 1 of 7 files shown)


integration_tests/steps/jira_issues_tests.py
```diff
@@ -327,3 +327,8 @@ def check_area_trade_and_bill(context):
     assert isclose(cb.cumulative_bills_results[context.name_uuid_map['Market Maker']]['earned'],
                    0.72)
     assert isclose(cb.cumulative_bills_results[context.name_uuid_map['Load']]['spent_total'], 0.84)
+
+
+@given("raising exceptions when running the simulation is disabled")
+def don_not_raise_exceptions(context):
```
spyrostz, line 333: Trivial, don_not -> do_not

integration_tests/jira_issues_tests.feature
```diff
@@ -52,6 +52,7 @@ Feature: Jira Issues Tests
      When we run the simulation with setup file jira.d3asim_871 and parameters [24, 60, 60, 1]
      Then there should be no unmatched loads
 
+  @disabled
```
spyrostz, line 55: Sorry for nitpicking, and it is really not that important since we probably will not need alternative pricing in the future, however why was this test disabled ?

spyrostz (commented):
One question, LGTM once discussed.

integration_tests/jira_issues_tests.feature
hannesdiedrich: @spyrostz There is a problem with the alternative pricing in the storage strategy in the one-sided market.
This is why i disabled this integrationtests: https://github.com/gridsingularity/d3a/blob/b49516f7ba99c9c4ca0f8b185d063698cb40f3f1/integration_tests/alternative_pricing.feature#L9 already some weeks ago.
I tried to fix it but gave up after some time because as you said: probably not needed any more.
The changes i did in the storage strategy originate from the try of fixing.

spyrostz: Thanks for clarifying this, it indeed looks like a waste of time and we should fix it if we plan to revamp alternative pricing.

spyrostz (approved):
LGTM

faizan2590 (approved):
LGTM

PR #1000, opened 2020-12-09
D3ASIM-3109: remove Solium linter and Ganache-CLI

fievelk:
Solium linter and Ganache were used when Solidity was used for smart contracts. As we deprecated Solidity (7fedb36), we should also get rid of these dependencies in tox, travis and in the requirements.

# Contents

- Remove solium configuration files
- remove fabric checks for solium
- remove reference to (unused) contracts folder from `setup.py`

# References

- [Add 'solium' solidity linter](8b060bac9859f6efc5daf4cbf7f3053c4dd689e6)
- [D3ASIM-2910: remove solidity installs from travis script](7fedb36)


fievelk:
Some notes: 
1. I deleted the `.support/solium_checker.py` module. One comment in it stated that the file would have been installed by fabric to `.git/hooks/checkers`. However, the `checkers` directory doesn't exist anymore, so I thought it would be safe to remove it. Let me know if I should restore it instead.

2. I think `captainhook` is not used anymore (see c4d094e5) but there are a lot of references to it. 
For example: https://github.com/gridsingularity/d3a/blob/f2baa06fcf05d7834fec841ab60b379df8b43a0e/setup.cfg#L16-L21 
What to do with it?

Hassan754 (dismissed):

hannesdiedrich (commented):
Thanks a lot for this PR!
I have a general question to @spyrostz @faizan2590 @lana-shanghai 
As we don't support solidity smart contracts any more, does it make sense to keep the ganache dependency?

fabfile.py
```diff
@@ -102,7 +83,6 @@ def _pre_check():
 
 def _post_check():
     if "Darwin" in system():
-        _ensure_solium()
         _ensure_ganache_cli()
```
spyrostz, line 86: Not part of this story, but it would be great if you also deleted the ganache references from here, thanks!

spyrostz (dismissed):
LGTM

faizan2590:
> Thanks a lot for this PR!
> I have a general question to @spyrostz @faizan2590 @lana-shanghai
> As we don't support solidity smart contracts any more, does it make sense to keep the ganache dependency?

As mentioned by @spyrostz in other [comment](https://github.com/gridsingularity/d3a/pull/1000#discussion_r539161527) too, would be good be to get rid of ganache.

faizan2590 (dismissed):
LGTM

fabfile.py
fievelk: I just pushed a new commit to remove Ganache 👍 :)

Hassan754 (approved):

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #1004, opened 2020-12-14
D3ASIM-3101: Adding simulation_area_market info while instantiating t…

faizan2590:
…rade contract.

### DIFF
--- a/src/d3a/blockchain/contracts/ink/lib.rs
+++ b/src/d3a/blockchain/contracts/ink/lib.rs
@@ -27,6 +27,7 @@ mod trade_storage {
     pub struct TradeStorage {
         /// The trades stored in the contract.
         trades: StorageHashMap<Hash, (AccountId, AccountId, Balance, i8)>,
+        area_market_id: StorageHashMap<u128, (u128, u32)>,
     }
 
     /// Storing a trade by a call from d3a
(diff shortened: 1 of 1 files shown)


src/d3a/blockchain/contracts/ink/lib.rs
```diff
@@ -27,6 +27,7 @@ mod trade_storage {
     pub struct TradeStorage {
         /// The trades stored in the contract.
         trades: StorageHashMap<Hash, (AccountId, AccountId, Balance, i8)>,
+        area_market_id: StorageHashMap<u128, (u128, u32)>,
```
spyrostz, line 30: Trivial, but I think this is better named as simulation_market_id or simulation_area_id, and stick to either using the term `market` or `area` for the whole smart contract (I do not mind which one of the 2, the term has to be consistent though).

spyrostz (approved):
One comment, once resolved LGTM

src/d3a/blockchain/contracts/ink/lib.rs
lana-shanghai: Possibly better to use DoubleStorage instead of Storage, as Storage is one-to-one kay-value mapping. https://substrate.dev/docs/en/knowledgebase/runtime/storage

faizan2590: As discussed offline `DoubleStorage ` is only for runtime storage and not for ink, I will close this  PR now and will look into the possibility of either importing runtime trade storage in ink or to implement via pallet contract.

faizan2590:
https://github.com/gridsingularity/d3a/pull/1004#discussion_r546585983

PR #1006, opened 2020-12-28
D3ASIM-3140: migration to github actions

mclrch:
Migration to github actions


hannesdiedrich (approved):
LGTM

.github/workflows/d3a_ci.yaml
```diff
@@ -0,0 +1,21 @@
...
+    - name: Set up Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      env:
+        TOXENV: travis
+      run: |
+        bash .support/travis-install.sh
+        bash .support/travis-run.sh
```
hannesdiedrich, line 21: I think it would be a good opportunity to combine `travis-install.sh` and `travis-run.sh` into one script called `run_ci.sh` (or similar) that only consists of the following code:
```
#!/usr/bin/env bash
pip install -U tox
tox -- --verbose
```
We could also rename the TOXENV to `github_ci` or something like that.
I am slo open to tackle this in another small story.

hannesdiedrich (approved):
One enhancement comment, but already LGTM

.github/workflows/d3a_ci.yaml
spyrostz: I also wrote a similar comment in the d3a-interface PR. I agree and I would actually push it even further and delete these files altogether, since tox is already installed as a dependency in requirements/tests.in

faizan2590 (dismissed):
LGTM

.github/workflows/d3a_ci.yaml
mclrch: changed to "ci"
thanks

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #1009, opened 2021-01-04
D3ASIM-3145: Catching exception if there is no external implementatio…

hannesdiedrich:
…n of strategy.

### DIFF
--- a/src/d3a/models/leaves.py
+++ b/src/d3a/models/leaves.py
@@ -15,6 +15,7 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+import logging
 from d3a.models.area import Area
 from d3a.models.strategy.commercial_producer import CommercialStrategy
 from d3a.models.strategy.market_maker_strategy import MarketMakerStrategy
@@ -48,6 +49,8 @@ forecast_strategy_mapping = {
     LoadHoursStrategy: LoadForecastExternalStrategy
 }
 
+log = logging.getLogger()
+
 
 class Leaf(Area):
     """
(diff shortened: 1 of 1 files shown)


Hassan754 (approved):

src/d3a/models/leaves.py
```diff
@@ -58,9 +61,17 @@ class Leaf(Area):
 
     def __init__(self, name, config=None, uuid=None, **kwargs):
         if kwargs.get("forecast_stream_enabled", False) is True:
-            self.strategy_type = forecast_strategy_mapping[self.strategy_type]
+            try:
+                self.strategy_type = forecast_strategy_mapping[self.strategy_type]
+            except KeyError:
+                log.error(f"ERROR: {self.strategy_type} could not be found in "
```
spyrostz, line 67: Minor, but you can skip the uppercase ERROR here.

Hassan754 (dismissed):

spyrostz (dismissed):
LGTM, but I think that I am late to the party on this one. If this is only for accomodating the market makers and other devices that do not have external connections, it would be much more elegant IMO to handle this in the d3a-web (either in the context of configuration_objects.py, or right before the data are sent to d3a in the db_to_d3a.py module)
However, this PR is ok and I agree that there is no harm in d3a to be flexible in the types it accepts, hence LGTM

hannesdiedrich:
> LGTM, but I think that I am late to the party on this one. If this is only for accomodating the market makers and other devices that do not have external connections, it would be much more elegant IMO to handle this in the d3a-web (either in the context of configuration_objects.py, or right before the data are sent to d3a in the db_to_d3a.py module)
> However, this PR is ok and I agree that there is no harm in d3a to be flexible in the types it accepts, hence LGTM

I agree with you @spyrostz and created the following task https://github.com/gridsingularity/d3a/pull/1009#pullrequestreview-561600999 and marked it as an onboarding task. 
I hope you agree. :-)

faizan2590 (approved):
LGTM

PR #1011, opened 2021-01-05
Feature/d3 asim 3084

hannesdiedrich:
Unfortunately, it is only kind of working. 
The following runtimes appear with the corresponding settings:
(slot_length = 15m)

**slot_length_realtime |  tick_length |  what it really took**
60s                             |  1s                 |  62.11s
60s                             |  15s               |  60.17s
120s                           |  1s                 |  122.2s
120s                           |  15s               |  120.19s

Apparently, there is an overhead in each tick that I didn't catch. But i really can not find it.
@spyrostz I remember that we had something similar when you implemented the realtime running for the CN, but i though We fixed that by adding the `realtime_tick_length_s` and subtracting it from the sleep time.

I am happy about any feedback. Thanks!


src/d3a/d3a_core/simulation.py
```diff
@@ -375,13 +381,15 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
                 self.simulation_config.external_redis_communicator.\
                     publish_aggregator_commands_responses_events()
 
-                realtime_tick_length = time.time() - simulation_time_counter
```
spyrostz, line 378: Please move all these to their own private method, since all these are related to the slowdown management.

spyrostz:
> Unfortunately, it is only kind of working.
> The following runtimes appear with the corresponding settings:
> (slot_length = 15m)
> 
> **slot_length_realtime | tick_length | what it really took**
> 60s | 1s | 62.11s
> 60s | 15s | 60.17s
> 120s | 1s | 122.2s
> 120s | 15s | 120.19s
> 
> Apparently, there is an overhead in each tick that I didn't catch. But i really can not find it.
> @spyrostz I remember that we had something similar when you implemented the realtime running for the CN, but i though We fixed that by adding the `realtime_tick_length_s` and subtracting it from the sleep time.
> 
> I am happy about any feedback. Thanks!

I suspect that this deviation is due to the fact that only the tick time is measured, and not the market_cycle time. I would first try to measure time from the beginning of the market cycle by moving this line https://github.com/gridsingularity/d3a/pull/1011/files#diff-00472b9024eb273db74d246aa396bcbb49a369f86334c3413c71a505a4994036R360 so that the market cycle is incorporated. 
Also, this does not take into account the area activation time, but I think it is a good thing to have.

spyrostz:
One general comment: I recall from discussions some time ago that we would use one of the 2 methods in d3a (either a slowdown percentage or a slowdown time) and convert to one another in d3a-web. That way we would not have the same method implemented twice. Despite that, in this PR the slowdown time was just added on top of the slowdown percentage. Is it something that we cannot live without? Because if not, having extra parameters unnecessary bloats the code (especially after looking at the d3a-interface PR)

src/d3a/d3a_core/simulation.py
```diff
@@ -375,13 +381,15 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
+                realtime_tick_length_s = time.time() - simulation_time_counter
                 if d3a.constants.RUN_IN_REALTIME:
-                    sleep(abs(tick_lengths_s - realtime_tick_length))
-                elif self.slowdown and realtime_tick_length < tick_lengths_s:
+                    sleep(abs(tick_lengths_s - realtime_tick_length_s))
+                elif self.slot_length_realtime:
+                    sleep(abs(tick_length_realtime_s - realtime_tick_length_s))
```
faizan2590, line 388: It gets me confusing with  `tick_length_realtime_s ` and `realtime_tick_length_s ` with two closely named variable and both having `realtime` in it

faizan2590 (commented):
Stopping the review as the closely related variable names confusing me

hannesdiedrich:
> One general comment: I recall from discussions some time ago that we would use one of the 2 methods in d3a (either a slowdown percentage or a slowdown time) and convert to one another in d3a-web. That way we would not have the same method implemented twice. Despite that, in this PR the slowdown time was just added on top of the slowdown percentage. Is it something that we cannot live without? Because if not, having extra parameters unnecessary bloats the code (especially after looking at the d3a-interface PR)

*sight I didnt read the story title properly. It clearly says "replace". I double check with Chris and then adapt accordingly. Thanks!

hannesdiedrich:
@spyrostz now I deprecated slowdown and tried to adapt for the time that the startup pf the slot takes.
But there is still an overhead that is not accounted for (which i dont understand) that is increasing with ticks per slot (as before):
`d3a -l WARN run --setup default_2a --slot-length-realtime 1m -t 60s -s 60m`  (60 ticks per slot) --> runtime: 24.05 minutes
`d3a -l WARN run --setup default_2a --slot-length-realtime 1m -t 1s -s 60m` (3600 ticks per slot) --> runtime: 26.9 minutes
Which results in an overhead of around 0.002 seconds per tick
Do you have an idea?

hannesdiedrich:
@spyrostz does he/she have a point? https://stackoverflow.com/a/11657786/6835791

spyrostz:
> @spyrostz does he/she have a point? https://stackoverflow.com/a/11657786/6835791

He does have a point, however a deviation of 2.9 minutes is way too much to blame the lack of a realtime OS. If you try to find out how much deviation is per tick, you will see that there is a 50ms deviation on every tick. This is way too much to blame on the lack of a realtime OS IMO.

hannesdiedrich:
> > @spyrostz does he/she have a point? https://stackoverflow.com/a/11657786/6835791
> 
> He does have a point, however a deviation of 2.9 minutes is way too much to blame the lack of a realtime OS. If you try to find out how much deviation is per tick, you will see that there is a 50ms deviation on every tick, which is way too much IMO

Actually, the deviation per tick is 0.002 seconds (170s/3600ticks_per_slot/24_slots). So it is in the same magnitude, right?

hannesdiedrich:
I gave it another 10 looks and dont see a point were we forget to account for runtime. In case your research reveals the same, I propose to add a fixed offset (of 0.002s) to the sleep to account for this.
In my tests this offset was always in this magnitude per tick, independant of the slot and tick length or slot_length_realtime. 
What do you think?

spyrostz:
> > > @spyrostz does he/she have a point? https://stackoverflow.com/a/11657786/6835791
> > 
> > 
> > He does have a point, however a deviation of 2.9 minutes is way too much to blame the lack of a realtime OS. If you try to find out how much deviation is per tick, you will see that there is a 50ms deviation on every tick, which is way too much IMO
> 
> Actually, the deviation per tick is 0.002 seconds (170s/3600ticks_per_slot/24_slots). So it is in the same magnitude, right?

Darn you are correct, forgot the extra division by 24. This could be it then yes, 2 ms are a bit high to blame on this, nevertheless it is possible (especially since I can not really see any other place where we could have missed). 
There is a solution for that, to amortize this deviation in the next tick / slot (by shortening the sleep time of the following ticks). However this will make the implementation a bit more complex without a lot of benefit.

Hassan754 (approved):

faizan2590 (dismissed):
LGTM

src/d3a/d3a_core/simulation.py
```diff
@@ -326,18 +325,24 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
-            log.warning(
-                "Slot %d of %d (%2.0f%%) - %s elapsed, ETA: %s",
-                slot_no + 1,
-                slot_count,
-                self.progress_info.percentage_completed,
-                self.progress_info.elapsed_time,
-                self.progress_info.eta
-            )
+            slot_log_str = f"Slot {slot_no + 1} of {slot_count} - " \
+                           f"({round(self.progress_info.percentage_completed, 1)}%) " \
```
spyrostz, line 339: Should be exactly the same as the one before, right ? Specifically for the floating point part, you can avoid the `round` by doing the following:
`f"({self.progress_info.percentage_completed:.1f}%) " \`
assuming you want one decimal point.

src/d3a/d3a_core/simulation.py
```diff
@@ -326,18 +325,24 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
+                           f"{self.progress_info.elapsed_time} elapsed, " \
+                           f"ETA: {self.progress_info.eta}"
+            if self.slot_length_realtime:
+                slot_log_str += f" (slowdown applied: {slow_down_applied})"
```
spyrostz, line 343: I am not sure that just one boolean flag, which will be the same for all slots and printed for all of them, is that helpful for debugging. IMO it is not needed here

src/d3a/d3a_core/simulation.py
```diff
@@ -351,20 +356,15 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
                 tick_resume = 0
-                log.trace(
-                    "Tick %d of %d in slot %d (%2.0f%%)",
-                    tick_no + 1,
-                    config.ticks_per_slot,
-                    slot_no + 1,
-                    (tick_no + 1) / config.ticks_per_slot * 100,
-                )
+                log.trace(f"Tick {tick_no + 1} of {config.ticks_per_slot} in slot "
+                          f"{slot_no + 1} ({(tick_no + 1) / config.ticks_per_slot * 100})")
```
spyrostz, line 367: Trivial since it is a trace log, but in order to mimic the same behavior you need to limit the floating point precision.

src/d3a/d3a_core/simulation.py
spyrostz: I still support that btw, all these could be a small private method that would handle the delays in case of either slowdown or realtime.

src/d3a/d3a_core/simulation.py
```diff
@@ -375,21 +375,20 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
-                        self._handle_input(console, diff_slowdown)
+                    tick_runtime_s = time() - tick_time_counter
+                    sleep(abs(tick_length_s - tick_runtime_s))
+                elif self.slot_length_realtime:
+                    current_expected_time = \
+                        current_expected_time.add(seconds=tick_length_realtime_s)
+                    sleep_time = current_expected_time.timestamp() - now().timestamp()
+                    if sleep_time > 0:
+                        slow_down_applied = True
+                        sleep(sleep_time)
```
spyrostz, line 387: I think it would be great to have a debug log here, that would denote the amount of sleep time that was forced to be used in order to sync with the desired time.

spyrostz (dismissed):
Some minor comments, but the algorithm LGTM.

src/d3a/d3a_core/simulation.py
hannesdiedrich: ok! Agreed, I would omit the flag than (https://github.com/gridsingularity/d3a/pull/1011#discussion_r559716207)
OK?

spyrostz: I'd very much prefer that, thanks!

spyrostz (approved):
LGTM

PR #1012, opened 2021-01-05
D3ASIM-2829: Adding current market fee for dso_market_stats

faizan2590:


Hassan754 (dismissed):

src/d3a/models/area/stats.py
```diff
@@ -132,6 +132,8 @@ def get_last_market_stats(self, dso=False):
             self.current_market.time_slot, default_trade_stats_dict))
         out_dict[current_market_time_slot_str]["market_bill"] = \
             self._get_current_market_bills()
+        out_dict[current_market_time_slot_str]["market_fee"] = \
```
hannesdiedrich, line 135: In order to not get confused with the grid_fee setting, i suggest to rename this to something like:
`earned_market_fee` or `market_fee_revenue`

faizan2590: Hi @chrisd12 , would you mind suggesting which one those names would be better in terms of API simplification, `market_fee , earned_market_fee  or market_fee_revenue `?

chrisd12: @faizan2590 I think **market_fee_revenue** is the most comprehensive name for API users. One question : is it the cumulative revenue over the simulation or the exact revenue of that specific market slot ?

faizan2590: @chrisd12 Its not the cumulative revenue but the exact revenue during that specific market slot.

hannesdiedrich (approved):
LGTM

src/d3a/models/area/stats.py
chrisd12: @faizan2590 ok great perfect !

Hassan754 (approved):

PR #1016, opened 2021-01-12
D3ASIM-3206: Fixed conditions to restore a simulation from state. Sho…

spyrostz:
…uld restore only if there are area data, and the simulation status is running or paused. Added explicit exception on the PV strategy in order to capture cases where the PredefinedPV strategy or its subclasses are instantiated without a power profile.

### DIFF
--- a/src/d3a/d3a_core/simulation.py
+++ b/src/d3a/d3a_core/simulation.py
@@ -622,8 +622,8 @@ def run_simulation(setup_module_name="", simulation_config=None, simulation_even
         raise click.BadOptionUsage(ex.args[0])
 
     if saved_sim_state is not None and \
-        (saved_sim_state["areas"] != {} or
-         saved_sim_state["general"]["sim_status"] != "initializing"):
+        (saved_sim_state["areas"] != {} and
+         saved_sim_state["general"]["sim_status"] in ["running", "paused"]):
         simulation.restore_global_state(saved_sim_state["general"])
         simulation.restore_area_state(saved_sim_state["areas"])
         simulation.run(initial_slot=saved_sim_state["general"]["slot_number"])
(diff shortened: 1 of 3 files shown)


src/d3a/d3a_core/simulation.py
```diff
@@ -622,8 +622,8 @@ def run_simulation(setup_module_name="", simulation_config=None, simulation_even
         raise click.BadOptionUsage(ex.args[0])
 
     if saved_sim_state is not None and \
-        (saved_sim_state["areas"] != {} or
-         saved_sim_state["general"]["sim_status"] != "initializing"):
+        (saved_sim_state["areas"] != {} and
+         saved_sim_state["general"]["sim_status"] in ["running", "paused"]):
```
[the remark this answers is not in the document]
spyrostz: indeed, thanks!

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -92,6 +93,10 @@ def set_produced_energy_forecast_kWh_future_markets(self, reconfigure=True):
             self._read_predefined_profile_for_pv()
         for market in self.area.all_markets:
             slot_time = market.time_slot
+            if not self.power_profile:
+                raise D3AException(
+                    f"PV {self.owner.name} tries to set its energy forecast without a "
+                    f"power profile.")
```
hannesdiedrich, line 99: Hmm, i wonder when this case can happen. Only when `read_arbitrary_profile` returns None, in line 124, right?

spyrostz: Either when it returns None or empty dict, yes. Also, if `area_reconfigure_event` is called with an empty dict, which was the case that we saw in the traceback.

hannesdiedrich: I see, then I propose to add this check in the predefined load, too:
https://github.com/gridsingularity/d3a/blob/e49b04bfcb4ddb73782b8defdb4598a8ee0d8f10/src/d3a/models/strategy/predefined_load.py#L108

spyrostz: Agreed, adding it

faizan2590 (approved):

Hassan754 (approved):

hannesdiedrich (approved):
LGTM

PR #1017, opened 2021-01-13
D3ASIM-2845 Prevent Duplicate area naming in the same Area

Hassan754:
### Description
This is part 2 of the restriction to have duplicate area names inside the same parent area
### References
[D3A web side](https://github.com/gridsingularity/d3a-web/pull/1122)

### DIFF
--- a/src/d3a/models/area/__init__.py
+++ b/src/d3a/models/area/__init__.py
@@ -117,7 +117,7 @@ class Area:
         self.uuid = uuid if uuid is not None else str(uuid4())
         self.slug = slugify(name, to_lower=True)
         self.parent = None
-        self.__children = AreaChildrenList(self, children) if children is not None\
+        self.children = AreaChildrenList(self, children) if children is not None\
             else AreaChildrenList(self)
         for child in self.children:
             child.parent = self
@@ -142,18 +142,6 @@ class Area:
         self.redis_ext_conn = RedisMarketExternalConnection(self) \
             if external_connection_available is True else None
         self.should_update_child_strategies = False
-
-    @property
-    def children(self):
-        return self.__children
-
-    @children.setter
-    def children(self, new_children: AreaChildrenList):
-        if type(new_children) == list:
-            self.__children = AreaChildrenList(self, new_children)
-        elif type(new_children) == AreaChildrenList:
-            new_children.parent_area = self
-            self.__children = new_children
 
     @property
     def name(self):
--- a/src/d3a/models/area/__init__.py
+++ b/src/d3a/models/area/__init__.py
@@ -59,6 +59,39 @@ DEFAULT_CONFIG = SimulationConfig(
 )
 
 
+def check_area_name_exists_in_parent_area(parent_area, name):
+    """
+    Check the children of parent of area , iterate through its children and
+        check if the name to be appended does not exist
+    Note: this check is to be called before adding a new area of changing its name
+    :param parent_area: Parent Area
+    :param name: New name of area
+    :return: boolean
+    """
+    for child in parent_area.children:
+        if child.name == name:
+            return True
+    return False
+
+
+class AreaChildrenList(list):
+    def __init__(self, parent_area, *args, **kwargs):
+        self.parent_area = parent_area
+        super(AreaChildrenList, self).__init__(*args, **kwargs)
+
+    def _validate_before_insertion(self, item):
+        if check_area_name_exists_in_parent_area(self.parent_area, item.name):
+            raise AreaException("Area name should be unique inside the same Parent Area")
+
+    def append(self, item: "Area") -> None:
+        self._validate_before_insertion(item)
+        super(AreaChildrenList, self).append(item)
+
+    def insert(self, index, item):
+        self._validate_before_insertion(item)
+        super(AreaChildrenList, self).insert(index, item)
+
+
 class Area:
 
     def __init__(self, name: str = None, children: List["Area"] = None,
@@ -79,12 +112,13 @@ class Area:
         self.active = False
         self.log = TaggedLogWrapper(log, name)
         self.current_tick = 0
-        self.name = name
+        self.__name = name
         self.throughput = throughput
         self.uuid = uuid if uuid is not None else str(uuid4())
         self.slug = slugify(name, to_lower=True)
         self.parent = None
-        self.children = children if children is not None else []
+        self.__children = AreaChildrenList(self, children) if children is not None\
+            else AreaChildrenList(self)
         for child in self.children:
             child.parent = self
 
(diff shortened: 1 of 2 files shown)


src/d3a/models/area/__init__.py
```diff
@@ -109,6 +143,29 @@ def __init__(self, name: str = None, children: List["Area"] = None,
...
+    @property
+    def name(self):
+        return self.__name
+
+    @name.setter
+    def name(self, new_name):
+        if not check_area_name_exists_in_parent_area(self.parent, new_name):
+            self.__name = new_name
+        else:
+            raise AreaException("Area name should be unique inside the same Parent Area")
```
hannesdiedrich, line 167: Minor comment: AFAIK we dont plan to have an area_renaming functionality, but leave it

src/d3a/models/area/__init__.py
```diff
@@ -109,6 +143,29 @@ def __init__(self, name: str = None, children: List["Area"] = None,
...
+    def children(self):
+        return self.__children
+
+    @children.setter
+    def children(self, new_children: AreaChildrenList):
+        if type(new_children) == list:
+            self.__children = AreaChildrenList(self, new_children)
+        elif type(new_children) == AreaChildrenList:
+            new_children.parent_area = self
+            self.__children = new_children
```
hannesdiedrich, line 156: you probably mean:
```suggestion
            self.children.append(new_children)
```
right?

Hassan754: @hannesdiedrich I made the children attribute protected `__children` , so I'm setting it in here .

hannesdiedrich: No wait, I a confused. Why do you need the whole block L146-156?

Hassan754: This is the case when we have another list of children and we set it to a different parent , I had to set their parent area in here

hannesdiedrich (commented):
Left a comment, and here a general comment:
Looks to me like AreaChildrenList is bit of an overkill for this, but maybe you can convince me that this is really needed.

src/d3a/models/area/__init__.py
Hassan754: Let's suppose a parent A has Children A , if we had parent B and we set parent B.children = parentA.children , here I'm taking into consideration that we should change the parent of the list

Hassan754:
> Left a comment, and here a general comment:
> Looks to me like AreaChildrenList is bit of an overkill for this, but maybe you can convince me that this is really needed.

I think this is required as I'm overriding the append and insert methods , in a normal list you cannot check for duplication before appending .

src/d3a/models/area/__init__.py
hannesdiedrich: IMO this is not a case we will ever have. We only change children of parents, never parents of children. 
(you dont Houses from Community to Community). But maybe the others can comment on that.

faizan2590: Thinking from the electrical network perspective, it indeed is a possibility to `change parent of children`. Grid Operator could do so to balance two neighbouring radial feeder or to balance load in between transformers or simply to switch tail-end houses to other feeders because they were experience poor voltage regulation.

TLDR: Its indeed a possibility but I wouldn't be in favour of incorporating the targeted code changes so silently and would make a ticket as well as discuss with research team if any immediate need to do so.

faizan2590 (commented):
Left a comment, ping me to review once resolved.

src/d3a/models/area/__init__.py
Hassan754: @faizan2590 right , I'm open to opening a new ticket for this matter
@spyrostz can we have your opinion on this

Hassan754: @faizan2590 I created a ticket and assigned it to the research team https://gridsingularity.atlassian.net/browse/D3ASIM-3229

Hassan754: @faizan2590 https://gridsingularity.atlassian.net/browse/D3ASIM-3229?focusedCommentId=21143

hannesdiedrich: @Hassan754 AFAIK, protected variables are only prepended with one underscore, not with two, right?

Hassan754: > The double underscore “__” (called “dunder“) is used to make an instance attribute or method private (cannot be accessed from outside the class) — when used as leading dunder. When used as enclosing dunder (e.g. “__init__”) it indicates that it is a special method in Python (called “magic method”).

spyrostz: Interesting approach in general, but there are 2 major problems with this:
1. there is no need to keep track of the parent in the children list. Each Area instance has a parent member, which already keeps a reference to the parent of this area (it is None for the root of the tree)
2. I highly doubt that we are ever going to move areas from one parent to another, because this might introduce cycles in the tree, and the whole code operates under the assumption that the graph is acyclic (tree is by definition acyclic). If we introduce a graph with cycles, I am certain that multiple parts of d3a will not work (first and foremost, offer/bid forwarding will endlessly forward the same offers and bids in the graph cycle), therefore if we ever need to implement this, we should do this much more carefully. 
For now I personally do not see a strong reason to implement it, even from the electricity network perspective, it will be much easier and far less error prone to destroy and recreate the area objects rather than swap their references. Therefore I would recommend to delete this property altogether.

src/d3a/models/area/__init__.py
```diff
@@ -59,6 +59,39 @@
...
+    :param name: New name of area
+    :return: boolean
+    """
+    for child in parent_area.children:
+        if child.name == name:
+            return True
+    return False
+
+
+class AreaChildrenList(list):
```
spyrostz, line 77: I am always intimidated when extending the default Python types, because of any potential side effects that these might have. I think a much better approach would be a validator on the constructor of the class, which already has the `children` member populated. IMO this will be much simpler and less error-prone than a new datatype.

Hassan754: @spyrostz The problem I am trying to address is the `append` or `insert` or changing a list's parent ( We can eliminate this option for now ) . So for the formers I don't see any other fix

src/d3a/models/area/__init__.py
Hassan754: @spyrostz Removed the property

src/d3a/models/area/__init__.py
spyrostz: Have a look at the Area class constructor. The children constructor argument is a list of children for this area. Therefore, you can easily scan the list names for duplicates directly in the constructor of the area class, and raise the exception there. There is no need to override the insert / append of the list itself.

Hassan754: @spyrostz This is only when creating the Area object , we initialize its children , and when we later do `area.children.append(child)` no constructor call is being made

src/d3a/models/area/__init__.py
```diff
@@ -79,12 +112,13 @@ def __init__(self, name: str = None, children: List["Area"] = None,
...
         self.current_tick = 0
-        self.name = name
+        self.__name = name
         self.throughput = throughput
         self.uuid = uuid if uuid is not None else str(uuid4())
         self.slug = slugify(name, to_lower=True)
         self.parent = None
-        self.children = children if children is not None else []
+        self.children = AreaChildrenList(self, children) if children is not None\
+            else AreaChildrenList(self)
```
spyrostz, line 121:
```suggestion
        child_names = [c.name for c in children]
        if len(set(child_names)) !=len(child_names):
            raise Exception("Duplicate names in children")
        self.children = children
```
Even better, you can abstract it away on a validation function in order to not have a lot of things in the constructor body. IMO it will be way cleaner than overriding the Python list class.

Hassan754: This makes sense to me in the creation case , but again if we try to append a child later , how can we validate the uniqueness of name

spyrostz: OK now I understand the rationale behind this, thanks for bearing with me. The validation could be ported to the activate() method of the Area class (which already includes references to the parent) however this will be more messy than the solution you have implemented (because activate will also contain validation code which is irrelevant to the strategy activation). Therefore LGTM, leave it as it is.

spyrostz (approved):
LGTM

faizan2590 (approved):

PR #1021, opened 2021-01-19
D3ASIM-2465: Enabling nested area to be added via live_event.

faizan2590:


src/d3a/models/area/event_dispatcher.py
```diff
@@ -79,6 +79,7 @@ def _broadcast_notification(self, event_type: Union[MarketEvent, AreaEvent], **k
         # Broadcast to children in random order to ensure fairness
         for child in sorted(self.area.children, key=lambda _: random()):
             child.dispatcher.event_listener(event_type, **kwargs)
+        kwargs.pop('current_tick', None)
```
spyrostz, line 82: I understand the need to pass the current_tick when broadcasting activate, so that the child areas can immediately sync with their root parent area. However, does it really hurt to remove it from here ? I think it would be fine to leave it as it is.

src/d3a/models/area/event_dispatcher.py
```diff
@@ -108,13 +109,14 @@ def _should_dispatch_to_strategies(self, event_type):
             return self.area.events.is_connected and self.area.events.is_enabled
 
     def event_listener(self, event_type: Union[MarketEvent, AreaEvent], **kwargs):
+        current_tick = kwargs.pop('current_tick', None)
```
spyrostz, line 112: Same as above, I do not think you really need this.

src/d3a/models/area/event_dispatcher.py
```diff
@@ -108,13 +109,14 @@ def _should_dispatch_to_strategies(self, event_type):
...
         if event_type is AreaEvent.TICK and \
                 self._should_dispatch_to_strategies(event_type):
             self.area.tick_and_dispatch()
         if event_type is AreaEvent.MARKET_CYCLE:
             self.area.cycle_markets(_trigger_event=True)
         elif event_type is AreaEvent.ACTIVATE:
-            self.area.activate()
+            self.area.activate(current_tick=current_tick)
```
spyrostz, line 119: I think you can pass kwargs here instead of the current_tick. Assuming that it works, it will make it more future-proof because it will work if we ever need to add more arguments to the broadcast_activate method.

src/d3a/models/state.py
```diff
@@ -48,16 +48,16 @@ def __init__(self):
         self._energy_production_forecast_kWh = {}
 
     def get_energy_production_forecast_kWh(self, time_slot, default_value=None):
-        assert self._energy_production_forecast_kWh[time_slot] >= -FLOATING_POINT_TOLERANCE
+        assert self._energy_production_forecast_kWh.get(time_slot, 0.) >= -FLOATING_POINT_TOLERANCE
```
spyrostz, line 51: Can you elaborate a bit on why this is needed ? Reason why I am asking is because this might mask some real problems with the code.

spyrostz (approved):
Some small comments, LGTM once resolved

src/d3a/models/area/event_dispatcher.py
faizan2590: I did it else i will get exception of unexpected keyword argument
`TypeError: event_activate() got an unexpected keyword argument 'current_tick'`

src/d3a/models/area/event_dispatcher.py
faizan2590: https://github.com/gridsingularity/d3a/pull/1021#discussion_r562698227

src/d3a/models/area/event_dispatcher.py
faizan2590: https://github.com/gridsingularity/d3a/pull/1021#discussion_r562698227

src/d3a/models/state.py
faizan2590: https://github.com/gridsingularity/d3a/pull/1021/commits/811f96b603fb3f0a1a8cf9232f9a53ce5f77d03d

src/d3a/models/area/event_dispatcher.py
hannesdiedrich: The, IMO, the `area.activate()` method should be made uniform and have the optional keyword "current_tick" , instead of poping it here.

src/d3a/models/state.py
```diff
@@ -51,13 +51,13 @@ def get_energy_production_forecast_kWh(self, time_slot, default_value=None):
         assert self._energy_production_forecast_kWh[time_slot] >= -FLOATING_POINT_TOLERANCE
         if default_value is not None:
             return self._energy_production_forecast_kWh.get(time_slot, default_value)
-        return self._energy_production_forecast_kWh[time_slot]
+        return self._energy_production_forecast_kWh.get(time_slot, 0.)
```
hannesdiedrich, line 54: didnt you check this by moving https://github.com/gridsingularity/d3a/pull/1021/files#diff-00472b9024eb273db74d246aa396bcbb49a369f86334c3413c71a505a4994036R343 (this is what i got from the slack discussion)

src/d3a/models/area/event_dispatcher.py
faizan2590: `current_tick` is an optional argument in `area.activate` https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/area/__init__.py#L258

The issue is that if I won't pop it here, then I need to comply all respective callback methods to accept arbitrary keyword arguments like `def cycle_markets(self, .., **kwargs), def activate(self, **kwargs),  def event_tick(self, **kwargs), def event_activate(self, **kwargs)`
Let me know if this is preferable?

src/d3a/models/state.py
faizan2590: Hmm, will do it too but its not important at this point as at L51 at this point of commit would have already raised an exception

src/d3a/models/area/event_dispatcher.py
hannesdiedrich: Sorry, but you only need it in the `event_activate` and `activate` methods, right? If so, I would vote for adding this.

src/d3a/models/state.py
hannesdiedrich: ah, yes so you dont need this change then. I am being relentless here to show the reader of the code, that a `time_slot` should be indeed in `self._energy_production_forecast_kWh` at this point. RIght?

src/d3a/models/area/event_dispatcher.py
faizan2590: I would be using **kwargs for `event_activate ` which itself also don't need it so IMO all callback should be homogenous in terms of either all accepting `**kwargs` or none

hannesdiedrich: Hmm, I dont agree here. As we handle all events differently, `event_activate` and `activate` could contain kwargs and the others don't. But if you like, add it to all of them.
Also tagging @spyrostz here to maybe add what is best practice here. Thanks!

spyrostz: I agree with Hannes here. First of all, you do not need to pass kwargs on every event, you only need to do so to the event_activate. Second, I think that the correct solution would be to pass it as a normal argument. Have a look at other events like evet_offer, we constantly pass different arguments to different events. This is going to produce a lot of changes, however IMO it is the correct solution, because you really want to pass something via the event here.

src/d3a/models/area/event_dispatcher.py
```diff
@@ -79,6 +79,7 @@ def _broadcast_notification(self, event_type: Union[MarketEvent, AreaEvent], **k
         # Broadcast to children in random order to ensure fairness
         for child in sorted(self.area.children, key=lambda _: random()):
             child.dispatcher.event_listener(event_type, **kwargs)
+        # kwargs.pop('current_tick', None)
```
spyrostz, line 82: Should be deleted

spyrostz (dismissed):
One cosmetic comment, LGMT

hannesdiedrich (approved):
Once Spyros comment is addressed, LGTM

spyrostz (approved):

PR #1024, opened 2021-01-20
D3ASIM-3243: Readded area_reconfigure_event for devices in UpdateArea…

hannesdiedrich:
…Event.

### DIFF
--- a/src/d3a/d3a_core/live_events.py
+++ b/src/d3a/d3a_core/live_events.py
@@ -55,19 +55,21 @@ class UpdateAreaEvent:
         if area_type is not None and area_type != "Area":
             if area.strategy is None:
                 return False
+            reactivate = False
             if area_type == "MarketMaker":
                 area.strategy = MarketMakerStrategy(**self.area_params)
+                reactivate = True
             elif area_type == "InfiniteBus":
                 # TODO: After hack to move this parameter casting at the web side
                 self.area_params['energy_sell_rate'] = self.area_params.pop('energy_rate', None)
                 area.strategy = InfiniteBusStrategy(**self.area_params)
-            else:
-                return False
-            area.activate()
-            area.strategy.event_activate()
-            area.strategy.event_market_cycle()
-        else:
-            area.area_reconfigure_event(**self.area_params)
+                reactivate = True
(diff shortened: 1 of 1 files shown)


src/d3a/d3a_core/live_events.py
```diff
@@ -55,19 +55,21 @@ def apply(self, area):
...
                 area.strategy = InfiniteBusStrategy(**self.area_params)
-            else:
-                return False
-            area.activate()
-            area.strategy.event_activate()
-            area.strategy.event_market_cycle()
-        else:
-            area.area_reconfigure_event(**self.area_params)
+                reactivate = True
+            if reactivate:
```
Hassan754, line 67: I think you have some extra spaces here

hannesdiedrich: Sorry, I dont understand. Should I remove indention here?

Hassan754: Ah sorry I saw it indented to the right

Hassan754 (approved):
LGTM

spyrostz (approved):
LGTM

faizan2590 (approved):
LGTM

src/d3a/d3a_core/live_events.py
```diff
@@ -55,19 +55,21 @@ def apply(self, area):
         if area_type is not None and area_type != "Area":
```
fievelk, line 55: Not part of this PR, but I've seen it in several lines of the codebase so I'll ask. Are we specifically checking for `None` values, or are we checking for broader false values instead? I would usually prefer `if area_type and area_type != 'Area'`, to consider cases in which `area_type` is something like `False`, `""` or `[]`.

fievelk (approved):
Just a small general question unrelated to the specific PR. Besides that, LGTM!

src/d3a/d3a_core/live_events.py
hannesdiedrich: Here we are actually checking for None. But you are right, there are cases, where these checks should test for   `False`, `""` or `[]`, too. PLease help us to make this consistent. Thanks!

PR #1025, opened 2021-01-20
Added burst mode to the rq workers. Added None as default value on sl…

spyrostz:
…ot_length_realtime, since this is not a part of the GlobalConfig.

### DIFF
--- a/src/d3a/d3a_core/d3a_jobs.py
+++ b/src/d3a/d3a_core/d3a_jobs.py
@@ -111,7 +111,7 @@ def start(scenario, settings, events, aggregator_device_mapping, saved_state):
         validate_global_settings(config_settings)
 
         slot_length_realtime = duration(seconds=settings['slot_length_realtime'].seconds) \
-            if 'slot_length_realtime' in settings else GlobalConfig.slot_length_realtime
+            if 'slot_length_realtime' in settings else None
 
         config = SimulationConfig(**config_settings)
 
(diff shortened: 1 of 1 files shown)


mclrch (approved):
LGTM

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -156,7 +156,7 @@ def main():
         Worker(
             [get_simulation_queue_name()],
             name=f'simulation.{getpid()}.{now().timestamp()}', log_job_description=False
-        ).work(max_jobs=1)
+        ).work(max_jobs=1, burst=True)
```
faizan2590, line 159: Can you help me understand the intuition behind enabling burst mode, as far as I know it dequeues all in one go till the queue is completely empty

hannesdiedrich (approved):
LGTM

src/d3a/d3a_core/d3a_jobs.py
spyrostz: First of all, each worker is single-threaded, meaning that it does not dequeue everything in parallel in one go, but one after another. Since our jobs are long-running, if there are jobs in the queue each worker will only dequeue one job, even with burst mode enabled. 
The rationale behind this is to avoid having d3a_jobs processes that have started, they are waiting for new jobs in the queue, but never terminated. We continuously see this problem in the live environment, where there are multiple d3a-jobs processes that do not run a canary network, but still they allocate memory, growing unnecessarily our resource consumption. For more information please visit d3a-devops channel on slack :D there is a detailed description of the issue.

PR #1026, opened 2021-01-20
D3ASIM-2876: Enabled toggling of allow_external_connection via live e…

hannesdiedrich:
…vent in external strategies.

I tried it with PV and Load strategy and the it worked:

Screenshot 2021-01-20 at 16 28 03

### DIFF
--- a/src/d3a/models/strategy/external_strategies/__init__.py
+++ b/src/d3a/models/strategy/external_strategies/__init__.py
@@ -114,7 +114,7 @@ class ExternalMixin:
     @property
     def should_use_default_strategy(self):
         return self._use_template_strategy or \
-               (not self.connected and not self.is_aggregator_controlled)
+               not (self.connected or self.is_aggregator_controlled)
 
     @property
     def _dispatch_tick_frequency(self):
--- a/src/d3a/models/strategy/external_strategies/__init__.py
+++ b/src/d3a/models/strategy/external_strategies/__init__.py
@@ -86,10 +86,19 @@ class ExternalMixin:
     def __init__(self, *args, **kwargs):
         self._connected = False
         self.connected = False
+        self.use_template_strategy = False
         super().__init__(*args, **kwargs)
         self._last_dispatched_tick = 0
         self.pending_requests = []
         self.lock = Lock()
+
+    @property
+    def channel_dict(self):
+        return {
+            f'{self.channel_prefix}/register_participant': self._register,
+            f'{self.channel_prefix}/unregister_participant': self._unregister,
+            f'{self.channel_prefix}/device_info': self._device_info,
+        }
 
     @property
     def channel_prefix(self):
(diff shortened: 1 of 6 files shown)


src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -86,11 +86,20 @@ class ExternalMixin:
     def __init__(self, *args, **kwargs):
         self._connected = False
         self.connected = False
+        self.use_template_strategy = False
```
spyrostz, line 89: Would be better to make this parameter private, only in order to discourage other people from accessing it from outside of this class.

spyrostz (approved):
One minor comment, LGTM otherwise

faizan2590 (approved):
LGTM

hannesdiedrich:
@spyrostz @faizan2590 In order to enable switching the API ON for a device that uses the template strategy at startup, i would have to move the self._use_template_strategy flag and its assignment into the load_hours, pv, storage strategy. 
I am not a big fan of that and wanted to have your opinion, before I ask Sarah whether this is really needed.

spyrostz:
> @spyrostz @faizan2590 In order to enable switching the API ON for a device that uses the template strategy at startup, i would have to move the self._use_template_strategy flag and its assignment into the load_hours, pv, storage strategy.
> I am not a big fan of that and wanted to have your opinion, before I ask Sarah whether this is really needed.

I would also not like that at all, and I frankly would prefer to deal with allowing live data + template strategies (if it is required of course) on its own story and not in this PR.

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -104,7 +113,8 @@ def is_aggregator_controlled(self):
 
     @property
     def should_use_default_strategy(self):
-        return not self.connected and not self.is_aggregator_controlled
+        return self._use_template_strategy or \
+               (not self.connected and not self.is_aggregator_controlled)
```
Hassan754, line 117:
```suggestion
        return self._use_template_strategy or \
               not (self.connected or self.is_aggregator_controlled)
```
Minor

Hassan754 (approved):
LGTM , but i could not follow the trace of toggling ( where is the endpoint or redis channel and doesn't it need to be configured from the `d3a-web` side ?

hannesdiedrich:
> LGTM , but i could not follow the trace of toggling ( where is the endpoint or redis channel and doesn't it need to be configured from the `d3a-web` side ?

@Hassan754 The toggling is done by the `allow_external_connection` flag. Have look into here: https://github.com/gridsingularity/d3a/pull/1026/files#diff-bc1fcffcd59eb05e5d8f357d9c2e6d095de871c1ab57e983392af8e51f1ac54eR136

Hassan754 (approved):

spyrostz (approved):
LGTM

PR #1030, opened 2021-01-25
Added logs when terminating a simulation via the Redis job metadata. …

spyrostz:
…Disabled stopping the simulation in case the current job cannot be found in Redis.


Hassan754 (approved):

hannesdiedrich (commented):
LGTM

PR #1031, opened 2021-01-25
feature/D3ASIM-3203: Made the RUN_ON_D3A_WEB controlled via cli and Environ

Hassan754:
### Summary
Made the `RUN_ON_D3A_WEB` constant True by default, + It can be controlled from the cli or the environment variable
If True, the API client will use the Rest API, otherwise it will use the Redis Client.

### Implementation Notes 
The default value is set in the src/d3a/constants.py

Added the cli option as flag `--use-redis-client`, if used it will set the aforementioned constant as False in the environment.

In the src/d3a/d3a_core/util.py, added a helper function `get_is_run_on_d3a_web`, this function will check the environment variable for the `RUN_ON_D3A_WEB` constant, if it does not exist, it will get the default value added in src/d3a/constants.py

### Usage
The setup files were not adapted yet, and the `RUN_ON_D3A_WEB` variable is set explicitly now.
However, to adapt any of them, it is enough to replace the `RUN_ON_D3A_WEB = True/False with RUN_ON_D3A_WEB = get_is_run_on_d3a_web()`


src/d3a/d3a_core/util.py
```diff
@@ -33,11 +33,10 @@
 from functools import wraps
 from logging import LoggerAdapter, getLogger, getLoggerClass, addLevelName, setLoggerClass, NOTSET
 
-import d3a.constants
```
Hassan754, line 36: This import was removed as it is not effective

spyrostz (commented):
This PR is actually redundant. RUN_ON_D3A_WEB is already correctly populated on d3a. The ticket is referring to the api client parameter.

PR #1032, opened 2021-01-25
D3ASIM-3231: Added: Make the job fail if the job in RQ can not be fou…

hannesdiedrich:
…nd any more (redis restart). Also removed redundant send of results when initialising slot.

### DIFF
--- a/src/d3a/d3a_core/launcher.py
+++ b/src/d3a/d3a_core/launcher.py
@@ -33,13 +33,9 @@ MAX_JOBS = os.environ.get('D3A_MAX_JOBS_PER_POD', 2)
 
 
 class Launcher:
-    def __init__(self,
-                 queue=None,
-                 max_jobs=None,
-                 max_delay_seconds=2):
-
+    def __init__(self, max_jobs=None, max_delay_seconds=2):
         self.redis_connection = StrictRedis.from_url(REDIS_URL, retry_on_timeout=True)
-        self.queue = queue or Queue(get_simulation_queue_name(), connection=self.redis_connection)
+        self.queue = Queue(get_simulation_queue_name(), connection=self.redis_connection)
         self.max_jobs = max_jobs if max_jobs is not None else int(MAX_JOBS)
         self.max_delay = timedelta(seconds=max_delay_seconds)
         python_executable = sys.executable \
--- a/src/d3a/d3a_core/launcher.py
+++ b/src/d3a/d3a_core/launcher.py
@@ -37,8 +37,9 @@ class Launcher:
                  queue=None,
                  max_jobs=None,
                  max_delay_seconds=2):
-        self.queue = queue or Queue(get_simulation_queue_name(), connection=StrictRedis.from_url(
-            REDIS_URL, retry_on_timeout=True))
+
+        self.redis_connection = StrictRedis.from_url(REDIS_URL, retry_on_timeout=True)
+        self.queue = queue or Queue(get_simulation_queue_name(), connection=self.redis_connection)
         self.max_jobs = max_jobs if max_jobs is not None else int(MAX_JOBS)
         self.max_delay = timedelta(seconds=max_delay_seconds)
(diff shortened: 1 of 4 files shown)


src/d3a/d3a_core/simulation.py
```diff
@@ -303,7 +303,6 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
 
         config.external_redis_communicator.sub_to_aggregator()
         config.external_redis_communicator.start_communication()
-        self._update_and_send_results()
```
Hassan754, line 306: LGTM, but please clarify why this line was removed

faizan2590, line 306: Can you help me understand the intuition behind it?

src/d3a/d3a_core/launcher.py
```diff
@@ -37,8 +37,9 @@ def __init__(self,
                  queue=None,
                  max_jobs=None,
                  max_delay_seconds=2):
-        self.queue = queue or Queue(get_simulation_queue_name(), connection=StrictRedis.from_url(
-            REDIS_URL, retry_on_timeout=True))
+
+        self.redis_connection = StrictRedis.from_url(REDIS_URL, retry_on_timeout=True)
+        self.queue = queue or Queue(get_simulation_queue_name(), connection=self.redis_connection)
```
faizan2590, line 42: How splitting it helps?

faizan2590: Nevermind, I think I got it, you intend to minimise number of redis connection?

faizan2590 (dismissed):
One comment for clarification, other than that LGTM

src/d3a/d3a_core/launcher.py
hannesdiedrich: Yes, in order to reuse the same connection.

src/d3a/d3a_core/simulation.py
hannesdiedrich: This is redundant because between the call of `_update_and_send_results` in `_init` and inside the slot nothing happened to the results (no updates or at least none that the d3a-web is interested in.

hannesdiedrich: https://github.com/gridsingularity/d3a/pull/1032#discussion_r563853049

src/d3a/d3a_core/launcher.py
```diff
@@ -37,8 +37,9 @@ def __init__(self,
                  queue=None,
```
spyrostz, line 37: How about taking the opportunity and removing this queue parameter ?

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -166,9 +166,9 @@ def _handle_redis_job_metadata(self):
                 self._simulation.stop()
 
         except NoSuchJobError:
-            log.error(f"Redis job {self._simulation_id} cannot be found in the Redis job queue. "
-                      f"get_current_job failed. Continue to operate as a daemon job, "
-                      f"without being monitored by RQ.")
+            raise(f"Redis job {self._simulation_id} cannot be found in the Redis job queue. "
```
spyrostz, line 169: Hmm, good idea to raise an exception with this message, however it is better to raise an exception of a specific type (e.g. D3AException). Also, please remove the `Continue to operate as a daemon job ... ` part, because it does not continue to operate anymore, the process terminates.

spyrostz (approved):
2 comments, LGTM otherwise

src/d3a/d3a_core/redis_connections/redis_communication.py
hannesdiedrich: Yes, you are right, I changed this without testing.

Hassan754 (approved):

spyrostz (approved):
LGTM

PR #1034, opened 2021-01-27
Enhancing cumulative grid trades.

faizan2590:
To be merged before https://github.com/gridsingularity/d3a/pull/1033

### DIFF
--- a/integration_tests/steps/jira_issues_tests.py
+++ b/integration_tests/steps/jira_issues_tests.py
@@ -24,6 +24,7 @@ from d3a.constants import TIME_ZONE
 from d3a.d3a_core.export import EXPORT_DEVICE_VARIABLES
 from d3a_interface.sim_results.market_price_energy_day import MarketPriceEnergyDay
 from d3a_interface.sim_results.bills import CumulativeBills
+from d3a_interface.sim_results.cumulative_grid_trades import CumulativeGridTrades
 
 
 def get_areas_from_2_house_grid(context):
@@ -299,10 +300,14 @@ def storage_decreases_bid_rate(context):
 
 @then('cumulative grid trades correctly reports the external trade')
 def area_external_trade(context):
-    cumulative_trade = context.simulation.endpoint_buffer.cumulative_grid_trades.current_trades
+    cumulative_trade = context.simulation.endpoint_buffer.cumulative_grid_trades.accumulated_trades
     house1 = list(filter(lambda x: x.name == "House 1", context.simulation.area.children))[0]
-    ext_trade = list(filter(lambda x: x['areaName'] == "External Trades",
-                            cumulative_trade[house1.uuid]))[0]['bars'][0]['energy']
+    cal_cum_trades = CumulativeGridTrades.generate_cumulative_grid_trades_target_area(
+            house1.uuid, {'cumulative_grid_trades': cumulative_trade.get(house1.uuid, None)}
+    )
+    ext_trade = sum(info['bars'][0]['energy']
+                    for info in cal_cum_trades[house1.uuid]
+                    if info['areaName'] == "External Trades")
     assert isclose(ext_trade, -1 * 0.666, rel_tol=1e-05)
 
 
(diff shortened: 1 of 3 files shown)


integration_tests/steps/jira_issues_tests.py
```diff
@@ -299,10 +300,14 @@ def storage_decreases_bid_rate(context):
...
+    cumulative_trade = context.simulation.endpoint_buffer.cumulative_grid_trades.accumulated_trades
     house1 = list(filter(lambda x: x.name == "House 1", context.simulation.area.children))[0]
-    ext_trade = list(filter(lambda x: x['areaName'] == "External Trades",
-                            cumulative_trade[house1.uuid]))[0]['bars'][0]['energy']
+    cal_cum_trades = CumulativeGridTrades.generate_cumulative_grid_trades_target_area(
+            house1.uuid, {'cumulative_grid_trades': cumulative_trade.get(house1.uuid, None)}
+    )
+    ext_trade = sum(info['bars'][0]['energy']
+                    for info in cal_cum_trades[house1.uuid]
+                    if info['areaName'] == "External Trades")
```
hannesdiedrich, line 310: Sorry, I dont understand why the `sum` is needed. Shouldn't there only be one `"External Trades"` member for the house1?

faizan2590: No, there could be many
Screen Shot 2021-01-27 at 08 57 05

hannesdiedrich: Ah, ok, you are checking the UI formated accumulated_trades here and in line 306 you are mimicking a return from the DB. Its ok to keep it, but i would re-add the check whether the `context.simulation.endpoint_buffer.cumulative_grid_trades.current_trades` data point is correct. 
BTW; why was this change needed? Did the test fail?

faizan2590: > but i would re-add the check whether the context.simulation.endpoint_buffer.cumulative_grid_trades.current_trades data point is correct

There is nothing inside cumulative_grid_trades.current_trades  after recent results endpoint enhancement, if you recall it was decided that only `accumulated_trades` would be calculated and saved to DB and `current_trades` the UI friendly with labels would be calculated on the fly

> BTW; why was this change needed? Did the test fail?
Yes, the recent result enhancement was tested via UI but not in standalone d3a run, fortunately only one failed as well also a self proving point that the recent results enhancement didn'T break anything major

hannesdiedrich:
```
There is nothing inside cumulative_grid_trades.current_trades after recent results endpoint enhancement, if you recall it was decided that only accumulated_trades would be calculated and saved to DB and current_trades the UI friendly with labels would be calculated on the fly
```
OK, but please get rid of the occurrences of `cumulative_grid_trades.current_trades` in the `generate_json_report` (here maybe double check with Colin or Chris whether this is really OK to deprecate, sorry, we should have thought about this also before) and also in the integrationtests.
Also, you can delete the `self.current_trades` in `CumulativeGridTrades` in the interface.
```
Yes, the recent result enhancement was tested via UI but not in standalone d3a run, fortunately only one failed as well also a self proving point that the recent results enhancement didn'T break anything major
```
OK, makes sense. Thanks

lana-shanghai (dismissed):
LGTM

integration_tests/steps/jira_issues_tests.py
faizan2590: Made a ticket: https://gridsingularity.atlassian.net/browse/D3ASIM-3263

All the other occurences are part of balancing market which is unused ATM, so I would rather not touch it

lana-shanghai (approved):
Would appreciate a comment in slack later on what exactly is the diff between current_trades and accumulated_trades! Otherwise LGTM

faizan2590:
> Would appreciate a comment in slack later on what exactly is the diff between current_trades and accumulated_trades! Otherwise LGTM

pinged on slack!

spyrostz (approved):
LGTM

PR #1035, opened 2021-01-27
added_pytz_to_requirements

hannesdiedrich:
This is needed to deserialize timezone aware timestamps send by the d3a-web.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -17,6 +17,7 @@ pip-tools==5.4.0          # via -r requirements/base.in
 plotly==4.14.1            # via -r requirements/base.in
 python-dateutil==2.8.1    # via pendulum
 python-rex==0.4           # via -r requirements/base.in
+pytz==2020.4              # via -r requirements/base.in
 pytzdata==2020.1          # via pendulum
 redis==3.5.3              # via -r requirements/base.in, rq
 regex==2020.11.13         # via awesome-slugify
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -17,7 +17,7 @@ pip-tools==5.4.0          # via -r requirements/base.in
 plotly==4.14.1            # via -r requirements/base.in
 python-dateutil==2.8.1    # via pendulum
 python-rex==0.4           # via -r requirements/base.in
-pytz==2020.4              # via -r requirements/base.in
+pytz==2020.5              # via -r requirements/base.in
 pytzdata==2020.1          # via pendulum
 redis==3.5.3              # via -r requirements/base.in, rq
 regex==2020.11.13         # via awesome-slugify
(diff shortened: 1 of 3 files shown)


requirements/base.txt
```diff
@@ -17,6 +17,7 @@ pip-tools==5.4.0          # via -r requirements/base.in
 plotly==4.14.1            # via -r requirements/base.in
 python-dateutil==2.8.1    # via pendulum
 python-rex==0.4           # via -r requirements/base.in
+pytz==2020.4              # via -r requirements/base.in
```
faizan2590, line 20: Whats the background of injecting this dependency?

hannesdiedrich: https://github.com/gridsingularity/d3a/pull/1035#issue-562341305

faizan2590: Okay, but you have added it as dependency without using it anywhere in code

hannesdiedrich: It is used internally in pickle, when deserialising timezone aware timestamps. Pickle is (also) used when sending jobs via RQ.

faizan2590: Okay, I guess it could be removed from [pandapower](https://github.com/gridsingularity/d3a/blob/master/requirements/pandapower.txt#L22) dependency then as if powerflow is executed it would be executed on top of base dependencies

spyrostz: To provide some more feedback, it is used not only in pickle, but in Python in general https://docs.python.org/3/library/zoneinfo.html#module-zoneinfo. In case Python fails to detect the timezone in a system, it will try to find this package (pytz) in order to set the correct timezone for the system. You can check also the description of the package here https://github.com/python/tzdata . 
I would recommend to add it even as a failsafe, despite the fact that it is not used by our code (it will be used internally by Python if there is no timezone in a system). Keep in mind that this is a first-party package, therefore it will probably be quite painless to keep it in our requirements.

hannesdiedrich: @faizan2590 pytz must be a requirement of pandapower itself, because `pip-compile requirements/pandapower.in` adds it to `pandapower.txt` automatically. But now I upgraded pytz from 2020.4 to 2020.5

faizan2590: @hannesdiedrich I think this could be avoided via adding 
`-r base.txt`
 at the top of `pandapower.txt`

hannesdiedrich: @faizan2590 but then I have all the dependencies of base.txt also in pandapower.txt

faizan2590: I think thats okay as pandower dependency would work with d3a on top of base dependencies BUT lets leave it for now as we are not using it ATM.

faizan2590 (approved):
LGTM

spyrostz (approved):
LGTM

PR #1037, opened 2021-01-28
Check first if the area exists in the simulation state before trying …

spyrostz:
…to restore the state for it, and log a warning if the state is not there, to avoid crashing the whole simulation.


Hassan754 (approved):

hannesdiedrich (approved):
LGTM

faizan2590 (approved):

PR #1038, opened 2021-01-28
D3ASIM-3220: Add replace_existing flag

fievelk:
## Content

- Implement `replace_existing` flag to allow the client to delete previously existing bids and offers.
- Use `deque` instead of a normal `list` to process bids in the order in which they have been made by the client


fievelk:
Some questions about the current work in progress:
1. I set `replace_existing=False` as default, because all the existing implementation and tests were relying on this behavior. Changing it would mean refactoring all the existing calls: do you think it would be better to do that?
2. I only changed the `market.bid()` call in the `post_bid` method. However, the same call is made in many other parts like:
    - [`split_bid`](https://github.com/gridsingularity/d3a/blob/9251833fb15274e9bbc8d1ac487840b46c6fe531/src/d3a/models/market/two_sided_pay_as_bid.py#L107-L114)
    - [`_bid_aggregator`](https://github.com/gridsingularity/d3a/blob/2219a0839fe68e7f3ff2a25601d13b9dc028483e/src/d3a/models/strategy/external_strategies/storage.py#L531-L536)
    - [`_update_bid_aggregator`](https://github.com/gridsingularity/d3a/blob/e2540ccb844810adc5ad7d8947e43c32df434545/src/d3a/models/strategy/external_strategies/load.py#L274-L275)
    - [`_forward_bid`](https://github.com/gridsingularity/d3a/blob/6aed935ce4a55831b69dd61a8e05d0d1528a6cce/src/d3a/models/strategy/area_agents/two_sided_pay_as_bid_engine.py#L54-L61)
  
    Should they also be updated to use the `replace_bid` argument?

tests/test_strategy_load_hours.py
```diff
@@ -112,9 +112,9 @@ def get_bids(self):
         return deepcopy(self.bids)
 
     def bid(self, price: float, energy: float, buyer: str, original_bid_price=None,
-            buyer_origin=None) -> Bid:
+            buyer_origin=None, replace_existing=False) -> Bid:
```
fievelk, line 115: I believe that this testing strategy is not optimal: we are currently creating a mock object that behaves as we want, but this doesn't necessarily mirrors the real implementation of a market. It might become material for a future task.

src/d3a/models/strategy/__init__.py
```diff
@@ -454,14 +455,28 @@ def __init__(self):
...
+        for bid in self.get_posted_bids(market):
+            assert bid.buyer == self.owner.name
+            if bid.id in market.bids.keys():
+                bid = market.bids[bid.id]
+
+            self.remove_bid_from_pending(market.id, bid.id)
+
+    def post_bid(self, market, price, energy, replace_existing=False, buyer_origin=None):
+        if replace_existing:
+            self._remove_existing_bids(market)
```
hannesdiedrich, line 470: Unfortunately, this is the one step: Here, in the BaseStrategy, the devices track their bids. Its right to replace them here. However, they also need to be removed from the markets.
So i suggest to get the uuid of the bid here and then delete them from the market (from the external strategy, not from within here).

src/d3a/models/strategy/__init__.py
```diff
@@ -454,14 +455,28 @@ def __init__(self):
...
+
         bid = market.bid(
             price,
             energy,
             self.owner.name,
             original_bid_price=price,
-            buyer_origin=buyer_origin
-        )
+            buyer_origin=buyer_origin,
+            replace_existing=replace_existing)
```
hannesdiedrich, line 478: I dont think this is needed and could also lead to confusion.
IMO you "only" need to do the following when the `replace_existing` flag was set via the api_client:
1. remove all bids from the BaseStrategy tracker and collect their uuids
2. use these uuids and delete them from the market (see how it is implemented in delete_bid).
Let me know if you need more information. Thanks!

src/d3a/models/strategy/__init__.py
fievelk: Thanks for the help! 
Isn't this what `self.remove_bid_from_pending(market.id, bid.id)` is doing? It gets the current market, then calls `market.delete_bid(b_id)` for every bid ID.

src/d3a/models/strategy/__init__.py
fievelk: I agree that this is redundant and not that optimal.
However, we want to find a way to add the value of `replace_existing` inside the `Bid` object, because that's what will be returned to the client. And the `Bid` object is created inside of `market.bid()` at the moment:
https://github.com/gridsingularity/d3a/pull/1038/files/9143d4cc7a88278692e1fe523fbc973967098cf5#diff-e048ea2bf894f98b35d1c8798a95454a8dd7507f5b1d03e592d8ad65345c85b2R81-R85

Unfortunately, I couldn't come up with a cleaner idea to do that yet.

src/d3a/models/strategy/__init__.py
hannesdiedrich: You are indeed right! Sorry, I was too quick with my answer (and thought that there is no market interaction in the BaseStrategy, but there is).
Ok, then all good!

src/d3a/models/strategy/__init__.py
hannesdiedrich: This is probably a design question and that is why I would like to consult @spyrostz here.
However IMO, the market (and neither the Blockchain) does not care about whether the bid is a "replace_existing" one and frankly this could lead to some privacy issues too, because this way others can see your bidding strategy very easily.
I still dont get why you want to track it. What does the client need to know? IMO only that the new bid was places and that the old ones where deleted. If useful, you can even return the uuids of the deleted bids to the client.
What do you think?

spyrostz: IMO it is far better to manually append it on the `self.redis.publish_json` call, than adding it to the Bid namedtuple. In general, d3a (the exchange itself) is not interested in this flag, the API user that sent this command might be interested in it though. 
For instance, for the bid case, the response is transmitted via this call:
`            self.redis.publish_json(
                bid_response_channel,
                {"command": "bid", "status": "ready", "bid": bid.to_JSON_string(),
                 "transaction_id": arguments.get("transaction_id", None)})`

The actual response of the command is this: `"bid": bid.to_JSON_string()`, therefore you could add one more key-value pair to the bid JSON to reflect the value of `replace_existing`. Hope these all make sense.

fievelk: @hannesdiedrich I hadn't thought about the privacy concerns. That's extremely interesting, thanks!

@spyrostz Just to be sure: you are suggesting to add it to `bid.to_JSON_string()`, and not as a separate key-value to the response (like `transaction_id`). Is this correct?

spyrostz: @fievelk yes, I am suggesting to add it to bid.to_JSON_string(), but as an optional method argument (for instance, by converting the method to def to_JSON_string(replace_existing=None). In this case you can pass the replace_existing variable and populate the bid json string with its value if the replace_existing argument is used, but if it is not used you can entirely omit it from the json string)

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -127,13 +134,16 @@ def _bid(self, payload):
             return
         try:
             arguments = json.loads(payload["data"])
-            assert set(arguments.keys()) == {'price', 'energy', 'transaction_id'}
+            assert set(arguments.keys()) == {
+                'price', 'energy', 'replace_existing', 'transaction_id'}
```
fievelk, line 138: Important question here: should we always expect the `replace_existing` parameter to be provided? It will be set to `True` by default on the api client (so it will always be provided from that source) but let me know if I should make it optional here and omit this strict check.

spyrostz: I think it is better to make it optional

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -127,15 +134,17 @@ def _bid(self, payload):
             return
         try:
             arguments = json.loads(payload["data"])
-            assert set(arguments.keys()) == {'price', 'energy', 'transaction_id'}
+            assert all(arg in self.ALLOWED_PARAMETERS for arg in arguments.keys())
```
fievelk, line 137: This check allows us to have optional arguments (like `replace_existing`). However, this is semantically different than the previous check:
```python
assert set(arguments.keys()) == {'price', 'energy', 'transaction_id'}
```
Because now these 3 arguments are not mandatory anymore.

Let me know if this needs to be changed: in that case, I will create a separate function or class to validate the input arguments for the bid/offer commands.

src/d3a/models/strategy/__init__.py
```diff
@@ -454,14 +455,27 @@ def __init__(self):
...
         self._traded_bids = {}
 
-    def post_bid(self, market, price, energy, buyer_origin=None):
+    def _remove_existing_bids(self, market: Market) -> None:
+        """Remove all existing bids in the market"""
+
+        for bid in self.get_posted_bids(market):
+            assert bid.buyer == self.owner.name
+            if bid.id in market.bids.keys():
+                bid = market.bids[bid.id]
```
hannesdiedrich, line 464: Can you please explain what you are trying to achieve here? I wonder how this works, as `bid` is the loop variable.

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: IMO you dont need this, as `_get_transaction_id` (line 131) will already fail if there is no `transaction_id` in the payload.

hannesdiedrich: And as this is done in every callback function, there is no need for `ALLOWED_PARAMETERS` and `ALLOWED_PARAMETERS` , as these are only applicable for commands like `bid` or `offer` . E.g. for _delete_offer, we need the `offer_id` as input parameter

hannesdiedrich (commented):
Left some comments, will review again after they are addressed. Thanks

src/d3a/models/strategy/__init__.py
fievelk: My mistake: this is something I took from another portion of the code and I later forgot to double-check:
https://github.com/gridsingularity/d3a/blob/89e521bf5af64ed05f0dd1e43cc2209df128713e/src/d3a/models/strategy/update_frequency.py#L197-L207

I am actually a bit confused here: what's the difference between `Strategy._bids` and `Market.bids`?

hannesdiedrich: You are right, apart from posting bids (or offers) in the markets directly (`Market.bids`), the devices also track the posted and bought/sold bids/offers also in the BaseStrategy (`Strategy._bids`). This enables the strategy to always have knowledge about these information without having to read this from the market. 
Let me know if you need more information. Thanks!

fievelk: In that case, I will leave it as it is 👍 

So the device (`Strategy`) tracks its own bids/offers. Then, at some point, they are posted on the market. Why is there a need to decouple these two lists? 

What I can imagine is something like: the Load has a list of bids (A) that it wants to post. Then the load actually posts them on the market, so the list of market bids (B) is filled with bids from (A). At the same time, (A) is emptied. Could this be the scenario?

hannesdiedrich: The list of posted bids in the strategy and in the market should be in sync when posting a bid.
This is the workflow (if you look into `post_bid`)
1. `market.bid` is called which posts the bid to be traded into the market
2. `add_bid_to_posted` is called that keeps track of the bid in the strategy side in order for the strategy to know how much it posted already, because the bid in the market could be instantly traded (and consequently deleted from the market)
Does it make more sense now?

src/d3a/models/strategy/external_strategies/load.py
fievelk: Ok, since the `ALLOWED_PARAMETERS` would be different depending on the method (bid/offer/delete etc) it doesn't make that much sense to have them as a class constant. I suppose it would be nice to have a separate validator class that defines these parameters for each method, maybe we can think of it for a different task!

I'll then replace the `ALLOWED_PARAMETERS` with a standard set defined in the function itself, for the moment :)

src/d3a/models/strategy/__init__.py
fievelk: Now I understand the process, thank you. My next question is now more abstract: why does the strategy need to know how much it has already posted, if the bid has already been processed by the market?

hannesdiedrich: The tracking of bids and offers on the strategy side was used for accounting (producing the results statistics) in earlier times of the d3a. It turned out to be handy when we added the external strategies. Here are two examples:
1. When posting a bid: the strategy can already check the total posted energy on the strategy side without interacting with the market (keep in mind that in the future, the markets will be totally decoupled from the strategies and only accessible via APIs)
2. Let's say the user tries to delete a bid via the external strategy. As the posted bids are tracked in the BaseStrategy class, the ExternalStrategy does not have to read from the market in order to get a list of the bids that have been posted.

But really, I find it hard to find a strong argument for tracking the posted bids and offers also on the strategy side.
@spyrostz Do you know one and help me? Is it maybe time to get rid of this extra accounting?

hannesdiedrich (approved):
LGTM

hannesdiedrich (changes requested):
Sorry, approved wrong PR

src/d3a/models/strategy/__init__.py
spyrostz: The original rationale behind this was that the markets would eventually end up in the blockchain. Therefore each asset would have to maintain its own copy of the posted bids and offers, in order to avoid requesting the bids / offers from the blockchain, which is potentially a much more expensive and slower operation than keeping a copy of the bids and offers in the strategies. 
Having said that though, we have already decided that even if we decentralize our market classes 100%, we are still going to have one agent for each market in order to avoid such expensive operations. Therefore we will be able to directly access the bids and offers of a market. So this reason is not valid anymore. 
One reason I see in favor of keeping this tracking though, is that in markets with lots of bids and offers, it is going to be much slower to select all the offers and bids from one specific device. For instance, if one market includes 1 million offers, iterating over this list and selecting only the offers for one device would be far more expensive than buffering the offers of each device and having the offers for this device already filtered out. It is premature optimization I know, still I think it is going to be more performant in large markets. 
TLDR I do not have a strong opinion, but I think that it is more optimal to keep the tracking of the bids and offers as it is.

fievelk: Thank you! I think I have a clearer picture now, even though the details of the exact use-case are still a bit abstract 😄 I think it makes sense to keep the two things separate. Adding some more documentation might help to identify more quickly the two separate contexts (market bids vs device bids), so I'll keep this in mind in case I work on those parts of the code.

fievelk:
I made some changes after the reviews:
- made `replace_existing` optional
- returned `replace_existing` value as part of the JSON response
- removed `ALLOWED_PARAMETERS`

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -45,6 +48,13 @@ def channel_dict(self):
                 f'{self.channel_prefix}/list_bids': self._list_bids,
                 }
 
+    @property
+    def filtered_bids(self):
```
hannesdiedrich, line 52: Just to be more verbose:
```suggestion
    def filtered_bids_next_market(self):
```

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -127,15 +134,17 @@ def _bid(self, payload):
...
             self.redis.publish_json(
                 bid_response_channel,
                 {"command": "bid",
-                 "error": "Incorrect bid request. Available parameters: (price, energy).",
-                 "transaction_id": transaction_id}
-            )
+                 "error": (
+                     "Incorrect offer request. "
+                     f"Available parameters: {'price', 'energy', 'replace_existing'}."),
+                 "transaction_id": transaction_id})
```
hannesdiedrich, line 147: Why this change? It is indeed a bid request, not an offer request.

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -127,15 +134,17 @@ def _bid(self, payload):
             return
         try:
             arguments = json.loads(payload["data"])
-            assert set(arguments.keys()) == {'price', 'energy', 'transaction_id'}
+            assert all(arg in {'price', 'energy', 'replace_existing'} for arg in arguments.keys())
```
hannesdiedrich, line 137: And this works? How about the `transaction_id` key?

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -262,19 +272,23 @@ def _bid(self, payload):
     def _bid_impl(self, arguments, bid_response_channel):
         try:
             assert arguments["energy"] <= self.state.energy_to_buy_dict[self.next_market.time_slot]
+
+            replace_existing = arguments.get('replace_existing', True)
```
hannesdiedrich, line 276: IMO you can skip this, as you already asserted for all arguments in line 256.

hannesdiedrich (commented):
Some minor comments, once addressed, LGTM.
I would be in favour of first opening the d3a-api-client PR for this that contains an integrationtest before merging this.

src/d3a/models/strategy/external_strategies/load.py
fievelk: 👍 Mistake on my side, probably while replicating the logic

src/d3a/models/strategy/external_strategies/load.py
fievelk: That shouldn't be working, actually. Tests are passing though, which means that something needs to be corrected there too.

hannesdiedrich: This is why i added the general comment https://github.com/gridsingularity/d3a/pull/1038#pullrequestreview-582141575 . We currently dont test the external strategies here in the d3a. Thats why the tests are passing.

src/d3a/models/strategy/external_strategies/storage.py
fievelk: The assertion on line `256` doesn't check whether all arguments are in the request, but rather whether all the arguments that are in the request are allowed. This is something that changed compared to the previous behavior.

`replace_existing` is not mandatory, so we should allow requests that do not contain it. In that case, the default value I set is `True`.

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -246,14 +253,17 @@ def _bid(self, payload):
             return
         try:
             arguments = json.loads(payload["data"])
-            assert set(arguments.keys()) == {'price', 'energy', 'transaction_id'}
+            assert all(arg in {'price', 'energy', 'replace_existing'} for arg in arguments.keys())
```
hannesdiedrich, line 256: I know that this is starting the discussion of ALLOWED and NEEDED parameters again, but `price` and `energy` are actually mandatory parameters, otherwise no bid can be placed.

src/d3a/models/strategy/external_strategies/storage.py
hannesdiedrich: yes, sorry, I didnt understand tha, but left another comment there https://github.com/gridsingularity/d3a/pull/1038/files/761256609509ec4fc92a817b14c9bf2cbcb9bcba#r569274825

src/d3a/models/strategy/external_strategies/load.py
fievelk: Will do! Out of curiosity: why is it called `next_market` rather than `current_market`? Even the docstring in the method refer to it as "current market":
https://github.com/gridsingularity/d3a/blob/cc260bac3905416325ee72b5253f21f15ef0e308/src/d3a/models/area/__init__.py#L502-L507

fievelk:
@hannesdiedrich 
> I would be in favour of first opening the d3a-api-client PR for this that contains an integrationtest before merging this.

I have a PR ready on d3a-api-client, but tests are failing since they point to the master branch of `d3a`. Is there a way to change that?

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: Yes, this is a common confusion for "beginners" that we couldn't solve until today. Happy about any suggestions!

There are two "nows" in energy markets and this what is confusing from my experience.
 Ok, let's say it is 11:08, so the market_slot, where the assets are trading in is 11:15-11:30. This is in the future, so it could be called "next_market" although it is really the "current_market" because the marke where all trades where done for the current time (11:00-11:15) is already closed. 
So there is this chain of markets
[..., past_market_1, _past_market_2_, **current_market**, future_market_1, future_market_2, ...]
with the times:
[..., 10:45-11:00, _11:00-11:15_, **11:15-11:30**, 11:30-11:45, 11:45-12:00, ...]
The market at the current time is the _past_market_2_ and the market that is currently held is  **current_market**
So the latter was formally named next_market because it is in the future, but really it is the currently held market.

Sorry to be so verbose. May even confuse you more.
Maybe the others have a better way of explaining

src/d3a/models/strategy/external_strategies/storage.py
fievelk: I could add something like this inside the `_bid` method (rather than as class variables):
```python
required_args = {'price', 'energy'}
allowed_args = required_args.union({'replace_existing', 'transaction_id'})

# Check that all required arguments have been provided
assert all(arg in arguments.keys() for arg in required_args)
# Check that every provided argument is allowed
assert all(arg in allowed_args for arg in arguments.keys())
```
What do you think?

In any case, I think this logic should go in a separate validator class (or similar) in the future. Maybe we could open a ticket for it.

src/d3a/models/strategy/external_strategies/load.py
fievelk: Ok, so if I understand correctly there is something like:
- "ongoing" market (`past_market_2` in your example): that is the market where the trades I've already posted in the past are being applied right now . 
- "next" market (`current_market` in your example): the market in which I can place my bids/offers now (but it will start when the ongoing market's execution will end).

It's called "current" because the backend is taking bids/offers for it "now", even though it will start at the end of the current slot. I hope this was correct :)

fievelk:
I made some changes after the review commits:

- fixed naming mistake ("bid" and "offer") in the response strings
- fixed assertion to expect `transaction_id` in bid requests
- differentiated between _required_ and _allowed_ bid parameters
- renamed and refactored `filtered_bids` method

src/d3a/models/strategy/external_strategies/load.py
faizan2590: > Will do! Out of curiosity: why is it called `next_market` rather than `current_market`? Even the docstrings in the method refer to it as "current market":
> https://github.com/gridsingularity/d3a/blob/cc260bac3905416325ee72b5253f21f15ef0e308/src/d3a/models/area/__init__.py#L502-L507

the docstring seems to be decoupled with method name, what actually has to be settled is if for the most immediate future open market which is accepting offers/bids and simultaneously clearing it `trades` should be called `current_market` or the `next_market`. If the term `current_market` to be used for the most immediate future open market then what should the market where all settled trades are currently getting realised be called

faizan2590:
> @hannesdiedrich
> 
> > I would be in favour of first opening the d3a-api-client PR for this that contains an integrationtest before merging this.
> 
> I have a PR ready on d3a-api-client ([gridsingularity/d3a-api-client#89](https://github.com/gridsingularity/d3a-api-client/pull/89)), but tests are failing since they point to the master branch of `d3a`. Is there a way to change that?

Can route d3a temporarily to this branch to validate all test targetting this code has been passed via changing it [here](https://github.com/gridsingularity/d3a-api-client/blob/master/integration_tests/build_test_containers.sh#L17)

src/d3a/models/market/market_structures.py
```diff
@@ -110,15 +110,19 @@ def offer_from_JSON_string(offer_string, current_time):
...
                 buyer_origin=None, energy_rate=None):
         if energy_rate is None:
             energy_rate = price / energy
         # overridden to give the residual field a default value
-        return super(Bid, cls).__new__(cls, str(id), time, price, energy, buyer,
-                                       original_bid_price, buyer_origin, energy_rate)
+
+        return super(Bid, cls).__new__(
+            cls, str(id), time, price, energy, buyer, original_bid_price, buyer_origin,
+            energy_rate)
```
faizan2590, line 125: Curious if any specific code convention is followed here?

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -400,8 +422,10 @@ def event_tick(self):
             if req.request_type == "set_energy_forecast":
                 self._set_energy_forecast_impl(req.arguments, req.response_channel)
 
-        self.pending_requests = [req for req in self.pending_requests
-                                 if req.request_type not in "set_energy_forecast"]
+        self.pending_requests = deque(
```
faizan2590, line 425: sorry if its already discussed somewhere, can you explain the specific reason of having a deque here?

faizan2590 (commented):
Left some comments

src/d3a/models/market/market_structures.py
fievelk: The change is actually due to two separate commits that were adding some new parameter and then removing it. The style had changed in the meantime.

In terms of PEP8, the two versions are equally good: https://www.python.org/dev/peps/pep-0008/#indentation
I usually prefer using a _hanging indent_ because it saves a lot of space on the right side. If the method name is long, you might end up with something like this:
```python
# Without hanging indent you would waste a lot of space on the left, below the
# name of the method. And you might not have enough space on the right
# because of the 100 characters line limit
def a_method_with_a_very_long_name_and_accepting_lots_of_parameters(self,
                                                                    foo,
                                                                    bar,
                                                                    *args,
                                                                    **kwargs):

# With hanging indent you could use the space on the left, so you'd have:
def a_method_with_a_very_long_name_and_accepting_lots_of_parameters(
    self, foo, bar, *args, **kwargs):

```

src/d3a/models/strategy/external_strategies/load.py
fievelk: Suppose I (as a user) send 4 bids (1 is the oldest and 4 is the most recent):
- Bid 1 (replace_existing: True)
- Bid 2 (replace_existing: True)
- Bid 3 (replace_existing: False)
- Bid 4 (replace_existing: True)

I expect that the most recent bid I sent (4) will replace all the others because it has `replace_existing=True` and it's the last order I sent. However, using a normal list we would have a LIFO behavior and what will happen is that:
- (4) will be the first bid request to be parsed, so it won’t find any existing bids in the market (and it won’t replace anything).
- Then (3) will be parsed, and it won’t replace anything because it has replace_existing=False (so now we have bids (4) and (3) in the market).
- Then (2) will be parsed. Since it has replace_existing=True, it will delete the bids (4) and (3) from the market. However, those bids are the most recent ones, so I would be expecting the opposite behavior.

That's why using a FIFO object like deque helps to keep the right order.

faizan2590: Thanks a verbose explanation, totally makes sense now

faizan2590 (dismissed):

fievelk:
- Added the `replace_existing` parameter for offers as well (not just bids)
- implemented and refactored tests

hannesdiedrich (dismissed):
LGTM, please dont push before the deploy to production. Thanks!

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -321,7 +338,14 @@ def _update_offer_aggregator(self, arguments):
...
 
     def _offer_aggregator(self, arguments):
-        assert set(arguments.keys()) == {'price', 'energy', 'transaction_id', 'type'}
+        required_args = {'price', 'energy', 'type', 'transaction_id'}
+        allowed_args = required_args.union({'replace_existing'})
+
+        # Check that all required arguments have been provided
+        assert all(arg in arguments.keys() for arg in required_args)
+        # Check that every provided argument is allowed
+        assert all(arg in allowed_args for arg in arguments.keys())
```
Hassan754, line 347: I see that this is used more than 5 times in the PR, might be a good candidate for a reusable function

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -33,6 +35,15 @@ class StorageExternalMixin(ExternalMixin):
...
 
+    @property
+    def filtered_bids_next_market(self) -> List[Dict]:
+        """Get a representation of each of the device's bids from the next market."""
+
+        return [
+            {'id': bid.id, 'price': bid.price, 'energy': bid.energy}
+            for _, bid in self.next_market.get_bids().items()
+            if bid.buyer == self.device.name]
+
```
Hassan754, line 46: IMO it would be great to add this property in the `ExternalMixin` class, as the users of it are direct children of it.
I would also vote for making it a bit more generic ( if it can return either bids or offers )

Hassan754 (commented):
Left some comments, LGTM in general

src/d3a/models/strategy/external_strategies/pv.py
fievelk: I agree: this should be refactored. However, I think that the refactoring should do more than just moving these things around: it would be good to have a validator class (or something similar) that can separate this logic from the external strategies classes. That's why I think it would be a candidate for a future task.

src/d3a/models/strategy/external_strategies/storage.py
fievelk: It's true that all the classes that use `filtered_bids_next_market` inherit from `ExternalMixin`. However, not all classes that inherit from `ExternalMixin` use `filtered_bids_next_market`.

- `LoadExternalMixin`
- `PVExternalMixin`  # This class doesn't make bids
- `StorageExternalMixin`

The same applies for the offers.
I agree that having this duplication isn't nice nor DRY. But I don't think that adding it to `ExternalMixin` would be correct (since it doesn't apply to all the children classes). In theory, this would probably belong to a new class (something like `ExternalBidMixin`) but it would probably be a bit too much for one method. I'm not sure, to be honest.

faizan2590 (dismissed):
LGTM

src/d3a/models/strategy/__init__.py
hannesdiedrich: @fievelk sorry, this has not been change. You can delete line 492 and 493. `remove_bid_from_pending` is taking care of the deletion in the market and in the strategy.

src/d3a/models/strategy/external_strategies/storage.py
spyrostz: I think we have discussed it again, but we were under the same dilemma when implementing these classes. The most straightforward way to do this is to implement separate ExternalBidMixin and ExternalOfferMixin mixins, as @fievelk suggested. However, this would bloat our class hierarchy, because the class hierarchy tree would become more complex. Therefore we decided to bite the bullet and do the code a bit less DRY than to complicate our class hierarchy further. 
Having said that, we can revisit it in the future and see if it makes sense now to create such mixins. Not in the content of this PR though.

spyrostz (approved):
LGTM

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -122,39 +133,68 @@ def _delete_offer_impl(self, arguments, response_channel):
...
+            # Do not consider previously offered energy, since those offers would be deleted
+            return (
+                offer_arguments['energy'] <=
+                self.state.energy_to_sell_dict[time_slot] +
+                self.state.offered_sell_kWh[time_slot])
+        else:
+            return (
+                offer_arguments['energy'] <=
+                self.state.energy_to_sell_dict[time_slot])
+
```
hannesdiedrich, line 180: so you are overwriting the bastrategy function here for the storage, right?
I suggest to rename this function to something like `can_storage_offer_be_posted`, because this is specific to the storage. Also because of the following point.
I realise that we dont check for the `has_battery_reached_max_power` in the external storage strategy (have a look into Storage.sell_energy and Storage.buy_energy.
This should be added in here. However, As this PR is getting long, we could add it in a separate to-be-added task.  FYI @spyrostz @faizan2590

hannesdiedrich (commented):
Two last comments, once addressed, LGTM

src/d3a/models/strategy/external_strategies/storage.py
fievelk: Yes, as you say this is specific to the Storage external strategy. Actually, I named it in the same way as the one in `BaseStrategy` exactly to override that method, since that method would cause an exception if called from a `StorageExternalStrategy`.
I have the feeling that the relation between mixins and strategies is a bit problematic, or at least I sometimes confuse the two. In theory, I would assume that a class that uses a mixin should be able to call all of its methods (or required to implement its own version), but this is not the case for `StorageExternalStrategy`.

Please let me know if you still think I should proceed and rename `can_offer_be_posted` to `can_storage_offer_be_posted`.

faizan2590: Since in this case `can_offer_be_posted ` is already rewritten inside `strategy/external_strategies/storage.py`, to me its  actually not really needed to explicitly name it `can_storage_offer_be_posted `

faizan2590 (approved):
LGTM

src/d3a/models/strategy/external_strategies/storage.py
faizan2590: > I realise that we don't check for the has_battery_reached_max_power in the external storage strategy (have a look into Storage.sell_energy and Storage.buy_energy.

Sorry missed that part earlier. Yes, I agree `has_battery_reached_max_power ` has to be checked too, better make a separate enhancement ticket as suggested.

spyrostz: Sorry for jumping in here, but I disagree, as far as I see `energy_to_sell_dict` is already clamped by `_battery_energy_per_slot` (which is the energy that the battery can consume / produce by its max power) , have a look [here](https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/state.py#L321). Therefore I think that the calculation is ok and no further checks need to be done.

hannesdiedrich: Yes, you are of course right, and we call `clamp_energy_to_sell_kWh` and `clamp_energy_to_buy_kWh` every time and offer or bid is posted. So all good. I actually was surprised when i thought that this is not being taken care of.
But we went through a lot of iterations about that already indeed.
SO, all good. ignore me. :-) Thanks a lot!

hannesdiedrich: Also, the renaming is not needed, as Faizan correctly mentioned.

hannesdiedrich (approved):
LGTM

PR #1039, opened 2021-01-28
D3ASIM_3216: Fixed: Only track one of two trade events in _publish_tr…

hannesdiedrich:
…ade_event when cleared in the same market.


Hassan754 (dismissed):

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -242,7 +242,7 @@ def _publish_trade_event(self, trade, is_bid_trade):
             return
 
         if ConstSettings.IAASettings.MARKET_TYPE != 1 and \
-                trade.buyer == self.device.name and \
+                trade.seller == self.device.name and \
```
spyrostz, line 245: I am not sure that this is going to do the trick here. First of all, this will make the Load receive 2 trade events at the same time, therefore it moves the issue to the demand side. Furthermore, I suspect that this change will not report trade events to devices that only have offer trades and no bid trades (keep in mind that 2 trade events are emitted only in the market that the bid and offer are matched, all other markets emit only one event).
My suggestion would be to add a second condition here:
`if ConstSettings.IAASettings.MARKET_TYPE != 1 and ((trade.buyer == self.device.name and isinstance(trade.offer, Offer)) or (trade.seller == self.device.name and isinstance(trade.offer, Bid)))`

spyrostz (commented):
One comment, will look again once dealt with, and also some unit tests will be nice to have here, especially since unit tests that check the event_trade behavior already exist (e.g. test_dispatch_event_trade_to_external_agent)

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: Yeah, looks like I was too quick. Let me look into it again an also test your suggestion (will also add tests)

spyrostz (approved):
LGTM

Hassan754 (approved):
LGTM

PR #1042, opened 2021-02-02
DD-14: added the energy and rate from the actual trade

lana-shanghai:


src/d3a/models/market/blockchain_interface.py
```diff
@@ -101,8 +101,13 @@ def track_trade_event(self, trade):
...
             'buyer': ALICE_STASH_ADDRESS,
             'seller': BOB_STASH_ADDRESS,
-            'amount': default_amount,
-            'rate': default_rate
+            'amount': {
+                'free': trade.offer.energy * BC_NUM_FACTOR,
+                'reserved': 0,
+                'miscFrozen': 0,
+                'feeFrozen': 0
+            },
```
faizan2590, line 109:
```suggestion
           'energy': trade.offer.energy,
```
Once this PR(https://github.com/gridsingularity/canvas-node/pull/7) from canvas-node side is merged

faizan2590 (commented):
Left a comment

src/d3a/models/market/blockchain_interface.py
lana-shanghai: Already had the commit lined up :)

src/d3a/models/market/blockchain_interface.py
```diff
@@ -101,8 +101,8 @@ def track_trade_event(self, trade):
             'trade_id': trade.id,
             'buyer': ALICE_STASH_ADDRESS,
             'seller': BOB_STASH_ADDRESS,
-            'amount': default_amount,
-            'rate': default_rate
+            'energy': trade.offer.energy * BC_NUM_FACTOR,
```
faizan2590, line 104:
```suggestion
            'energy': trade.offer.energy,
```
I think `BC_NUM_FACTOR ` could be skipped

lana-shanghai: The reason why ETH is calculated in Wei is because Weis are always whole numbers, and ETH are not. So they use 10**18. Here they have the 10**12 factor, I would rather leave it to avoid any precision or rounding problems

faizan2590: `energy` isn't now in  ETH or any currency like `AccountData` but its now simply in `u16`

lana-shanghai: OK makes sense! Edited

faizan2590 (approved):
LGTM

spyrostz (approved):
LGTM

PR #1043, opened 2021-02-02
Result exports bug fix

Hassan754:


hannesdiedrich (dismissed):
LGTM

spyrostz (commented):
Change looks good to me indeed, however it would be great to repeat the behavior in the self.max_trade_price (use None value if there is no trades in the market) for consistency.

spyrostz (approved):
LGTM, thanks for the extra fix on the min/max_offer_price!

hannesdiedrich (approved):
LGTM

PR #1046, opened 2021-02-08
Explicitly kill workhorse in case Redis connection fails. That way it…

spyrostz:
… is guaranteed that the workhorse processes will terminate before the parent process.


mclrch (approved):
LGTM

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #1050, opened 2021-02-09
Adapted the state restore to be more tolerant and update the existing…

spyrostz:
… fields instead of replacing them. That way the strategy will be more flexible and automatically fall back to defaults in case the state that is restored is missing some of the expected datetimes.


hannesdiedrich (approved):
LGTM
(thanks for the explanation, needed it)!

Hassan754 (approved):

PR #1055, opened 2021-02-10
D3ASIM-3279: Added more unit test coverage for the energy bills to co…

spyrostz:
…ver individual methods.


setup.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 
-d3a_interface_branch = os.environ.get("BRANCH", "master")
+d3a_interface_branch = os.environ.get("BRANCH", "bug/D3ASIM-3279")
```
hannesdiedrich, line 5: Dont forget to revert before pushing, thanks!

tests/helpers.py
```diff
@@ -0,0 +1,9 @@
+from deepdiff import DeepDiff
+
+
+def compare_dicts(dict1, dict2):
```
hannesdiedrich, line 4: Minor, but how about giving a hint of the return value type already in the method name?
```suggestion
def area_dicts_identical(dict1, dict2):
```

hannesdiedrich: Or you add the assert already here, just like `_compare_bills `

hannesdiedrich (dismissed):
One very minor comment.
LGTM

tests/helpers.py
spyrostz: I think that I prefer your second solution for uniformity and rename these methods to `assert_dicts_identical` and `assert_lists_contain_same_elements`. Fixing it

tests/helpers.py
```diff
@@ -0,0 +1,9 @@
+from deepdiff import DeepDiff
+
+
+def assert_dicts_identical(dict1, dict2):
+    assert len(DeepDiff(dict1, dict2, ignore_order=True)) == 0
+
+
+def assert_lists_contain_same_elements(list1, list2):
+    assert set(list1) == set(list2)
```
faizan2590, line 9: good candidates for `d3a-interface`

hannesdiedrich (approved):
LGTM

faizan2590 (dismissed):
one suggestion, other than that LGTM

tests/helpers.py
spyrostz: Only reason why I didn't port them, was because I thought it was too premature to create a utils module in the d3a-interface for helpers in unit tests. Since you also mentioned that though, you changed my mind, adapting the PR. Thanks!

PR #1058, opened 2021-02-12
D3ASIM-3301: Fix bug energy value = None

Hassan754:


hannesdiedrich (approved):
LGTM

src/d3a/models/strategy/__init__.py
```diff
@@ -311,6 +311,8 @@ def accept_offer(self, market_or_id, offer, *, buyer=None, energy=None,
             buyer = self.owner.name
         if not isinstance(offer, Offer):
             offer = market_or_id.offers[offer]
+        if energy is None:
+            energy = offer.energy
```
spyrostz, line 315: I doubt that this is the root cause of the error. Even if energy is passed as None in the accept_offer method, the accept_offer and accept_bid will both populate the energy value correctly, see here for instance https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/market/one_sided.py#L211. I would have a more detailed look in 3 places to find the culprit: 
1. Pay a closer look at the external strategies, and more specifically the way they transmit trade events and how they populate these messages
2. Trade class serialization code, to see if there is an evident problem there
3.Have a look at the trades in the DB around that time, to see if there are any trades with incorrect values.

hannesdiedrich: This one goes on me, I suggested this change. I was too quick. You are right, the accept_offer method in the markets deals with that. Will help you find the culprit in the other suggested places @Hassan754

hannesdiedrich: #3 can be ruled out, because I fetched all bids/offers/trades from 12 till 7PM this day (11.2.2021) and there are no energy values set to 0.

fievelk: Do you think that writing a test for this would be useful? I mean creating a failing scenario that reproduces the current situation and then making sure that the scenario passes once the bug is fixed.

hannesdiedrich: But I just realised, that we can also have the case, that there is no energy member sent at all. In this case we would get the same error. Ill have another look in the data (this will be hard to detect).

PR #1062, opened 2021-02-15
D3ASIM-3302: Added to state the information whether an API client was…

spyrostz:
… already connected to a specific area. Also saved information about whether to use the template strategy or the external strategy.


hannesdiedrich (approved):
LGTM

Hassan754 (approved):

PR #1063, opened 2021-02-16
D3ASIM-3314: Add not empty current_market value in the results only i…

spyrostz:
…f the initial market slot has passed on CNs. Checked by the time difference between the current and the next market slot.

### DIFF
--- a/src/d3a/d3a_core/sim_results/endpoint_buffer.py
+++ b/src/d3a/d3a_core/sim_results/endpoint_buffer.py
@@ -33,6 +33,7 @@ from d3a.models.strategy.load_hours import LoadHoursStrategy
 from d3a.models.strategy.finite_power_plant import FinitePowerPlant
 from d3a.models.strategy.infinite_bus import InfiniteBusStrategy
 from d3a.models.strategy.market_maker_strategy import MarketMakerStrategy
+import d3a.constants
 
 _NO_VALUE = {
     'min': None,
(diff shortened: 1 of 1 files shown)


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -228,7 +229,10 @@ def _populate_core_stats_and_sim_state(self, area):
     def update_stats(self, area, simulation_status, progress_info, sim_state):
         self.area_result_dict = self._create_area_tree_dict(area)
         self.status = simulation_status
-        if area.current_market is not None:
+        is_initial_current_market_on_cn = d3a.constants.IS_CANARY_NETWORK and \
+            (area.next_market is None or
+             area.next_market.time_slot - area.current_market.time_slot > area.config.slot_length)
```
hannesdiedrich, line 234: This is when the CN gets restarted and the `area.current_market.time_slot` is 00:00 right?

hannesdiedrich (approved):
One clarification, LGTM

src/d3a/d3a_core/sim_results/endpoint_buffer.py
spyrostz: When the Cn starts, the market slots are the following: 
initializing: current_market == None, next_market == None
first cycle: current_market == 00:00 , next_market == <current_time>
second cycle: current_market == <current_time>, next_market == <current_time> + 15 

Therefore, the clause `area.next_market is None` catches the initializing state, and the `area.next_market.time_slot - area.current_market.time_slot > area.config.slot_length` catches the first cycle. Essentially my point was to find out that, if the current and the next slot have a time difference of more than 15 mins (which the expected one) then it is the case where the current market is at the start of the day, and the next market is at the current time.

There is a cornercase here that I have deliberately not handled, which is if someone starts a CN at e.g. 00:02, the CN fails, and reinitialized in the same market slot (before 00:15). In this case this check will not work, because the diff between the market slots is going to be less than the slot length. Nevertheless it does not matter, because in this case the current market is indeed 00:00 , therefore its value will be correct. 

Hope this clarifies the fix, it hides more than it looks :D

hannesdiedrich: Thanks a lot!

faizan2590 (approved):
LGTM

PR #1070, opened 2021-02-19
D3ASIM-3210: Fixed bug in redis.externaal_market_connection.event_mar…

hannesdiedrich:
…ket_cycle.

### DIFF
--- a/src/d3a/models/area/redis_external_market_connection.py
+++ b/src/d3a/models/area/redis_external_market_connection.py
@@ -173,13 +173,15 @@ class RedisMarketExternalConnection:
         market_info["self_sufficiency"] = \
             self.area.stats.kpi.get("self_sufficiency", None)
         market_info["area_uuid"] = self.area.uuid
-        data = {"status": "ready",
-                "event": "market",
-                "market_info": market_info}
         if self.is_aggregator_controlled:
+            market_info["event"] = "market"
+            market_info["status"] = "ready"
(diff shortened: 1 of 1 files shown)


src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -173,13 +173,15 @@ def event_market_cycle(self):
         market_info["self_sufficiency"] = \
             self.area.stats.kpi.get("self_sufficiency", None)
         market_info["area_uuid"] = self.area.uuid
-        data = {"status": "ready",
-                "event": "market",
-                "market_info": market_info}
         if self.is_aggregator_controlled:
+            market_info["event"] = "market"
+            market_info["status"] = "ready"
```
Hassan754, line 178: if the market_info are not needed in case controlled by an aggregator ( I did not understand this ), you can skip building up the object, maybe you can put the check at the top of function

hannesdiedrich: If the aggregator is only controlling one market, the market_info is indeed needed to be sent.
I know, there is a lot of duplicated data sent. But this we need to handle in the following API stories.
Do you agree?

Hassan754: @hannesdiedrich right. I'm actually looking forward for all these deprecations/adaptations tickets

Hassan754 (approved):
LGTM

spyrostz (approved):
LGTM

PR #1071, opened 2021-02-22
Increase Redis external connection and ResettableCommunicator base cl…

spyrostz:
…ass thread sleep time, in order to not spam the Canary Network CPU.


mclrch (approved):
LGTM

PR #1074, opened 2021-02-23
D3ASIM-3325: Added remove_from_aggregaor when unregister of external …

hannesdiedrich:
…startegy is called.

In order to be able to unregister a device also from the d3a-web side, we need to decouple it also from the aggregator that might be trading in behalf of it. 
In this PR the device_uuid is removed from the aggregator mapping that makes the external strategy automatically using the template strategy for trading.


src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -159,8 +165,13 @@ def _register(self, payload):
     def _unregister(self, payload):
         self._connected = unregister_area(self.redis, self.channel_prefix, self.connected,
                                           self._get_transaction_id(payload))
+        if self.is_aggregator_controlled:
+            self._remove_from_aggregator = True
 
     def register_on_market_cycle(self):
+        if self._remove_from_aggregator:
```
spyrostz, line 172: In order to avoid keeping track of the extra `_remove_from_aggregator` variable in the whole class, you could reuse the `connected` and `_connected` variables to detect if an unregister happened, by this clause: `if self.connected is True and self._connected is False`. This implies that the strategy was registered, but about to unregister. Of course, it is better to move this clause in a separate method in order for it to be self-documenting.

spyrostz (commented):
One improvement suggestion, once resolved LGTM

faizan2590 (approved):
LGTM

Hassan754 (approved):
LGTM

PR #1076, opened 2021-02-24
DD-10: Adding simulation_id, market_id and slot_info for trade tracki…

faizan2590:
…ng on substrate.


src/d3a/models/area/markets.py
```diff
@@ -117,6 +119,8 @@ def create_future_markets(self, current_time, is_spot_market, area):
             if timeframe not in markets:
                 # Create markets for missing slots
                 market = market_class(
+                    simulation_id=str(uuid.uuid4()),
```
faizan2590, line 122: To be replaced with actual `simulation_id`.

lana-shanghai: Why can't it be replaced with the actual simulated id?

lana-shanghai (approved):
LGTM

src/d3a/models/area/markets.py
```diff
@@ -27,12 +27,14 @@
 from d3a_interface.constants_limits import ConstSettings
 from collections import OrderedDict
 from d3a.d3a_core.util import is_timeslot_in_simulation_duration
-from d3a import constants
+import d3a.constants
 
 
 class AreaMarkets:
-    def __init__(self, area_log):
+    def __init__(self, market_id, area_log):
```
spyrostz, line 34: You do not need this now

tests/test_market.py
```diff
@@ -53,7 +54,8 @@
 
 class FakeTwoSidedPayAsBid(TwoSidedPayAsBid):
     def __init__(self, bids=[], m_id=123, time_slot=now()):
-        super().__init__(grid_fees=transfer_fees, time_slot=time_slot)
+        super().__init__(bc=NonBlockchainInterface(str(uuid4())),
```
spyrostz, line 57: I would rather use a Mock here to avoid importing the NonBlockchainInterface

src/d3a/d3a_core/rq_job_handler.py
```diff
@@ -25,6 +25,7 @@ def launch_simulation_from_rq_job(scenario, settings, events, aggregator_device_
                                   saved_state, job_id):
     logging.getLogger().setLevel(logging.ERROR)
     scenario = decompress_and_decode_queued_strings(scenario)
+    d3a.constants.SIMULATION_ID = job_id
```
spyrostz, line 28: I think you can adapt it if you pass it as an extra argument in the activate() method of the area class. After all, this is the only place where the blockchain objects are constructed.

spyrostz (commented):
Some more comments, but in general looks good. One more thing that needs to be checked out is if it works correctly with deep hierarchies. The reason why I am mentioning this is that the Area class' activate method broadcasts the activate to children like this `self.dispatcher.broadcast_activate(current_tick=self.current_tick)`, which makes me think that the bc argument is never forwarded to children. Therefore, if this does not work for child areas, changing it like this `self.dispatcher.broadcast_activate(bc=bc, current_tick=self.current_tick)` would do the trick in order to force the broadcast of the bc argument.

spyrostz (dismissed):
LGTM

lana-shanghai (approved):
LGTM! Looking forward to the motivation behind the refactoring :)

src/d3a/models/market/blockchain_interface.py
```diff
@@ -95,14 +99,17 @@ def handle_blockchain_trade_event(self, offer, buyer, original_offer, residual_o
...
+            'simulation_id': str(self.simulation_id),
+            'market_id': str(self.market_id),
+            'market_slot': datetime.timestamp(time_slot),
             'trade_id': trade.id,
             'buyer': ALICE_STASH_ADDRESS,
             'seller': BOB_STASH_ADDRESS,
-            'energy': trade.offer.energy,
-            'rate': trade.offer.energy_rate
+            'energy': int(trade.offer.energy),
+            'rate': int(trade.offer.energy_rate)
```
spyrostz, line 112: Why do you need to cast these as int ? Especially for the energy value, this usually is between 0-1, which means that this casting would immediately convert it to zero. I would omit it actually.

spyrostz (approved):
One comment, LGTM once resolved.

src/d3a/models/market/blockchain_interface.py
faizan2590: The reason I need to cast them to `int`is due to pallet don't accept floating point data type.
If you agree I will do the conversion of energy from kWh to Wh so integer casting doesn't round it to zero.

spyrostz (approved):
LGTM

PR #1079, opened 2021-02-26
D3ASIM-3331: Added root-level market_info / WIP

Hassan754:


src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -183,11 +185,16 @@ def event_market_cycle(self):
...
-                market_info.update(default_market_info)
+            current_tick = self.device.current_tick % self.device.config.ticks_per_slot
+            market_info = MarketInfo(
+                event="market",
+                simulation_id=d3a.constants.COLLABORATION_ID if
+                d3a.constants.EXTERNAL_CONNECTION_WEB else None,
+                num_ticks=self.device.config.total_ticks,
+                market_slot=self.area.current_slot.time_slot_str,
+                slot_completion=f"{int((current_tick / self.device.config.ticks_per_slot) * 100)}%"
+            ).__dict__
```
hannesdiedrich, line 196: yes, let me create my PR first.
But i can already comment: You didnt select only the root_area here . this structure is returned for every market now. But lets deal with this later.
Also; can you please tell my why a dataclass is good practice here? I dont see the benefit, as you convert it into a dict right afterwards.

Hassan754: @hannesdiedrich it looks like they are referring to the first node in dict https://gridsingularity.atlassian.net/browse/D3ASIM-3331?focusedCommentId=21807, not the root area
Regarding the dataclass, I wanted to centralize the market_info and tick_info somewhere instead of having them as dicts in different files

spyrostz: It would be great if you tried to isolate the construction of the dataclass in a separate function inside the `dataclasses.py`, in order to be called by other places in the code if needed. 
I personally like the dataclass approach because, contrary to the dict, there can be 2 different types for the Tick and Market Info, thus retaining the type information, while with a raw dict we cannot do that. However, I would advocate instead of using the dunder method, to expose a serialization method in these dataclasses (for instance, named as_dict()). This will of course be more verbose, but then we will not pollute the caller code with the dunder method, but hide it inside the implementation of the dataclass.

PR #1082, opened 2021-03-01
Bug/d3 asim 3321: Tracking buyer/sellers with their uuids

faizan2590:


src/d3a/models/market/market_structures.py
```diff
@@ -33,14 +33,17 @@ def my_converter(o):
 
 class Offer:
     def __init__(self, id, time, price, energy, seller,
-                 original_offer_price=None, seller_origin=None):
+                 original_offer_price=None, seller_origin=None, seller_origin_id=None,
+                 seller_id=None):
```
faizan2590, line 37: May be the higher order objects be better now as arguments for classes and methods with too many arguments:
https://stackoverflow.com/questions/5899185/class-with-too-many-parameters-better-design-strategy

Using `dataclasses ` sounds like a good approach for class/method with too many arguments.

hannesdiedrich (approved):
LGTM

hannesdiedrich:
@faizan2590 Can you please check how much the amount of data in the DB increases with this change? We might have to think about stripping away the uuid parts in the Bids/Offers/Trades.
Also I just thought: is the seller_id really needed? Because as of the current implementation, the seller_name is unique by design (as there are only unique names in one parent area)

spyrostz:
> @faizan2590 Can you please check how much the amount of data in the DB increases with this change? We might have to think about stripping away the uuid parts in the Bids/Offers/Trades.
> Also I just thought: is the seller_id really needed? Because as of the current implementation, the seller_name is unique by design (as there are only unique names in one parent area)

In order to be able to recreate the stats if needed I think we cannot avoid saving the uuids to the DB for the trades. For the bids and the offers though, we can definitely get away by removing the uuids completely.

spyrostz (dismissed):
LGTM

faizan2590:
> > @faizan2590 Can you please check how much the amount of data in the DB increases with this change? We might have to think about stripping away the uuid parts in the Bids/Offers/Trades.
> > Also I just thought: is the seller_id really needed? Because as of the current implementation, the seller_name is unique by design (as there are only unique names in one parent area)
> 
> In order to be able to recreate the stats if needed I think we cannot avoid saving the uuids to the DB for the trades. For the bids and the offers though, we can definitely get away by removing the uuids.

After offline discussion uuid part of Bids/Offers/Trades to be kept in DB for later to decide.

> Also I just thought: is the seller_id really needed? Because as of the current implementation, the seller_name is unique by design (as there are only unique names in one parent area)

Yes they are needed in order to avoid double counting of trade in intermediate hierarchy(https://github.com/gridsingularity/d3a-interface/pull/176/files#diff-4fe53216bd5f83d932a02c99f5062b7714e165902d3aad6274ef33e009b5b2cbR61-R63)

integration_tests/steps/integration_tests.py
```diff
@@ -390,6 +390,17 @@ def test_export_data_csv(context, scenario):
...
 
+@then('we get the area tree summary')
+def area_tree(context):
+    get_simulation_raw_results(context)
+    # tree_path = glob.glob(os.path.join(context.export_path, os.listdir(context.export_path)[0],
+    #                                    'area_tree_summary.json'))
+    # with open(tree_path[0]) as tree_summary:
+    #     context.tree_data = json.loads(tree_summary.read())
+    # assert context.tree_data
+    # context.area_name_uuid_map = get_area_name_uuid_mapping(context.tree_data)
```
spyrostz, line 401: Please remove these before merging

spyrostz (dismissed):
LGTM

spyrostz (dismissed):
LGTM

integration_tests/steps/integration_tests.py
```diff
@@ -390,6 +390,11 @@ def test_export_data_csv(context, scenario):
                                                                     file=data_fn))
 
 
+@then('we get the area tree summary')
+def area_tree(context):
+    get_simulation_raw_results(context)
```
fievelk, line 395: If this is not asserting anything and it's just performing some operations, I'd suggest using a `@when` step (and using it in the `.feature` files as `and`).

fievelk: What happens if this step is not performed? If all `test_export_of_kpi_result` must necessarily use this step, I think it would make sense to directly add it inside of `test_export_of_kpi_result`. Please let me know if I'm missing something!

src/d3a/models/market/market_structures.py
fievelk: I agree. Just by glancing at it, I have the feeling we could benefit from a `Seller` class to store some of those arguments.

src/d3a/d3a_core/rq_job_handler.py
```diff
@@ -23,7 +23,7 @@ def decompress_and_decode_queued_strings(queued_string):
 
 def launch_simulation_from_rq_job(scenario, settings, events, aggregator_device_mapping,
                                   saved_state, job_id):
-    logging.getLogger().setLevel(logging.ERROR)
+    logging.getLogger().setLevel(logging.INFO)
```
hannesdiedrich, line 26: please revert

integration_tests/kpi.feature
```diff
@@ -4,76 +4,87 @@ Feature: KPI integration tests
      Given we have a scenario named kpi.house_not_self_sufficient
      And d3a is installed
      When we run the d3a simulation on console with kpi.house_not_self_sufficient for 24 hrs
+     Then we get the area tree summary
```
hannesdiedrich, line 7: Can you please explain, why this is needed?

hannesdiedrich: explained offline. thanks

integration_tests/steps/integration_tests.py
hannesdiedrich: I agree with Pierpaolo. 
Please add these lines to test_export_of_kpi_result
```
    area_tree_summary = glob.glob(os.path.join(context.export_path, "*", "area_tree_summary.json"))
    with open(area_tree_summary[0], "r") as sf:
        context.area_tree_summary_data = json.load(sf)
```

hannesdiedrich (dismissed):
Two more comments, LGTM once resolved

spyrostz (dismissed):
LGTM

fievelk (approved):

PR #1083, opened 2021-03-02
Feature/d3 asim 3333 Adapted to new market_info and tick_info event dict format.

hannesdiedrich:


src/d3a/d3a_core/util.py
```diff
@@ -572,3 +573,41 @@ def wrapper(*args, **kwargs):
...
+
+def get_market_maker_rate_from_global_setting(next_market):
+    if isinstance(GlobalConfig.market_maker_rate, dict):
+        if next_market is None:
+            return GlobalConfig.market_maker_rate[next(iter(GlobalConfig.market_maker_rate))]
+        else:
+            return find_object_of_same_weekday_and_time(GlobalConfig.market_maker_rate,
+                                                        next_market.time_slot)
+    else:
+        return GlobalConfig.market_maker_rate
```
spyrostz, line 613: I think that this does not cover the case of islanded mode. I recall from a discussion with Chris, that he was interested in the feed-in-tariff, which is not exactly the market maker rate (we were supposed to report a None value if the market maker is not an infinite bus or commercial producer, but a simple market maker that sets a market maker rate and does not place offers and bids)

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -181,26 +182,8 @@ def event_market_cycle(self):
...
-                get_current_market_maker_rate(self.area.current_market.time_slot) \
-                if self.area.current_market else None
-            market_info['last_market_stats'] = \
-                self.market_area.stats.get_price_stats_current_market()
-            self.redis.publish_json(market_event_channel, market_info)
-            if self.is_aggregator_controlled:
-                self.redis.aggregator.add_batch_market_event(self.device.uuid,
-                                                             market_info,
-                                                             self.area.global_objects)
+            self.external_tick_timer.reset_event_tick_counter()
```
spyrostz, line 185: I particularly like refactoring the tick algorithm to a separate class (ExternalTickTimer). There are 2 improvements I see:
1. this class is not exactly a Timer in the way Python perceives the term (a thread that performs some actions at a definite time). I think it is better to be renamed to ExternalTickCounter or similar. Then you can rename the `reset_event_tick_counter()` to just `reset()`
2. I think that the access to this class can only live in the ExternalMixin, in order to not repeat the `reset_event_tick_counter` calls

spyrostz (commented):
Looks great for now, looking forward to the finalized version of it.

src/d3a/d3a_core/util.py
hannesdiedrich: Actually I forgot about the `feed_in_tariff_rate`. Will add it and of course. Thanks for reminding.
The above was just a refactoring of getting the market_maker as it was called in the load, the pv and now also in the global_objects.

src/d3a/d3a_core/redis_connections/aggregator_connection.py
```diff
@@ -30,8 +23,15 @@ def __init__(self, redis_db):
...
         self.grid_buffer = {}
 
-    def set_aggregator_device_mapping(self, aggregator_device_mapping):
-        self.aggregator_device_mapping = aggregator_device_mapping
+    @property
+    def _aggregator_info(self):
+        """
+        to be implemented
+        """
+        return {}
```
hannesdiedrich, line 31: Didnt read the discussion closely, will remove this as it is not needed any more.

src/d3a/d3a_core/util.py
hannesdiedrich: Actually, no there is a ticket D3ASIM-3350 for that. So, lets add this later.

src/d3a/d3a_core/global_objects.py
```diff
@@ -15,29 +15,39 @@
...
-                area_dict = {"last_market_slot": area.current_market.time_slot_str,
-                             "last_market_bill": area.stats.get_last_market_stats_for_grid_tree(),
-                             "last_market_stats": area.stats.get_price_stats_current_market(),
-                             "last_market_fee": area.current_market.fee_class.grid_fee_rate,
-                             "next_market_fee": area.get_grid_fee(),
-                             "children": {}}
+                area_dict = {'market_maker_rate': get_market_maker_rate_from_config(
+                                                  area.current_market),
+                             'last_market_bill': area.stats.get_last_market_stats_for_grid_tree(),
+                             'last_market_stats': area.stats.get_price_stats_current_market(),
```
faizan2590, line 39: Would be a good opportunity to settle on terms `last_market` vs `current_market`

faizan2590: Please make a task for research/testing team to settle on nomenclature when to use `last_market/current_market/next_market`

candrews42: If the time is 10:13, energy is being delivered for the 10:00 market slot, and the current spot market is the range 10:15-10:30. The active spot market is considered 'current'.

last_market = 10:00-10:15
current_market = 10:15-10:30
next_market = 10:30-10:45

src/d3a/d3a_core/simulation.py
```diff
@@ -169,7 +170,11 @@ def _init(self, slot_length_realtime, seed, paused, pause_after, redis_job_id):
             log.info("Random seed: {}".format(random_seed))
 
         self.area = self.setup_module.get_setup(self.simulation_config)
+        # TODO: discuss whether the following circular reference is a good idea (probably not)
+        #  it is needed for being able to update the whole grid_tree from the devices
+        self.global_objects = GlobalStatistics(self.area)
```
faizan2590, line 175:
```suggestion
        self.global_stats = GlobalStatistics(self.area)
```

minor but since you renamed the class

faizan2590:
In general looks good, will give another look once comments are addressed.

src/d3a/d3a_core/simulation.py
```diff
@@ -169,7 +170,11 @@ def _init(self, slot_length_realtime, seed, paused, pause_after, redis_job_id):
             log.info("Random seed: {}".format(random_seed))
 
         self.area = self.setup_module.get_setup(self.simulation_config)
+        # TODO: discuss whether the following circular reference is a good idea (probably not)
+        #  it is needed for being able to update the whole grid_tree from the devices
```
spyrostz, line 174: I am fairly certain that we'll unfortunately regret the circular reference in the future. In this case, even though it is an antipattern, I think that having a "singleton" is much better than having to deal with strong references. I want to see the calling code of this class though, if I find a less intrusive way than a singleton I'll edit this comment.
If you decide to keep this version for whatever reason, I would at least make the references weak, in order to not prohibit areas that are deleted to be garbage-collected https://docs.python.org/3/library/weakref.html.

src/d3a/d3a_core/simulation.py
```diff
@@ -363,13 +371,19 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
 
                 self.area.tick_and_dispatch()
                 self.area.update_area_current_tick()
+                if self.simulation_config.external_connection_enabled and \
+                        self.external_tick_counter.is_it_time_for_external_tick(tick_no):
```
spyrostz, line 375: So the external_tick_counter in the simulation.py is only useful for calculating if the global_objects need to be updated? There is an alternative solution to that, which is to make the external_tick_counter a part of the `GlobalStatistics` class , and rename the class to `ExternalConnectionGlobalStatistics`. That way you will have only one counter for the whole simulation, and not separate counters per strategy. 
Also, this has the advantage that the `if self.simulation_config.external_connection_enabled and self.external_tick_counter.is_it_time_for_external_tick(tick_no)` checks can be moved inside the update() method, therefore all you will have to do from the caller code is `update()`, and the simulation.py code will be much cleaner.

src/d3a/d3a_core/simulation.py
```diff
@@ -363,13 +371,19 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
+                    self.global_objects.update()
 
                 self.simulation_config.external_redis_communicator.\
                     publish_aggregator_commands_responses_events()
 
                 self.handle_slowdown_and_realtime(tick_no)
                 self.tick_time_counter = time()
 
+            if self.simulation_config.external_connection_enabled:
+                self.external_tick_counter.reset()
```
spyrostz, line 385: As a followup from the comment above, if the GlobalStatistics class includes the external_tick_counter, then this reset() method could be populated automatically by the update() method (by adding one extra parameter which will indicate when does this update take place, on a tick event or a market event )

src/d3a/d3a_core/simulation.py
hannesdiedrich: @spyrostz unfortunately not, the `external_tick_counter` is also used in the `ExternalMixin._dispatch_event_tick_to_external_agent` were it was originating from. Now i also needed for deciding whether the global updates need to be updated.

src/d3a/d3a_core/simulation.py
hannesdiedrich: Interesting read. Thanks! But I fear that if not implemented correctly we still run into the memory leak hole (talking about the implementation of the `finalize` function) .
I would rather try to implement the Global_Statitics as a singleton. Should be straight forward and I already thought about that but remembered ("No, singletons are bad!").

src/d3a/d3a_core/simulation.py
spyrostz: Yes, but this is easy to circumvent if you implement the singleton `ExternalConnectionGlobalStatistics`. You can expose a public method `is_it_time_for_external_tick` on the singleton (or, even better, a property. Since you update the global statistics object on every tick, it is rather easy to buffer the 'is this tick supposed to update external agents' information) and use the buffered version, instead of directly calling the `is_it_time_for_external_tick`. 
In general, I think this singleton would be a very good first step to a change that I have been thinking of for quite some time. At the moment, we have scattered across our code various objects that contain state for the whole simulation. The ideal way to deal with this is to split this state into different objects (e.g. ConfigSettings for configuration, ExternalConnectionState for external connection-specific variables) and then create a wrapper class that will instantiate all these objects, with this class only being a singleton (or, even better, constructed in the Simulation class, which we know it is only one per simulation). Before we do something like this though, we have to isolate the state, that's why I am insisting to move the external_tick_counter inside the same class that handles other parts of external connection state. Even if you think about it conceptually, even if we instantiate the ExternalTickCounter multiple times, the ticks are essentially centralized (tick has the same value for all areas, before or after the event_tick)

hannesdiedrich: Ok, understood and agreed. I will add another singleton then (on top of global_statistics that i am implementing right now)

spyrostz: Awesome, just one last comment, the ExternalTickCounter does not need to be a singleton, since it is only going to be constructed by the ExternalConnectionGlobalStatistics singleton it can be a normal class.

spyrostz (commented):
Pending the implementation of the singleton, rest of the PR looks good though

src/d3a/d3a_core/simulation.py
hannesdiedrich: Thanks for the clarification! Would have done it wrong.

src/d3a/d3a_core/redis_connections/aggregator_connection.py
```diff
@@ -43,32 +37,47 @@ def is_controlling_device(self, device_uuid):
...
-        if aggregator_uuid not in batch_event_dict:
-            batch_event_dict[aggregator_uuid] = []
+    def _delete_not_owned_devices(self, indict, outdict):
+        for area_uuid, area_dict in indict.items():
+            if 'children' in area_dict:
+                self._delete_not_owned_devices(indict[area_uuid]['children'],
+                                               outdict[area_uuid]['children'])
+            else:
+                if area_uuid not in self.device_aggregator_mapping:
+                    outdict[area_uuid] = {'area_name': area_dict['area_name']}
```
Hassan754, line 54: I did not get how exactly you're deleting the not owned devices

src/d3a/d3a_core/global_objects.py
```diff
@@ -15,29 +15,72 @@
...
+                             'current_market_fee': area.get_grid_fee(),
+                             'area_name': area.name,
+                             'children': {}}
             else:
-                area_dict = {"children": {}}
-            outdict[area.name].update(area_dict)
+                area_dict = {'children': {},
+                             'area_name': area.name}
+            outdict[area.uuid].update(area_dict)
             for child in area.children:
```
Hassan754, line 81: A side note for the future, we should decide on whether we should use recursion or use the area traversal function for such cases

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -270,31 +275,60 @@ def _publish_trade_event(self, trade, is_bid_trade):
...
+                                   'time': trade.time.isoformat(),
+                                   'trade_price': trade.offer.price,
+                                   'traded_energy': trade.offer.energy,
+                                   'total_fee': trade.fee_price,
+                                   'local_market_fee':
+                                       self.area.current_market.fee_class.grid_fee_rate
+                                       if self.area.current_market is not None else "None",
+                                   'attributes': {},
+                                   'seller': trade.seller
+                                   if trade.seller_id == self.device.uuid else 'anonymous',
```
Hassan754, line 291: In [here](https://gridsingularity.atlassian.net/browse/D3ASIM-3199?focusedCommentId=21906)

> The trade only displays trades as “bought” energy which means that if a PV sells energy it logs “anonymous bought…”. This should not be the case, it should say “PV Sold…”

 Chris wanted to show the seller name, do we have a reason for returning anonymous ?

spyrostz (approved):
LGTM

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -270,31 +275,60 @@ def _publish_trade_event(self, trade, is_bid_trade):
...
-        event_response_dict["event_type"] = "buy" \
-            if trade.buyer == self.device.name else "sell"
-        event_response_dict[bid_offer_key] = trade.offer.id
-        trade_event_channel = f"{self.channel_prefix}/events/trade"
-        if self.connected:
-            self.redis.publish_json(trade_event_channel, event_response_dict)
-
         if self.is_aggregator_controlled:
+            event_response_dict = {'event': 'trade',
+                                   'asset_id': self.device.uuid,
```
Hassan754, line 280: we can build the common keys outside the conditional if, but let's leave it for now

Hassan754 (approved):
LGTM

src/d3a/d3a_core/global_objects.py
hannesdiedrich: Please keep in mind that I have to create an outdict in the same structure as the indict here. Not sure whether this is possible with a traversal function

src/d3a/d3a_core/redis_connections/aggregator_connection.py
hannesdiedrich: Normally the `outdict[area_uuid]` would contain a lot of information, but here we only expose the name.

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: Yeah, this is from the time when i thought that the direct connection will be deprecated.

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: @Hassan754 yes, I thought you agreed with him to change it after 3333 is pushed. Right?

PR #1088, opened 2021-03-09
D3ASIM-3348: Hotfixes for d3a-api-client.

faizan2590:


src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -295,6 +295,8 @@ def _bid(self, payload):
 
             arguments['buyer'] = self.device.name
             arguments['buyer_origin'] = self.device.name
+            arguments['buyer_origin_id'] = self.device.uuid
+            assert self.can_bid_be_posted(self.next_market.time_slot, **arguments)
```
hannesdiedrich, line 299: I agree that is should be here. But please remove it from `_bid_impl `

src/d3a/setup/strategy_tests/external_ess_bids.py
```diff
@@ -31,7 +31,7 @@ def get_setup(config):
                 'House 1',
                 [
                     Area('PV', strategy=PVStrategy(
-                        max_panel_power_W=2000, initial_selling_rate=30, final_selling_rate=30.0)
+                        max_panel_power_W=2000, initial_selling_rate=40, final_selling_rate=30.0)
```
hannesdiedrich, line 34: Why is this needed?

src/d3a/models/strategy/external_strategies/storage.py
hannesdiedrich: Same goes for `can_offer_be_posted` actually.

hannesdiedrich: Or shall we check only shortly before the bid is added?
In any case, this needs to be homogenised.

hannesdiedrich (commented):
Left a comment and a question.

src/d3a/setup/strategy_tests/external_ess_bids.py
faizan2590: This was discovered while testing `integration_tests/test_ess_bid_connection.py` on `d3a-api-client` side when the bids placed by ess was instantly matched by `pv` leading to [offered_buy_kWh](https://github.com/gridsingularity/d3a-api-client/pull/113/commits/26893ff1560aa95a4dcbfc57091f7f74ad93ac57#diff-9487872865372fcdd22510280a797481a6ccc35e7cc7f9071cd63006516c2631R32) not equivaled to cumulative placed bid energy as parts of it was already matched by pv

src/d3a/models/strategy/external_strategies/storage.py
faizan2590: I would prefer to be in `_bid/_offer` than `_bid_impl/_offer_impl` as `can_bid_be_posted /can_offer_be_posted ` inside `_bid/_offer` shows the external user that their bids/offer can no more be posted for that slot

src/d3a/setup/strategy_tests/external_ess_bids.py
hannesdiedrich: Ah, yes, I just ran into this in my tests too. Weird that this only happened now.

fievelk: I'm not sure whether this might be related, but also keep in mind that the `replace_existing` parameter is `True` by default. So each offer/bid will always replace the previous ones unless you specify `replace_existing=False`.

hannesdiedrich: @fievelk yes, i just realise that you havent pushed your `replace_existing` -PR here yet. Why is that?

fievelk (approved):

src/d3a/models/strategy/external_strategies/storage.py
hannesdiedrich: I agree with you, however, if you look into the load and pv, these checks are done in `bid_impl` `offer_impl`.
There mus be a reason for that, but i can not remember. @spyrostz Can you help us please?

spyrostz: In order to avoid race conditions with other bids / offers. _bid/_offer are executed in the context of the redis messaging thread (therefore not during the tick event, but between ticks). Therefore if you do the validation in these methods, there might be more bids/offers that are received before the event tick, and the energy limits of the device can be surpassed (even though the can_bid_be_posted allows us, it only takes into account bids that are already placed in the market, not bids that live in the command buffer). Thus we are obliged to do this check in the bid_impl/offer_impl, which are executed on the main thread, in the tick event. Hope it clarifies things.

fievelk:
@faizan2590 Is the `arguments['seller_origin_id']` also needed in PV, or does it work differently than `buyer_origin_id`?

fievelk (commented):
@faizan2590 Is the `arguments['seller_origin_id']` also needed in PV's `_offer`, or does it work differently than `buyer_origin_id`?

spyrostz (dismissed):
Once the `can_offer_be_posted` and `can_bid_be_posted` are moved in the `_offer_impl` and `_bid_impl` LGTM

src/d3a/models/strategy/external_strategies/storage.py
faizan2590: Do you think its better to have `can_bid_be_posted/can_offer_be_posted`  in both `_bid/_offer` as well as in `_bid_impl/_offer_impl` . Although the later is for sure better(after your explanation) to prevent state violation inside d3a but external client isn’t notified as the recently posted bid would have already violated the state(considering that `replace_existing=False`)?

faizan2590:
> @faizan2590 Is the `arguments['seller_origin_id']` also needed in PV's `_offer`, or does it work differently than `buyer_origin_id`?

Yes, it would be needed there too, will add it in incoming commit.

src/d3a/models/strategy/external_strategies/storage.py
spyrostz: I think it makes thing much harder if we have them in the _bid / _offer methods, because we have to think about bid / offer deletions too. Therefore, we might reject a bid / offer that could in reality be accepted (because a bid was deleted before that, thus decreasing the total bid energy)

hannesdiedrich (dismissed):
LGTM

src/d3a/models/strategy/__init__.py
```diff
@@ -506,8 +511,8 @@ def post_bid(self, market, price, energy, replace_existing=True, buyer_origin=No
             energy,
             self.owner.name,
             original_bid_price=price,
-            buyer_origin=buyer_origin,
-            buyer_origin_id=buyer_origin_id,
+            buyer_origin=self.owner.name,
+            buyer_origin_id=self.owner.uuid,
```
hannesdiedrich, line 515: Great!
please remove `buyer_origin` and `buyer_origin_id` from the input parameters. AFAIS; they are not needed any more.

src/d3a/setup/strategy_tests/external_ess_bids.py
hannesdiedrich: @faizan2590 could you please add a TODO comment in here that logs the change, why we did it and that this should be reverted once the `replace_existing`-PR is pushed? Thanks!

src/d3a/setup/strategy_tests/external_ess_bids.py
```diff
@@ -30,6 +30,9 @@ def get_setup(config):
             Area(
                 'House 1',
                 [
+                    # initial_selling_rate is adjusted from 30 to 40, in order to avoid getting
```
hannesdiedrich, line 33: Please prepend with `TODO:` so it is highlighted in pycharm. thanks!

hannesdiedrich (approved):
One minor comment. LGTM once addressed

hannesdiedrich (approved):
LGTM

fievelk (approved):

Hassan754 (approved):

PR #1089, opened 2021-03-09
D3ASIM-3324: Deprecated unmatched loads

Hassan754:


integration_tests/jira_issues_tests.feature
```diff
@@ -20,8 +20,7 @@ Feature: Jira Issues Tests
      And d3a is installed
      And export is_needed
      When we run the simulation with setup file strategy_tests.home_cp_ess_load and parameters [24, 60, 60, 1]
-     Then there should be no unmatched loads
```
hannesdiedrich, line 23: I should have mentioned this earlier, sorry, but the check if the load received all the energy that it demanded is important for most of the times this stop was implemented. 
I doubt that there is a handy step function implemented that loops over all loads and checks if the energy demand was always fulfilled. 
It should not be that hard actually, because the `_energy_requirement_Wh` member of the load state is buffering this information.
TLDR;
Please replace `there should be no unmatched loads` with a to-be-implemented function `all load demand in setup was fulfilled on every market slot`
Thanks a lot!

hannesdiedrich (commented):
One important comment, LGTM once addressed.

integration_tests/jira_issues_tests.feature
Hassan754: @hannesdiedrich thanks for reviewing, do you think that I should create a ticket for it too. ?

hannesdiedrich: No, this should be part of this ticket.
1. to not get forgotten or pushed back 
2. We need these checks, especially for the setup.jira integrationtests.
@spyrostz correct me if I am overly cautious.

Hassan754: @hannesdiedrich Yes I meant a ticket for implementing the function

hannesdiedrich: @Hassan754 yes, i meant to do it in the frame of this ticket.

spyrostz: I agree with @hannesdiedrich that even though we deprecate the unmatched loads endpoint, in the tests we still need to detect if a device did not consume as much energy as the test dictates it too, and I think it would be good to be done in the context of this story because IIRC this is a check in many tests.

integration_tests/steps/load_tests.py
```diff
@@ -76,22 +76,17 @@ def check_user_pv_dict_profile(context):
 @then('LoadHoursStrategy does not buy energy with rates that are higher than the provided profile')
 def check_user_rate_profile_dict(context):
     house = next(filter(lambda x: x.name == "House 1", context.simulation.area.children))
+    load1 = next(filter(lambda x: x.name == "H1 General Load 1", house.children))
+    # load2 = next(filter(lambda x: x.name == "H1 General Load 2", house.children))
+    # print(load2.strategy.final_buying_rate)
```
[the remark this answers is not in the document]
Hassan754: Not really, these were meant to be removed.

integration_tests/jira_issues_tests.feature
Hassan754: Done

integration_tests/steps/jira_issues_tests.py
```diff
@@ -113,6 +113,8 @@ def step_impl(context):
 
 @then('on every market slot there should be matching trades on grid and house markets')
 def check_matching_trades(context):
+    from integration_tests.steps.integration_tests import get_simulation_raw_results
+    get_simulation_raw_results(context)
     for time_slot, core_stats in context.raw_sim_data.items():
```
hannesdiedrich, line 118: Didn't we discuss to change this function to have a return value?

Hassan754: We agreed to adopt this style from now on, there are lots of examples that I think we can keep them atm.

integration_tests/jira_issues_tests.feature
```diff
@@ -72,13 +72,6 @@ Feature: Jira Issues Tests
     When we run a multi-day d3a simulation with jira.d3asim_962 [2019-01-01, 48, 60, 60]
     Then the device statistics are correct
 
-  @slow
-  Scenario: D3ASIM-1139, no unmatched loads on setup with many loads and only one CEP
-    Given we have a scenario named jira/d3asim_1139
-    And d3a is installed
-    When we run the simulation with setup file jira.d3asim_1139 and parameters [24, 60, 60, 1]
-    Then there should be no unmatched loads
-
```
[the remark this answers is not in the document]
Hassan754: After a discussion with Spyros we agreed to create a ticket regarding this test
https://gridsingularity.atlassian.net/browse/D3ASIM-3377

integration_tests/jira_issues_tests.feature
```diff
@@ -72,6 +72,15 @@ Feature: Jira Issues Tests
     When we run a multi-day d3a simulation with jira.d3asim_962 [2019-01-01, 48, 60, 60]
     Then the device statistics are correct
 
+  @slow
+  # TODO: this is to be fixed as it is always failing
```
hannesdiedrich, line 76: Hmm, so no easy fix for that? Did you try to find out what goes wrong here?

Hassan754: @hannesdiedrich it was a problem with the floating point precision, it was not equal to 0 but very close to it, fixed now

hannesdiedrich: great!

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #1090, opened 2021-03-09
Feature/d3 asim 3329 Implemented new trade_info format for aggregator connections.

hannesdiedrich:


src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -285,32 +285,60 @@ def _publish_trade_event(self, trade, is_bid_trade):
...
+                                   'traded_energy': trade.offer.energy,
+                                   'total_fee': trade.fee_price,
+                                   'local_market_fee':
+                                       self.area.current_market.fee_class.grid_fee_rate
+                                       if self.area.current_market is not None else "None",
+                                   'attributes': {},
+                                   'seller': trade.seller
+                                   if trade.seller_id == self.device.uuid else 'anonymous',
+                                   'buyer': trade.buyer
+                                   if trade.buyer_id == self.device.uuid else 'anonymous',
```
faizan2590, line 303: `seller_id` & `buyer_id ` should always be equal to `device_uuid` for offers/bids of strategies. Have you already encountered where they were not equal?

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -285,32 +285,60 @@ def _publish_trade_event(self, trade, is_bid_trade):
...
-                               "residual_id": trade.residual.id
-                               if trade.residual is not None else "None"}
-
-        bid_offer_key = 'bid_id' if is_bid_trade else 'offer_id'
-        event_response_dict["event_type"] = "buy" \
-            if trade.buyer == self.device.name else "sell"
-        event_response_dict[bid_offer_key] = trade.offer.id
-
         if self.is_aggregator_controlled:
+            event_response_dict = {'event': 'trade',
```
spyrostz, line 289: In general, it would be great instead of creating 2 long dicts, one for the aggregator and one for the device connection, that have some common key-value pairs, I would suggest to refactor this and create one dictionary for both cases, and then the aggregator/device can add the fields that are unique to them.

src/d3a/models/strategy/external_strategies/__init__.py
spyrostz: If I am not mistaken trade seller and buyer are not always equal to the device uuid. Seller/buyer can either be another device on the same market, or the market itself, according to the usecase.

spyrostz (approved):
One refactoring comment, LGTM otherwise

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: Yeah, my motivation for this was that we will deprecate the direct sending of the trade event soonish.
Or do you think that this will be some time until then and stand by your comment?

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: > `seller_id` & `buyer_id ` should always be equal to `device_uuid` for offers/bids of strategies. Have you already encountered where they were not equal?

Sorry, dont get your question. Do you mean why we return "anonymous" for if the device is not the seller or buyer? I really can not remember. I guess that this was due to a privacy issue.

hannesdiedrich: > If I am not mistaken trade seller and buyer are not always equal to the device uuid. Seller/buyer can either be another device on the same market, or the market itself, according to the usecase.

Ah yes, but why is it not good to return it in any case? I understand that the buyer_origin should be hidden, but in the same market, it should be ok to expose the seller or buyer right?

spyrostz: Depends on the usecase, for houses this is definitely ok. For the higher levels in the hierarchy though, this might be a security concern (for instance, if a power plant sells energy to an area). At least that was the original though behind that, when we were much stricter about privacy than we unfortunately are now.

src/d3a/models/strategy/external_strategies/__init__.py
spyrostz: In this case I agree, feel free to leave it as is!

src/d3a/models/strategy/external_strategies/__init__.py
faizan2590: > Sorry, dont get your question. Do you mean why we return "anonymous" for if the device is not the seller or buyer? I really can not remember. I guess that this was due to a privacy issue.

Earlier, I skipped the case where strategies market having external traded bids/offers BUT still always using `anonymous` for external agent may not be needed. For `buyer_origin/seller_origin` may be okay to mask it as anonymous but at least for `buyer/seller` as they show only the immediate connected market masking them as `anonymous` could be skipped, anyways will leave it for future feedback of users

faizan2590 (approved):
LGTM

PR #1091, opened 2021-03-10
D3ASIM-3350: Added feed_in_tariff_rate to global event dict for aggre…

hannesdiedrich:
…gator event dispatching.

### DIFF
--- a/src/d3a/d3a_core/global_objects.py
+++ b/src/d3a/d3a_core/global_objects.py
@@ -16,7 +16,9 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from d3a.models.strategy.external_strategies import ExternalMixin
-from d3a.d3a_core.util import get_market_maker_rate_from_config
+from d3a.d3a_core.util import get_market_maker_rate_from_config, \
+    find_object_of_same_weekday_and_time
+from d3a.models.strategy.infinite_bus import InfiniteBusStrategy
 
 
 class GlobalStatistics:
@@ -24,10 +26,25 @@ class GlobalStatistics:
     def __init__(self, root_area):
         self.area_stats_tree_dict = {}
         self.root_area = root_area
+        self.feed_in_tariff = None
 
-    def update(self):
+    def update(self, current_market_slot=None):
         if self.root_area.current_market:
             self._create_grid_tree_dict(self.root_area, self.area_stats_tree_dict)
+        if current_market_slot is not None:
+            self._buffer_feed_in_tariff(self.root_area, current_market_slot)
+
+    def _buffer_feed_in_tariff(self, area, current_market_slot):
+        """
+        This simplified recursion is sufficient as the infinite bus is expected to be in the
+        uppermost level of the tree
+        """
+        for child in area.children:
+            if isinstance(child.strategy, InfiniteBusStrategy):
+                self.feed_in_tariff = \
+                    find_object_of_same_weekday_and_time(child.strategy.energy_buy_rate,
+                                                         current_market_slot)
+                return
 
     def _create_grid_tree_dict(self, area, outdict):
         outdict[area.uuid] = {}
--- a/src/d3a/d3a_core/global_objects.py
+++ b/src/d3a/d3a_core/global_objects.py
@@ -26,7 +26,7 @@ class GlobalStatistics:
     def __init__(self, root_area):
         self.area_stats_tree_dict = {}
         self.root_area = root_area
-        self.feed_in_tariff = None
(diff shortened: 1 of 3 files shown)


src/d3a/d3a_core/global_objects.py
```diff
@@ -16,18 +16,35 @@
...
+        """
+        This simplified recursion is sufficient as the infinite bus is expected to be in the
+        uppermost level of the tree
+        """
+        for child in area.children:
+            if isinstance(child.strategy, InfiniteBusStrategy):
+                self.feed_in_tariff = \
+                    find_object_of_same_weekday_and_time(child.strategy.energy_buy_rate,
+                                                         current_market_slot)
+                return
```
Hassan754, line 47: Is this progressive ? I mean in every iteration you are setting a new value for the `self.feed_in_tariff`

src/d3a/d3a_core/global_objects.py
```diff
@@ -16,18 +16,35 @@
...
+            self._buffer_feed_in_tariff(self.root_area, current_market_slot)
+
+    def _buffer_feed_in_tariff(self, area, current_market_slot):
+        """
+        This simplified recursion is sufficient as the infinite bus is expected to be in the
+        uppermost level of the tree
+        """
+        for child in area.children:
+            if isinstance(child.strategy, InfiniteBusStrategy):
+                self.feed_in_tariff = \
```
spyrostz, line 44: I would rename this to `_current_feed_in_tariff` and expose a `feed_in_tariff_rate` property in order to have a clean API in this class, because I predict that it will grow quite soon.

src/d3a/d3a_core/global_objects.py
spyrostz: It is indeed, the goal is to keep track of the current Infinite Bus energy_buy_rate at every moment in the simulation, which can change on every market slot.

src/d3a/d3a_core/simulation.py
```diff
@@ -345,7 +345,8 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
             self.area.cycle_markets()
 
             if self.simulation_config.external_connection_enabled:
-                self.global_objects.update()
+                self.global_objects.update(current_market_slot=self.area.current_market.time_slot
+                                           if self.area.current_market is not None else None)
```
spyrostz, line 349: Because I have seen this repeating check a lot in our code, how about exposing a small property in the area, namely `current_market_time_slot` that contains this None check and does not leave the responsibility to the caller code ?

spyrostz (approved):
2 small comments, already LGTMing

src/d3a/d3a_core/simulation.py
hannesdiedrich: good idea! Will add it.

hannesdiedrich: Actually, I had a look. There are a lot of checks if the current_market is None, but the consequence is different every time. The time_slot is not needed apart from here. Maybe I missed this, in which case, please point me to the places. Thanks!

Hassan754 (approved):
LGTM

src/d3a/d3a_core/global_objects.py
```diff
@@ -25,16 +25,34 @@ def __init__(self):
...
     def __call__(self, root_area, ticks_per_slot):
         self.root_area = root_area
         self.external_tick_counter = ExternalTickCounter(ticks_per_slot)
 
+    def _buffer_feed_in_tariff(self, area, current_market_slot):
+        """
+        This simplified recursion is sufficient as the infinite bus is expected to be in the
+        uppermost level of the tree
+        """
+        # the lazy is needed in order to avoid circular imports
```
spyrostz, line 39: After converting this object to a singleton this should not be needed anymore. I have not checked the code, but mentally I foresee this object being imported by simulation.py and external strategies only, and I cannot visualise any circular import in these places.

src/d3a/d3a_core/simulation.py
spyrostz: In this case then you can skip this, all good.

src/d3a/d3a_core/global_objects.py
hannesdiedrich: @spyrostz i exerienced this, that is why I added it like this
It is also the reason for changing the check for an external strategy to 
`if area.strategy is not None and area.external_connection_available`

hannesdiedrich: BTW; i merged 3333 into here already

spyrostz: Interesting, then leave it as is

PR #1092, opened 2021-03-15
D3ASIM-3378: Adding add_entry_in_hrs_per_day for external load strate…

faizan2590:
…gies.


Hassan754 (approved):

hannesdiedrich (approved):
Took ma a while to understand why it is needed, but now i get it. LGTM

PR #1104, opened 2021-03-30
D3ASIM-1877: Separate bids and offers in updatefrequencymixin

Hassan754:


src/d3a/models/strategy/update_frequency.py
```diff
@@ -181,21 +181,10 @@ def update_energy_price(self, market, strategy):
...
-                self.update_energy_price(market, strategy)
 
-    def update_market_cycle_bids(self, strategy):
-        # decrease energy rate for each market again, except for the newly created one
-        for market in strategy.area.all_markets[:-1]:
-            self.update_counter[market.time_slot] = 0
-            self._post_bids(market, strategy)
+class BidsUpdateFrequencyMixin(UpdateFrequencyMixin):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
spyrostz, line 187: Should not be needed, the default constructor should be automatically invoked

src/d3a/models/strategy/pv.py
```diff
@@ -85,9 +85,10 @@ def _init_price_update(self, update_interval, initial_selling_rate, final_sellin
         validate_pv_device_price(fit_to_limit=fit_to_limit,
                                  energy_rate_decrease_per_update=energy_rate_decrease_per_update)
 
-        self.offer_update = UpdateFrequencyMixin(initial_selling_rate, final_selling_rate,
-                                                 fit_to_limit, energy_rate_decrease_per_update,
-                                                 update_interval)
+        self.offer_update = OffersUpdateFrequencyMixin(initial_selling_rate, final_selling_rate,
```
spyrostz, line 88: Since we are in this part of the code, and the Mixin term is obviously wrong here (the PV/Load/StorageStrategy classes are not inheriting the Offers/BidsUpdateFrequencyMixin, thus this class is obviously not a Mixin but a normal class), how about renaming it to TemplateStrategyOfferUpdater and TemplateStrategyBidUpdater? @hannesdiedrich @fievelk feel free to pitch in if you disagree to the new naming.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -209,7 +198,28 @@ def _post_bids(self, market, strategy):
...
+
     def update_posted_bids_over_ticks(self, market, strategy):
         if self.time_for_price_update(strategy, market.time_slot):
             if strategy.are_bids_posted(market.id):
                 self._post_bids(market, strategy)
+
+
+class OffersUpdateFrequencyMixin(UpdateFrequencyMixin):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
```
spyrostz, line 215: Same here, no need for the explicit constructor.

spyrostz (commented):
Some general comments:
- update_energy_price should also be moved to the Offer class 
- Should homogenize the interface of the class (more comments related to that in the PR)
- Potentially scan through the Strategy classes to find occurrences where members of this class are directly accessed, and potentially add them to the class interface. That way we'll have more visibility over what the strategies require from these classes.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -209,7 +198,28 @@ def _post_bids(self, market, strategy):
...
+
+    def update_market_cycle_offers(self, strategy):
```
spyrostz, line 217: I think we can rename this to just `reset` because all this method does is to reset the update_counter and post updated bids / offers, so essentially to start from scratch. This happens during the market cycle event, that's why the name crept in this class.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -30,25 +30,78 @@ def __init__(self, initial_rate, final_rate, fit_to_limit=True,
...
         self.rate_limit_object = rate_limit_object
 
+    @property
+    def fit_to_limit(self):
+        return self._fit_to_limit
+
+    @fit_to_limit.setter
+    def fit_to_limit(self, fit_to_limit):
+        self._fit_to_limit = fit_to_limit
+
```
hannesdiedrich, line 59: Can you please explain me the rationale behind going from a self member to this property?
I see more code and an additional `self._fit_to_limit` and can not see the benefit of this. Thanks!

hannesdiedrich: This applies to all the other properties, too. A lot of boilerplate code.

Hassan754: @hannesdiedrich this was requested by Hannes 

> Potentially scan through the Strategy classes to find occurrences where members of this class are directly accessed, and potentially add them to the class interface. That way we'll have more visibility over what the strategies require from these classes.
I'm not sure if he meant something else

fievelk: Just hopping in to ask: if you have setters, shouldn't they be used in these cases as well? https://github.com/gridsingularity/d3a/pull/1104/files/2d8a5e7b7f350cf7c421b0b957a4e6de32499f88#diff-bafdd7ead350b771695da62be3b1e612a379422c602b64f7f65a22746a2a78b6R33

Hassan754: I added setters for the members that were set in the project ( For members that were only read, I only added getters )

hannesdiedrich:
> * update_energy_price should also be moved to the Offer class

I like to remind you on this comment from Spyros. I think the reason for this is that this function only calls  `BaseStrategy.Offers` code . 
You could move it to the `Offers` class and rename it to `update_offer_price(market, updated_price)` (see that the API has changed and that you dont need to pass the strategy class any more. 

BTW, this should be also done for the `_post_bids` method (--> `update_bid_price(market, updated_price)`)

Let me know if you have questions here.

src/d3a/models/strategy/update_frequency.py
hannesdiedrich: OK, I see. I am not sure whether this is what Spyros wanted, so leaving it to him to decide.

hannesdiedrich:
> > * update_energy_price should also be moved to the Offer class
> 
> I like to remind you on this comment from Spyros. I think the reason for this is that this function only calls `BaseStrategy.Offers` code .
> You could move it to the `Offers` class and rename it to `update_offer_price(market, updated_price)` (see that the API has changed and that you dont need to pass the strategy class any more.
> 
> BTW, this should be also done for the `_post_bids` method (--> `update_bid_price(market, updated_price)`)
> 
> Let me know if you have questions here.

For the bid case, it has to reside in `BidEnabledStrategy` i think

Hassan754:
> Offers

@hannesdiedrich I don't see the rational behind moving these inside the `Offers` || `BidEnabledStrategy` classes
The `update_energy_price` is only used within the class' `update` method where we are looping over the strategy.area.markets and updating the energy price for the respective market .

src/d3a/models/strategy/update_frequency.py
spyrostz: This was only part of what I actually intended for this comment unfortunately :D To my eyes I do not have a problem with the extra boilerplate code for the setters (even though in this case they are no more useful than having a public member variable, since they contain no logic at all). The main purpose of my comment was to actually check where these are called from, and to see if we can group multiple property assignments in one method, in order to simplify the interface of the class. For instance, `initial_rate_profile_buffer` and `final_rate_profile_buffer` are both updated only in 2 places (in the constructor and in the area_reconfigure methods of the strategies), therefore we could expose a single method here that assigns both at once, instead of assigning them in the caller code.

hannesdiedrich:
> > Offers
> 
> @hannesdiedrich I don't see the rational behind moving these inside the `Offers` || `BidEnabledStrategy` classes
> The `update_energy_price` is only used within the class' `update` method where we are looping over the strategy.area.markets and updating the energy price for the respective market .

I only saw this now, sorry.
Please have a closer look. `update_energy_price` is code that solely manages offers in the basestrategy (for the tracking of offers inside the BaseStrategy) and in the markets. So, it makes sense to move them to the respective class IMO (and apparently also in Spyros'. I hope that I explained it the way he wanted.

Hassan754:
> > > Offers
> > 
> > 
> > @hannesdiedrich I don't see the rational behind moving these inside the `Offers` || `BidEnabledStrategy` classes
> > The `update_energy_price` is only used within the class' `update` method where we are looping over the strategy.area.markets and updating the energy price for the respective market .
> 
> I only saw this now, sorry.
> Please have a closer look. `update_energy_price` is code that solely manages offers in the basestrategy (for the tracking of offers inside the BaseStrategy) and in the markets. So, it makes sense to move them to the respective class IMO (and apparently also in Spyros'. I hope that I explained it the way he wanted.

@hannesdiedrich I think @spyrostz meant that it should be moved to the Offer's class `TemplateStrategyOfferUpdater`

src/d3a/models/strategy/update_frequency.py
Hassan754: @spyrostz Should I export a generic update interface that can receive named parameters and update them ? because there are several combinations where we update different members

spyrostz:
> > > > Offers
> > > 
> > > 
> > > @hannesdiedrich I don't see the rational behind moving these inside the `Offers` || `BidEnabledStrategy` classes
> > > The `update_energy_price` is only used within the class' `update` method where we are looping over the strategy.area.markets and updating the energy price for the respective market .
> > 
> > 
> > I only saw this now, sorry.
> > Please have a closer look. `update_energy_price` is code that solely manages offers in the basestrategy (for the tracking of offers inside the BaseStrategy) and in the markets. So, it makes sense to move them to the respective class IMO (and apparently also in Spyros'. I hope that I explained it the way he wanted.
> 
> @hannesdiedrich I think @spyrostz meant that it should be moved to the Offer's class `TemplateStrategyOfferUpdater`

I should have been more clear here indeed and define the exact class. Hassan got it right, I actually meant the `TemplateStrategyOfferUpdater` class, which was not in the PR yet, that's why I did not mention it by name. Having said that, I understand Hannes' idea about moving the update_energy_price method to the BaseStrategy class, and I think it makes sense. The only tricky part about it is that we need to also pass the energy rate to this method, because it is calculated by the TemplateStrategyOfferUpdater.get_updated_rate (and I think that it is right to remain in this class, because the main purpose of this class is to calculate the energy rate of the template strategies on each tick). 
TLDR I agree with Hannes that it is better to move this class to the BaseStrategy or Offers method, depending on convenience (I think BaseStrategy is more appropriate, however let me know if you disagree).

Hassan754:
@spyrostz @hannesdiedrich this is the invoker of `update_energy_price`, The `time_for_price_update` uses the members `elapsed_seconds`, `update_interval` and `update_counter` which are not available in the BaseStrategy
```python
        for market in strategy.area.all_markets:
            if self.time_for_price_update(strategy, market.time_slot):
                self.update_energy_price(market, strategy)
```

hannesdiedrich:
> @spyrostz @hannesdiedrich this is the invoker of `update_energy_price`, The `time_for_price_update` uses the members `elapsed_seconds`, `update_interval` and `update_counter` which are not available in the BaseStrategy
> 
> ```python
>         for market in strategy.area.all_markets:
>             if self.time_for_price_update(strategy, market.time_slot):
>                 self.update_energy_price(market, strategy)
> ```

@Hassan754 We are talking about the `update_energy_price` method. 
If you have a look in the code below, all you need to do is to exchange `self.get_updated_rate(market.time_slot)` with an input parameter and it can live in the basestrategy. 
```
    def update_energy_price(self, market, strategy):
        if market.id not in strategy.offers.open.values():
            return

        for offer, iterated_market_id in strategy.offers.open.items():
            iterated_market = strategy.area.get_future_market_from_id(iterated_market_id)
            if market is None or iterated_market is None or iterated_market.id != market.id:
                continue
            try:
                iterated_market.delete_offer(offer.id)
                updated_price = round(offer.energy * self.get_updated_rate(market.time_slot), 10)
                new_offer = iterated_market.offer(
                    updated_price,
                    offer.energy,
                    strategy.owner.name,
                    original_offer_price=updated_price,
                    seller_origin=offer.seller_origin,
                    seller_origin_id=offer.seller_origin_id,
                    seller_id=strategy.owner.uuid
                )
                strategy.offers.replace(offer, new_offer, iterated_market.id)
            except MarketException:
                continue
```

hannesdiedrich:
> > > > > Offers
> > > > 
> > > > 
> > > > @hannesdiedrich I don't see the rational behind moving these inside the `Offers` || `BidEnabledStrategy` classes
> > > > The `update_energy_price` is only used within the class' `update` method where we are looping over the strategy.area.markets and updating the energy price for the respective market .
> > > 
> > > 
> > > I only saw this now, sorry.
> > > Please have a closer look. `update_energy_price` is code that solely manages offers in the basestrategy (for the tracking of offers inside the BaseStrategy) and in the markets. So, it makes sense to move them to the respective class IMO (and apparently also in Spyros'. I hope that I explained it the way he wanted.
> > 
> > 
> > @hannesdiedrich I think @spyrostz meant that it should be moved to the Offer's class `TemplateStrategyOfferUpdater`
> 
> I should have been more clear here indeed and define the exact class. Hassan got it right, I actually meant the `TemplateStrategyOfferUpdater` class, which was not in the PR yet, that's why I did not mention it by name. Having said that, I understand Hannes' idea about moving the update_energy_price method to the BaseStrategy class, and I think it makes sense. The only tricky part about it is that we need to also pass the energy rate to this method, because it is calculated by the TemplateStrategyOfferUpdater.get_updated_rate (and I think that it is right to remain in this class, because the main purpose of this class is to calculate the energy rate of the template strategies on each tick).
> TLDR I agree with Hannes that it is better to move this class to the BaseStrategy or Offers method, depending on convenience (I think BaseStrategy is more appropriate, however let me know if you disagree).

Sorry that I didnt wait for your comment and only "assumed".
When you wrote `TLDR I agree with Hannes that it is better to move this class to the BaseStrategy or Offers method`
you really meant `TLDR I agree with Hannes that it is better to move this method (update_energy_price/_post_bid) to the BaseStrategy or Offers method` right?

spyrostz:
> > > > > > Offers
> > > > > 
> > > > > 
> > > > > @hannesdiedrich I don't see the rational behind moving these inside the `Offers` || `BidEnabledStrategy` classes
> > > > > The `update_energy_price` is only used within the class' `update` method where we are looping over the strategy.area.markets and updating the energy price for the respective market .
> > > > 
> > > > 
> > > > I only saw this now, sorry.
> > > > Please have a closer look. `update_energy_price` is code that solely manages offers in the basestrategy (for the tracking of offers inside the BaseStrategy) and in the markets. So, it makes sense to move them to the respective class IMO (and apparently also in Spyros'. I hope that I explained it the way he wanted.
> > > 
> > > 
> > > @hannesdiedrich I think @spyrostz meant that it should be moved to the Offer's class `TemplateStrategyOfferUpdater`
> > 
> > 
> > I should have been more clear here indeed and define the exact class. Hassan got it right, I actually meant the `TemplateStrategyOfferUpdater` class, which was not in the PR yet, that's why I did not mention it by name. Having said that, I understand Hannes' idea about moving the update_energy_price method to the BaseStrategy class, and I think it makes sense. The only tricky part about it is that we need to also pass the energy rate to this method, because it is calculated by the TemplateStrategyOfferUpdater.get_updated_rate (and I think that it is right to remain in this class, because the main purpose of this class is to calculate the energy rate of the template strategies on each tick).
> > TLDR I agree with Hannes that it is better to move this class to the BaseStrategy or Offers method, depending on convenience (I think BaseStrategy is more appropriate, however let me know if you disagree).
> 
> Sorry that I didnt wait for your comment and only "assumed".
> When you wrote `TLDR I agree with Hannes that it is better to move this class to the BaseStrategy or Offers method`
> you really meant `TLDR I agree with Hannes that it is better to move this method (update_energy_price/_post_bid) to the BaseStrategy or Offers method` right?

I should really take more time when formulating PR comments, of course you are right, only the respective method (update_energy_price/post_bid) should be moved to BaseStrategy class.

hannesdiedrich:
> > > > > > > Offers
> > > > > > 
> > > > > > 
> > > > > > @hannesdiedrich I don't see the rational behind moving these inside the `Offers` || `BidEnabledStrategy` classes
> > > > > > The `update_energy_price` is only used within the class' `update` method where we are looping over the strategy.area.markets and updating the energy price for the respective market .
> > > > > 
> > > > > 
> > > > > I only saw this now, sorry.
> > > > > Please have a closer look. `update_energy_price` is code that solely manages offers in the basestrategy (for the tracking of offers inside the BaseStrategy) and in the markets. So, it makes sense to move them to the respective class IMO (and apparently also in Spyros'. I hope that I explained it the way he wanted.
> > > > 
> > > > 
> > > > @hannesdiedrich I think @spyrostz meant that it should be moved to the Offer's class `TemplateStrategyOfferUpdater`
> > > 
> > > 
> > > I should have been more clear here indeed and define the exact class. Hassan got it right, I actually meant the `TemplateStrategyOfferUpdater` class, which was not in the PR yet, that's why I did not mention it by name. Having said that, I understand Hannes' idea about moving the update_energy_price method to the BaseStrategy class, and I think it makes sense. The only tricky part about it is that we need to also pass the energy rate to this method, because it is calculated by the TemplateStrategyOfferUpdater.get_updated_rate (and I think that it is right to remain in this class, because the main purpose of this class is to calculate the energy rate of the template strategies on each tick).
> > > TLDR I agree with Hannes that it is better to move this class to the BaseStrategy or Offers method, depending on convenience (I think BaseStrategy is more appropriate, however let me know if you disagree).
> > 
> > 
> > Sorry that I didnt wait for your comment and only "assumed".
> > When you wrote `TLDR I agree with Hannes that it is better to move this class to the BaseStrategy or Offers method`
> > you really meant `TLDR I agree with Hannes that it is better to move this method (update_energy_price/_post_bid) to the BaseStrategy or Offers method` right?
> 
> I should really take more time when formulating PR comments, of course you are right, only the respective method (update_energy_price/post_bid) should be moved to BaseStrategy class.

@Hassan754 please be aware of the above. Let us know when you have adapted or if you have questions.

Hassan754:
> > > > > > > > Offers
> > > > > > > 
> > > > > > > 
> > > > > > > @hannesdiedrich I don't see the rational behind moving these inside the `Offers` || `BidEnabledStrategy` classes
> > > > > > > The `update_energy_price` is only used within the class' `update` method where we are looping over the strategy.area.markets and updating the energy price for the respective market .
> > > > > > 
> > > > > > 
> > > > > > I only saw this now, sorry.
> > > > > > Please have a closer look. `update_energy_price` is code that solely manages offers in the basestrategy (for the tracking of offers inside the BaseStrategy) and in the markets. So, it makes sense to move them to the respective class IMO (and apparently also in Spyros'. I hope that I explained it the way he wanted.
> > > > > 
> > > > > 
> > > > > @hannesdiedrich I think @spyrostz meant that it should be moved to the Offer's class `TemplateStrategyOfferUpdater`
> > > > 
> > > > 
> > > > I should have been more clear here indeed and define the exact class. Hassan got it right, I actually meant the `TemplateStrategyOfferUpdater` class, which was not in the PR yet, that's why I did not mention it by name. Having said that, I understand Hannes' idea about moving the update_energy_price method to the BaseStrategy class, and I think it makes sense. The only tricky part about it is that we need to also pass the energy rate to this method, because it is calculated by the TemplateStrategyOfferUpdater.get_updated_rate (and I think that it is right to remain in this class, because the main purpose of this class is to calculate the energy rate of the template strategies on each tick).
> > > > TLDR I agree with Hannes that it is better to move this class to the BaseStrategy or Offers method, depending on convenience (I think BaseStrategy is more appropriate, however let me know if you disagree).
> > > 
> > > 
> > > Sorry that I didnt wait for your comment and only "assumed".
> > > When you wrote `TLDR I agree with Hannes that it is better to move this class to the BaseStrategy or Offers method`
> > > you really meant `TLDR I agree with Hannes that it is better to move this method (update_energy_price/_post_bid) to the BaseStrategy or Offers method` right?
> > 
> > 
> > I should really take more time when formulating PR comments, of course you are right, only the respective method (update_energy_price/post_bid) should be moved to BaseStrategy class.
> 
> @Hassan754 please be aware of the above. Let us know when you have adapted or if you have questions.

@hannesdiedrich Can you review it now , I pushed some changes

src/d3a/models/strategy/update_frequency.py
Hassan754: @spyrostz

hannesdiedrich (approved):
LGTM

src/d3a/models/strategy/update_frequency.py
spyrostz: Please have a look at the very end of each `_area_reconfigure_prices` method. You will see that this is the only place where the setters of these properties are invoked. This applies to all strategies (Load, PV, Storage).

spyrostz: > @spyrostz Should I export a generic update interface that can receive named parameters and update them ? because there are several combinations where we update different members

Also if you can add a link with the part of the code where this actually happens it would be great, because from a brief search I did not find any such occurrences, but I might have missed it.

Hassan754: @spyrostz Okay so we'll go with the one function updater and remove the property setters

hannesdiedrich (dismissed):
LGTM

src/d3a/models/strategy/storage.py
```diff
@@ -192,18 +194,20 @@ def _area_reconfigure_prices(self, **kwargs):
...
-        self.bid_update.final_rate_profile_buffer = final_buying_rate
-        self.bid_update.energy_rate_change_per_update_profile_buffer = \
-            energy_rate_increase_per_update
-        self.offer_update.energy_rate_change_per_update_profile_buffer = \
-            energy_rate_decrease_per_update
-        self.bid_update.fit_to_limit = bid_fit_to_limit
-        self.offer_update.fit_to_limit = offer_fit_to_limit
-        self.bid_update.update_interval = update_interval
-        self.offer_update.update_interval = update_interval
+        self.offer_update.update_members(
```
spyrostz, line 197: Minor, but I think that a name like `set_parameters()` would be much more appropriate.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -157,59 +156,32 @@ def time_for_price_update(self, strategy, time_slot):
...
-                continue
-
-    def update_market_cycle_offers(self, strategy):
-        for market in strategy.area.all_markets[:-1]:
-            self.update_counter[market.time_slot] = 0
-            self.update_energy_price(market, strategy)
+    def update_members(self, **kwargs):
+        for key, value in kwargs.items():
+            if hasattr(self, key):
+                setattr(self, key, value)
```
spyrostz, line 162: Interesting approach, but unfortunately this is not foolproof because someone can easily break the operation of the class by assigning arguments like `number_of_available_updates` or `update_counter`. It would be much better if there were concrete arguments and an explicit assignment.

spyrostz (commented):
Looks good in general, but it would be great if you reinstated the properties in the `TemplateStrategyOfferUpdater` class, I am not sure why these were removed in the last iteration. Probably a miscommunication in the PR comments, but I was only opposed to the individual property setters, the property getters were good and a nice addition. 
Once the comments are dealt with LGTM though.

Hassan754:
> Looks good in general, but it would be great if you reinstated the properties in the `TemplateStrategyOfferUpdater` class, I am not sure why these were removed in the last iteration. Probably a miscommunication in the PR comments, but I was only opposed to the individual property setters, the property getters were good and a nice addition.
> Once the comments are dealt with LGTM though.

@spyrostz I was facing some issues and after some investigations long story short we can't create getters properties for dicts without also adding setters , so what do you think we should do in this case

spyrostz:
> > Looks good in general, but it would be great if you reinstated the properties in the `TemplateStrategyOfferUpdater` class, I am not sure why these were removed in the last iteration. Probably a miscommunication in the PR comments, but I was only opposed to the individual property setters, the property getters were good and a nice addition.
> > Once the comments are dealt with LGTM though.
> 
> @spyrostz I was facing some issues and after some investigations long story short we can't create getters properties for dicts without also adding setters , so what do you think we should do in this case

In this case it is alright to skip this comment, thanks!

spyrostz (approved):
LGTM

src/d3a/models/strategy/update_frequency.py
```diff
@@ -156,10 +156,20 @@ def time_for_price_update(self, strategy, time_slot):
         return self.elapsed_seconds(strategy) >= \
                self.update_interval.seconds * self.update_counter[time_slot]
 
-    def update_members(self, **kwargs):
-        for key, value in kwargs.items():
-            if hasattr(self, key):
-                setattr(self, key, value)
+    def set_parameters(self, *, initial_rate_profile_buffer=None, final_rate_profile_buffer=None,
```
hannesdiedrich, line 159: Thanks a lot for this. This skipped my attention and could have lead to some issues.

hannesdiedrich (approved):
LGTM

PR #1109, opened 2021-04-07
D3ASIM-3213: Fix can_offer_be_posted | fix update_offer

Hassan754:


src/d3a/models/strategy/__init__.py
```diff
@@ -144,7 +144,7 @@ def can_offer_be_posted(
 
         if replace_existing:
             # Do not consider previous offers, since they would be replaced by the current one
-            posted_offer_energy = 0.0
+            posted_offer_energy = self.sold_offer_energy(market.id)
```
fievelk, line 147: I think it would be good to update the comment above to document this change. 
It could be also great to write in the PR description the reason why the previous version was wrong, and what this change is fixing. This might help in the future.

fievelk (approved):
LGTM.
Minor: some more documentation would be great. Using double-quotes for new code should be the new standard after our poll.

src/d3a/models/strategy/__init__.py
Hassan754: @fievelk the `posted_offer_energy` consists of open and traded offers, In the case of replace existing  we want to exclude the open offers but the take the sold ones into consideration

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -316,6 +315,8 @@ def _update_offer_aggregator(self, arguments):
                     offer_arguments['energy'] = offer.energy
                     offer_arguments['price'] = \
                         (offer_arguments['price'] / offer_arguments['energy']) * offer.energy
+                    offer_arguments["seller"] = offer.seller
```
hannesdiedrich, line 318: Please also add this in the external Storage

hannesdiedrich (commented):

src/d3a/models/strategy/external_strategies/pv.py
Hassan754: @hannesdiedrich Done

PR #1115, opened 2021-04-20
Fix dependencies between requirements environments

fievelk:
Some environments were relying on reciprocally incompatible dependencies.

For example, `ptpython` was definied in both `tests.in` and `dev.in`. In order to keep it internally compatible with the other packages of each environment, `pip-tool` will install a different version of its sub-dependency `pygments` in the two environments.
`fab sync` therefore tried to sync all the requirement files (`base.txt`, `dev.txt` and `tests.txt`) and failed.

In order to fix this problem (and a similar problem with `pandapower.txt`) we refactor the inheritance between requirements files. We make both `tests.in` and `pandapower.in` depend on `dev.txt`.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -4,29 +4,63 @@
 #
 #    pip-compile --max-rounds=100 --no-emit-index-url requirements/base.in
 #
-attrs==20.3.0             # via -r requirements/base.in
-awesome-slugify==1.6.5    # via -r requirements/base.in
-base58==2.1.0             # via -r requirements/base.in
-cached-property==1.5.2    # via -r requirements/base.in
-click-default-group==1.2.2  # via -r requirements/base.in
-click==7.1.2              # via -r requirements/base.in, click-default-group, pip-tools, rq
-colorlog==4.7.2           # via -r requirements/base.in
-numpy==1.19.4             # via -r requirements/base.in
-pendulum==2.1.2           # via -r requirements/base.in
-pip-tools==5.4.0          # via -r requirements/base.in
-plotly==4.14.1            # via -r requirements/base.in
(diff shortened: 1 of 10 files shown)


requirements/base.txt
```diff
@@ -4,29 +4,63 @@
...
+    # via -r requirements/base.in
+base58==2.1.0
+    # via -r requirements/base.in
+cached-property==1.5.2
+    # via -r requirements/base.in
+click-default-group==1.2.2
+    # via -r requirements/base.in
+click==7.1.2
+    # via
+    #   -r requirements/base.in
```
Hassan754, line 19: The line length here isn't a problem, if possible revert them

fievelk: I'm not sure I understood what you mean. Would you like me to manually remove the newlines added by pip-tools?

fievelk: Ah, I think I understand now. I didn't add those newlines, it was automatically done by pip-tools. It might be a new format they use (which adds more information, not just the newline). 

I don't think there's an option on pip-tools to avoid this (it's been asked, but not yet implemented if I'm not wrong). I think it is better to keep it as it is, otherwise we would have to manually remove the newlines every time we compile the requirements.

hannesdiedrich (approved):
LGTM

spyrostz (commented):
One general comment, you need to also adapt the targets in tox.ini in order to not pip-sync from multiple files.

fievelk:
> One general comment, you need to also adapt the targets in tox.ini in order to not pip-sync from multiple files.

Thanks! I've updated the requirements in tox.ini: b4c9611c

spyrostz (approved):
LGTM

faizan2590 (approved):
LGTM

PR #1119, opened 2021-04-21
D3ASIM-2540: Set global market maker rate in instantiation of Infinit…

Hassan754:
The bug was because the `src/d3a/models/strategy/load_hours.py:232` 
```python
    def event_activate_price(self):
            self._area_reconfigure_prices(
                final_buying_rate=get_market_maker_rate_from_config(
                    self.area.next_market, 0) + self.owner.get_path_to_root_fees(), validate=False) #here
```
The get_market_maker_rate_from_config is accessing the `GlobalConfig.market_maker_rate` which is a str path atm
loading it in the instantiation phase solves it


faizan2590 (approved):

spyrostz (approved):
LGTM, however unit tests that validate this behavior and remind us of the expected behavior in the future would be great to have

spyrostz (approved):
LGTM

faizan2590 (approved):

PR #1121, opened 2021-04-22
Feature/d3 asim 3355v1: Decoupling matching algo from the market.

faizan2590:


src/d3a/models/market/__init__.py
```diff
@@ -124,6 +124,10 @@ def _create_fee_handler(self, grid_fee_type, grid_fees):
     def _is_constant_fees(self):
         return isinstance(self.fee_class, ConstantGridFees)
 
+    @property
+    def get_open_stock(self):
```
spyrostz, line 128: Because bad names tend to stick around for a while, I would be precise and not invent nomenclature here, `get_open_bids_and_offers` or `list_open_bids_and_offers` are way more appropriate

src/d3a/models/market/basic_market.py
```diff
@@ -0,0 +1,71 @@
...
+from d3a.models.market.market_structures import TradeBidOfferInfo
+from d3a_interface.constants_limits import ConstSettings
+
+log = getLogger(__name__)
+
+DEFAULT_PRECISION = 8
+FLOATING_POINT_TOLERANCE = 0.00001
+
+
+class BasicMarket(TwoSidedPayAsBid):
```
spyrostz, line 30: Instead of having a separate market for that, why not making Mycos the default behavior for all markets ? That way we can deprecate 2 sided pay as clear market, and have only 2 market types (one sided, two sided) and 2 myco types (pay as bid , pay as clear)

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -0,0 +1,53 @@
+from d3a.constants import FLOATING_POINT_TOLERANCE
+
+
+class MycoMatcher:
```
spyrostz, line 4: Pay-as-clear is missing, there needs to be a second Myco class that deals with it

src/d3a/models/area/event_dispatcher.py
```diff
@@ -136,7 +136,7 @@ def create_agent_object(owner, higher_market, lower_market, is_spot_market):
                     return OneSidedAlternativePricingAgent(**agent_constructor_arguments)
                 else:
                     return OneSidedAgent(**agent_constructor_arguments)
-            elif ConstSettings.IAASettings.MARKET_TYPE == 2:
+            elif ConstSettings.IAASettings.MARKET_TYPE in [2, 4]:
```
spyrostz, line 139: I think that it is a good opportunity to get rid of the two-sided pay as clear agent, since it is more or less the same as the two sided pay as bid agent

src/d3a/models/area/__init__.py
```diff
@@ -388,13 +390,16 @@ def _consume_commands_from_aggregator(self):
...
+        if ConstSettings.IAASettings.MARKET_TYPE in [2, 3]:
             if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
                 self.dispatcher.publish_market_clearing()
             else:
                 for market in self.all_markets:
                     market.match_offers_bids()
+        elif ConstSettings.IAASettings.MARKET_TYPE == 4:
+            if self.next_market is not None:
+                self.matcher.calculate_recommendation(*self.next_market.get_open_stock)
+                self.next_market.match_recommendation(self.matcher.bid_offer_pairs)
```
spyrostz, line 402: Similar to a comment below, we can make this behavior default and not create a separate market type for this.

src/d3a/models/myco_matcher/__init__.py
faizan2590: Added!

src/d3a/models/market/basic_market.py
faizan2590: Adapted, please give it a look: https://github.com/gridsingularity/d3a/pull/1121/commits/30d621c9303d6bcb5c5ef9e9a8516a8e787fd440

src/d3a/models/area/__init__.py
```diff
@@ -388,13 +390,14 @@ def _consume_commands_from_aggregator(self):
...
+        if ConstSettings.IAASettings.MARKET_TYPE == 2:
             if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
                 self.dispatcher.publish_market_clearing()
             else:
-                for market in self.all_markets:
-                    market.match_offers_bids()
+                if self.next_market is not None:
+                    self.matcher.calculate_recommendation(
+                        *self.next_market.get_open_bids_and_offers)
+                    self.next_market.match_recommendation(self.matcher.bid_offer_pairs)
```
spyrostz, line 400: Why not repeating the same process for all future markets ? Essentially you can do the following and avoid the if:

```suggestion
                for market in self.all_markets:
                    self.matcher.calculate_recommendation(
                        *market.get_open_bids_and_offers)
                    market.match_recommendation(self.matcher.bid_offer_pairs)
```

src/d3a/models/area/markets.py
```diff
@@ -98,15 +97,14 @@ def select_market_class(is_spot_market):
...
-            elif ConstSettings.IAASettings.MARKET_TYPE == 3:
-                return TwoSidedPayAsClear
+                return TwoSidedMarket
         else:
             return BalancingMarket
 
     def create_future_markets(self, current_time, is_spot_market, area):
         markets = self.markets if is_spot_market else self.balancing_markets
         market_class = self.select_market_class(is_spot_market)
+        print(f'market_class: {market_class}')
```
spyrostz, line 107: Minor, reminder to remove before merging

src/d3a/models/market/market_structures.py
```diff
@@ -352,9 +352,9 @@ def _to_csv(self):
 
 class MarketClearingState:
     def __init__(self):
-        self.cumulative_offers = dict()  # type: Dict[DateTime, dict()]
```
spyrostz, line 355: I think the type info does not hurt to have, however it would be great to homogenize the constructors. `self.clearing = dict()` would be better IMO.

src/d3a/models/market/two_sided.py
```diff
@@ -254,18 +228,23 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
         return bid_trade, trade
 
     def match_offers_bids(self):
```
spyrostz, line 230: This method can be removed from the markets now.

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -0,0 +1,31 @@
+from d3a.constants import DOUBLE_SIDED_ALGO
+from d3a.models.myco_matcher.pay_as_bid import PayAsBidMatch
+from d3a.models.myco_matcher.pay_as_clear import PayAsClear
+
+
+class MycoMatcher:
+    def __init__(self):
+        self.bid_offer_pairs = []
+        if DOUBLE_SIDED_ALGO == 1:
```
spyrostz, line 9: minor, but instead of ALGO I would use ALGORITHM

src/d3a/models/myco_matcher/pay_as_clear.py
```diff
@@ -15,33 +15,27 @@
...
-                         in_sim_duration=in_sim_duration)
+class PayAsClear(PayAsBidMatch):
+    def __init__(self):
         self.state = MarketClearingState()
         self.sorted_bids = []
-        self.mcp_update_point = \
-            GlobalConfig.ticks_per_slot / \
-            ConstSettings.GeneralSettings.MARKET_CLEARING_FREQUENCY_PER_SLOT
 
     def __repr__(self):  # pragma: no cover
```
spyrostz, line 40: This is not needed anymore

src/d3a/models/myco_matcher/pay_as_clear.py
```diff
@@ -15,33 +15,27 @@
...
 
-class TwoSidedPayAsClear(TwoSidedPayAsBid):
 
-    def __init__(self, time_slot=None, bc=None, notification_listener=None, readonly=False,
-                 grid_fee_type=ConstSettings.IAASettings.GRID_FEE_TYPE,
-                 grid_fees=None, name=None, in_sim_duration=True):
-        super().__init__(time_slot, bc, notification_listener, readonly,
-                         grid_fee_type, grid_fees, name,
-                         in_sim_duration=in_sim_duration)
+class PayAsClear(PayAsBidMatch):
```
spyrostz, line 35: I think that a common abstract base class for the matcher would be great to have, in order to define methods that all matcher classes will need to implement, and to not need to inherit the PayAsClear from PayAsBid. That way you can put the sorting methods in the abc, and PayAsClear and PayAsBid can both inherit from the same base class. That way the class hierarchy will be flat.

spyrostz (approved):
Looks good to me, added some enhancement comments. Adding LGTM though because once those are resolved the PR is ready to my eyes

src/d3a/constants.py
```diff
@@ -44,3 +44,5 @@
 CN_PROFILE_EXPANSION_DAYS = 7
 
 RUN_IN_REALTIME = False
+
+DOUBLE_SIDED_ALGO = 1  # 1--PAB; 2--PAC
```
fievelk, line 48: What do you think about defining this as Enum? In my opinion that would be clearer and safer than relying on the comment.

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -0,0 +1,31 @@
+from d3a.constants import DOUBLE_SIDED_ALGO
+from d3a.models.myco_matcher.pay_as_bid import PayAsBidMatch
+from d3a.models.myco_matcher.pay_as_clear import PayAsClear
+
+
+class MycoMatcher:
```
fievelk, line 6: Nit: it would be good to have docstrings documenting what this class is.

src/d3a/models/area/__init__.py
```diff
@@ -144,6 +145,7 @@ def __init__(self, name: str = None, children: List["Area"] = None,
         self.redis_ext_conn = RedisMarketExternalConnection(self) \
             if external_connection_available and self.strategy is None else None
         self.should_update_child_strategies = False
+        self.matcher = MycoMatcher()
```
fievelk, line 148: Nit: I think the name `matcher` is pretty broad, so it's not immediately clear what it represents. Maybe `offers_bids_matcher` would be clearer (but quite long). Do you have any ideas about possible alternatives?

src/d3a/models/area/event_dispatcher.py
```diff
@@ -137,15 +136,13 @@ def create_agent_object(owner, higher_market, lower_market, is_spot_market):
...
-                )
-            elif ConstSettings.IAASettings.MARKET_TYPE == 3:
-                return TwoSidedPayAsClearAgent(
+                return TwoSidedAgent(
                     **agent_constructor_arguments,
                     min_bid_age=ConstSettings.IAASettings.MIN_BID_AGE
                 )
+            else:
+                assert False, f'Wrong market type setting flag ' \
+                              f'{ConstSettings.IAASettings.MARKET_TYPE}'
```
fievelk, line 145: Rather than relying on `assert False` to raise an Exception, I would recommend to explicitly raise a custom exception. For example:
```suggestion
                raise WrongMarketTypeException(
                    f"Wrong market type setting flag {ConstSettings.IAASettings.MARKET_TYPE}")
```

src/d3a/models/area/__init__.py
```diff
@@ -388,13 +390,14 @@ def _consume_commands_from_aggregator(self):
     def tick(self):
         self._consume_commands_from_aggregator()
 
-        if ConstSettings.IAASettings.MARKET_TYPE == 2 or \
-                ConstSettings.IAASettings.MARKET_TYPE == 3:
+        if ConstSettings.IAASettings.MARKET_TYPE == 2:
```
fievelk, line 393: Nit: this was not implemented in your PR, but I think it might also be a candidate to use enums.

src/d3a/models/market/__init__.py
```diff
@@ -124,6 +124,10 @@ def _create_fee_handler(self, grid_fee_type, grid_fees):
     def _is_constant_fees(self):
         return isinstance(self.fee_class, ConstantGridFees)
 
+    @property
+    def get_open_bids_and_offers(self):
```
fievelk, line 128: I think that, when defining properties' names, nouns are more suitable than verbs. So what would you think about renaming this to something like `open_bids_and_offers`?

src/d3a/models/area/__init__.py
spyrostz: Since it caught my attention, I would not do that, because IIRC these variables can be sent from d3a-web, therefore they will be JSON-serialized. Therefore in order to avoid creating custom JSON encoders/decoders just for enums I would say that integers are fine specifically for this case.

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+        elif DOUBLE_SIDED_ALGO == 2:
+            self.match_algo = PayAsClear()
+
+    @staticmethod
+    def sorting(obj, reverse_order=False):
+        if reverse_order:
+            # Sorted bids in descending order
+            return list(reversed(sorted(
+                obj.values(),
+                key=lambda b: b.energy_rate)))
```
fievelk, line 20:
```suggestion
            return list(sorted(
                obj.values(),
                key=lambda b: b.energy_rate, reverse=True))
```

fievelk: - What kind of object is `obj`? I would add this information as type hint.
- If `obj` is a `dict`, you can directly use `sorted()` without casting to `list`.
- `sorted()` has a nice `reverse` parameter that you can use to replace `reversed`.

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -0,0 +1,31 @@
...
+
+        else:
+            # Sorted bids in ascending order
+            return list(sorted(
+                obj.values(),
+                key=lambda b: b.energy_rate))
```
fievelk, line 26:
```suggestion
def sorting(obj, reverse_order=False):
    # Sorted bids in descending order
    return sorted(
        obj.values(),
        key=lambda b: b.energy_rate, reverse=reverse_order)
```

fievelk: I would also suggest renaming the method to something different (as `sorting` isn't an imperative verb). For example `sort_by_energy_rate()`.

src/d3a/models/myco_matcher/pay_as_bid.py
```diff
@@ -0,0 +1,45 @@
...
+            return list(reversed(sorted(
+                obj.values(),
+                key=lambda b: b.energy_rate)))
+
+        else:
+            # Sorted bids in ascending order
+            return list(sorted(
+                obj.values(),
+                key=lambda b: b.energy_rate))
+
```
fievelk, line 21: This method seems to be an exact duplicate of [the one defined in `MycoMatcher`](https://github.com/gridsingularity/d3a/blob/2f5abe0b27603a78e6a8ae5d79587a729e913059/src/d3a/models/myco_matcher/__init__.py#L14-L26). I would suggest refactoring it and using one common implementation.

src/d3a/models/myco_matcher/pay_as_bid.py
```diff
@@ -0,0 +1,45 @@
...
+    def calculate_match_recommend(self, bids, offers):
+        # Pay as bid first
+        # There are 2 simplistic approaches to the problem
+        # 1. Match the cheapest offer with the most expensive bid. This will favor the sellers
+        # 2. Match the cheapest offer with the cheapest bid. This will favor the buyers,
+        #    since the most affordable offers will be allocated for the most aggressive buyers.
```
fievelk, line 27: - Nit: this can be a docstring (`"""`)

src/d3a/models/myco_matcher/pay_as_bid.py
```diff
@@ -0,0 +1,45 @@
...
+                obj.values(),
+                key=lambda b: b.energy_rate)))
+
+        else:
+            # Sorted bids in ascending order
+            return list(sorted(
+                obj.values(),
+                key=lambda b: b.energy_rate))
+
+    def calculate_match_recommend(self, bids, offers):
```
fievelk, line 22: I think this might be a typo for `calculate_match_recommendation`. Is this correct?

src/d3a/models/myco_matcher/pay_as_bid.py
```diff
@@ -0,0 +1,45 @@
...
+        already_selected_bids = set()
+        bid_offer_pairs = []
+        for offer in sorted_offers:
+            for bid in sorted_bids:
+                if bid.id not in already_selected_bids and \
+                        (offer.energy_rate - bid.energy_rate) <= \
+                        FLOATING_POINT_TOLERANCE and offer.seller != bid.buyer:
+                    already_selected_bids.add(bid.id)
+                    bid_offer_pairs.append(tuple((bid, offer, bid.energy_rate)))
+                    break
```
fievelk, line 44: Negative conditionals can sometimes make the code a bit trickier to understand. What would you think of something like the following?
```suggestion
for offer in sorted_offers:
    for bid in sorted_bids:
        if offer.seller != bid.buyer or bid.id in already_selected_bids:
            continue
       
        if (offer.energy_rate - bid.energy_rate) <= FLOATING_POINT_TOLERANCE:
            already_selected_bids.add(bid.id)
            bid_offer_pairs.append(tuple((bid, offer, bid.energy_rate)))
            break
```
What's your opinion?

src/d3a/models/myco_matcher/pay_as_clear.py
```diff
@@ -15,33 +15,27 @@
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+
+from d3a.models.myco_matcher.pay_as_bid import PayAsBidMatch
+
```
fievelk, line 20: Give the position of this import, I think they still need to be optimized.

src/d3a/models/myco_matcher/pay_as_clear.py
```diff
@@ -110,42 +104,53 @@ def _clearing_point_from_supply_demand_curve(self, bids, offers):
             if len(clearing) > 0:
                 return clearing[-1].rate, clearing[-1].energy
 
-    def _perform_pay_as_clear_matching(self):
-        self.sorted_bids = self.sorting(self.bids, True)
+    def calculate_match_recommend(self, bids, offers):
```
fievelk, line 107: Same question as above: https://github.com/gridsingularity/d3a/pull/1121/files#r620242625

fievelk (commented):
Some questions and some minor comments.

Please double check single quotes and imports optimization.

src/d3a/models/area/__init__.py
fievelk: My suggestion would have been to replace the simple `== 2` with something like ` == MarketTypeEnum.DOUBLE_SIDED_PAY_AS_BID.value`. This would make it a bit clearer to understand what `2` represents. This enum could be defined in d3a-interface so that it is shared by both d3a and d3a-web.
Am I missing something?

spyrostz: In that case I agree with you, it will be much better that way! I only disagreed with changing the type of this variable from int to enum, but a separate enum that will compare its value (like the one you described) would be good to have here!

fievelk: Yes sorry, my original comment was probably too generic!

src/d3a/models/area/__init__.py
faizan2590: I don't have any recommendation, I will then take `offers_bids_matcher ` as you proposed.

src/d3a/models/myco_matcher/__init__.py
faizan2590: It no more has a usage, I will remove it and adapt your suggestion [here](https://github.com/gridsingularity/d3a/pull/1121#discussion_r620240996)

src/d3a/models/market/two_sided.py
faizan2590: I kept it for redis market communication, since we haven't used area markets communication via redis, I assume we can safely deprecate it ?

src/d3a/models/myco_matcher/pay_as_bid.py
faizan2590: > Negative conditionals can sometimes make the code a bit trickier to understand.

Sorry, not able to think myself of cases where the negative conditional statement may make it tricky, but based on your suggestion, I will adapt it as its not much big of a change.

fievelk: It's just one common "rule", so of course we're free to decide based on the case whether it makes sense or not. I personally think it makes sense (also in daily spoken language). If you're interested, you can google for "negative conditionals programming". Here's a short explanation: https://csharpdeveloper.wordpress.com/2016/01/20/roberts-rules-of-coders-8-avoid-negative-conditionals-and-method-names/

faizan2590: Thanks for the link, makes sense for future to read and understand the logic easily, accpepting your suggestion with slight modification
```suggestion
        for offer in sorted_offers:
            for bid in sorted_bids:
                if bid.id in already_selected_bids or offer.seller == bid.buyer:
                    continue
                if (bid.energy_rate - offer.energy_rate) > FLOATING_POINT_TOLERANCE:
                    already_selected_bids.add(bid.id)
                    bid_offer_pairs.append(tuple((bid, offer, bid.energy_rate)))
```

src/d3a/models/myco_matcher/pay_as_bid.py
```diff
@@ -0,0 +1,37 @@
...
+        for offer in sorted_offers:
+            for bid in sorted_bids:
+                if bid.id in already_selected_bids or offer.seller == bid.buyer:
+                    continue
+                if (offer.energy_rate - bid.energy_rate) <= FLOATING_POINT_TOLERANCE:
+                    already_selected_bids.add(bid.id)
+                    selected_energy = min(bid.energy, offer.energy)
+                    bid_offer_pairs.append(BidOfferMatch(
+                        bid=bid, offer=offer, bid_energy=selected_energy,
+                        offer_energy=selected_energy, trade_rate=bid.energy_rate))
```
Hassan754, line 35: In both pay_as_bid and pay_as_clear we passed bid_energy and offer_energy where they were equal 
If this is the case we can use 1 variable instead of two

src/d3a/models/market/two_sided.py
```diff
@@ -254,18 +228,48 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
...
+                bid.original_bid_price / bid.energy
+
+            trade_bid_info = TradeBidOfferInfo(
+                original_bid_rate=original_bid_rate,
+                propagated_bid_rate=bid.price/bid.energy,
+                original_offer_rate=offer.original_offer_price/offer.energy,
+                propagated_offer_rate=offer.price/offer.energy,
+                trade_rate=recommended_pair.trade_rate)
+
+            bid_trade, trade = self.accept_bid_offer_pair(
```
Hassan754, line 251: I don't see where the validation is residing

src/d3a/models/myco_matcher/base_matcher.py
```diff
@@ -0,0 +1,24 @@
...
+            # Sorted bids in descending order
+            return list(reversed(sorted(
+                obj.values(),
+                key=lambda b: b.energy_rate)))
+
+        else:
+            # Sorted bids in ascending order
+            return list(sorted(
+                obj.values(),
+                key=lambda b: b.energy_rate))
```
Hassan754, line 20:
```suggestion
            return list(sorted(
                obj.values(),
                key=lambda b: b.energy_rate, reversed=reverse_order))

```

Hassan754 (commented):
Left some comments, probably we'll need to update our wiki to describe the changes done in this PR

src/d3a/models/area/__init__.py
```diff
@@ -386,13 +388,14 @@ def _consume_commands_from_aggregator(self):
...
             if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
                 self.dispatcher.publish_market_clearing()
             else:
                 for market in self.all_markets:
-                    market.match_offers_bids()
+                    self.offers_bids_matcher.calculate_recommendation(
+                        *market.open_bids_and_offers, self.now)
+                    market.match_recommendation(self.offers_bids_matcher.bid_offer_pairs)
```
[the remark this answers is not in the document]
faizan2590: Okay, curious if there is a code style that prefers that?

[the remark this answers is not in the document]
fievelk: I remember we had a conversation about this a few weeks ago on slack. In general, it is clearer when methods return values, and variables are assigned these values. I'll send you the reference of the conversation on slack.

src/d3a/models/myco_matcher/base_matcher.py
faizan2590:
```suggestion
        return list(sorted(obj.values(), key=lambda b: b.energy_rate, reverse=reverse_order))
```
Adapted with slight modification.

src/d3a/d3a_core/export.py
```diff
@@ -132,8 +132,9 @@ def export(self, export_plots=True, power_flow=None):
                 self.plot_device_stats(self.area, [])
             if ConstSettings.GeneralSettings.EXPORT_ENERGY_TRADE_PROFILE_HR:
                 self.plot_energy_trade_profile_hr(self.area, self.plot_dir)
-            if ConstSettings.IAASettings.MARKET_TYPE == 3 and \
-                    ConstSettings.GeneralSettings.EXPORT_SUPPLY_DEMAND_PLOTS:
+            if ConstSettings.IAASettings.MARKET_TYPE == 2 and \
+                    ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == 2 and \
```
hannesdiedrich, line 136:
```suggestion
                    ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == BidOfferMatchEnum.PAY_AS_CLEAR.value and \
```

hannesdiedrich: This should be changed more places. Thanks!

src/d3a/models/myco_matcher/pay_as_bid.py
faizan2590: `BidOfferMatch` was there before this PR, though I agree that both `bid_energy/offer_energy` could be merged to `selected_energy`

src/d3a/constants.py
```diff
@@ -44,3 +45,8 @@
 CN_PROFILE_EXPANSION_DAYS = 7
 
 RUN_IN_REALTIME = False
+
+
+class BidOfferMatchEnum(Enum):
```
hannesdiedrich, line 50: As there is already a `BidOfferMatch` object that holds the actual match, this naming here could lead to confusion (just experienced it ;-) ) 
I suggest to rename to: BidOfferMatchAlgoEnum

src/d3a/models/myco_matcher/pay_as_bid.py
```diff
@@ -0,0 +1,37 @@
+from d3a.constants import FLOATING_POINT_TOLERANCE
+from d3a.models.market.market_structures import BidOfferMatch
+from d3a.models.myco_matcher.base_matcher import BaseMatcher
+
+
+class PayAsBidMatch(BaseMatcher):
```
hannesdiedrich, line 6: Also here, I would propose:
```suggestion
class PayAsBidMatcher(BaseMatcher):
```

src/d3a/models/myco_matcher/pay_as_clear.py
```diff
@@ -15,48 +15,26 @@
...
 
-class TwoSidedPayAsClear(TwoSidedPayAsBid):
 
-    def __init__(self, time_slot=None, bc=None, notification_listener=None, readonly=False,
-                 grid_fee_type=ConstSettings.IAASettings.GRID_FEE_TYPE,
-                 grid_fees=None, name=None, in_sim_duration=True):
-        super().__init__(time_slot, bc, notification_listener, readonly,
-                         grid_fee_type, grid_fees, name,
-                         in_sim_duration=in_sim_duration)
+class PayAsClear(BaseMatcher):
```
hannesdiedrich, line 34:
```suggestion
class PayAsClearMatcher(BaseMatcher):
```

src/d3a/models/area/__init__.py
```diff
@@ -142,6 +143,7 @@ def __init__(self, name: str = None, children: List["Area"] = None,
         self.redis_ext_conn = RedisMarketExternalConnection(self) \
             if external_connection_available and self.strategy is None else None
         self.should_update_child_strategies = False
+        self.offers_bids_matcher = MycoMatcher()
```
hannesdiedrich, line 146: Sorry to leave only renaming comments for now. But IMO a consistent naming scheme is helping understanding, also later when working with the code.
Here I would vote to be consistent and rename to:
```suggestion
        self.bid_offer_matcher = MycoMatcher()
```

hannesdiedrich: Also: There are two steps now: `recommend` and `match` . I haven't found a perfect solution for naming this yet, but really the `MycoMatcher` is really a `MycoRecommendator` (Please dont name it like this :-)) This is just a comment which you maybe remember later on. Thanks

integration_tests/grid_fee.feature
```diff
@@ -16,6 +16,7 @@ Feature: GridFee integration tests
      |       5       |     60      |     10      |
      |       9       |     60      |     5       |
 
+  @slow
```
hannesdiedrich, line 19: Please use `@disabled` here, because I guess that these aren't suddenly slow but they are not working right?

integration_tests/steps/integration_tests.py
```diff
@@ -214,8 +214,10 @@ def one_sided_market(context, market_type):
         ConstSettings.IAASettings.MARKET_TYPE = 1
     elif market_type == "two-sided-pay-as-bid":
         ConstSettings.IAASettings.MARKET_TYPE = 2
-    if market_type == "two-sided-pay-as-clear":
-        ConstSettings.IAASettings.MARKET_TYPE = 3
+        ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE = 2
```
hannesdiedrich, line 217: This is wrong right? It should be 1 or:
```suggestion
        ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE = BidOfferMatchEnum.PAY_AS_BID.value
```

hannesdiedrich: Maybe the reason for the reduced SS/CS?

integration_tests/two_sided_pay_as_clear_market.feature
```diff
@@ -18,35 +20,42 @@ Feature: Two sided pay_as_clear market tests
...
      Then all traded energy have finite value
 
   Scenario: Test offers, bids and market-clearing-rate are exported
      Given we have a scenario named two_sided_pay_as_clear/default_2a
      And d3a is installed
+     And d3a uses an two-sided-pay-as-clear market
      When we run the d3a simulation on console with two_sided_pay_as_clear.default_2a for 2 hrs
      Then there are files with offers, bids & market_clearing_rate for every area
 
+  @slow
```
hannesdiedrich, line 34: Can you please explain why this test is slow?

integration_tests/two_sided_pay_as_clear_market.feature
```diff
@@ -18,35 +20,42 @@ Feature: Two sided pay_as_clear market tests
...
      Given we have a scenario named two_sided_pay_as_clear/one_offer_one_bid
      And d3a is installed
+     And d3a uses an two-sided-pay-as-clear market
      When we run the simulation with setup file two_sided_pay_as_clear.one_offer_one_bid and parameters [24, 60, 60, 1]
      Then one-on-one matching of offer & bid in PAC happens at bid rate
 
   Scenario: One Offer and multiple Bids are matched at the last clearing bid rate
      Given we have a scenario named two_sided_pay_as_clear/one_offer_multiple_bids
      And d3a is installed
+     And d3a uses an two-sided-pay-as-clear market
```
hannesdiedrich, line 52: Why you need this extra step and do not change it in the setup file, like we did before? These setup files are already put into separate folders that indicate their market type. I suggest either to remove these folders and only have one setup file for all market types and change the market type with this extra step OR to change the market type in the setup file. Thanks!

src/d3a/models/myco_matcher/pay_as_clear.py
```diff
@@ -110,92 +88,63 @@ def _clearing_point_from_supply_demand_curve(self, bids, offers):
...
-            return self._clearing_point_from_supply_demand_curve(
+            clearing = self._clearing_point_from_supply_demand_curve(
                 ascending_rate_bids, cumulative_offers)
         elif ConstSettings.IAASettings.PAY_AS_CLEAR_AGGREGATION_ALGORITHM == 2:
             cumulative_bids = self._discrete_point_curve(self.sorted_bids, math.floor)
             cumulative_offers = self._discrete_point_curve(self.sorted_offers, math.ceil)
             max_rate = self._populate_market_cumulative_offer_and_bid(cumulative_bids,
                                                                       cumulative_offers)
-            return self._get_clearing_point(max_rate)
+            clearing = self._get_clearing_point(max_rate)
```
hannesdiedrich, line 112: AFAIS, `_get_clearing_point` returns a tuple. You only use the rate here, right?
I suggest to refactor it to only return the rate. Also, it would be good to adapt `_clearing_point_from_supply_demand_curve` and `_get_clearing_point` to return a single value, not a list, so you dont need to select always the first value (in line 114)

integration_tests/steps/two_sided_pay_as_clear.py
```diff
@@ -26,15 +26,16 @@
 def test_traded_energy_rate(context):
     def has_one_of_clearing_rates(trade, market):
         return any(isclose((trade.offer.price / trade.offer.energy), clearing_rate)
-                   for clearing_rate in market.state.clearing[trade.time])
+                   for clearing_rate in list(market.state.clearing.values()))
```
hannesdiedrich, line 29: really minor, but `values()` already returns an iterable object, no need convert it into a list.

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -0,0 +1,25 @@
...
+        self.bid_offer_pairs = []
+        if ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == BidOfferMatchEnum.PAY_AS_BID.value:
+            self.match_algorithm = PayAsBidMatch()
+        elif ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == \
+                BidOfferMatchEnum.PAY_AS_CLEAR.value:
+            self.match_algorithm = PayAsClear()
+        else:
+            raise WrongMarketTypeException(f'Wrong market type setting flag '
+                                           f'{ConstSettings.IAASettings.MARKET_TYPE}')
+
```
hannesdiedrich, line 20: So, how do I integrate a custom matching algorithm now? Lets say I have a
```
class HannesMatcher(BaseMatcher):
```
I need to add it here, two, right? Wouldn't it be better to add a `CUSTOM` to the `BID_OFFER_MATCH_TYPE` enum ? Or is this not part of the story?

hannesdiedrich (commented):
Left some comments, will review again once addressed.

src/d3a/models/myco_matcher/pay_as_clear.py
faizan2590: Its used by `_create_bid_offer_matchings` where both `clearing_rate` and `clearing_energy` is used.

src/d3a/models/myco_matcher/__init__.py
faizan2590: Yes, to be extended as part of https://gridsingularity.atlassian.net/browse/D3ASIM-3447. For now restricted to only two types of double sided market to be sure if no loopholes are there at the moment of accidental introduction of any third market type.

src/d3a/models/myco_matcher/base_matcher.py
faizan2590: Sorry, just tested, neither mine nor your suggestion worked.

integration_tests/two_sided_pay_as_clear_market.feature
faizan2590: Yes, just retested, changing the market type in setup file is enough.

src/d3a/models/market/two_sided.py
faizan2590: Added in recent commit: https://github.com/gridsingularity/d3a/blob/feature/D3ASIM-3355v1/src/d3a/models/market/__init__.py#L266

src/d3a/models/myco_matcher/pay_as_clear.py
faizan2590: Added: https://github.com/gridsingularity/d3a/pull/1121/commits/8e7aa92b935c04751d47fe6e1b16b41579354e2f

integration_tests/steps/integration_tests.py
faizan2590: Don't think so as the tageted `kpi integration test` explicitly targeted PAB

src/d3a/models/myco_matcher/__init__.py
hannesdiedrich: Thanks for explaining!

src/d3a/models/myco_matcher/pay_as_clear.py
hannesdiedrich: Ahh, you pass it as a tuple. OK.

integration_tests/kpi.feature
```diff
@@ -56,12 +56,18 @@ Feature: KPI integration tests
      Then self_sufficiency of {'Grid': 1.0, 'Community': 0.666, 'House 1': 0.0, 'House 2': 0.5, 'House 3': 1.0} are correctly reported
      And self_consumption of {'Grid': 1.0, 'Community': 0.8, 'House 1': None, 'House 2': 1.0, 'House 3': 0.5} are correctly reported
 
+  """
+  Feedback required:
+  Earlied PAB was contiuously matching and announcing trade till it has been exhausted, but now it simply forwards the
+  recommended pair and rematch the derivative bids/offers in the next tick, due to which there are chances of unmatched
+  loads at the last tick leading to reduction in SS/SC
+  """
```
hannesdiedrich, line 64: I think I need to understand this before I can approve. Can you maybe elaborate a bit more? Maybe with an example?

faizan2590: Earlier PAB algo was simultaneously creating a match and entering into trade until all bids and offers have exhausted

```
    def match_offers_bids(self):
        while len(self._perform_pay_as_bid_matching()) > 0:
            for bid, offer in self._perform_pay_as_bid_matching():
                selected_energy = bid.energy if bid.energy < offer.energy else offer.energy
                original_bid_rate = bid.original_bid_price / bid.energy
                matched_rate = bid.energy_rate

                trade_bid_info = TradeBidOfferInfo(
                    original_bid_rate=original_bid_rate,
                    propagated_bid_rate=bid.price/bid.energy,
                    original_offer_rate=offer.original_offer_price/offer.energy,
                    propagated_offer_rate=offer.price/offer.energy,
                    trade_rate=original_bid_rate)

                self.accept_bid_offer_pair(bid, offer, matched_rate,
                                           trade_bid_info, selected_energy)
```
Now it only sends back the recommended pair, and the derived offers/bids from residual trade is recommended at the next tick

hannesdiedrich: I see. I wonder why the other KPI tests dont fail then. Why only this one where there are multiple levels?

faizan2590: What I understood of the reason is its the one without pv/power_producing device in one area, so the residual offers from other area also need time to propagate too

hannesdiedrich: Sorry, do not understand `the one without pv/power_producing device in one area` It has indeed PVs in all Houses. Can you please investigate more, because I dont like to shrug it of that easily. Backward compatibility should either be given or at least the reason for the change in behaviour should be known.

faizan2590: Please give `kpi.d3asim_2103_nopv1_2_sided_pab` being used in this scenario and then re-read my last comment.

hannesdiedrich: Ah yes, sorry, I had a look at the wrong setup file. OK, But lets look at the numbers here. I may understand why the self-sufficiency of the House2 is reduced (sending PV energy to somewhere else). However, by definition, the self-consumption of the Grid should be 1 in any case, right? There is no external entity that provides energy. 
Or am I missing something here?

src/d3a/models/myco_matcher/base_matcher.py
Hassan754: @faizan2590 I tested this and it worked fine

Hassan754: + I suggest we change this function to be a bit more generic, to sort depending on any attribute the user sends
For example it could be :
```python
    def sort_bids_offers(obj, attribute,  reverse_order=False):
       if hasattr(obj, attribute):
             return list(sorted(
                obj.values(),
                key=lambda b: getattr(b, attribute), reverse=reverse_order))
```

src/d3a/models/myco_matcher/pay_as_bid.py
```diff
@@ -0,0 +1,37 @@
...
+        1. Match the cheapest offer with the most expensive bid. This will favor the sellers
+        2. Match the cheapest offer with the cheapest bid. This will favor the buyers,
+           since the most affordable offers will be allocated for the most aggressive buyers.
+        """
+
+        return self._perform_pay_as_bid_match(bids, offers)
+
+    def _perform_pay_as_bid_match(self, bids, offers):
+        # Sorted bids in descending order
+        sorted_bids = self.sort_by_energy_rate(bids, True)
```
Hassan754, line 21: sort_by_energy_rate should not be called with self as it it a static method

spyrostz: I do not agree on that. Static methods can be called using self , exactly like normal methods. It also makes the code much less verbose than using `PayAsBidMycoMatcherClassName.sort_by_energy_rate`

integration_tests/kpi.feature
faizan2590: @hannesdiedrich Please give this commit a look https://github.com/gridsingularity/d3a/pull/1121/commits/d7ddb9f38cab1706fd2e98791f112b42378d1291

hannesdiedrich: Took me a while to understand but, great! Nice catch. This was falsely added when we added the seller_id to the bids and trades, right?
Why wasn't this an issue before i wonder? Do you have an explanation?
This proves that the integrationtests we have at the moment should not be changed (apart from something major was changed) . If an integrationtest fails, there must be something wrong.

integration_tests/kpi.feature
```diff
@@ -56,12 +56,18 @@ Feature: KPI integration tests
...
   Scenario: D3ASIM-2103, self sufficiency reports correct values on levels higher than houses with energy deficit for PAB
      Given we have a scenario named kpi.d3asim_2103_nopv1_2_sided_pab
      And d3a is installed
      When we run the d3a simulation on console with kpi.d3asim_2103_nopv1_2_sided_pab for 24 hrs (30, 30)
-     Then self_sufficiency of {'Grid': 1.0, 'Community': 0.833, 'House 1': 0.0, 'House 2': 0.5, 'House 3': 1.0} are correctly reported
-     And self_consumption of {'Grid': 1.0, 'Community': 1.0, 'House 1': None, 'House 2': 1.0, 'House 3': 0.5} are correctly reported
+     Then self_sufficiency of {'Grid': 1.0, 'Community': 0.833, 'House 1': 0.0, 'House 2': 0.25, 'House 3': 1.0} are correctly reported
+     And self_consumption of {'Grid': 1.0, 'Community': 1.0, 'House 1': None, 'House 2': 0.5, 'House 3': 0.5} are correctly reported
```
hannesdiedrich, line 70: So the self consumption increased in House 2, do you have an explanation for that?

integration_tests/grid_fee.feature
```diff
@@ -16,6 +16,7 @@ Feature: GridFee integration tests
      |       5       |     60      |     10      |
      |       9       |     60      |     5       |
 
+  @disabled
```
hannesdiedrich, line 19: Did you try to enable the newly disabled tests after your fix?

integration_tests/kpi.feature
faizan2590: > Took me a while to understand but, great! Nice catch. This was falsely added when we added the seller_id to the bids and trades, right?

Yes, from commit history, its been like that since I introduced `seller_id`

> Why wasn't this an issue before i wonder? Do you have an explanation?

`seller_id/buyer_id/buyer_origin_id` all are only relevant for ´kpi´, only explanation I can have is that it appeared now due to `self-consumption` ended-up being less than 1 for root area

integration_tests/kpi.feature
faizan2590: Same that now [seller_id](https://github.com/gridsingularity/d3a/commit/d7ddb9f38cab1706fd2e98791f112b42378d1291#diff-e504bd14031965b45bf3234fa47da69d52b61bc12220260f1a82a64aae8784b0R155) is properly tracked.

integration_tests/grid_fee.feature
faizan2590: They failed due to percentage fee and still failing.

integration_tests/kpi.feature
hannesdiedrich: Sorry, i miswrote. Self_sufficiency and self_consumption decrease for House 2 with your changes (reafctoring the matching algorithms). You blame the unmatched loads due to the lack of continues matching for this?

faizan2590: > You blame the unmatched loads due to the lack of continues matching for this?

Yes, that's my observation/ understanding so far and dropped a doc-string comment for other to gain attention while reviewing

src/d3a/models/myco_matcher/base_matcher.py
faizan2590: Im not sure, how you tested but its still not working for me.

https://user-images.githubusercontent.com/33482427/117308799-4a05f380-ae9b-11eb-9573-3cfa2c29231a.mov

integration_tests/kpi.feature
hannesdiedrich: I would suggest to give it another look to be sure that this is really the reason. 
Will review again, once other comment on this or just say: This is fine.

integration_tests/steps/two_sided_pay_as_clear.py
```diff
@@ -103,23 +104,23 @@ def test_offer_bid_market_clearing_rate_files(context):
...
 
 @then('clearing rate is equal to the bid_rate')
 def clearing_rate_at_bid_rate(context):
-    assert all(isclose(clearing[0], 25.0)
+    assert all(isclose(clearing, 25.0)
                for child in context.simulation.area.children
-               for market in child.past_markets
-               for clearing in list(market.state.clearing.values()))
+               for clearing in
+               list(child.bid_offer_matcher.match_algorithm.state.clearing.values()))
```
fievelk, line 126: Nit: you can iterate over `.values()` without converting to list:
```suggestion
               child.bid_offer_matcher.match_algorithm.state.clearing.values())
```

integration_tests/kpi.feature
faizan2590: Okay, after double checking House2's load isn't getting matched to its own PV for some market as House1's Load too bidding at the same rate, and then for the next tick House2's PV energy is already sold leading to reduce selfsufficiency/consumption at House2 level

integration_tests/steps/two_sided_pay_as_clear.py
faizan2590: Yes sure, will adapt it

integration_tests/kpi.feature
hannesdiedrich: Thanks for explaining. If everybody is aware of this change and agrees, then all good.

hannesdiedrich (dismissed):
Once everybody agrees on https://github.com/gridsingularity/d3a/pull/1121#discussion_r627375693, LGTM

integration_tests/kpi.feature
spyrostz: Agreed and I think this is an unavoidable aftermath of decoupling the matching from the clearing processes. For the time being though (until the real API is added) in order to satisfy the tests I would add a while loop that retries the matching and clearing until there are no further matches, similar to what the 2-sided PAB market did. I think it is worth to add it, not just for satisfying the tests, but also to satisfy the collaborations that we used in the Hackathon, which depended on this behavior.

faizan2590: Adapted as per last comment https://github.com/gridsingularity/d3a/pull/1121/commits/a30de754640733711a85fe07861968887db0e4ac

src/d3a/models/area/__init__.py
```diff
@@ -386,13 +388,16 @@ def _consume_commands_from_aggregator(self):
...
+        if ConstSettings.IAASettings.MARKET_TYPE == 2:
             if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
                 self.dispatcher.publish_market_clearing()
             else:
                 for market in self.all_markets:
-                    market.match_offers_bids()
+                    while bid_offer_pairs:=self.bid_offer_matcher.calculate_recommendation(  # NOQA
+                            *market.open_bids_and_offers, self.now):  # NOQA
+                        # bid_offer_pairs = self.bid_offer_matcher.calculate_recommendation(
+                        #     *market.open_bids_and_offers, self.now)
```
spyrostz, line 399: You can remove this

spyrostz (approved):
One comment, LGTM

PR #1128, opened 2021-04-29
D3ASIM-3450: remove validation of area names

fievelk:
The simulation results that can be downloaded by the users contain a directory structure based on the names of the areas in the scenario. If these names contain invalid characters (like `/`) the export process will either fail or be compromised.

Our previous solution was to restrict the allowed characters in area names. However, this created incompatibilities with several simulations that had already been created by users in the past. We therefore decided to allow those characters again, but to sanitize them when generating the paths for the downloaded results (see https://github.com/gridsingularity/d3a-web/pull/1326).

### DIFF
--- a/src/d3a/models/area/__init__.py
+++ b/src/d3a/models/area/__init__.py
@@ -18,32 +18,31 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import traceback
 from logging import getLogger
 from typing import List  # noqa
+from uuid import uuid4
+
 from cached_property import cached_property
+from d3a_interface.area_validator import validate_area
+from d3a_interface.constants_limits import ConstSettings, GlobalConfig
+from d3a_interface.utils import key_in_dict_and_not_none
 from pendulum import DateTime, duration, today
 from slugify import slugify
-from uuid import uuid4
-from d3a.constants import TIME_ZONE
+
+import d3a.constants
+from d3a.d3a_core.device_registry import DeviceRegistry
 from d3a.d3a_core.exceptions import AreaException
-from d3a.models.config import SimulationConfig
+from d3a.d3a_core.util import TaggedLogWrapper
 from d3a.events.event_structures import TriggerMixin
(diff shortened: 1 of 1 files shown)


hannesdiedrich (approved):
LGTM

src/d3a/d3a_core/rq_job_handler.py
```diff
@@ -31,11 +31,7 @@ def launch_simulation_from_rq_job(scenario, settings, events, aggregator_device_
         GlobalConfig.IS_CANARY_NETWORK = scenario.pop("is_canary_network", False)
         d3a.constants.RUN_IN_REALTIME = GlobalConfig.IS_CANARY_NETWORK
     saved_state = decompress_and_decode_queued_strings(saved_state)
-    log.error(f"Scenario: {scenario}\n"
-              f"Settings: {settings}\n"
-              f"Events: {events}\n"
-              f"Aggregator device mapping: {aggregator_device_mapping}\n"
-              f"Previous simulation state: {saved_state}")
```
faizan2590, line 38: I assume the intuition may be that these info already exist in DB, so no need to clutter the logs with redundant infos

fievelk: This is part of @hannesdiedrich's D3ASIM-3451. The changes ended up here when I targeted `staging` instead of `master`, but we decided to keep them and merge them together as part of this PR.

spyrostz (approved):
LGTM

faizan2590:
>  We therefore decided to allow those characters again, but to sanitize them when generating the paths for the downloaded results.

Sorry, can you point me where did you do the sanitization when generating the paths for the downloaded results.

fievelk:
> > We therefore decided to allow those characters again, but to sanitize them when generating the paths for the downloaded results.
> 
> Sorry, can you point me where did you do the sanitization when generating the paths for the downloaded results.

It's on the d3a-web repo. I'll add the reference in the description!

faizan2590 (approved):
lgtm

PR #1131, opened 2021-05-03
D3ASIM-1873: Update one sided market rate threshold

Hassan754:


integration_tests/steps/infinite_bus_tests.py
```diff
@@ -35,6 +35,7 @@ def check_buy_behaviour_ib(context):
     grid = context.simulation.area
     bus = list(filter(lambda x: x.name == "Infinite Bus", grid.children))[0]
     for time_slot, core_stats in context.raw_sim_data.items():
+        print(core_stats[name_uuid_map['Grid']]['trades'])
```
spyrostz, line 38: Reminder to remove before merging

spyrostz (approved):
LGTM, would be great to write an outcome of the discussion on the Jira ticket so that the research team understands what is implemented here.

hannesdiedrich (approved):
LGTM once Spyros comments are addressed.

src/d3a/setup/strategy_tests/load_const_price_increase.py
```diff
@@ -0,0 +1,66 @@
...
+
+You can expect 5 to 6 updates per market slot (with 15 sec ticks). See below:
+
+Considering tick_length = 15s, and max_offer_traversal_length = 10 (in order to propagate
+offer from one end to the other extreme end). So, the minimum waiting time for offer update
+would be offer_update_wait_time = tick_length * max_offer_traversal_length (15 * 10 = 150s)
+Considering, time_slot =  15m -> 900s
+The max_possible_offer_update_per_slot = time_slot / offer_update_wait_time (900/150=6).
+However, due to some reason, max_possible_offer_update_per_slot is made one unit less.
+Once Spyros is back, it has to be discussed.
```
hannesdiedrich, line 34: Please remove this and also from the file were you copy pasted this from. thanks!

integration_tests/load_tests.feature
```diff
@@ -33,3 +33,9 @@ Feature: Load Tests
      And export is_needed
     When we run the simulation with setup file strategy_tests.user_rate_profile_load_dict and parameters [24, 30, 30, 4]
     Then LoadHoursStrategy does not buy energy with rates that are higher than the provided profile
+
+  Scenario: LOAD CONSTANT based price increase
```
hannesdiedrich, line 37:
```suggestion
  Scenario: Load strategy increases its accepted offers price in one sided market
```

integration_tests/steps/load_tests.py
```diff
@@ -51,6 +52,28 @@ def check_traded_energy_rate(context):
...
+    rate_list = \
+        [ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE - i *
+         energy_rate_change_per_update
+         for i in range(number_of_available_updates)]
+    if load.strategy.bid_update.fit_to_limit is False:
+        for market in house.past_markets:
+            for trade in market.trades:
+                if trade.buyer == load.name:
+                    assert any([isclose(trade.offer.price / trade.offer.energy,
+                                        rate) for rate in rate_list])
```
hannesdiedrich, line 72: This does not check whether the offer price really got increased. There is only one trade in each market and probably they have all the same price. 
I think you can only check it with a unittest. Please correct me, if I overlook something. Thanks!

hannesdiedrich (commented):
Added some comments, LGTM after addressed

src/d3a/models/strategy/storage.py
```diff
@@ -513,8 +513,7 @@ def buy_energy(self, market, offer=None):
             return
         if self.state.has_battery_reached_max_power(-FLOATING_POINT_TOLERANCE, market.time_slot):
             return
-        max_affordable_offer_rate = min(self.bid_update.get_updated_rate(market.time_slot),
-                                        self.bid_update.final_rate[market.time_slot])
+        max_affordable_offer_rate = self.bid_update.get_updated_rate(market.time_slot)
```
spyrostz, line 516: Does not hurt to change that, however I am curious why this was needed in the first place. Essentially, what this did was a sanity check in order to always make sure that the bid rate will never surpass the final rate (in order to be able to cover potential problems with the rate calculation like e.g. border conditions due to floating point precision) , therefore I think it would be better if you left it at peace unless there is a specific reason to change it.

spyrostz (approved):
1 comment, LGTM once this and Hannes' comments are fixed.

src/d3a/models/strategy/storage.py
faizan2590: sorry to jump in, Spyros AFAI, the sanity check is already part of [bid_update.get_updated_rate](https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/strategy/update_frequency.py#L135) implementation

spyrostz: Then it is all good here, leave it as it is

spyrostz (approved):
LGTM

tests/test_strategy_load_hours.py
```diff
@@ -497,12 +504,22 @@ def test_predefined_load_strategy_rejects_incorrect_rate_parameters(use_mmr, ini
...
     load.area = FakeArea()
+    load.owner = load.area
     load.event_activate()
+    assert load.state._energy_requirement_Wh[TIME] == 25.0
+    offer = Offer('id', now(), 1, (MIN_BUY_ENERGY/500), 'A', market_test1)
+    load._one_sided_market_event_tick(market_test1, offer)
+    assert load.bid_update.get_updated_rate(TIME) == 0
+    assert load.state._energy_requirement_Wh[TIME] == 25.0
+    load.event_tick()
+    assert load.bid_update.get_updated_rate(TIME) == 10
```
hannesdiedrich, line 520: Hmm, sorry to be relentless here. But this only checks for the functionality of `load.bid_update.get_updated_rate`, but not the correct bid placement. 
What happened to my suggestion, to access the `market.bids` member in the integrationtests?

hannesdiedrich: Ah, yes, you are right. And now I understand your unittest. Agreed.

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #1133, opened 2021-05-06
D3ASIM-3442: Added tolerance to floating point comparison for the off…

spyrostz:
…er energy and the to-be-accepted energy.

### DIFF
--- a/src/d3a/models/market/one_sided.py
+++ b/src/d3a/models/market/one_sided.py
@@ -18,6 +18,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 from typing import Union  # noqa
 from logging import getLogger
 from pendulum import DateTime
+from math import isclose
 
 from d3a.events.event_structures import MarketEvent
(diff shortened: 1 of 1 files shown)


hannesdiedrich (approved):
Did you validate that it is a floating point issue or is this PR just for checking out whether this is the issue ?

spyrostz:
> Did you validate that it is a floating point issue or is this PR just for checking out whether this is the issue ?

Of course, I tested it by connecting to a clone of the dev DB, which had another configuration that exhibited the same problem (configuration name is GZ, and you can see the same logs as the bug reports in Kibana - reason why I wanted the dev DB was because the DB there is much smaller in size and quicker to clone compared to production). Reproduced the issue first, the issue indeed appeared, then checked out the energy values, validated that they were within the range of the floating point precision, applied the fix, rerun the same configuration again and no errors appeared anymore.   
TLDR, this fix should solve the issue in production as is.

faizan2590 (approved):

PR #1137, opened 2021-05-10
Feature/d3 asim 3453: Replace simulation results service Redis pubsub with Kafka

faizan2590:


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -61,6 +78,21 @@ def __init__(self, job_id, initial_params, area, should_export_plots):
                 ConstSettings.GeneralSettings.EXPORT_ENERGY_TRADE_PROFILE_HR:
             self.offer_bid_trade_hr = OfferBidTradeGraphStats()
 
+        if constants.KAFKA_MOCK:
+            self.producer = Mock()
```
fievelk, line 82: Can you please clarify this? Production code is generally agnostic of testing functionalities (that's one reason why we wanted to remove all the `ast.literal_eval()` calls for scenarios).

fievelk: One possible way to achieve this would be to add `producer` to the `__init__` arguments. Then in tests you could do something like
```python
epb = SimulationEndpointBuffer("1", {"seed": 0}, grid, True, producer=Mock())
```

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -33,6 +39,17 @@
     'max': None
 }
 
+BOOTSTRAP_SERVERS = os.environ.get('BOOTSTRAP_SERVERS')
+SASL_PLAIN_USERNAME = os.environ.get('SASL_PLAIN_USERNAME')
+SASL_PLAIN_PASSWORD = os.environ.get('SASL_PLAIN_PASSWORD')
+SECURITY_PROTOCOL = os.environ.get('SECURITY_PROTOCOL', 'SASL_SSL')
+SASL_MECHANISM = os.environ.get('SASL_MECHANISM', 'SCRAM-SHA-512')
```
fievelk, line 46: Nit: please use `"` over `'`

fievelk: (This also applies to the rest of the PR)

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -61,6 +78,21 @@ def __init__(self, job_id, initial_params, area, should_export_plots):
                 ConstSettings.GeneralSettings.EXPORT_ENERGY_TRADE_PROFILE_HR:
             self.offer_bid_trade_hr = OfferBidTradeGraphStats()
 
+        if constants.KAFKA_MOCK:
+            self.producer = Mock()
+        elif BOOTSTRAP_SERVERS is not None:
```
fievelk, line 83:
```suggestion
        elif BOOTSTRAP_SERVERS:
```

src/d3a/d3a_core/sim_results/endpoint_buffer.py
fievelk, line 46: It would be great if you could add some documentation about these variables. For example, what does `BOOTSTRAP_SERVERS` do?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -61,6 +78,21 @@ def __init__(self, job_id, initial_params, area, should_export_plots):
...
+            kwargs = {'bootstrap_servers': BOOTSTRAP_SERVERS,
+                      'sasl_plain_username': SASL_PLAIN_USERNAME,
+                      'sasl_plain_password': SASL_PLAIN_PASSWORD,
+                      'security_protocol': SECURITY_PROTOCOL,
+                      'ssl_context': context, 'sasl_mechanism': SASL_MECHANISM,
+                      'api_version': (0, 10), 'retries': 5, 'buffer_memory': 2048000000,
```
fievelk, line 89: Where does the `api_version` come from?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -61,6 +78,21 @@ def __init__(self, job_id, initial_params, area, should_export_plots):
...
+                      'max_request_size': 2048000000}
+            self.producer = KafkaProducer(**kwargs)
+        else:
+            kwargs = {'bootstrap_servers': 'localhost:9092'}
```
fievelk, line 93: I think it would be good to have this value as an environment value (ideally defined in an `.env` file or something similar)

integration_tests/environment.py
```diff
@@ -72,3 +73,9 @@ def before_all(context):
     constants.D3A_TEST_RUN = True
     if os.environ.get("DISPATCH_EVENTS_BOTTOM_TO_TOP") == "False":
         d3a.constants.DISPATCH_EVENTS_BOTTOM_TO_TOP = False
+    os.system('docker-compose up -d')
+    sleep(60)
```
spyrostz, line 77: This will delay the integration tests needlessly IMO, because even at the current implementation the integration tests do not use Redis at all, and I think the same should apply for Kafka

spyrostz (commented):
In general I think instead of having the Kafka connection as a member of the endpoint buffer, it would be a much better approach to create a separate class that hides all the Kafka connection, similar to the RedisSimulationCommunication class. This will have 2 advantages:
1. easier to mock for the integration and unit tests, that way you will be able to deal with @fievelk 's correct comment easily
2. easier to extend if we want to add topics other than the d3a-results.

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -188,6 +188,9 @@ def publish_results(self, endpoint_buffer):
 
         self._handle_redis_job_metadata()
         self.redis_db.publish(self.result_channel, results)
+        endpoint_buffer.producer.send(
+            'd3a-results', value=json.dumps(result_report).encode('ascii'),
```
spyrostz, line 192: 2 comments here:
1. it would be better to have a separate constant for 'd3a-results' , for instance KAFKA_SIM_RESULTS_TOPIC
2. is the ascii encoding required here ? If not, why not using utf-8 as the encoding type?

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -188,6 +188,9 @@ def publish_results(self, endpoint_buffer):
 
         self._handle_redis_job_metadata()
         self.redis_db.publish(self.result_channel, results)
+        endpoint_buffer.producer.send(
```
hannesdiedrich, line 191: You can get rid of L186-190 now, right?

docker-compose.yml
```diff
@@ -0,0 +1,46 @@
+---
```
hannesdiedrich, line 1: I would vote to rename this file to show that this only starts a kafka server, e.g. `docker-compose-kafka` in order to not let the user confuse it for a d3a docker-compose file.
Also, I see this only being used for the integrationtests. How about moving it into the integration_test folder?

hannesdiedrich:
I left my 5 cents. Happy to review again, once Spyros' comments are addressed. Thanks!

integration_tests/steps/integration_tests.py
```diff
@@ -512,38 +512,24 @@ def mock_publish_json(*args):
...
-def final_results(context):
-    context.final_results_count = 0
-
-    def final_res_count(_):
-        context.final_results_count += 1
-    context.simulation.redis_connection.publish_results = final_res_count
+    context.simulation.kafka_connection.publish = interm_res_count
 
 
 @when('the redis_connection is enabled')
```
Hassan754, line 524:
```suggestion
@when('the kafka_connection is enabled')
```

integration_tests/steps/integration_tests.py
```diff
@@ -512,38 +512,24 @@ def mock_publish_json(*args):
...
+@then('intermediate results are transmitted on every slot and final results once')
 def interm_res_report(context):
     # Add an extra result for the start of the simulation
-    assert context.interm_results_count == 12
-
-
-@then('final results are transmitted once')
-def final_res_report(context):
-    assert context.final_results_count == 1
+    assert context.interm_results_count == 13
```
Hassan754, line 532: Please add a final results assertion if needed here

Hassan754 (approved):
Left some comments, LGTM

integration_tests/steps/integration_tests.py
```diff
@@ -512,38 +512,24 @@ def mock_publish_json(*args):
     context.simulation.redis_connection._sub_callback_dict[channel](payload)
 
 
-@when('the simulation is able to transmit intermediate results')
+@when('the simulation is able to transmit intermediate and final results')
 def interm_results(context):
     context.interm_results_count = 0
 
-    def interm_res_count(_):
+    def interm_res_count(_, _1):
```
hannesdiedrich, line 519: Can you please explain the `_1` ?

src/d3a/models/market/two_sided.py
```diff
@@ -252,10 +252,10 @@ def match_recommendation(self, recommended_list):
                 propagated_bid_rate=bid.price/bid.energy,
                 original_offer_rate=offer.original_offer_price/offer.energy,
                 propagated_offer_rate=offer.price/offer.energy,
-                trade_rate=recommended_pair.trade_rate)
+                trade_rate=original_bid_rate)
```
hannesdiedrich, line 255: Can you please explain this, change too? Why is always the `original_bid_rate` used for the trade_rate? I would expect the `bid.energy_rate`  for the pay-as-bid and the `clearing_rate` for the pay-as-clear market. Consequently, I would expect that the original implementation is correct. Right?

src/d3a/models/market/two_sided.py
```diff
@@ -252,10 +252,10 @@ def match_recommendation(self, recommended_list):
...
 
             bid_trade, trade = self.accept_bid_offer_pair(
-                bid, offer, trade_bid_info.trade_rate, trade_bid_info, selected_energy
+                bid, offer, bid.energy_rate, trade_bid_info, selected_energy
```
hannesdiedrich, line 258: The same applies here of course. But probably I am missing something here. Please fill me in. Thanks a lot!

hannesdiedrich (commented):
Left a question concerning the grid fees fix. The Kafka part LGTM once the corresponding question on d3a-web is answered.

src/d3a/models/market/two_sided.py
spyrostz: Well, this is far more straightforward than the former comment. In our markets, we always clear based on the bid rate and not the offer rate. Therefore the trade rate should be the bid rate. However, the myco should have already handled that, therefore I think it is much better to use the `recommended_pair.trade_rate` instead. Will apply the change now, thanks!

integration_tests/steps/integration_tests.py
spyrostz: I have removed the final results, because now there is only one function that handles all results without distinction between intermediate and final

integration_tests/steps/integration_tests.py
spyrostz: Will apply it thanks!

integration_tests/steps/integration_tests.py
spyrostz: Yes, the signature of the `publish` method has changed for Kafka and requires one extra argument (the partition name, which is the simulation_id). Therefore the mock of the publish method should include 2 positional arguments, for which we do not care for either of them. Therefore the Pythonic way of setting arguments that you actually do not need is this: _, _1, _2, _3 ....

docker-compose.yml
spyrostz: I'll move it in another directory since I removed the dependency of Kafka for the integration tests.

src/d3a/models/market/two_sided.py
hannesdiedrich: OK, thanks a lot! Indeed, I wasn't fully aware of the concept of `TradeBidOfferInfo is a structure that is referring to the end-to-end trade chain [..] and not to the trade of this market` (any more, probably self-protection of my brain ;-) ).
Nevertheless. I took some time and looked into the implementation again. I also remember the discussion about the nomenclature now and that I was not a fan of naming it just 'trade_rate'. Anyway. Got it now (i think ;-)) .Thanks!

src/d3a/models/market/two_sided.py
hannesdiedrich: OK, I am a bid confused by your comment `In our markets, we always clear based on the bid rate and not the offer rate`. What about the pay-as-clear market? Isn't the clearing rate applied However, `recommended_pair` is set correctly for this in the PAC case AFAIS, so your proposed change is adapting for this. Thanks!

spyrostz: You are right, my comment only applies for the pay as bid case, for the pay as clear we really need the clearing rate. Nevertheless should be fixed now, thanks!

hannesdiedrich (dismissed):
LGTM

PR #1138, opened 2021-05-10
D3ASIM-3442: Added the offer fix for floating point precision errors …

spyrostz:
…on energy values to the bids.


hannesdiedrich (approved):
Too bad! Thanks a lot for fixing!

Hassan754 (approved):

PR #1141, opened 2021-05-11
D3ASIM-3447: first myco api client release

Hassan754:


src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,71 @@
...
+        self.publish_offers_bids(data)
+
+    def publish_offers_bids(self, data):
+        channel = f"{self.channel_prefix}get_offers_bids/response/"
+        self.myko_ext_conn.publish(channel, json.dumps(data))
+
+    def match_recommendations(self, payload):
+        for record in payload["data"]:
+            market = self.markets_mapping[record["market_id"]]
+            validate_authentic_bid_offer_pair(
```
faizan2590, line 59: `validate_authentic_bid_offer_pair ` needs to be adapted as at the moment it validates `bid/offer` based on objects derived from `class Bid` & `class Offer` where as here AFAIS it is received in the form of json_dict

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,71 @@
...
+        self.recommendations = []
+
+    def _setup_redis_connection(self):
+        self.myko_ext_conn = RedisCommunicator()
+        self.myko_ext_conn.sub_to_channel(f"{self.channel_prefix}get_offers_bids/",
+                                          self.get_offers_bids)
+        self.myko_ext_conn.sub_to_channel(f"{self.channel_prefix}post_recommendations/",
+                                          self.match_recommendations)
+
+    def get_offers_bids(self, payload):
```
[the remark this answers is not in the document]
Hassan754: Yes, will rename it to `publish_offers_bids`

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -8,13 +9,17 @@
 
 class MycoMatcher:
     def __init__(self):
-        self.bid_offer_pairs = []
+        ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE = \
+            BidOfferMatchAlgoEnum.CUSTOM.value  # For testing
```
Hassan754, line 13: Reminder to revert before merging

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,76 @@
...
+    def publish_offers_bids(self, message):
+        """
+        Function that queries publishes open offers and bids
+        published data are of the following format
+        market_offers_bids_list_mapping = {"market_id" : {"bids": [], "offers": [] }, }
+        """
+        # TODO: message can contain filters
+        data = {"event": "offers_bids_response"}
+        market_offers_bids_list_mapping = {}
+        for area_uuid, markets in self.area_uuid_markets_mapping.items():
```
hannesdiedrich, line 39: You are not using the `area_uuid` here. I dont see a danger of hash collisions, so the market_ids should be enough to distinguish between the markets, right?

src/d3a/models/myco_matcher/__init__.py
hannesdiedrich: This should be set inside a setup file for an integrationtest.

hannesdiedrich (commented):
Left some comments. Will review once all comments on all repos are addressed.

src/d3a/d3a_core/simulation.py
```diff
@@ -388,7 +388,8 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
         self.deactivate_areas(self.area)
         self.simulation_config.external_redis_communicator.\
             publish_aggregator_commands_responses_events()
-
+        if self.simulation_config.external_connection_enabled:
```
hannesdiedrich, line 391: Non-market areas should not publish these events AFAIS.
I propose to move the `publish_event_finish_myco` and `publish_event_tick_myco` and `publish_event_market_cycle_myco ` into the area and check whether the strategy is None. 

Also as a side note, in the future we might have to think about only sending these events for areas that have an active myco client connected to them. But I guess for these initial cases, we can leave it as is.

src/d3a/models/myco_matcher/external_matcher.py
hannesdiedrich: Any opinion here @Hassan754 . If you are not using the area_uuid, please use `.values()` instead. Thanks

src/d3a/d3a_core/simulation.py
Hassan754: I'm publishing ticks / markets from the simulation not from areas
Also I contacted Colin and he mentioned that we will only allow 1 myco client to get open offers/bids and post recommendations (and he did not deny that they may leave it open for clients to compete) so he also insisted that I should not put constraints now because they need more time to agree on the process of registration, this is why I;'m only validating if the client is loggedin

hannesdiedrich: You mixed something up here with my comment on d3a-web.

Here I am just saying that it does not make sense to send the e.g. `publish_event_market_cycle_myco` for a load. Only for a market it makes sense. And This is why i am advocating to move this into the area where you have access to the check `if area.strategy` (if it is None, then it is a market area).

Hassan754: @hannesdiedrich you're missing the fact the the Myco client is not aware of areas, it is for the whole simulation
pay attention to the difference between it and the api-client , here the client subscribes to the simulation channel whereas in the api-client we used to instantiate objects for each area where it subscribes to this specific area events

hannesdiedrich: 🤦  I am sorry. You are right of course. If this is settled, then some of my other comments dont make sense as well. Let me re-review. Thanks for bering with me!

src/d3a/d3a_core/simulation.py
```diff
@@ -362,10 +361,11 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
                     approve_aggregator_commands()
 
                 current_tick_in_slot = tick_no % config.ticks_per_slot
-                if self.simulation_config.external_connection_enabled and \
-                        external_global_statistics.is_it_time_for_external_tick(
+                if self.simulation_config.external_connection_enabled:
```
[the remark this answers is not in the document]
Hassan754: Sure, makes sense !

src/d3a/d3a_core/simulation.py
```diff
@@ -362,13 +364,18 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
-                    self.redis_connection.publish_event_tick_myco()
                     if external_global_statistics.is_it_time_for_external_tick(
                             current_tick_in_slot):
                         external_global_statistics.update()
 
                 self.area.tick_and_dispatch()
                 self.area.update_area_current_tick()
+                if self.simulation_config.external_connection_enabled and \
+                        ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == \
+                        d3a.constants.BidOfferMatchAlgoEnum.CUSTOM.value and \
```
hannesdiedrich, line 375: This is getting kind of long and I suggest to refactor this into a function (can be reused in L398 too)

hannesdiedrich (commented):
One last refactoring comment. LGTM after addressed.

src/d3a/global_utils.py
```diff
@@ -0,0 +1,9 @@
+from d3a.constants import BidOfferMatchAlgoEnum
+from d3a_interface.constants_limits import ConstSettings
+
+
+def check_is_exact_match_type(bid_offer_match_algo: BidOfferMatchAlgoEnum):
+    """Compares the matching algorithm set in IAASettings with the passed bid_offer_match_algo
+    Returns True if both are matched
+    """
+    return ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == bid_offer_match_algo.value
```
hannesdiedrich, line 9: IMO this way of implementing it does not make the check shorter or more readable. You actually only want to check if the match type is custom right? There will not be any other custom match algorithms in the future, so I propose:
```suggestion
def is_custom_matching_enabled():
    return ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == d3a.constants.BidOfferMatchAlgoEnum.CUSTOM and self.simulation_config.external_connection_enabled
```
I also added the check for the `external_connection_enabled` to make the call in the simulation.py even shorter

spyrostz (dismissed):
LGTM once Hannes' comments are dealt with

src/d3a/models/myco_matcher/external_matcher.py
hannesdiedrich: This still needs to be dealt with.

src/d3a/global_utils.py
```diff
@@ -0,0 +1,10 @@
+from d3a.constants import BidOfferMatchAlgoEnum
```
hannesdiedrich, line 1: Why does this need to be a separate module?

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: @hannesdiedrich in the Test 2 we need to pass the area_uuid to the Myco client, this is why I left it here

src/d3a/global_utils.py
Hassan754: This is to solve some imports issue, it used to reside in a directory 2 levels below this one, and we had to import it from different directories in the same level

src/d3a/global_utils.py
hannesdiedrich: What happened to my comment `I also added the check for the external_connection_enabled to make the call in the simulation.py even shorter` ? Didnt you like that?

src/d3a/models/myco_matcher/external_matcher.py
hannesdiedrich: OK! Understood.

src/d3a/global_utils.py
hannesdiedrich: I am not a fan of adding yet another util module. And I really dont understand the import issues. 
Can you try to add it to `d3a/d3a-core/util.py` please?

src/d3a/global_utils.py
Hassan754: @hannesdiedrich the `external_connection_enabled` attribute exists in the `self.simulation_config` object which does not exists in the function, we can pass it to it but this won't make the code shorter at all , it's basically the same

hannesdiedrich: (For future PR reviews: This comment would have bee great even before I asked for it. Thanks)

I understand. Then leave it out. Sorry for dragging this cosmetic issue so long. I thought is is a simple change and didnt thing it is such a hassle.

src/d3a/d3a_core/rq_job_handler.py
```diff
@@ -96,6 +96,10 @@ def launch_simulation_from_rq_job(scenario, settings, events, aggregator_device_
                 ConstSettings.IAASettings.MARKET_TYPE = 2
                 ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE = \
                     d3a.constants.BidOfferMatchAlgoEnum.PAY_AS_CLEAR.value
+            if spot_market_type == 4:
```
fievelk, line 99: Where can I find documentation about the `4`?

fievelk: Found it! https://github.com/gridsingularity/d3a-web/blob/41bcce49f00dedf83723ac868f0e393ac141236a/simulation/graphql/graphql_enums.py#L28-L31

I really think we should turn these integers into Enums.

src/d3a/d3a_core/simulation.py
```diff
@@ -16,42 +16,42 @@
...
-from time import sleep, time, mktime
-from numpy import random
 from importlib import import_module
 from logging import getLogger
+from time import sleep, time, mktime
 
+import click
+import d3a.constants
+import psutil
+# noinspection PyUnresolvedReferences
```
fievelk, line 31: I would personally not add this, since it just applies to PyCharm.

src/d3a/d3a_core/simulation.py
```diff
@@ -339,6 +339,9 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
             if self.simulation_config.external_connection_enabled:
                 external_global_statistics.update(market_cycle=True)
                 self.area.publish_market_cycle_to_external_clients()
+                if is_custom_matching_enabled():
+                    self.simulation_config.bid_offer_matcher.\
+                        match_algorithm.publish_market_cycle_myco()
```
fievelk, line 344: Nit: to follow PEP-8 and avoid using backslashes you could use parentheses:
```suggestion
                if is_custom_matching_enabled():
                    (self.simulation_config.bid_offer_matcher
                        .match_algorithm.publish_market_cycle_myco())
```

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -8,13 +10,14 @@
...
     def __init__(self):
-        self.bid_offer_pairs = []
         if ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == \
                 BidOfferMatchAlgoEnum.PAY_AS_BID.value:
             self.match_algorithm = PayAsBidMatcher()
         elif ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == \
                 BidOfferMatchAlgoEnum.PAY_AS_CLEAR.value:
             self.match_algorithm = PayAsClearMatcher()
+        elif is_custom_matching_enabled():
+            self.match_algorithm = ExternalMatcher()
```
fievelk, line 20: I would suggest to uniform this and the other checks (here you use a function, while the other `if` clauses don't). Maybe it would even possible to use a dictionary for the matching, like:
```python
MARKET_TO_MATCHING_ALGORITHM_MAPPING = {
    BidOfferMatchAlgoEnum.PAY_AS_CLEAR.value: PayAsClearMatcher,
    BidOfferMatchAlgoEnum.PAY_AS_BID.value: PayAsBidMatcher,
    BidOfferMatchAlgoEnum.CUSTOM.value: ExternalMatcher,
}
```
and then you could just do
```python
self.match_algorithm = MARKET_TO_MATCHING_ALGORITHM_MAPPING[ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE]
```
Or something similar.

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,137 @@
...
+
+import d3a.constants
+from d3a.d3a_core.redis_connections.redis_area_market_communicator import ResettableCommunicator
+from d3a.models.market import validate_authentic_bid_offer_pair
+from d3a.models.market.market_structures import BidOfferMatch, offer_from_JSON_string, \
+    bid_from_JSON_string
+from d3a.models.myco_matcher.base_matcher import BaseMatcher
+
+
+class ExternalMatcher(BaseMatcher):
```
fievelk, line 11: To describe this matching strategy, you've always used the `custom` keyword. I think we should uniform the terminology and either use `custom` or `external`, not both.
So I would vote for either changing this class name to `CustomMatcher` (which I'm not so sure about) or changing the `BidOfferMatchAlgoEnum` to `EXTERNAL`. Or a third option, if you have it!

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,137 @@
...
+    def __init__(self):
+        super(ExternalMatcher, self).__init__()
```
fievelk, line 13:
```suggestion
        super().__init__()
```
This is the python 3 version.

src/d3a/models/myco_matcher/__init__.py
fievelk: If the logic you want to enforce is more complex, you could create a class for it (like `MatchingAlgorithmMapper.get_matching_algorithm()`)

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,137 @@
...
+        Myco API
+        """
+        channel = f"external-myco/{d3a.constants.COLLABORATION_ID}/response/events/"
+        data = {"event": "tick"}
+        self.myco_ext_conn.publish_json(channel, data)
+
+    def publish_market_cycle_myco(self):
+        """
+        Myco API
+        """
```
fievelk, line 123: - Nit: I'm not convinced that this docstring says much about the method (especially since it's the same docstring repeated for multiple methods).
- Nit: PEP 257 (the one for docstrings) recommends using a single line when possible. So this would be `"""Myco API"""`. Anyway, this is very minor and I appreciate the fact that you added documentation 👍

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,137 @@
...
+    def _setup_redis_connection(self):
+        self.myco_ext_conn = ResettableCommunicator()
+        self.myco_ext_conn.sub_to_multiple_channels(
+            {"external-myco/get_simulation_id": self.get_simulation_id,
+             f"{self.channel_prefix}get_offers_bids/": self.publish_offers_bids,
+             f"{self.channel_prefix}post_recommendations/": self.match_recommendations})
+
+    def publish_offers_bids(self, message):
+        """
+        Function that queries publishes open offers and bids
```
fievelk, line 31: Nit, if you're intersted: if you want to follow the PEP recommendations (from [PEP 257](https://www.python.org/dev/peps/pep-0257/#one-line-docstrings)):
> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

So this could become:
```suggestion
        """Query published open offers and bids.
        
        Published data are of the following format: ...
```
(The blank line in between is also part of the PEP for multiline docstrings).

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,137 @@
...
+        published data are of the following format
+        market_offers_bids_list_mapping = {"market_id" : {"bids": [], "offers": [] }, }
+        """
+        # TODO: message can contain filters
```
fievelk, line 35: Is this a `TODO` or a regular comment?

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,137 @@
...
+                bids, offers = market.open_bids_and_offers
+                market_offers_bids_list_mapping[market.id]["bids"].extend(
+                    list(bid.serializable_dict() for bid in bids.values()))
+                market_offers_bids_list_mapping[market.id]["offers"].extend(
+                    list(offer.serializable_dict() for offer in offers.values()))
+        data.update({
+            "market_offers_bids_list_mapping": market_offers_bids_list_mapping,
+        })
+
+        channel = f"{self.channel_prefix}response/get_offers_bids/"
```
fievelk, line 51: Sorry for noticing this late. I think that it would actually be nice to call the channel `offers_bids` (without the `get_`). The reason is that `get` describes only one part of the interactions with the channel, which is used both for publishing and for subscribing. What do you think?

fievelk: Especially for this method, the inconsistency seems evident to me: it is `publish_offers_bids()`, but it's using a channel that is called `get_offers_bids`

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,137 @@
...
+        channel = f"{self.channel_prefix}response/matched_recommendations/"
+        response_dict = {"event": "match", "status": "success"}
+        data = json.loads(message.get("data"))
+        recommendations = data.get("recommended_matches", [])
+        validated_records = {}
+        for record in recommendations:
+            market = self.markets_mapping.get(record.get("market_id"), None)
+            if market is None or market.readonly:
+                # The market is already finished or doesn't exist
+                del record
```
fievelk, line 67: Why is this needed?

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -0,0 +1,137 @@
...
+                validated_records[record.get("market_id")].append(BidOfferMatch(
+                    bid,
+                    record.get("selected_energy"),
+                    offer,
+                    record.get("trade_rate")))
+            except AssertionError:
+                # If validation fails or offer/bid were consumed
+                response_dict["status"] = "fail"
+                response_dict["message"] = "Validation Error"
+                break
```
fievelk, line 95: I would highly recommend replacing assertions with custom exceptions. As a best practice, it's recommended to use assertions only to check for code issues that need to be spotted during the development/debug phase (for coding mistakes, for example). 
For cases in which the program can fail because of external input (like in this case) the best thing is to use exceptions and handle them.

One more reason to use exceptions (that doesn't apply to our current case, but might in the future) is that `assert` statements can be automatically removed if optimizations are requested by Python.

fievelk: Some sources that explain it clearly (in my opinion):
- https://softwareengineering.stackexchange.com/a/225996
- https://stackoverflow.com/questions/40182944/whats-the-difference-between-raise-try-and-assert

fievelk: (So usually you want assertions to fail badly and not to be catched, because they're meant to be seen and fixed by the developer. Exceptions are meant to be catched instead, since they are closely related to the production code.)

fievelk (commented):
I left some comments. Some of them are minor (like docstring formatting) while other are more important (like assertions/exceptions and some refactoring). Will review again once the comments are addressed :)

src/d3a/models/myco_matcher/__init__.py
Hassan754: This can work for setting the Match Algorithm, however the `is_custom_matching_enabled` is intended as a shortcut for checking if the external matching is enabled, not only for setting the matching algorithm.

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: This is a helpful TODO comment for the Test 2

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: It is a remnant of previous implementation

src/d3a/models/myco_matcher/__init__.py
fievelk: I'm still not fully convinced. Right now, `is_external_matching_enabled()` has to be called just to decide if we're going to call some specific methods on one specific matcher, and the matcher class itself is instantiated based on this check. So I see some kind of redundancy.
I have the feeling that this could be refactored taking advantage of the standard interface for all matchers (the `BaseMatcher`). We could have one method shared by all children, and they would implement a different internal logic.

src/d3a/models/market/__init__.py
```diff
@@ -264,7 +264,8 @@ def get_bids_offers_trades(self):
...
 def validate_authentic_bid_offer_pair(bid, offer, clearing_rate, selected_energy):
-    assert bid.energy >= selected_energy
-    assert offer.energy >= selected_energy
-    assert (bid.energy_rate + FLOATING_POINT_TOLERANCE) >= clearing_rate
-    assert (bid.energy_rate + FLOATING_POINT_TOLERANCE) >= offer.energy_rate
+    if not (bid.energy >= selected_energy and
+            offer.energy >= selected_energy and
+            (bid.energy_rate + FLOATING_POINT_TOLERANCE) >= clearing_rate and
+            (bid.energy_rate + FLOATING_POINT_TOLERANCE) >= offer.energy_rate):
+        raise Exception
```
fievelk, line 271: This is a clear improvement over the previous assertions, that's great 👍 
I have another remark to this: it is highly encouraged to avoid raising/catching the base `Exception` class. It would be much better to create a specific exception for this and possibly catch it later on. It could be something generic like `MarketException` or something more specific like `BidOfferValidationException`, or anything really to avoid the base `Exception` :)

fievelk: Some more information: https://rules.sonarsource.com/python/RSPEC-112#:~:text=Raising%20instances%20of%20Exception%20and,prone%20and%20difficult%20to%20maintain.

Hassan754: The goal of this exception-handling here is to make sure no transaction is posted if any other one will fail, a runtime exception of any type can be thrown here for some other reason besides validation, so we either keep it the way it is now or add new Exception type and add another broad exception handler in the calling code, what do you think ?

src/d3a/models/myco_matcher/__init__.py
Hassan754: The problem that Hannes was pointing to is that in several places in the code we wanted to check if the external matching is enabled [See here](https://github.com/gridsingularity/d3a/blob/ceb0cd569ebd369540e69c5c502f993157bd94fb/src/d3a/d3a_core/simulation.py#L341)
I can encapsulate the algorithm selection in a function if you believe this is an addition, or I can create another function which we can pass the matching type and it verifies if it is set globally, and the `is_external_matching_enabled` will then call it with the corresponding argument.
If you have something else in mind let me know

src/d3a/models/market/__init__.py
fievelk: > a runtime exception of any type can be thrown here for some other reason besides validation

What kind of exceptions do you think might occur here? In theory, the method does not need to deal with all possible types of exceptions, because other parts of the code should have already dealt with those. For example, you will in general expect `bid.energy_rate` to exist and never raise an `AttributeError`. If you really want to catch all possible types of exceptions (which is not ideal, but it's an option sometimes) you could do something like:
```python
try:
    if not (...):
except Exception as ex:
    raise SomeCustomException(ex) from ex

class SomeCustomException(Exception):
    """Exception raised when this and that happens."""
```
So you would still raise a custom exception, and the other parts of the code will know where it comes from. Also, it will contain the message raised by the original exception, which is necessary to properly debug errors.
At this point, you can either catch it using its name, or with a broad `Exception` (not ideal, but sometimes not easily avoidable).

fievelk: I suppose that, to avoid crashes on running simulations, there's a main entry point for d3a where all exceptions are dealt with, catching a broad `Exception`.

src/d3a/models/myco_matcher/__init__.py
fievelk: My idea is to avoid having more small functions that we need to call multiple times across the code. Also, these functions might be populated with `if` statements, which complicates stuff. That's why I am suggesting alternative ways, more SOLID-friendly, that I believe would be more robust and easier to test.
However, you have the biggest knowledge of this code. So if you believe that my suggestion is not practical and the current implementation is a better way to proceed, I will give my LGTM 👍

Hassan754: I will keep working in the same area in Test 2, will try to tackle this then

src/d3a/d3a_core/redis_connections/aggregator_connection.py
```diff
@@ -68,6 +67,7 @@ def _delete_not_owned_devices(self, indict: dict, aggregator_uuid: str, outdict:
     def _create_grid_tree_event_dict(self, aggregator_uuid: str) -> dict:
         """Accumulate area_stats_tree_dict information and initiate a event dictionary
         to be sent to the client"""
+        from d3a.d3a_core.singletons import external_global_statistics
```
spyrostz, line 70: I assume the lazy import is required in order to be able to instantiate the Redis connection only if needed, right ?

Hassan754: Not really, there is here to solve a  circular import issue
I had the map of the imports on paper, I can draw you the whole setup if needed

fievelk: Nit: do you think there are other ways to solve this? Circular imports often point out some issues with the package structure so it can be good to deal with it. Not blocking though.

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -1,23 +1,26 @@
...
 from d3a.models.myco_matcher.pay_as_bid import PayAsBidMatcher
 from d3a.models.myco_matcher.pay_as_clear import PayAsClearMatcher
 from d3a.d3a_core.exceptions import WrongMarketTypeException
+from d3a_interface.enums import BidOfferMatchAlgoEnum
 
 
 class MycoMatcher:
-    def __init__(self):
-        self.bid_offer_pairs = []
+    def init(self):
```
fievelk, line 12: What was the reason to make this a normal method, rather than using `__init__` as before? I think it would be good to document it in the docstring.

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -1,23 +1,26 @@
...
             self.match_algorithm = PayAsBidMatcher()
         elif ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == \
                 BidOfferMatchAlgoEnum.PAY_AS_CLEAR.value:
             self.match_algorithm = PayAsClearMatcher()
+        elif is_external_matching_enabled():
+            self.match_algorithm = ExternalMatcher()
         else:
-            raise WrongMarketTypeException(f'Wrong market type setting flag '
-                                           f'{ConstSettings.IAASettings.MARKET_TYPE}')
+            raise WrongMarketTypeException(f"Wrong market type setting flag "
```
fievelk, line 22: Nit:
```suggestion
            raise WrongMarketTypeException("Wrong market type setting flag "
```

src/d3a/d3a_core/redis_connections/aggregator_connection.py
Hassan754: Yes sure I prefer restructuring the modules, maybe in the context of another ticket

src/d3a/models/myco_matcher/__init__.py
Hassan754: This is because we need the MycoMatcher singleton at startup, but the matcher should be set later on

src/d3a/models/area/__init__.py
```diff
@@ -393,13 +391,17 @@ def tick(self):
...
                 self.dispatcher.publish_market_clearing()
             else:
-                for market in self.all_markets:
-                    bid_offer_pairs = self.bid_offer_matcher.calculate_recommendation(
-                        *market.open_bids_and_offers, self.now)
-                    while bid_offer_pairs:
-                        bid_offer_pairs = self.bid_offer_matcher.calculate_recommendation(
+                if is_external_matching_enabled():
+                    bid_offer_matcher.match_algorithm.area_uuid_markets_mapping.\
+                        update({self.uuid: self.all_markets})
```
fievelk, line 396: Could you please explain what happens here?
I'm not very convinced about this part: it interacts a bit too deeply with the internals of the matcher (without using any interface method). This makes testing trickier, among other stuff. I think it would be cool if we could refactor this somehow, but not necessarily in the context of this PR. Maybe for the next test, if that's ok with you.

Hassan754: This is updating the mapping of areas: open offers / bids
I agree with you that this should be using an interface, I'm touching this part in my current test to only update if the Area is not a device, I'll also add the interface there too
Thanks

integration_tests/steps/two_sided_pay_as_clear.py
```diff
@@ -105,22 +106,19 @@ def test_offer_bid_market_clearing_rate_files(context):
 @then('one-on-one matching of offer & bid in PAC happens at bid rate')
 def one_on_one_matching_at_clearing_rate_at_bid_rate(context):
     assert all(isclose(clearing, 30.0)
-               for child in context.simulation.area.children
                for clearing in
-               list(child.bid_offer_matcher.match_algorithm.state.clearing.values()))
+               list(bid_offer_matcher.match_algorithm.state.clearing.values()))
```
fievelk, line 110: Nit:
```suggestion
               bid_offer_matcher.match_algorithm.state.clearing.values())
```

fievelk: Same applies below

fievelk (dismissed):
LGTM. 
I would like to have some more documentation in the singleton `init()` method: singletons are quite delicate beasts, so it would be great to explicitly document why we're using them and how.
Among other things, I like the refactoring of Offers/Bids as dataclasses, I think that's a cool improvement for the future.

spyrostz (approved):
One clarification comment, LGTM though

src/d3a/d3a_core/redis_connections/aggregator_connection.py
spyrostz: I also agree with Pierpaolo, that in the long run we should restructure the modules to avoid the circular import, not a blocker for now though. In general, we have the implicit rule to only allow lazy imports in the d3a-web integration tests (e.g. in order to avoid imports from models.py before the test suite started Django), we are not too strict on this rule though therefore it is okay.

spyrostz (approved):
LGTM

fievelk (approved):

PR #1147, opened 2021-05-12
Log error message on launcher failure

fievelk:
The `launcher.py` script is currently failing without showing errors on exit (for example when redis is not available). 
This PR adds logging to notify about the exception message.


Hassan754 (approved):
LGTM

src/d3a/d3a_core/d3a_jobs.py
```diff
@@ -41,7 +45,8 @@ def main():
         )
         try:
             worker.work(max_jobs=1, burst=True)
-        except Exception:
+        except Exception as ex:
+            logger.error(ex)
```
hannesdiedrich, line 49: I fear that this will spam our logs on dev/staging/production. 
Is this a nice to have or did you really experience a case where you would like to see this log?
AFAIK, all exceptions are shown when the launcher is called.

hannesdiedrich: Also, here you only log the kind of exception. Is this really helpful? 
A traceback would be better, right? I am not suggesting to add it, because again, i think all exceptions should be logged.

fievelk: > Is this a nice to have or did you really experience a case where you would like to see this log?
>AFAIK, all exceptions are shown when the launcher is called.

I've experienced this a few days ago. If you try to use `launcher.py` when Redis is not running, it will fail without any useful exception message:
```bash
$ python src/d3a/d3a_core/launcher.py
[1]    68047 killed     python src/d3a/d3a_core/launcher.py
```
It took me a while to figure out that it was related to redis (I had shut it down to run integration tests).

> Also, here you only log the kind of exception. Is this really helpful?
It also shows the exception message, so it would still be helpful for some cases (like the redis one). Logging the full traceback would be more expressive and useful for other kinds of bugs though. I can definitely add it if you wish.

> I fear that this will spam our logs on dev/staging/production.

Why is that? Do we have many failures on `launcher.py`?

hannesdiedrich: > Why is that? Do we have many failures on launcher.py? 
No probably not. I just remember vaguely that we did not log this error here for a reason. But I might be wrong. @spyrostz can maybe shine some light in here.

In general, you convinced me and will give a LGTM

spyrostz: Under normal circumstances this should not spam the logs, and it will be logged only in case of an error in the worker. However I am not sure if this solves @fievelk 's issue, because in case of a Redis connection error, the context manager `with Connection(...)` would have failed immediately and not log any message (because the try/except covers only the work() method call)

fievelk: The exception I am facing is raised by `worker.work()`, so after the `with Connection()` context has begun:
```shell
  File "src/d3a/d3a_core/d3a_jobs.py", line 47, in main
    worker.work(max_jobs=1, burst=True)
  File ".pyenv/versions/d3a/lib/python3.8/site-packages/rq/worker.py", line 535, in work
    self.register_birth()
  File ".pyenv/versions/d3a/lib/python3.8/site-packages/rq/worker.py", line 278, in register_birth
    if self.connection.exists(self.key) and \
  File ".pyenv/versions/d3a/lib/python3.8/site-packages/redis/client.py", line 1581, in exists
    return self.execute_command('EXISTS', *names)
  File ".pyenv/versions/d3a/lib/python3.8/site-packages/redis/client.py", line 898, in execute_command
    conn = self.connection or pool.get_connection(command_name, **options)
  File ".pyenv/versions/d3a/lib/python3.8/site-packages/redis/connection.py", line 1192, in get_connection
    connection.connect()
  File ".pyenv/versions/d3a/lib/python3.8/site-packages/redis/connection.py", line 563, in connect
    raise ConnectionError(self._error_message(e))
redis.exceptions.ConnectionError: Error 61 connecting to localhost:6379. Connection refused.
```

I compared the behavior from `master` and from the current branch:
```shell
(d3a) ➜  d3a git:(master) redis-cli shutdown
(d3a) ➜  d3a git:(master) python src/d3a/d3a_core/launcher.py
[1]    3178 killed     python src/d3a/d3a_core/launcher.py

(d3a) ➜  d3a git:(master) git checkout show-launcher-error
Switched to branch 'show-launcher-error'
(d3a) ➜  d3a git:(show-launcher-error) python src/d3a/d3a_core/launcher.py
Error 61 connecting to localhost:6379. Connection refused.
[1]    3365 killed     python src/d3a/d3a_core/launcher.py
```
As you can see, from this branch the exception message seems to be correctly shown:
> Error 61 connecting to localhost:6379. Connection refused.

spyrostz: I see, so probably the connection is established lazily (essentially when the worker needs it, and not when the actual object is created), then all good from my side!

spyrostz (approved):
LGTM

PR #1149, opened 2021-05-19
Urllib version sync

Hassan754:
Sync urllib version with that of d3a-api-client/ d3a-interface

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -14,4 +14,4 @@ redis
 rq
 pytz
 sortedcontainers
-urllib3==1.26.3
+urllib3==1.26.4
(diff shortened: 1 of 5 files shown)


hannesdiedrich:
Do you mind upgrading pyaml and py to (as dependabot suggested)? Thanks!

Hassan754:
> Do you mind upgrading pyaml and py to (as dependabot suggested)? Thanks!

I upgraded pyaml but left py as it will require extra PR in the d3a-interface and this PR is a bit urgent

Hassan754:
> In d3a we use `pip-tools` to manage dependencies, so this change would be undone next time we run `pip-compile` (or `fab compile`). I would suggest either upgrading both `d3a-interface` and `d3a` requirements (if they end up to the same version) or pinning these requirements in `base.in` (not `base.txt`).

It looks the version was already pinned, I updated it

fievelk (approved):

faizan2590 (approved):
LGTM

PR #1150, opened 2021-05-20
 Sync urllib3 version

Hassan754:


hannesdiedrich:
I think we need a workflow to adapt the requirements and their versions. I just had trouble in the other ticket too.
Can you please add a ticket that targets homogenising the dependencies between all repos and also to create a little wiki page that describes how to do that also in the future? Thanks!

hannesdiedrich (approved):
For a quick fix it is ok, that you adapted the version numers manually, however we should use pip-compile. LGTM

PR #1151, opened 2021-05-20
Fix market cycle event for loads

fievelk:
Reset price rates to be equal to the initial rate just once, before looping over each market.


src/d3a/models/strategy/load_hours.py
```diff
@@ -323,7 +324,7 @@ def post_or_update_bid(self):
                     if not self.are_bids_posted(market.id):
                         self.post_first_bid(market, bid_energy)
                     else:
-                        self.bid_update.reset(self)
+                        self.bid_update.update(market, self)
```
hannesdiedrich, line 327: I thin you can get rid of 
```
else:                        
     self.bid_update.reset(self)
```

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #1152, opened 2021-05-21
D3ASIM-3421: Added DB connection in order to read user profiles. Also…

hannesdiedrich:
… added rotation of all profiles.

Until now energy and price profiles were populated and saved in memory, even for yearly simulations. With this a lot of memory is used and could lead to OOM errors. Additionally, the initiation time can be long if one considers a lot of devices.
Consequently, it was decided to store the user profiles (e.g. PV power profiles) into a postgres DB that is solely designed to hold profiles. This feature is only used when the job is send from the d3a-web. Local runs of the d3a simulations will still load weekly chunks of the profiles into the memory. 

What was done:
1.  created connection to a postres profiles DB using pony ORM
2. added handling (rotation) of DB profiles
3. added rotation of all other used profiles (e.g. in update_frequency.py)
4. added integrationtest that creates a temporary DB in a docker container


src/d3a/models/area/__init__.py
```diff
@@ -356,6 +356,10 @@ def cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fal
         else:
             changed_balancing_market = None
 
+        if d3a.constants.CONNECT_TO_PROFILES_DB and not self.parent and self.next_market:
+            global_objects.profile_db_connection.buffer_profiles_from_db(
```
spyrostz, line 360: global_objects is also accessible from the simulation.py IIRC, therefore you can move this there and 1. not have to check the if block for every area, and remove the not self.parent clause 2. move the call of the method somewhere where it is more appropriate (because the Area class conceptually does not have any relation to profile management).
In addition to that, it would be much better to expose a method in the global_objects, called update_on_market_cycle, that would encapsulate the d3a.constants.CONNECT_TO_PROFILES_DB check. That way the check will be in only one place.

src/d3a/models/strategy/__init__.py
```diff
@@ -230,6 +234,45 @@ def __init__(self):
 
     parameters = None
 
+    def rotate_profile(self, profile_type: InputProfileTypes, profile,
```
spyrostz, line 237: I think that both methods can be moved to the global_objects.profile_db_connection. That way profile handling will be centralized to one class and the strategy code will remain clean from it.

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -15,29 +15,46 @@
...
+                 energy_rate_profile_uuid=None):
         super().__init__()
         self.energy_per_slot_kWh = INF_ENERGY
-        self.energy_buy_rate = energy_buy_rate
-        self.buying_rate_profile = buying_rate_profile
-        self.energy_rate = energy_sell_rate
-        self.energy_rate_profile = energy_rate_profile
+        self.energy_buy_rate = None
+
+        if d3a.constants.CONNECT_TO_PROFILES_DB and buying_rate_profile_uuid:
```
spyrostz, line 40: In order to not have if checks for this constant throughout the code, I think that a function called `should_read_profile_from_db(profile_uuid)` would be much better. It also has the advantage that it can check the cornercase where the profile_uuid does not exist in the DB.

src/d3a/models/strategy/market_maker_strategy.py
```diff
@@ -15,36 +15,63 @@
...
+
         self._grid_connected = grid_connected
-        super().__init__(energy_rate)
+
+        self._read_or_rotate_rate_profiles()
+
+    def event_activate_price(self):
+        pass
+
+    def _read_or_rotate_rate_profiles(self):
```
spyrostz, line 54: The more I see it in the code, the more I think that an empty implementation of this method should live in the BaseStrategy class in order to denote that it is needed for the implementation of any strategy.

src/d3a/models/strategy/update_frequency.py
```diff
@@ -30,24 +30,48 @@ def __init__(self, initial_rate, final_rate, fit_to_limit=True,
...
+        self.energy_rate_change_per_update_profile_buffer = {}
+        self._read_or_rotate_rate_profiles(GlobalConfig.start_date)
 
         self.energy_rate_change_per_update = {}
         self.update_interval = update_interval
         self.update_counter = {}
         self.number_of_available_updates = 0
         self.rate_limit_object = rate_limit_object
 
+    def _read_or_rotate_rate_profiles(self, current_timestamp: DateTime,
```
spyrostz, line 50: A general observation by looking at the signature of this method. I think that all methods that read or rotate profiles should accept the timestamp as input and not depend on the market timeslot. That way the method will be more decoupled from its class and will be possible to abstract away more easily in a dependent class, if this needs to become more complex in the future (even now, I cannot help but have the impression that the if/else blocks that check whether a profile or a value should be used could be simplified by having a separate class instead of a method, however I am not sure it is worth it)

spyrostz (commented):
Left some comments, will rereview once dealt with

src/d3a/models/strategy/__init__.py
hannesdiedrich: Here I am not convinced. This method is also dealing with rotating the profiles that are not in the DB. Only for the profiles stored in the DB, `global_objects.profile_db_connection.get_profile_from_db_buffer(profile_uuid)` is called. Happy to move it somewhere else, thought. But where?

src/d3a/models/strategy/update_frequency.py
hannesdiedrich: `(even now, I cannot help but have the impression that the if/else blocks that check whether a profile or a value should be used could be simplified by having a separate class instead of a method, however I am not sure it is worth it)` Sorry, I dont understand what you mean. 

For the latter: Yes, this is why i left the TODO comment: `# TODO: this needs to be implemented to except profile UUIDs and DB connection`
I left it out for now in order to wrap this story up eventually. The API of the strategies and the UpdateFrequencyMixin will have to change in order to also except the price profile uuids. 
Please let me know what you think: Shall I add it here or in a separate story.?

fievelk: Just hopping in regarding the `UpdateFrequencyMixin`: I would vote for adding it in a separate story since this one is already quite big :)

spyrostz: Regarding UpdateFrequencyMixin I also agree that this should be added in a new story, however keep in mind that with my d3a-web change I do not differentiate in price and energy profiles, they are all stored in the DB. Therefore if someone  uploads a price profile, it will be in the DB, and the UpdateFrequencyMixin won't be able to read it, therefore before I deploy the d3a-web changes, the to-be-created story needs to be implemented in order to not affect the users.

Regarding this `(even now, I cannot help but have the impression that the if/else blocks that check whether a profile or a value should be used could be simplified by having a separate class instead of a method, however I am not sure it is worth it)` the reason why I am stating this is because I see scattered throughout the strategy classes this check `if d3a.constants.CONNECT_TO_PROFILES_DB`. This makes me think that instead of if/else blocks, the decision for the strategy from where to retrieve the profile could be managed by a separate class, which would contain all the members of the strategy classes that can be user profiles. That way this if block (and in general profile management and decisions from where will the profile be read) would live only in the new class and not in the strategy class. I know it is probably overengineering, however the existence of all these if/else blocks is an indication that an extra abstraction layer would make the code easier to read.

hannesdiedrich: `Therefore if someone uploads a price profile, it will be in the DB` - ATM we dont have any user interface for that. I know we were discussing this in the many discussions before this PR, but really, we only support uploading the power profiled of the Load and PV and the buying and selling rate profiles of the InfiniteBusStrategy now. In order to support uploading price profiles for the template strategies, also the d3a-web will need to be changed. Or am I missing something that  you will add on the d3a-web part?

Thanks for your explanation of the latter. It makes sense and i will give it a go and try it. If it is too much hassle, I'll stop and keep the if blocks (or the mentioned function for this). OK?

src/d3a/models/strategy/__init__.py
spyrostz: I think the best approach is to create another class that wraps up this functionality. Therefore, instead of having only the ProfileDBConnectionHandler class, you could create another wrapper class, that also lives in the global_objects:
`class ProfileHandler:
    def __init__(self): 
        self.db = ProfileDBConnectionHandler()
`
That way you will have a single interface for profile management, that internally uses either the db connection or the profiles that are not in the DB.

src/d3a/models/strategy/update_frequency.py
spyrostz: My bad, I was under the impression that we also supported price profiles for the Load / PV strategies via the UI, but I was obviously mistaken, please ignore my comment.

src/d3a/models/area/__init__.py
hannesdiedrich: @spyrostz I remember now why this has to be exactly here (for this one I actually spent the most time): We need to rotate the profiles after the next_market is created, but before the market_cycle event is dispatched to all subareas. This is because of a hen-egg problem: If I dont check if the profile contains the `self.next_market.time_slot` before actually accessing it, an error will occur. But I can not rotate the file without knowing the `self.next_market.time_slot`. 
(I wanted to drop a in-code comment to show this, but forgot)
Do you have an idea how to fix this?
Just FYI: the profiles are only read once, because there is a check for "did i already read new profiles" in `buffer_profiles_from_db`. So even if you dont have an idea, it is ok to stay here. We only have to bite the bullet of the two unneeded if statements for each area. 
Looking forward to your comment.

hannesdiedrich: (ah, I remember now why I added the check for self.parent: In order to only run it for the root are and consequently only once per market slot)

spyrostz: Understandable, but you really do not need the market in order to calculate the timeslot. You can easily circumvent that by creating a small method in the area class that exposes the now_value (badly named, but in reality it is the current datetime of the market that will be created, irregardless if we have a simulation or CN). Therefore all you have to do is to create a method that exposes the now_value (and name it appropriately, eg current_slot_datetime or current_timeslot) and once you have this method you can use it from the Simulation class, since the Simulation class has access to the root area (which is all you need since this is the only area without a parent)

hannesdiedrich: Ahh, thanks a lot! will do that.

src/d3a/d3a_core/user_profile_handler.py
```diff
@@ -0,0 +1,253 @@
...
+from pony.orm.core import Query
+
+
+class ProfileDBConnectionHandler:
+    """
+    Handles connection and interaction with the user-profiles postgres DB via pony ORM
+    """
+    _db = Database()
+
+    class ProfileTimeSeries(_db.Entity):
```
Hassan754, line 41: Why don't we set the profile time series as JSON field in the ConfigurationAreaProfileUUIDs instead of creating 2 models and instances for all rows?

Hassan754 (commented):
Left some questions, will review again once answered

src/d3a/constants.py
```diff
@@ -32,7 +32,7 @@
 # 20% of the slot length
 DISPATCH_EVENT_TICK_FREQUENCY_PERCENT = 10
 
-COLLABORATION_ID = ""
+CONFIGURATION_ID = ""
```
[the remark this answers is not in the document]
hannesdiedrich: Before we only needed the config uuids from collaborations for the communication with the client API.
Now we need uuid even for normal simulation configs in order to retrieve the corresponding profiles from the profiles DB. That is why I renamed it, too.

src/d3a/d3a_core/rq_job_handler.py
```diff
@@ -25,8 +24,8 @@ def launch_simulation_from_rq_job(scenario, settings, events, aggregator_device_
                                   saved_state, job_id):
     logging.getLogger().setLevel(logging.ERROR)
     scenario = decompress_and_decode_queued_strings(scenario)
+    d3a.constants.CONFIGURATION_ID = scenario.pop("configuration_uuid")
     if "collaboration_uuid" in scenario:
```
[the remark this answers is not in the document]
hannesdiedrich: You are right, this is not thought through. We still need a flag for the type of simulation now.
@spyrostz Let us discuss this. I would vote for exposing the configuration.type from the d3a-web also to the d3a in order to make the right decisions here. 
However, there should only be one member that holds the uuids of the configuration ("configuration_uuid")

src/d3a/d3a_core/user_profile_handler.py
```diff
@@ -0,0 +1,253 @@
...
+        profile_uuid = Required(str)
```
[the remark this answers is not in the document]
hannesdiedrich: Yes, but this is not needed here. We only need to connect to the DB for reading. We dont actually need to create the tables (only for the integrationtests, but there we dont need a default value either).

src/d3a/d3a_core/user_profile_handler.py
```diff
@@ -0,0 +1,253 @@
...
+        value = Required(float)
+
+    class ConfigurationAreaProfileUuids(_db.Entity):
+        configuration_uuid = Required(str)
+        area_uuid = Required(str)
+        profile_uuid = Required(str)
+        profile_type = Required(int)  # values of InputProfileTypes
+
+    def __init__(self):
+        self._profile_uuid_type_mapping = {}
```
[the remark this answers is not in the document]
hannesdiedrich: You are right, ATM, we dont need the `_profile_uuid_type_mapping` especially after I added another buffer for the profile_uuids now. I thought at the beginning that the type of profile could be reused when the `rotate_profile ` is called, but really the caller of `rotate_profile` knows about the profile type already. So, I will remove the buffer here. 
Just FYI: saving the type of profile in the DB still makes sense for data-analysists.

src/d3a/d3a_core/rq_job_handler.py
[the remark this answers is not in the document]
spyrostz: I agree, there should be a single parameter called configuration_uuid regardless of the type of the config/collab/CN, and the type of the configuration as a separate field. That way we will be able to get rid of the IS_CANARY_NETWORK constant by only checking the type of the config .

src/d3a/models/strategy/update_frequency.py
```diff
@@ -159,20 +180,28 @@ def time_for_price_update(self, strategy, time_slot):
...
-                       energy_rate_change_per_update_profile_buffer=None, fit_to_limit=None,
-                       update_interval=None, ):
-        if initial_rate_profile_buffer is not None:
-            self.initial_rate_profile_buffer = initial_rate_profile_buffer
-        if final_rate_profile_buffer is not None:
-            self.final_rate_profile_buffer = final_rate_profile_buffer
-        if energy_rate_change_per_update_profile_buffer is not None:
-            self.energy_rate_change_per_update_profile_buffer = \
-                energy_rate_change_per_update_profile_buffer
+    def set_parameters(self, area, initial_rate=None, final_rate=None,
```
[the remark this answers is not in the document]
hannesdiedrich: Sorry, I had to google a bit to understand. Do you mean this solution:
https://stackoverflow.com/a/14302007/6835791 ?
Is it to prohibit implementations of the call of `set_parameters` like:
```
set_parameters(area, 1,2,3,4)
```
?

src/d3a/d3a_core/user_profile_handler.py
hannesdiedrich: We need to be able to filter for the timestamp. And this needs to be done fast. If you have to deserialise a JSON string first and then filter for the time, you first load the entire profile into memory (which is the reason for this story to be implemented: reducing the used memory) and second it needs too much time. 
We actually tried out some DB types that are dedicated (e.g. influxDB) but Spyros found out that using a simple postgres DB an storing the profiles in a timestamp-value manner is much more performant. 
This needs to be evaluated again once we have thousands of profiles.

src/d3a/models/strategy/update_frequency.py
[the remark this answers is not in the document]
hannesdiedrich: I realised that area is not needed here any more. I added the asterik too.

src/d3a/d3a_core/rq_job_handler.py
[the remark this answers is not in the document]
hannesdiedrich: Ok @spyrostz. Let's have a short off-github discussion what you will send and how I receive it on the d3a side.

integration_tests/environment.py
```diff
@@ -39,11 +40,36 @@
 """
 
 
+if d3a.constants.CONNECT_TO_PROFILES_DB is True:
```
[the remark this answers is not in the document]
Hassan754: reminder

Hassan754 (approved):
LGTM, thanks for the global config 🥇

integration_tests/environment.py
```diff
@@ -16,14 +16,15 @@
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import contextlib
-import shutil
 import os
+import shutil
+from time import sleep
 
 import d3a.constants
+from d3a import constants
```
faizan2590, line 24: Any specific reason to have both style of imports?

integration_tests/steps/user_profile.py
```diff
@@ -0,0 +1,122 @@
...
+def init_uuids(context):
+    context.config_uuid = str(uuid.uuid4())
+    context.pv_area_uuid = str(uuid.uuid4())
+    context.pv_profile_uuid = str(uuid.uuid4())
+    context.load_area_uuid = str(uuid.uuid4())
+    context.load_profile_uuid = str(uuid.uuid4())
+    GlobalConfig.slot_length = duration(minutes=60)
+
+
+@given("a yearly PV profile exist in the DB")
```
faizan2590, line 22:
```suggestion
@given("an yearly PV profile exist in the DB")
```
nitpicky

src/d3a/d3a_core/user_profile_handler.py
```diff
@@ -0,0 +1,247 @@
...
+        value = Required(float)
+
+    class ConfigurationAreaProfileUuids(_db.Entity):
+        configuration_uuid = Required(str)
+        area_uuid = Required(str)
+        profile_uuid = Required(str)
+        profile_type = Required(int)  # values of InputProfileTypes
+
+    def __init__(self):
+        # self._profile_uuid_type_mapping = {}
```
faizan2590, line 53: Assuming it to be removed before merging.

integration_tests/steps/user_profile.py
fievelk: I think the grammatically correct version is the original one:
> "a yearly ..."

That's because "Y" has a consonant sound in this word (see [Which one is grammatically correct: "an year" or "a year"?](https://www.quora.com/Which-one-is-grammatically-correct-an-year-or-a-year#:~:text=Originally%20Answered%3A%20What%20is%20grammatically,%22a%20year%22%20is%20correct.))

src/d3a/d3a_core/util.py
```diff
@@ -532,3 +530,7 @@ def _dispatch_tick_frequency(self) -> int:
 
     def is_it_time_for_external_tick(self, current_tick_in_slot) -> bool:
         return current_tick_in_slot % self._dispatch_tick_frequency == 0
+
+
+def should_read_profile_from_db(profile_uuid):
```
faizan2590, line 535: How about moving it to `src/d3a/d3a_core/user_profile_handler.py` module?

src/d3a/models/area/__init__.py
```diff
@@ -471,6 +470,10 @@ def now(self) -> DateTime:
             seconds=self.config.tick_length.seconds * self.current_tick
         )
 
+    # @property
+    # def next_market_slot(self):
+    #     return self.now.add(seconds=self.config.slot_length.seconds)
+
```
faizan2590, line 476: Assuming that to be removed safely before merging this PR

integration_tests/steps/user_profile.py
faizan2590: Wow, learned some english grammar today, was so used to of `an yearly`

integration_tests/environment.py
hannesdiedrich: Not really, will change it.

src/d3a/d3a_core/util.py
hannesdiedrich: It's really a matter of taste as it is used in a lot of places (mainly in the strategies).
So I would wait for another one voting for it.

src/d3a/models/strategy/home_meter.py
```diff
@@ -532,7 +538,7 @@ def _one_sided_market_event_tick(self, market, offer=None):
         """
         if not self.state.can_buy_more_energy(market.time_slot):
             return
-        if not offer and not market.offers:
+        if not offer or not market.offers:
```
hannesdiedrich, line 541: @fievelk IMO this was a bug. It revealed itself in the `Strategies follow yearly user profile saved in the DB` integration test. Do you agree on this change or was the `and` added on purpose?

src/d3a/models/strategy/home_meter.py
```diff
@@ -258,14 +271,13 @@ def area_reconfigure_event(self, *args, **kwargs):
...
+        self.bid_update.update_and_populate_price_settings(self.area)
 
         # Update the raw profile. It will be read later while setting the energy forecast.
         if kwargs.get("home_meter_profile") is not None:
             self.home_meter_profile = kwargs["home_meter_profile"]
-
-        self.offer_update.update_and_populate_price_settings(self.area)
-        self.bid_update.update_and_populate_price_settings(self.area)
-        self._set_energy_forecast_for_future_markets(reconfigure=True)
+            self._set_energy_forecast_for_future_markets(reconfigure=True)
```
fievelk, line 280: Just a doubt: what happens if the `home_meter_profile` is `None` here? Can this happen?

src/d3a/models/strategy/home_meter.py
fievelk: I'm not actually confident about this, since these `if`s are quite convoluted. This line was implemented after this comment: https://github.com/gridsingularity/d3a/pull/1159#discussion_r655984021
The original version was taken from the load implementation: https://github.com/gridsingularity/d3a/blob/153648206fcca5ccc86b3aa47e265ccaee01b1db/src/d3a/models/strategy/load_hours.py#L266-L274

We don't want to exit here if `offer` is `None`, because in that case we would never execute this line:
```python
acceptable_offer = offer if offer else self._find_acceptable_offer(market)
```
So it looks like we only want to exit if we have neither `offer` nor `market.offers`, because in that case we couldn't find an acceptable offer anywhere.
If we have at least one of them (either `offer` or `market.offers`) we need to go on and check if there's an `acceptable_offer` in the markets.

tests/test_home_meter.py
```diff
@@ -123,13 +123,13 @@ def test_event_activate_price_without_market_maker_rate(self):
         self.strategy.bid_update.set_parameters.assert_not_called()
         self.strategy.offer_update.set_parameters.assert_not_called()
 
-    @patch("d3a.models.strategy.home_meter.read_arbitrary_profile")
-    def test_set_energy_forecast_for_future_markets(self, read_arbitrary_profile_mock):
+    @patch("d3a.d3a_core.global_objects_singleton.global_objects.profiles_handler.rotate_profile")
```
fievelk, line 126: Are you sure that this shouldn't be the following?
```python
@patch("d3a.models.strategy.home_meter.global_objects.profiles_handler.rotate_profile")
```
I would double-check that the actual patched object is being used. You can read here for more details: [Unittest - Where to patch](https://docs.python.org/3.8/library/unittest.mock.html#id6)

If you find out that my suggestion is correct, it means that tests are probably not doing their job (because they should have failed).

fievelk: (The same applies to the rest of the commit)

src/d3a/models/strategy/home_meter.py
hannesdiedrich: yes, indeed, this can happen. But is caught by the if clause. We only have to run `self._set_energy_forecast_for_future_markets` when it is not None.

You gave me a hint for something that is missing still:
What if the user wants to change the profile (via the profile_uuid). 
I need to adapt for this case (not only here but also in the predefined load and pv. Thanks!

tests/test_home_meter.py
hannesdiedrich: You are totally right! Will change that.

src/d3a/models/strategy/home_meter.py
hannesdiedrich: I think i understand. Let me check what went wrong there.

hannesdiedrich: Actually, I found that the if statements are not translated correctly following the comment from Spyros.
This should be correct now:
https://github.com/gridsingularity/d3a/pull/1152/commits/316229f891477d722acf2512785273c5c547bd4c#diff-1187ba841fc3bdd8c4b0200dc3033dfc48bd59ba5c22485e605fd0d1293e1d78R541

In order to also catch the case when the offer is indeed None.

src/d3a/models/strategy/home_meter.py
hannesdiedrich: Did that here: https://github.com/gridsingularity/d3a/pull/1152/commits/845adce0cda917b1d0e1712cd6e39c4256d17e4d

PR #1153, opened 2021-05-26
Fix Load rate updates in two-sided markets

fievelk:
Load strategies were updating their bid rates in two-sided markets only when they could have bought more energy. Rates should be updated in every case instead (even when not buying energy).

I believe that the intended behaviour was to apply the check only in one-sided markets (post a bid only if possible) but the `if` statement got in the way of two-sided markets as well.


hannesdiedrich (commented):
Thanks! LGTM
I hate to be the bad cop here, but this needs at least a unittest for the event_tick in the load in the double_sided_market case. Thanks for adding it!

tests/test_strategy_load_hours.py
```diff
@@ -294,6 +294,24 @@ def test_active_markets(load_hours_strategy_test1):
...
+
+    # Test for all available market types (one-sided and two-sided markets)
+    available_market_types = [1, 2, 3]
+    for market_type_id in available_market_types:
+        ConstSettings.IAASettings.MARKET_TYPE = market_type_id
+        load_hours_strategy_test1.event_tick()
+
+    # Rates should be updated once for each event_tick call
+    assert (
+        load_hours_strategy_test1.bid_update.increment_update_counter_all_markets.call_count
```
hannesdiedrich, line 311: Sorry, but this check would have passed even before your change AFAIS.
Please check if the `_double_sided_market_event_tick` or the `_one_sided_market_event_tick` is called for all markets (depending on the market type). Thanks!

fievelk: If you want, I can also check if this works when `can_buy_more_energy` is `True`. It just requires one more loop.

fievelk: Doing it now

tests/test_strategy_load_hours.py
```diff
@@ -294,6 +294,30 @@ def test_active_markets(load_hours_strategy_test1):
...
+    available_market_types = (1, 2, 3)
+    # Bids' rates should be updated both when the load can buy energy and when it cannot do it
+    for can_buy_energy in (True, False):
+        load_hours_strategy_test1.state.can_buy_more_energy.return_value = can_buy_energy
+        for market_type_id in available_market_types:
+            ConstSettings.IAASettings.MARKET_TYPE = market_type_id
+            load_hours_strategy_test1.bid_update.update.reset_mock()
+            load_hours_strategy_test1.event_tick()
+            if market_type_id == 1:
+                # Bids' rates are not updated in one-sided markets
```
hannesdiedrich, line 314: How about increasing the coverage even more and add
```
load_hours_strategy_test1._one_sided_market_event_tick = Mock()
....
load_hours_strategy_test1._one_sided_market_event_tick.call_count == number_of_markets
```

hannesdiedrich (approved):
Once optional enhancement comment. LGTM

tests/test_strategy_load_hours.py
hannesdiedrich: OK.

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #1156, opened 2021-05-27
Refactor state and strategies

fievelk:
- Optimize imports
- Add some documentation (docstrings, comments, type hints)
- Refactor some methods (e.g. method names and convoluted `if` statements)


src/d3a/models/strategy/predefined_load.py
```diff
@@ -125,6 +125,11 @@ def _allowed_operating_hours(self, time):
         return True
 
     def area_reconfigure_event(self, **kwargs):
+        """Reconfigure the device properties at runtime using the provided arguments.
+
+        This method is triggered when the device strategy is updated while the simulation is
+        running. The update can happen via live events (triggered by the user) or scheduled events.
+        """
```
hannesdiedrich, line 132: Every strategy kind as well as the area has this function. I am not sure how to handle it. Either you add this docstring to all of them or just leave it. What do you think?

src/d3a/models/strategy/pv.py
```diff
@@ -68,22 +67,21 @@ def __init__(self, panel_count: int = 1,
         self.panel_count = panel_count
         self.max_panel_power_W = max_panel_power_W
         self.state = PVState()
+        self.use_market_maker_rate = use_market_maker_rate
+        self.final_selling_rate = final_selling_rate
```
hannesdiedrich, line 71: The external pv strategy is inheriting from `PVStrategy` where we dont need these variables. This is why they are initiated in a separate function. Not really important, but IMO, please revert.

src/d3a/models/strategy/load_hours.py
```diff
@@ -275,6 +287,9 @@ def _one_sided_market_event_tick(self, market, offer=None):
             self.log.exception("An Error occurred while buying an offer")
 
     def _double_sided_market_event_tick(self, market):
+        """
+        Define the behavior of the device on TICK events in double-sided markets (post bids).
+        """
```
hannesdiedrich, line 292: This is a perfect example for a redundant docstring IMO. Leave it please, but really, the function name is already telling what is done.
Also, the `(post bids)` addition is not really true, right? It updates the bids. Right?

src/d3a/models/strategy/load_hours.py
```diff
@@ -291,17 +306,28 @@ def event_tick(self):
         self.bid_update.increment_update_counter_all_markets(self)
 
     def event_offer(self, *, market_id, offer):
+        """Automatically react to offers in single-sided markets.
+
+        This method is triggered by the OFFER event.
+        """
```
hannesdiedrich, line 312: Here too. You would need to add a docstring vor all `event_offer` methods in the codebase to be consistent.
But there are a lot. So you can leave this here. But the coder has to know the architecture of the d3a anyway where he/she would know what this function does and how it gets triggered. 
Dont want to start a discussion and dont expect a pitch for docstrings from you here, just want to mention . :-)

src/d3a/models/strategy/pv.py
fievelk: That's interesting...I noticed this inconsistency because pylint was complaining about it. I would normally expect a class to declare all instance variables in `__init__()`, also because the `__init__()` usually tells which instance variables we can expect and use.

This might be a sign that the current inheritance structure could be somehow improved (maybe with a new class). Anyway, I can revert this change.

src/d3a/models/strategy/load_hours.py
fievelk: I think you're right, also from the moment that the function is private so it doesn't need to necessarily have a docstring.

src/d3a/models/strategy/load_hours.py
fievelk: In this case I don't completely agree with you, even though I understand your point. 
This method slowed me down a bit, as I could not immediately understand what it was doing, so I was following method calls, constant values and wiki articles. Some method names have the `event_` prefix even though they don't have an event trigger, which is sometimes confusing. 
I think that the coder often gains knowledge about the architecture by the code, and not just the opposite (especially for public repos). That's why I added this docstring in the end, hoping that when we will onboard new people it will help a bit :)

hannesdiedrich: Ok, convinced. But to be consistent, you would need to add this docstring to all `event_offer`, `event_activate`, `event_bid_traded`, `event_market_cycle` etc. methods, right?
Otherwise the uneducated user would have to find this docstring by accident. Or how would you propose to handle it?

src/d3a/models/strategy/predefined_load.py
fievelk: I found something that can help us here in [PEP 257](https://www.python.org/dev/peps/pep-0257/):
>The docstring for a class should summarize its behavior and list the public methods and instance variables. If the class is intended to be subclassed, and has an additional interface for subclasses, this interface should be listed separately (in the docstring). The class constructor should be documented in the docstring for its __init__ method. Individual methods should be documented by their own docstring.
>
> If a class subclasses another class and its behavior is mostly inherited from that class, its docstring should mention this and summarize the differences. Use the verb "override" to indicate that a subclass method replaces a superclass method and does not call the superclass method; use the verb "extend" to indicate that a subclass method calls the superclass method (in addition to its own behavior).

So I can move the full docstring into the `BaseStrategy` definition of the method, and just keep the first line in the child. What do you think? Leaving it without docstring would raise a pylint warning.

hannesdiedrich: OK! Agreed.

src/d3a/models/strategy/load_hours.py
fievelk: Ideally yes, I would add it to all those methods. In practice, this current PR is growing together with the implementation of the Home Meter device so I'm refactoring things as I go along.
We could either keep this refactoring "incomplete", or I can complete it now for all those methods. I would personally go for the second option, since this is a specific PR aimed at refactoring. If the refactoring was made inside a feature PR, maybe this would be less relevant.

fievelk: For the future, I'm thinking that even these refactoring PRs should be smaller so that reviewing them could be almost matter of seconds.

src/d3a/models/strategy/__init__.py
```diff
@@ -258,7 +258,11 @@ def event_response_was_received_callback():
             self.event_response_uuids.remove(data["transaction_uuid"])
 
     def area_reconfigure_event(self, *args, **kwargs):
-        pass
```
hannesdiedrich, line 261: dont you need the pass anyway?

hannesdiedrich (dismissed):
One last comment, LGTM

src/d3a/models/strategy/__init__.py
fievelk: Apparently, you don't need it if you add a docstring. If found that out because pylint complains if you have both a docstring and a `pass` statement in an "empty" method. The difference is that a docstring populates the `__doc__` special attribute of the method (useful to generate documentation) so it's not totally equivalent to the `pass` statement.

hannesdiedrich: Interesting!

src/d3a/models/state.py
```diff
@@ -55,10 +55,7 @@ def get_energy_production_forecast_kWh(self, time_slot, default_value=None):
         return self._energy_production_forecast_kWh[time_slot]
 
     def get_available_energy_kWh(self, time_slot, default_value=None):
-        if default_value is not None:
-            available_energy = self._available_energy_kWh.get(time_slot, default_value)
-        else:
-            available_energy = self._available_energy_kWh.get(time_slot)
+        available_energy = self._available_energy_kWh.get(time_slot, default_value)
```
Hassan754, line 58: This can raise an error in the line below if it fell down to the default None value

Hassan754 (approved):
LGTM, thanks btw :)

src/d3a/models/state.py
fievelk: That's true, even though it doesn't depend on this specific PR if I'm not mistaken

fievelk: Fixed!

Hassan754 (approved):

hannesdiedrich (approved):
LGTM

PR #1159, opened 2021-06-01
D3ASIM-3353: implement Home Meter device

fievelk:
Implement the new Home Meter device.

## Requirements

- The user can upload a meter profile (positive = consumption, negative = production)
- The user can set the following attributes:
  - a location for the asset
  - initial selling rate, final selling rate, and rate decrease (production)
  - initial buying rate, final buying rate, and rate increase (consumption)
  - update interval
- The default initial selling rate is the market maker rate (includes grid fees)
- The default final buying rate is the market maker rate (includes grid fees)


src/d3a/models/state.py
```diff
@@ -43,36 +47,154 @@
...
+    @abstractmethod
+    def restore_state(self, state_dict: Dict):
+        pass
+
+    @abstractmethod
+    def delete_past_state_values(self, time_slot):
+        pass
+
+
+class ConsumptionState(StateInterface):
```
hannesdiedrich, line 67: `ConsumptionState` and `ProductionState` both are not implementing `delete_past_state_values`, but they should, right?

hannesdiedrich (commented):
I had a brief look and LGTM as a draft.
I like the refactoring of the states!
I think all you need to add are intergationtests and this PR is good to go. Thanks!

src/d3a/models/state.py
fievelk: You're right. Since the implementation of `delete_past_state_values` is different in every device State class, I didn't add it to `Consumption/Production` states. But I could do it since it would make more sense to have it there, rather than on Load/PV states.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -160,6 +160,13 @@ def _populate_core_stats_and_sim_state(self, area):
             core_stats_dict['grid_fee_constant'] = area.current_market.const_fee_rate \
                 if area.current_market is not None else 0.
 
+        if isinstance(area.strategy, HomeMeterStrategy):
+            core_stats_dict["home_meter_profile_kWh"] = (
+                area.strategy.state.get_energy_at_market_slot(self.current_market_time_slot))
+            if area.parent.current_market is not None:
+                for trade in area.strategy.trades[area.parent.current_market]:
+                    core_stats_dict["trades"].append(trade.serializable_dict())
```
Hassan754, line 168: These 3 lines are repeated 6 times below, would be great if it can exit the if conditions and be set at the last, maybe inside a BaseStrategy condition to make sure it won't be called on invalid input

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -160,6 +160,13 @@ def _populate_core_stats_and_sim_state(self, area):
...
+
         if isinstance(area.strategy, PVStrategy):
```
Hassan754, line 170: Nit
```suggestion
        elif isinstance(area.strategy, PVStrategy):
```

src/d3a/d3a_core/sim_results/plotly_graph.py
```diff
@@ -62,7 +59,7 @@ def _invert(inlist: list): return [-1 * ll for ll in inlist]
 
 
 def _get_color(key, alpha):
-    return DEVICE_PLOT_COLORS[key].replace("alpha", str(alpha))
+    return DEVICE_PLOT_COLORS.get(key, green).replace("alpha", str(alpha))
```
Hassan754, line 62: What's the rational behind falling to this default value ?

src/d3a/models/state.py
```diff
@@ -43,131 +46,249 @@
...
 
 
-class PVState:
+class StateInterface(metaclass=ABCMeta):
+    """Interface containing methods that need to be defined by each State class."""
+
+    @abstractmethod
+    def get_state(self) -> Dict:
+        """Return the current state of the device."""
+        return {}
```
Hassan754, line 55: The implementation here should be removed as it's an abstract function

src/d3a/models/state.py
```diff
@@ -43,131 +46,249 @@
...
+        super().__init__()
+        # Energy that the load wants to consume (given by the profile or live energy requirements)
+        self._desired_energy_Wh = {}
+        # Energy that the load needs to consume. It's reduced when new energy is bought
+        self._energy_requirement_Wh = {}
+        self._total_energy_demanded_Wh = 0
+
+    def get_state(self) -> Dict:
+        """Return the current state of the device. Extends super implementation."""
+        state = super().get_state()
```
Hassan754, line 82: If the `get_state` is to stay abstract this line can be safely removed

src/d3a/models/state.py
```diff
@@ -43,131 +46,249 @@
...
+        consumption_state = {
+            "desired_energy_Wh": convert_pendulum_to_str_in_dict(self._desired_energy_Wh),
+            "total_energy_demanded_Wh": self._total_energy_demanded_Wh}
+        # The inherited state should not have keys with the same name (to avoid overwriting them)
+        conflicting_keys = state.keys() & consumption_state.keys()
+        assert not conflicting_keys, f"Conflicting state values found for {conflicting_keys}."
```
Hassan754, line 88: We should not rely on assertions in our production code, a custom exception would work instead

src/d3a/models/state.py
```diff
@@ -43,131 +46,249 @@
...
+    def restore_state(self, state_dict: Dict):
+        super().restore_state()
+
+        self._desired_energy_Wh.update(convert_str_to_pendulum_in_dict(
+            state_dict["desired_energy_Wh"]))
+        self._total_energy_demanded_Wh = state_dict["total_energy_demanded_Wh"]
+
+    def get_energy_requirement_Wh(self, time_slot, default_value=0.0):
+        if default_value is None:
+            return self._energy_requirement_Wh[time_slot]
```
Hassan754, line 102: IMO the default value should always be fallen to, what if the interface gets a None and the key doesn't exist ?

Hassan754 (commented):
Left some comments, will review again once addressed

src/d3a/d3a_core/sim_results/plotly_graph.py
fievelk: You can check here: https://github.com/gridsingularity/d3a/pull/1159/files/3b073c1caa090c0499cd74ce34df080d15bc80a1#diff-a112ac89fc68de1a5551c06f65c8d71598bfdf7b440ad5bd883d2242c0d05ceaL37-L44

Before my changes, most of the keys all had the same value. So I just provided that value (`green`) as default and removed those lines.

src/d3a/models/state.py
fievelk: This line is used to allow the collaboration between mixins (when using multiple inheritance). That's also the reason why the abstract `get_state()` method in `StateInterface` is returning `{}`. Removing that would break the code that follows (that expects to handle a dict object).

The method resolution order for the `HomeMeterState` is the following:
```python
HomeMeterState.__mro__
(<class 'HomeMeterState'>, <class 'ConsumptionState'>, <class 'ProductionState'>, <class 'StateInterface'>, <class 'object'>)
```
So calling `super().get_state()` would call the `get_state()` method on all siblings/parents in the following order:
```
HomeMeterState -> ConsumptionState -> ProductionState -> StateInterface
```
If the `StateInterface.get_state()` returns `None`, the `ProductionState.get_state()` method will fail. We also want to allow using mixins flexibly, so this shouldn't be fixed in every individual state that implements `StateInterface`, but rather at the source.
I see some alternatives:
1. Do not make the method abstract (but that would sort of defeat the purpose of the interface class)
2. Check for `None` in every class that implements the interface (`ConsumptionState` and `ProductionState`).
3. Leave it as it is (use `@abstractmethod` but also provide implementation)

What do you think? Do you have any other ideas?

src/d3a/models/state.py
fievelk: This is not supposed to be a check for production: it should break the system during the development stages.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
fievelk: I agree that it would be good to refactor this somehow. I'm not sure that the BaseStrategy should be changed though because it would mean allowing something like `strategy.trades[None]`, which I find counterintuitive.
I personally think that the entire `SimulationEndpointBuffer` class should be refactored and split into several classes (in particular for the `_populate_core_stats_and_sim_state` method). But if that's ok for you I would keep it out of the current PR since it might take some time.

src/d3a/models/state.py
fievelk: I agree. I had changed it for the other methods below but I missed this one.

src/d3a/models/state.py
fievelk: I've replied here: https://github.com/gridsingularity/d3a/pull/1159#discussion_r653696953

src/d3a/models/state.py
Hassan754: I see, the solution #2 looks the best imo

src/d3a/d3a_core/sim_results/plotly_graph.py
hannesdiedrich: Here for a change I would have voted for the more verbose version (the original version) for readability. But leave it. It is not important part of the code.

src/d3a/models/state.py
```diff
@@ -43,131 +46,247 @@
...
+    def __init__(self):
+        super().__init__()
 
-    def get_desired_energy(self, time_slot):
-        return self._desired_energy_Wh[time_slot]
+    def _validate_state(self):
+        production_market_slots = {
+            slot for slot, energy in self._available_energy_kWh.items() if energy}
+        consumption_market_slots = {
+            slot for slot, energy in self._energy_requirement_Wh.items() if energy}
```
hannesdiedrich, line 247: Every time the `market_slot` property is called, this validation is performed. IMO an overkill. 
AFAIU, you are already checking here https://github.com/gridsingularity/d3a/pull/1159/files#diff-1187ba841fc3bdd8c4b0200dc3033dfc48bd59ba5c22485e605fd0d1293e1d78R438 If produced and consumed both have non-zero numbers. This is enough IMO. For the sake of performance, i suggest to remove this additional validation here and only  return the union like you do in the market_slots property. 
This is especially valid for high number if future markets.
Or did I miss some corner case?

tests/test_home_meter.py
```diff
@@ -0,0 +1,242 @@
...
+from d3a_interface.exceptions import D3AException
+from pendulum import datetime, duration
+
+from d3a.models.area import Area
+from d3a.models.market.one_sided import OneSidedMarket
+from d3a.models.state import HomeMeterState
+from d3a.models.strategy.home_meter import HomeMeterStrategy
+
+
+# pylint: disable=protected-access
```
hannesdiedrich, line 34: 👍

hannesdiedrich (commented):
Left a few comments, LGTM once addressed.

src/d3a/d3a_core/sim_results/plotly_graph.py
fievelk: Yes, the old version was more explicit. It was also more error-prone (if the key didn't exist in the dictionary, it would have raised an exception). But this is not necessarily bad, actually.

src/d3a/models/state.py
fievelk: I think your concern is well-founded. This validation can become slow for cases where we have many markets, and in theory this situation shouldn't occur given the structure of the profiles. 
Just to be sure: you're suggesting to completely remove the `_validate_state()` method and not call it in the `market_slots` property, right?

hannesdiedrich: Yes

tests/test_home_meter.py
```diff
@@ -0,0 +1,242 @@
...
+        market_mock.offer = Mock(return_value=offer_mock)
+        self.strategy.offer_update.initial_rate = {market_mock.time_slot: 11}
+        self.strategy.offers.post = Mock()
+
+        self.strategy._post_offer(market_mock)
+        self.strategy.offers.post.assert_called_once_with(offer_mock, market_mock.id)
+
+    @staticmethod
+    def _create_profile_mock():
+        # NOTE: the datetimes should be the same that will be used by the market slots
```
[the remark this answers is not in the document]
fievelk: Sure. The energy profile must be filled with data that will be used by the market slots. If we create the market slots (mocks) using different timestamps, they will not use the energy readings provided in the profile.
Do you suggest rephrasing the comment somehow, or moving it elsewhere maybe?

src/d3a/models/strategy/home_meter.py
```diff
@@ -0,0 +1,577 @@
...
+from d3a.models.market.market_structures import Offer
+from d3a.models.state import HomeMeterState
+from d3a.models.strategy import BidEnabledStrategy
+from d3a.models.strategy.update_frequency import (
+    TemplateStrategyBidUpdater, TemplateStrategyOfferUpdater)
+
+log = getLogger(__name__)
+
+
+# pylint: disable=invalid-name
```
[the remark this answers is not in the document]
fievelk: Pylint will complain about all cases in which we use `Wh` in variable names. For example:
> Variable name "offer_energy_kWh" doesn't conform to snake_case naming style (380:8) [invalid-name]

So your suggestion of adding it to `.pylintrc` makes sense, since this is widely used in our code.

tests/test_home_meter.py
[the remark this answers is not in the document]
hannesdiedrich: Ok, I understand now. I was confused because that should be of course the case.
It is really minor and feel free to skip it, but if you want to avoid this comment, you could expose the `slot_times` list in self and re-use it in `_create_market_mocks` with this you can be sure to create the fitting markets. Again, feel free to skip it, it is just the test.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
spyrostz: I agree with all of the above, just for the future IMO the BaseStrategy should not offer a full implementation, just the method with an Exception to notify us in case it is not implemented for a strategy, and each of these if blocks should belong to the respective strategy classes (in order to avoid the isinstance checks)

src/d3a/models/state.py
spyrostz: I would actually personally prefer to leave it as it is. Reason for that is that I predict that these classes will be composed in ways that we cannot envision now (for instance, I can imagine once we add power flow to have a power state, which would then convert the MRO to HomeMeterState -> ConsumptionState -> ProductionState -> PowerState -> StateInterface) and this would make it inconvenient to have to check for None for every implemented state class. Furthermore, default implementations in abstractmethods / interfaces are even allowed in Java nowadays, therefore I would not consider this an antipattern.

src/d3a/models/state.py
spyrostz: To reiterate what @fievelk  says, I am a big fan of assertions in production code, especially when debugging. When the code stumbles upon a condition that will definitely crash the system, it is much more easily debugged and reproduced to check why the assertion fails than trying to find the root cause of the issue through logs. 
For instance, this is a perfect example of a programming error (the parent classes should not have the same state keys as the child classes). It is much easier even for the developer to get notified immediately that the implementation is incorrect during development, rather than let it slide and have runtime errors (or, for this case, errors in the values reported in the UI). 
To the exceptions / assertions discussion, an assert is a type of an exception (AssertionError) meant exactly for this purpose. It just has a convenient syntax.

src/d3a/models/state.py
```diff
@@ -43,131 +46,247 @@
...
+        super().__init__()
         self._available_energy_kWh = {}
         self._energy_production_forecast_kWh = {}
 
-    def get_energy_production_forecast_kWh(self, time_slot, default_value=None):
-        assert self._energy_production_forecast_kWh[time_slot] >= -FLOATING_POINT_TOLERANCE
-        if default_value is not None:
-            return self._energy_production_forecast_kWh.get(time_slot, default_value)
-        return self._energy_production_forecast_kWh[time_slot]
+    def get_state(self) -> Dict:
```
spyrostz, line 157: I see that this method is mostly repeating itself for the strategy classes, but unfortunately I do not have a good way off the top of my head to make it more DRY. Maybe when I stumble upon it in the future I'll also try to fight it to remove the repeating code

src/d3a/models/strategy/home_meter.py
```diff
@@ -0,0 +1,577 @@
...
+
+        This method is triggered by the MARKET_CYCLE event.
+        """
+        super().event_market_cycle()
+
+        self._reset_rates_and_update_prices()
+        self._set_energy_forecast_for_future_markets(reconfigure=False)
+
+        # TODO: should we always do both bids and offers, or should we first check the +/- of the
+        #  profile?
```
spyrostz, line 184: These approaches are more or less equal (since no bid/offerwill be placed if the profile is negative/positive, the check is just not here but in the _post_offer and _post_first_bid respectively). If you want to be more explicit and not call one of the 2 methods be my guest, however I would leave it as it is. After all, you do not gain much in performance (market cycles are relatively infrequent)

src/d3a/models/strategy/home_meter.py
```diff
@@ -0,0 +1,577 @@
...
+    def _one_sided_market_event_tick(self, market, offer=None):
+        """
+        Define the behavior of the device on TICK events in one-sided markets (react to offers).
+        """
+        if not self.state.can_buy_more_energy(market.time_slot):
+            return
+
+        try:
+            if offer is None:
+                if not market.offers:
```
spyrostz, line 540: I would move this to the start of the method, to exit immediately. Same goes for the `if offer.id not in market.offers` below. This would allow the if / else block to be replaced by this `acceptable_offer = self._find_acceptable_offer(market) if not offer else offer`

spyrostz (dismissed):
Some comments, but LGTM in general

src/d3a/models/strategy/home_meter.py
fievelk: Thanks. I'll replace the TODO comment then 👍

tests/test_home_meter.py
```diff
@@ -0,0 +1,244 @@
...
+        return OrderedDict([
+            (self.slot_times[0], 1), (self.slot_times[1], -0.5), (self.slot_times[2], -0.1)])
+
+    def _create_market_mocks(self, num_of_markets=3):
+        market_mocks = [create_autospec(OneSidedMarket) for _ in range(num_of_markets)]
+        slot_time = self.slot_times[0]
+        for market_mock in market_mocks:
+            market_mock.time_slot = slot_time
+            market_mock.id = uuid.uuid4()
+            slot_time += duration(minutes=15)
```
hannesdiedrich, line 242: Sorry, I meant to reuse it even more. Again, minor!
```suggestion
    def _create_market_mocks(self):
        market_mocks = [create_autospec(OneSidedMarket) for _ in range(self.slot_times)]
        for ii, market_mock in enumerate(market_mocks):
            market_mock.time_slot = self.slot_times[ii]
            market_mock.id = uuid.uuid4()
```
I just realise that you increase the slot_time after assigning it to market_mock.time_slot. Is this working due to the fact that is is a reference (pointer)?

hannesdiedrich (approved):
One minor comment, LGTM

tests/test_home_meter.py
fievelk: Hmm in Python it's not so straight-forward, so I hope I'll get this right. 
`slot_time = self.slot_times[0]` creates a new variable (`slot_time`) with the same reference as `self.slot_times[0]` (so to they refer to the same `datetime` object). When `slot_time += duration(minutes=15)` is executed, the reference of `slot_time` changes, but not the reference of `self.slot_times[0]`. So `self.slot_times[0]` does not change.
Python is neither pass-by-reference nor pass-by-value. When we re-assign `slot_time`, we are rebinding only its reference to another object (but not `self.slot_times[0]`'s reference). So the two references will point to two different objects.

fievelk: Talking about the specific case here: I wanted to have a flexible implementation of `_create_market_mocks` to create as many mocks as I wanted, unrelated to the number of time slots that I defined previously. That's why the time is changed dynamically (just adding 15 minutes every time). If you want, I can definitely add your suggested changes, since they wouldn't change the outcome of the current tests.

hannesdiedrich: Yeah, this is valid if you want to reuse the _create_market_mocks for a different test later on. So leave it be.

Hassan754 (approved):
LGTM

PR #1160, opened 2021-06-02
D3ASIM-3483: Corrected way of deleting prohibited asset info form the…

hannesdiedrich:
… area_stats_tree_dict that is sent to the aggregator client.

This issue was discovered when connecting two aggregators to the simulation. Each aggregator was receiving asset_infos from the devices that were controlled by the other aggregator too. Due to privacy issues, I added a filtering for the aggregator uuid two. The main fix is in `_delete_not_owned_devices`.

### DIFF
--- a/src/d3a/d3a_core/redis_connections/aggregator_connection.py
+++ b/src/d3a/d3a_core/redis_connections/aggregator_connection.py
@@ -39,32 +39,36 @@ class AggregatorHandler:
         aggregator_uuid = self.device_aggregator_mapping[device_uuid]
         create_subdict_or_update(batch_event_dict, aggregator_uuid, event)
 
-    def _delete_not_owned_devices_from_dict(self, area_stats_tree_dict):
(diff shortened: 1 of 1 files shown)


spyrostz (approved):
LGTM, I assume this was due to duplicate area uuids from cloned configurations ? Because IIRC only one aggregator can access an area at the moment.

hannesdiedrich:
> LGTM, I assume this was due to duplicate area uuids from cloned configurations ? Because IIRC only one aggregator can access an area at the moment.

Yes, `only one aggregator can access an area at the moment` is true, but in each event we send the area_stats_tree_dict that contains all of the assets and areas. We prohibited other aggregators to read the infos from other areas, only with a too weak check. We only checked `if area_uuid not in self.device_aggregator_mapping`. But this is true for all areas that are connected to any aggregators.

hannesdiedrich:
Sorry, I forgot to push my changes for being compliant to pylint. Just did.

Hassan754 (dismissed):
LGTM

Hassan754 (approved):

PR #1161, opened 2021-06-07
D3ASIM-3486: Fixed: Don't send results when run via cli.

hannesdiedrich:


Hassan754 (dismissed):
LGTM

tests/test_simulation.py
```diff
@@ -0,0 +1,85 @@
...
+        simulation.endpoint_buffer.results_handler.all_ui_results = \
+            {k: {} for k in results_mapping}
+
+        simulation._update_and_send_results()
+
+        assert not simulation.endpoint_buffer.prepare_results_for_publish.called
+        assert not simulation.kafka_connection.publish.called
+
+    @staticmethod
+    def test_results_are_send_via_kafka_if_not_started_from():
```
fievelk, line 61:
```suggestion
    def test_results_are_sent_via_kafka_if_not_started_from_cli():
```

tests/test_simulation.py
```diff
@@ -0,0 +1,85 @@
...
+from d3a_interface.kafka_communication.kafka_producer import (DisabledKafkaConnection,
+                                                              KafkaConnection)
+from d3a_interface.sim_results.all_results import ResultsHandler
+from pendulum import duration, today
+
+
+class SimulationTest(unittest.TestCase):
+
+    @staticmethod
+    def test_results_not_send_via_kafka_if_started_from_cli():
```
fievelk, line 34:
```suggestion
    def test_results_not_sent_via_kafka_if_started_from_cli():
```

fievelk: Extremely minor: I just suggested changing the spelling from `send` to `sent`.

tests/test_simulation.py
```diff
@@ -0,0 +1,85 @@
...
+        simulation = Simulation(
+            "default_2a", simulation_config, None, 0, False, duration(), False, False, None, None,
+            redis_job_id, False
+        )
+        simulation.endpoint_buffer = MagicMock(spec=SimulationEndpointBuffer)
+        results_mapping = ResultsHandler().results_mapping
+        simulation.endpoint_buffer.results_handler = MagicMock(spec=ResultsHandler)
+        simulation.kafka_connection = MagicMock(spec=DisabledKafkaConnection)
+        simulation.endpoint_buffer.results_handler.all_ui_results = \
+            {k: {} for k in results_mapping}
```
fievelk, line 53: If you want to follow PEP-8, you can avoid using the backslash (which is discouraged) and do this:
```suggestion
        simulation.endpoint_buffer.results_handler.all_ui_results = {
            k: {} for k in results_mapping}
```
I don't know why PyCharm doesn't enforce this yet. They have an open ticket for it anyway (https://youtrack.jetbrains.com/issue/PY-17193).

fievelk (dismissed):
LGTM. Only very minor comments that can be ignored.

src/d3a/d3a_core/simulation.py
```diff
@@ -265,10 +265,12 @@ def _update_and_send_results(self):
             self.file_stats_endpoint(self.area)
             return
 
-        results = self.endpoint_buffer.prepare_results_for_publish()
-        if results is None:
-            return
-        self.kafka_connection.publish(results, self._simulation_id)
+        # don't send results to kafka when running on cli even though kafka connection is enabled:
+        if not self._started_from_cli and self.kafka_connection.is_enabled():
```
faizan2590, line 269:
```suggestion
        if not self._started_from_cli and self.kafka_connection.is_enabled:
```
How about making `is_enabled` a property method?

faizan2590 (dismissed):
One suggestion, other than that LGTM

src/d3a/d3a_core/simulation.py
spyrostz, line 269: If you want to go this way, then you should add the self._started_from_cli to the should_export_stats method, in order to be applied for the whole simulation class and not just here.

spyrostz: I do not really recall why, but I think that I stumbled upon issues while trying to implement exactly this (probably with unit tests and mocking the property IIRC )

spyrostz (dismissed):
LGTM, but FYI, same as with Redis before introducing Kafka, you do not really need this, since disabling Kafka should allow the CLI to run (have a look at `should_export_results` as discussed in the comment too). Redis was also sending the results in the past if one had a running redis server while running via CLI, the most probably reason why it did not crash was because the results were not validated.
Having said that, it is okay to have this as an extra precaution

hannesdiedrich:
> LGTM, but FYI, same as with Redis before introducing Kafka, you do not really need this, since disabling Kafka should allow the CLI to run (have a look at `should_export_results` as discussed in the comment too). Redis was also sending the results in the past if one had a running redis server while running via CLI, the most probably reason why it did not crash was because the results were not validated.
> Having said that, it is okay to have this as an extra precaution

@spyrostz I had a look even before and the results where indeed validated before your change, have a look here: https://github.com/gridsingularity/d3a/pull/1137/files#diff-78571cbd782c75f4e8975fac86e013383e9b02c5c3bc1b63f7889a6c1c91423cL176
I thought the early return of `publish _results` when redis is not enabled was the difference. But you are right, if a redis server was running, we would have the same issue. But then I can not really understand why this fails locally. 
I would keep this change, as you agreed that it does not make sense to send results via kafka if you run locally.

src/d3a/d3a_core/simulation.py
hannesdiedrich: OK, great! Will change that.

hannesdiedrich: Actually, now that I tried it again (had before but forgot). 
We actually look for the case here: `KAFKA=True` + `LOCAL_CLI_RUN=False`.
Your proposal of `self.should_export_plots` would be: `KAFKA=False` + `LOCAL_CLI_RUN=True`
which is indeed the invert of the above.
However, we like to omit sending results for the case: `KAFKA=True` + `LOCAL_CLI_RUN=True`

spyrostz: My proposal was to actually use an or condition and not the and condition that you mentioned:
`return self._started_from_cli or not self.kafka_connection.is_enabled()`

hannesdiedrich: Ill have an idea how to fix it. Will propose something.

src/d3a/d3a_core/simulation.py
```diff
@@ -399,19 +399,15 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
-                self.export.export(export_plots=self.should_export_results,
-                                   power_flow=self.power_flow)
-            else:
-                self.export.export(self.should_export_results)
+            self.export.export(power_flow=self.power_flow if GlobalConfig.POWER_FLOW else None)
 
     @property
-    def should_export_results(self):
-        return not self.kafka_connection.is_enabled()
+    def should_send_results(self):
```
fievelk, line 409: Nit: the naming/documentation seems to be still a bit unclear to me (where should the results be sent? Why?)

fievelk (dismissed):
LGTM

src/d3a/d3a_core/simulation.py
spyrostz: I think that should_send_results_to_broker would help clarify this. I would suggest should_publish_result if we sticked to Redis, however we are not running on pubsub anymore, therefore it is a misleading name.

integration_tests/steps/integration_tests.py
```diff
@@ -603,6 +602,7 @@ def run_sim_market_count(context, scenario):
 @when('export {flag}')
 def export_logic(context, flag):
     context.no_export = True if flag == 'isnt_needed' else False
+    print(context.no_export)
```
spyrostz, line 605: Reminder to remove before merging

src/d3a/d3a_core/simulation.py
```diff
@@ -255,20 +254,21 @@ def _update_and_send_results(self):
...
 
-        results = self.endpoint_buffer.prepare_results_for_publish()
-        if results is None:
-            return
-        self.kafka_connection.publish(results, self._simulation_id)
+        if self.should_send_results:
+            results = self.endpoint_buffer.prepare_results_for_publish()
+            if results is None:
+                return
+            self.kafka_connection.publish(results, self._simulation_id)
```
spyrostz, line 271: I would further refactor all this to the following structure:
`if self.export_results_on_finish: .... `
`elif self.should_send_results: ....`
Reason for that is because these statements are mutually exclusive, therefore we are certain that if we are exporting the simulation results we should not send them via the message broker.

spyrostz (approved):
LGTM once comments are dealt with

fievelk (approved):

spyrostz (approved):
LGTM

PR #1163, opened 2021-06-14
D3ASIM-3464: Initial implementation of saving KPI.

faizan2590:


src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -146,6 +147,13 @@ def _populate_core_stats_and_sim_state(self, area):
                 core_stats_dict['trades'].append(trade.serializable_dict())
         if hasattr(area.current_market, 'market_fee'):
             core_stats_dict['market_fee'] = area.current_market.market_fee
+        if hasattr(area.current_market, 'const_fee_rate') is not None and \
+                area.current_market is not None:
```
Hassan754, line 151:
```suggestion
        if hasattr(area.current_market, 'const_fee_rate') is not None and \
                area.current_market.const_fee_rate is not None:
```

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -146,6 +147,13 @@ def _populate_core_stats_and_sim_state(self, area):
...
+            core_stats_dict['const_fee_rate'] = area.current_market.const_fee_rate \
+                if area.current_market.const_fee_rate is not None else 0.
+            core_stats_dict['feed_in_tariff'] = FEED_IN_TARIFF
+            core_stats_dict['market_maker_rate'] = \
+                ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE
```
spyrostz, line 156: I think that it is better to fetch the real current market maker rate from the strategy instead of the constant, reason for that is because there are plans in the future to add market maker profiles, and multiple market makers per simulation.

spyrostz (commented):
One important comment, LGTM once resolved

setup.py
```diff
@@ -1,8 +1,8 @@
 from setuptools import find_packages, setup
-import os
+# import os
 
 
-d3a_interface_branch = os.environ.get("BRANCH", "master")
+d3a_interface_branch = "feature/D3ASIM-3464"  # os.environ.get("BRANCH", "master")
```
fievelk, line 5: Reminder to restore the original version

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -146,6 +147,13 @@ def _populate_core_stats_and_sim_state(self, area):
                 core_stats_dict['trades'].append(trade.serializable_dict())
         if hasattr(area.current_market, 'market_fee'):
             core_stats_dict['market_fee'] = area.current_market.market_fee
+        if hasattr(area.current_market, 'const_fee_rate') is not None and \
```
fievelk, line 150: Please use `"` over `'`. The same comment applies to the rest of the PR.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -146,6 +147,13 @@ def _populate_core_stats_and_sim_state(self, area):
...
+                area.current_market is not None:
+            core_stats_dict['const_fee_rate'] = area.current_market.const_fee_rate \
```
fievelk, line 152: Nit: please use `( )` over `\` for line continuation.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
faizan2590: Can you please explain the motive behind the proposed change?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
faizan2590: How about using `GlobalConfig.market_maker_rate` ?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
Hassan754: In your code you are checking if `hasattr(area.current_market, 'const_fee_rate') is not None`, this is not a valid check, the `hasattr` will return a boolean either True or False, and there is no way it will return a None (The value you're comparing against)
Furthermore, the same check is intended to validate if area.current_market has an attrribute `const_fee_rate`, so the second condition `area.current_market is not None` doesn't make sense at this stage, if there is a possibility that this can be None it should be the first condition

faizan2590:
```suggestion
        if (getattr(area.current_market, "const_fee_rate", None) is not None and
                area.current_market is not None):
```
How about this? 

`area.current_market is not None` is needed when its called at the start of simulation and `current_market` is None at that point.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
spyrostz: This will work for now, however keep in mind that this is a dict, therefore you have to access the current market slot value of the market maker rate and not just add the whole dict.

integration_tests/savings_kpi.feature
```diff
@@ -0,0 +1,19 @@
...
+     Given we have a scenario named kpi.d3asim_3464_family_with_pv
+     And d3a is installed
+     When we run the d3a simulation on console with kpi.d3asim_3464_family_with_pv for 24 hrs
+     Then saving_percentage of {'Family 2 children with PV': -43.537} are correctly reported
+
+  Scenario: Family House with PV & ESS
+     Given we have a scenario named kpi.d3asim_3464_family_with_pv_ess
+     And d3a is installed
+     When we run the d3a simulation on console with kpi.d3asim_3464_family_with_pv_ess for 24 hrs
+     Then saving_percentage of {'Family 2 children with PV + ESS': -176.514} are correctly reported
```
fievelk, line 19: Why not using behave's [scenario outlines](https://behave.readthedocs.io/en/stable/tutorial.html#scenario-outlines) here?
We already used them in other tests and they're useful to reduce code repetition.

integration_tests/steps/kpi.py
```diff
@@ -36,6 +36,7 @@ def test_export_of_kpi_result(context, kpi, expected_kpis):
                                           "aggregated_results", "kpi.json"))
     with open(sim_data_csv[0], "r") as sf:
         kpi_data = json.load(sf)
+    print(f"kpi_data: {kpi_data}")
```
fievelk, line 39: Leftover from debug

setup.py
fievelk: Another reminder, since you've asked for a new review

src/d3a/constants.py
```diff
@@ -45,3 +45,6 @@
 CN_PROFILE_EXPANSION_DAYS = 7
 
 RUN_IN_REALTIME = False
+
+
+FEED_IN_TARIFF = 20
```
fievelk, line 50: Nit: why two blank lines?
```suggestion

FEED_IN_TARIFF = 20
```

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -135,31 +136,43 @@ def _populate_core_stats_and_sim_state(self, area):
...
+        if hasattr(area.current_market, "trades"):
             for trade in area.current_market.trades:
-                core_stats_dict['trades'].append(trade.serializable_dict())
-        if hasattr(area.current_market, 'market_fee'):
-            core_stats_dict['market_fee'] = area.current_market.market_fee
+                core_stats_dict["trades"].append(trade.serializable_dict())
+        if hasattr(area.current_market, "market_fee"):
+            core_stats_dict["market_fee"] = area.current_market.market_fee
+        if (getattr(area.current_market, "const_fee_rate", None) is not None and
+                area.current_market is not None):
```
fievelk, line 152: The order of the two operands should be switched:
```suggestion
        if area.current_market and getattr(area.current_market, "const_fee_rate", None) is not None:
```
Besides this: could you explain why we are using `getattr` and `hasattr` to check if the area has these attributes? What's the situation in which we don't have them?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -135,31 +136,43 @@ def _populate_core_stats_and_sim_state(self, area):
...
+            core_stats_dict["const_fee_rate"] = (area.current_market.const_fee_rate
+                                                 if area.current_market.const_fee_rate is not None
```
fievelk, line 154: You've already checked whether `area.current_market.const_fee_rate` is not `None` on the line above this. So I think this could be:
```suggestion
            core_stats_dict["const_fee_rate"] = area.current_market.const_fee_rate
```
Do you agree?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -135,31 +136,43 @@ def _populate_core_stats_and_sim_state(self, area):
...
+                                                 else 0.)
```
fievelk, line 155: You've already checked whether `area.current_market.const_fee_rate` is not `None` on the line above this. So I think this could be:
```suggestion
            core_stats_dict["const_fee_rate"] = area.current_market.const_fee_rate
```
Do you agree?

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -135,31 +136,43 @@ def _populate_core_stats_and_sim_state(self, area):
...
+            core_stats_dict["feed_in_tariff"] = FEED_IN_TARIFF
+            if isinstance(GlobalConfig.market_maker_rate, int):
+                core_stats_dict["market_maker_rate"] = GlobalConfig.market_maker_rate
+            elif isinstance(GlobalConfig.market_maker_rate, dict):
+                core_stats_dict["market_maker_rate"] = \
```
fievelk, line 160: Nit: please use `(` and `)` to replace `\` for line continuation.

src/d3a/setup/kpi/d3asim_3464_family_with_pv.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def get_setup(config):
+
+    ConstSettings.IAASettings.MARKET_TYPE = 2
+    ConstSettings.IAASettings.MIN_OFFER_AGE = 0
+    ConstSettings.IAASettings.MIN_BID_AGE = 0
+
+    area = Area(
+        'Grid',
```
fievelk, line 16: Nit: please use `"` over `'`.

fievelk: (This applies to the entire module).

fievelk:
Left some comments. Will review again once answered.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -135,31 +136,43 @@ def _populate_core_stats_and_sim_state(self, area):
...
+                    GlobalConfig.market_maker_rate[area.current_market.time_slot]
```
hannesdiedrich, line 161: There is a function that does exactly that: https://github.com/gridsingularity/d3a/blob/3c75998cb2b9963030bfebbea5ef4cda6cc625c3/src/d3a/d3a_core/util.py#L502
(BTW; you could refactor the input value `next_market` into `market`, thanks!)

src/d3a/constants.py
```diff
@@ -45,3 +45,5 @@
 CN_PROFILE_EXPANSION_DAYS = 7
 
 RUN_IN_REALTIME = False
+
+FEED_IN_TARIFF = 20
```
hannesdiedrich, line 49: We want to change this only in a blue moon or why is this not a GlobalConfig parameter? Not an expert here, but doesn't this price also change frequently?

hannesdiedrich (approved):
One global comment, but probably reasoned by design of this feature. 
LGTM

src/d3a/constants.py
faizan2590: > We want to change this only in a blue moon or why is this not a GlobalConfig parameter?

Looks better there, will move it even now

> Not an expert here, but doesn't this price also change frequently?

I haven't changed it yet as for all three referenced cases, feed-in tariff had the same rate.

hannesdiedrich (approved):
LGTM

hannesdiedrich:
@faizan2590 Just a general question: Is this KPI available to the user (UI) even now without a change in the d3a-web? The ticket suggests that they want to add a bar chart for that.

faizan2590:
> @faizan2590 Just a general question: Is this KPI available to the user (UI) even now without a change in the d3a-web? The ticket suggests that they want to add a bar chart for that.

Yes, give this commit a look: https://github.com/gridsingularity/d3a-interface/pull/223/commits/9371f0858d93922c2cb5e6bb24975c08baa4a080

spyrostz (dismissed):
LGTM, as Pierpaolo also says don't forget to restore the d3a-interface branch to master before merging !

PR #1164, opened 2021-06-14
Remove update offer bid

Hassan754:


spyrostz (approved):
LGTM, but have you already checked with Colin to validate that these methods are not used in any of their scripts ? If yes, then LGTM

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -255,40 +255,6 @@ def event_offer(self, *, market_id, offer):
...
-                    "command": "update_bid", "status": "ready",
-                    "bid": updated_bid.to_json_string(),
-                    "area_uuid": self.device.uuid,
-                    "transaction_id": arguments.get("transaction_id", None)}
-            else:
-                return {
-                    "command": "update_bid", "status": "error",
-                    "area_uuid": self.device.uuid,
-                    "error_message": "Updated bid would only work if the old exist in market.",
-                    "transaction_id": arguments.get("transaction_id", None)}
```
fievelk, line 290: - `update_bid` (`Commands.UPDATE_BID` in d3a-api-client): when there is no bid to update, the command will return an error response. 
- `bid` (`Commands.BID` in d3a-api-client): when there is no bid to update, the command will create a new bid without returning any error.

I suppose the same applies to offers. So the behaviour of the two methods is actually different. I'm not against deprecating the update methods, but since there is no equivalent method for it we should probably ask the research team before proceeding.

Hassan754:
Tagging Colin in the PR

Hassan754:
After an offline discussion this PR can be safely merged after dealing with [this ticket](https://gridsingularity.atlassian.net/browse/D3ASIM-3488)

fievelk (approved):

hannesdiedrich (approved):
Please also add a PR for d3a-web and adapt the `valid_command_types` there. Thanks a lot!
LGTM

Hassan754:
> valid_command_types

Thanks for pointing me out to it, adding it!

PR #1167, opened 2021-06-18
D3ASIM-2540: Enabling InfiniteBus to read csv profile for local run.

faizan2590:


Hassan754 (approved):
LGTM, need extra attention to the back slashes + these parsing look messy and should be refactored , but it's fine for now

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -44,9 +44,12 @@ def __init__(self, energy_sell_rate=None, energy_rate_profile=None, energy_buy_r
         self._set_market_maker_rate()
 
     def _set_market_maker_rate(self):
-        if self.energy_rate_profile is not None:
+        if self.energy_rate_profile is not None and isinstance(self.energy_rate_profile, dict):
```
fievelk, line 47: General question: `self.energy_rate_profile` is one of the elements in the `parameters` set. Doesn't this mean that the `InfiniteBusStrategy` can't be instantiated without it?

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -57,17 +60,27 @@ def _set_market_maker_rate(self):
                 ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE
 
     def event_activate(self, **kwargs):
-        if self.energy_rate_profile is not None:
+        if self.energy_rate_profile is not None and isinstance(self.energy_rate_profile, dict):
```
fievelk, line 63: Nit: we can save some complexity if we switch the order:
```python
if not self.energy_rate_profile:
   # do something

if isinstance(self.energy_rate_profile, dict):
    # do something
elif isinstance(self.energy_rate_profile, str):
    # do something else
```

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -57,17 +60,27 @@ def _set_market_maker_rate(self):
...
             self.energy_rate = read_and_convert_identity_profile_to_float(self.energy_rate_profile)
             del self.energy_rate_profile
+        elif self.energy_rate_profile is not None and isinstance(self.energy_rate_profile, str):
+            self.energy_rate = read_arbitrary_profile(InputProfileTypes.IDENTITY,
+                                                      self.energy_rate_profile)
+            del self.energy_rate_profile
```
fievelk, line 69: Why do we need to delete this attribute?

fievelk: The same question applies to all other deletions.

fievelk (approved):
LGTM in general. I left some questions and refactoring suggestions.

hannesdiedrich (commented):
Left a question. Will review again when it and Pierpaolos comments are addressed.

src/d3a/models/strategy/infinite_bus.py
faizan2590: Its not a mandatory argument in constructor, so can be instantiated without it.

src/d3a/models/strategy/infinite_bus.py
faizan2590: As buying rate irrespective of single value or profile is casted to `self.energy_buy_rate` at the time of `event_activate` and till than it(buying_rate_profile) is saved as instance variable and same goes for selling rate

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -57,17 +60,27 @@ def _set_market_maker_rate(self):
...
+        if self.buying_rate_profile is not None and \
+                isinstance(self.buying_rate_profile, (float, dict)):
             self.energy_buy_rate = \
                 read_and_convert_identity_profile_to_float(self.buying_rate_profile)
             del self.buying_rate_profile
+        elif self.buying_rate_profile is not None and isinstance(self.buying_rate_profile, str):
+            self.energy_buy_rate = \
+                read_arbitrary_profile(InputProfileTypes.IDENTITY, self.buying_rate_profile)
+            del self.buying_rate_profile
+
```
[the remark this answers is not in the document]
faizan2590: If its an `str` it contains file path to csv where profile is for local run, and in other case it could be ~float~`dict` being passed from d3a-web.

src/d3a/models/strategy/infinite_bus.py
fievelk: Sorry, I couldn't understand the explanation. `self.buying_rate_profile` contains a string/dict/path that defines a profile file. We process its content and we store the results in `self.energy_buy_rate`.  Why do we need to delete the original `self.buying_rate_profile`?

src/d3a/models/strategy/infinite_bus.py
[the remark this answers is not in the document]
fievelk: I think that `read_arbitrary_profile` already handles all these cases. Please make sure you double-check so that we could possibly simplify these `if` statements :)

tests/test_strategy_infinite_bus.py
```diff
@@ -333,3 +333,66 @@ def testing_event_market_cycle_posting_bids(bus_test4, area_test4):
...
+@pytest.fixture()
+def bus_test5(area_test5):
+    c = InfiniteBusStrategy(energy_rate_profile="src/d3a/resources/SAM_SF_Summer.csv")
+    c.area = area_test5
+    c.owner = area_test5
+    return c
+
+
+def test_global_market_maker_rate_profile(bus_test5, area_test5):
+    assert isinstance(GlobalConfig.market_maker_rate, dict)
```
fievelk, line 355: Sorry, I just noticed this part. What is this test doing? Why does it need fixtures, and what does it have to do with the Infinite Bus strategy?

tests/test_strategy_infinite_bus.py
```diff
@@ -333,3 +333,66 @@ def testing_event_market_cycle_posting_bids(bus_test4, area_test4):
...
+
+@pytest.fixture()
+def bus_test6(area_test6):
+    c = InfiniteBusStrategy(energy_rate_profile="src/d3a/resources/SAM_SF_Summer.csv")
+    c.area = area_test6
+    c.owner = area_test6
+    return c
+
+
+def test_infinite_bus_selling_rate_set_as_profile(bus_test6, area_test6):
```
fievelk, line 374: Is `area_test6` needed as a parameter?

tests/test_strategy_infinite_bus.py
```diff
@@ -333,3 +333,66 @@ def testing_event_market_cycle_posting_bids(bus_test4, area_test4):
...
+    bus_test6.event_activate()
+    assert isinstance(bus_test6.energy_rate, dict)
+
+
+"""TEST7"""
+
+
+@pytest.fixture()
+def area_test7():
+    return FakeArea(0)
```
fievelk, line 384: `area_test5/6/7` are all the same. Why do we need to create separate fixtures for them?

src/d3a/models/strategy/infinite_bus.py
faizan2590: > I mean, I understand that its content might not be used in the remaining operations, but using del on it seems a bit strange (since we never do it in any other situation). Also, this would cause an exception if I try to re-activate the device I suppose.

Can you please shed some clarity where/how you meant to re-activate the device?

fievelk:
Sorry for the late comments about tests, I must have missed them for some reason during my first review.
It seems like you're only checking that the profile of the Infinite Bus is a `dict` object, but you're not checking its contents. Why is that? 
Also, you're not testing individual methods but rather a long-distance side-effect of the instantiation of an `InfiniteBus` strategy; this might give us some insight, but it's a bit different from a unit test I think. Anyway, you don't need to address this now.

Nit: as discussed on slack, using classes in tests would be better to create a clearer distinction between the objects being tested.

tests/test_strategy_infinite_bus.py
faizan2590: I followed the same pattern as before, may be thought to have isolation of area object for each unittest, also I tested no now with common `area_test` fixtures and nothing complained, so will adapt your suggestion.

src/d3a/models/strategy/infinite_bus.py
fievelk: I'm not planning on doing it, that was just a note on the side. (However, it shows that deleting an instance variable might lead to unexpected effects that are not being handled with proper exceptions).

tests/test_strategy_infinite_bus.py
fievelk: If I'm not wrong, the fixtures created by pytest have a default `scope="function"`, so they should be reset before each test function runs.

src/d3a/models/strategy/infinite_bus.py
faizan2590: > I'm not planning on doing it, that was just a note on the side. (However, it shows that deleting an instance variable might lead to unexpected effects that are not being handled with proper exceptions).

My intuition was to delete un-needed instance variable(specially this one which contains profile leading to higher memory usage at run-time), but lets be safe than sorry of not thinking about any possible corner case that may fail/crash the simulation

tests/test_strategy_infinite_bus.py
faizan2590: No, remnants of copy/paste, removing it now, thanks for catching it.

src/d3a/models/strategy/infinite_bus.py
fievelk: I think your intuition is good: if the content of this variable can noticeably increase the memory consumption - and if this can be problematic - then it makes sense to provide ways to avoid it. However, if this is not the case, I think that using `del` is not necessary most of the time. So please proceed as you feel would make more sense 👍

tests/test_strategy_infinite_bus.py
fievelk: Are you using pylint to check these files? I think it would have caught it.

tests/test_strategy_infinite_bus.py
faizan2590: > Why does it need fixtures, and what does it have to do with the Infinite Bus strategy?

InfiniteBus strategy also sets the global market maker rate that is used by other strategies as a reference.

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -45,33 +44,41 @@ def __init__(self, energy_sell_rate=None, energy_rate_profile=None, energy_buy_r
...
+            GlobalConfig.market_maker_rate = (
+                ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE)
 
-    def event_activate(self, **kwargs):
+    def _populate_selling_rate(self):
         if self.energy_rate_profile is not None:
-            self.energy_rate = read_and_convert_identity_profile_to_float(self.energy_rate_profile)
+            self.energy_rate = read_arbitrary_profile(InputProfileTypes.IDENTITY,
+                                                      self.energy_rate_profile)
+            # TODO: to be checked via deleting in case increased memory is observed during runtime
```
hannesdiedrich, line 59: I will take care of it when adapting to this in this PR:
https://github.com/gridsingularity/d3a/pull/1152/files#diff-b181cf173660ff13072297b6d1be9ab11df92dca436269dbef7da3ea0a523909
(src/d3a/models/strategy/infinite_bus.py) Or did you experience a large memory usage?

src/d3a/models/strategy/commercial_producer.py
```diff
@@ -63,6 +63,7 @@ def event_market_cycle(self):
 
     def offer_energy(self, market):
         energy_rate = find_object_of_same_weekday_and_time(self.energy_rate, market.time_slot)
+        print(f"offer_energy: {energy_rate}")
```
hannesdiedrich, line 66: Please remove

tests/test_strategy_infinite_bus.py
```diff
@@ -306,30 +306,88 @@ def testing_event_market_cycle_post_offers(bus_test3, area_test3):
...
     ConstSettings.IAASettings.MARKET_TYPE = 1
+
+
+def test_global_market_maker_rate_single_value(bus_test4):
+    assert isinstance(GlobalConfig.market_maker_rate, int)
+    assert (GlobalConfig.market_maker_rate ==
+            ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE)
+
+
+"""TEST5"""
```
hannesdiedrich, line 339: Why are these needed?

src/d3a/models/strategy/infinite_bus.py
faizan2590: As discussed in other [thread](https://github.com/gridsingularity/d3a/pull/1167#discussion_r665194267), its my intuition

tests/test_strategy_infinite_bus.py
faizan2590: Still following the convention in continuity to old test

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -45,33 +44,41 @@ def __init__(self, energy_sell_rate=None, energy_rate_profile=None, energy_buy_r
...
+        elif self.energy_rate is not None:
             GlobalConfig.market_maker_rate = self.energy_rate
-        elif isinstance(self.energy_rate, (str, dict)):
-            GlobalConfig.market_maker_rate = \
-                read_arbitrary_profile(InputProfileTypes.IDENTITY, self.energy_rate)
         else:
-            GlobalConfig.market_maker_rate = \
-                ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE
+            GlobalConfig.market_maker_rate = (
+                ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE)
```
hannesdiedrich, line 53: I know this is not new code, and I only see it now.
Imagine a user sets the `GlobalConfig.market_maker_rate` in a setup file, you would overwrite it by adding an infinite bus and don't provide an `energy_sell_rate`. Is this wanted? 
If you agree that it is not please remove this block. Thanks!

tests/test_strategy_infinite_bus.py
```diff
@@ -306,30 +306,88 @@ def testing_event_market_cycle_post_offers(bus_test3, area_test3):
...
+    return c
+
+
+def test_infinite_bus_selling_rate_set_as_profile(bus_test6):
+    bus_test6.event_activate()
+    assert isinstance(bus_test6.energy_rate, dict)
+    assert len(bus_test6.energy_rate) == 96
+    assert list(bus_test6.energy_rate.values())[0] == 516.0
+    assert list(bus_test6.energy_rate.values())[-1] == 595.0
+
```
hannesdiedrich, line 374: IMO you could combine these methods. If you want to separate the check for the `energy_rate` and the `market_maker_rate`, ok, but you can definitely reuse the `bus_test5` fixture in `test_infinite_bus_selling_rate_set_as_profile `

hannesdiedrich (commented):

tests/test_strategy_infinite_bus.py
faizan2590: Okay, will do so BUT please try to catch/drop such comments in earlier rounds of review, so could be adapted within lesser number review rounds

src/d3a/models/strategy/infinite_bus.py
faizan2590: As far as I know `InfiniteBus` is meant to supersede for setting the `GlobalConfig.market_maker_rate`. Have you already seem a setup where user didn'T set `energy_sell_rate ` but goes the other way via setting `GlobalConfig.market_maker_rate` ?

tests/test_strategy_infinite_bus.py
hannesdiedrich: Yes, sorry, my bad. I thought I'd look into the tests once the overall code is ready. I will do it right away next time.

src/d3a/models/strategy/infinite_bus.py
hannesdiedrich: I haven't, I just thought about this corner case. And yes, the `InfiniteBus is meant to supersedes` the setting of the `GlobalConfig.market_maker_rate` even with my proposal. But if you "defuse" it and don't provide a energy_rate it should not overwrite the global config that the user set before.

tests/test_strategy_infinite_bus.py
hannesdiedrich: Yeah. I think at some point we can get rid of them. They don't seem to have any use.
Leave it here though.

tests/test_strategy_infinite_bus.py
hannesdiedrich: But I have to say: https://github.com/gridsingularity/d3a/pull/1167#discussion_r665177470 Is basically the same comment and you could have adapted ti the bus fixtures too.

faizan2590: Not really AFAIS, https://github.com/gridsingularity/d3a/pull/1167#discussion_r665177470 was to reuse the same fixtures whereas this thread is to merge two test into one targeting different assertions

src/d3a/models/strategy/infinite_bus.py
hannesdiedrich: IMO this comment is not very descriptive (now we know what it means, but in 2 weeks we will not remember). Leave it though, as I will refactor the profile handling in D3ASIM-3421 anyways.

hannesdiedrich (approved):
LGTM

src/d3a/models/strategy/infinite_bus.py
```diff
@@ -45,33 +44,38 @@ def __init__(self, energy_sell_rate=None, energy_rate_profile=None, energy_buy_r
...
-                read_arbitrary_profile(InputProfileTypes.IDENTITY, self.energy_rate)
-        else:
-            GlobalConfig.market_maker_rate = \
-                ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE
 
-    def event_activate(self, **kwargs):
+    def _populate_selling_rate(self):
         if self.energy_rate_profile is not None:
-            self.energy_rate = read_and_convert_identity_profile_to_float(self.energy_rate_profile)
+            self.energy_rate = read_arbitrary_profile(InputProfileTypes.IDENTITY,
```
Hassan754, line 54: I'm wondering why we're dealing with energy_rate as a profile when there is a dedicated variable `energy_rate_profile`, what do you think ?

Hassan754 (approved):
Left one general question, LGTM

src/d3a/models/strategy/infinite_bus.py
faizan2590: First it was like that even before this PR, second from the external users' perspectve  its more clear to keep both separate but internally why not cast both to a single variable than keeping both?

PR #1168, opened 2021-06-22
D3ASIM-3461: Check test coverage compliance using Codecov

fievelk:
Add checks for test coverage.

- Compliance to test coverage will be performed using [Codecov](https://about.codecov.io/) through GitHub Actions. 
- We want checks to fail if the coverage of the proposed PRs is lower than the coverage of the `master` branch.

## Reference

- [Codecov docs](https://docs.codecov.com/docs)


fievelk:
Please let me know if you'd like me to change the content of the automatic comment (see [docs here](https://docs.codecov.com/docs/pull-request-comments)) or if you want me to completely remove it.

tox.ini
```diff
@@ -79,6 +79,7 @@ commands =
     flake8
     coverage run -m py.test {posargs:tests}
     coverage combine
+    coverage xml
```
fievelk, line 82: This is needed because Codecov requires a coverage output in XML format

Hassan754 (approved):
LGTM

.github/workflows/d3a_ci.yaml
```diff
@@ -22,3 +22,9 @@ jobs:
         export LD_LIBRARY_PATH=$HOME/solc/usr/lib:$LD_LIBRARY_PATH
         pip install tox
         tox -e $TOXENV --verbose
+
+    - name: Check coverage with Codecov
+      uses: codecov/codecov-action@v1
+      with:
+        fail_ci_if_error: true
```
hannesdiedrich, line 29: So If I add code and no unittest the coverage will be reduced and the CI will fail?
For you change it did not fail. This is because this is configuration and not code, right?

hannesdiedrich (approved):
One clarification, LGTM

.github/workflows/d3a_ci.yaml
fievelk: Yes, if you add untested lines the coverage should fail, in theory. It's possible to apply a % threshold to define how much the coverage can decrease without failing.

I think that this PR doesn't fail because it's not adding/removing lines in code that's "relevant" for the coverage. This is defined in the `.coveragerc` file: https://github.com/gridsingularity/d3a/blob/4247c16b859616a88ab226e263037c0c41b45d4b/.coveragerc

In any case, I think we will need to further tune the checks based on our needs in the future, but the important thing is that we start getting a bit more used to them.

fievelk: You can find some more info about the configuration (and its final status) here: https://docs.codecov.com/docs/commit-status

spyrostz (approved):
LGTM

PR #1171, opened 2021-06-23
D3ASIM-3356: Add  attributed bids and offers

Hassan754:
Hint while reviewing:
The test_market module doesn't really have any changes, I just moved all the two sided market tests to the test_two_sided_market module class , so it's enough to review the first two tests in the class (these are the only added tests)


src/d3a/models/area/__init__.py
```diff
@@ -211,8 +219,8 @@ def area_reconfigure_event(self, **kwargs):
                             export_capacity_kVA=export_capacity_kVA
                         )
 
-        except Exception as e:
-            log.error(str(e))
+        except Exception as ex:
+            log.error(str(ex))
```
fievelk, line 223: Just FYI: you can also avoid converting to string. This would be equivalent as far as I know:
```suggestion
        except Exception as ex:
            log.error(ex)
```

src/d3a/models/area/__init__.py
```diff
@@ -374,26 +380,33 @@ def publish_market_cycle_to_external_clients(self):
...
+        if (ConstSettings.IAASettings.MARKET_TYPE != SpotMarketTypeEnum.ONE_SIDED.value
+                and not self.strategy):
             if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
                 self.dispatcher.publish_market_clearing()
             else:
                 if is_external_matching_enabled():
-                    bid_offer_matcher.match_algorithm.area_uuid_markets_mapping.\
-                        update({self.uuid: self.all_markets})
+                    (bid_offer_matcher.match_algorithm.update_area_uuid_markets_mapping(
+                        {self.uuid: self.all_markets}))
```
fievelk, line 409: Nit:
```suggestion
                    bid_offer_matcher.match_algorithm.update_area_uuid_markets_mapping(
                        {self.uuid: self.all_markets})
```

src/d3a/models/area/__init__.py
```diff
@@ -414,6 +427,7 @@ def update_area_current_tick(self):
             child.update_area_current_tick()
 
     def tick_and_dispatch(self):
+        """Invokes tick handler and broadcast the event to children."""
```
fievelk, line 430: Nit: [PEP 257](https://www.python.org/dev/peps/pep-0257/) recommends using the imperative form for docstrings:
> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

So this would be 
```suggestion
        """Invoke tick handler and broadcast the event to children."""
```
Minor anyway.

src/d3a/models/market/balancing.py
```diff
@@ -51,13 +51,15 @@ def __init__(self, time_slot=None, bc=None, notification_listener=None, readonly
 
     def offer(self, price: float, energy: float, seller: str, offer_id=None,
               original_offer_price=None, dispatch_event=True, seller_origin=None,
-              adapt_price_with_fees=True, seller_origin_id=None):
+              adapt_price_with_fees=True, seller_origin_id=None,
+              attributes: dict = None, requirements: List[Dict] = None):
```
fievelk, line 55: Nit: as I pointed out in the other PR (on d3a-api-client), the bid/offer methods are using too many arguments. This means that we probably can/should figure out a way to refactor them and use a more OOP-friendly approach. The risk of keeping adding new arguments is that the complexity will increase, it'll be hard to extend and refactor the methods, and testing them will become a nightmare.

fievelk: For example, `seller`, `seller_origin` and `seller_origin_id` conceptually belong to the same entity (I would call it `Seller`). The rest of the parameters might belong to a `OfferParameter` class, but this is  just a quick idea.

src/d3a/models/area/__init__.py
hannesdiedrich: Thanks for this comment! I always wondered that and always did i wrong. Will change it from now on.

src/d3a/models/market/two_sided.py
```diff
@@ -19,21 +19,31 @@
...
 
 
 class TwoSidedMarket(OneSidedMarket):
+    """Extends One sided market class and adds support for bidding functionality.
+
+    A market type that allows producers to place energy offers to the markets
+    (exactly the same way as on the one-sided market case), but also allows the consumers
+    to place energy bids on their respective markets.
+    Contrary to the one sided market, where the offers are selected directly by the consumers,
+     the offers and bids are matched by the inter area agents.
```
hannesdiedrich, line 45: No, the offers and bids are not matched by the IAA. They are matched by the Mycos. The IAA are only respnsible for forwarding bids and offers and keeping track of those.

tests/test_market_redis_connection.py
```diff
@@ -208,7 +208,6 @@ def test_subscribes_to_market_channels(self):
                 "id/DELETE_BID": self.subscriber._delete_bid,
                 "id/ACCEPT_BID": self.subscriber._accept_bid,
                 "id/BID": self.subscriber._bid,
-                "id/CLEAR": self.subscriber._clear_market,
```
fievelk, line 211: I see you're removing this functionality. What was it doing and why is it not required anymore? Maybe I missed the discussion in the slack channel.

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +277,70 @@ def match_recommendation(self, recommended_list):
...
 
+    @staticmethod
+    def validate_requirements_satisfied(obj1: Union[Offer, Bid], obj2: Union[Offer, Bid]) -> None:
+        """Validate if both trade parties satisfy each other's requirements.
+
+        :raises:
+            InvalidBidOfferPair: Bid offer pair failed the validation
+        """
+        def _validate_requirements_satisfied(_obj1: Union[Offer, Bid], _obj2: Union[Offer, Bid]):
+            if _obj1.requirements:
```
hannesdiedrich, line 288: In order to reduce the number of indentions here I would rather return early here
```suggestion
            if not _obj1.requirements:
                return
```

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +277,70 @@ def match_recommendation(self, recommended_list):
...
+                            return
+                # If no requirement dict is satisfied
+                raise InvalidBidOfferPair
+
+        _validate_requirements_satisfied(obj1, obj2)
+        _validate_requirements_satisfied(obj2, obj1)
+
+    @classmethod
+    def validate_authentic_bid_offer_pair(
+            cls, bid: Bid, offer: Offer, clearing_rate: float, selected_energy: float) -> None:
```
fievelk, line 328: What is `selected_energy` here?

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +277,70 @@ def match_recommendation(self, recommended_list):
...
+                        if (key == "hashed_identity" and
+                                (not _obj2.attributes or
+                                 _obj2.attributes.get("hashed_identity") != value)):
+                            assert isinstance(value, str)
+                            is_satisfied = False
+                            continue
+
+                        if is_satisfied:
+                            """Requirement dict is satisfied."""
+                            return
```
hannesdiedrich, line 319: It is important that True is returned here, right?
```suggestion
                            return True
```

hannesdiedrich: Ignore me, understood now.

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +277,70 @@ def match_recommendation(self, recommended_list):
                     recommended_list, index+1, trade, bid_trade
                 )
 
+    @staticmethod
+    def validate_requirements_satisfied(obj1: Union[Offer, Bid], obj2: Union[Offer, Bid]) -> None:
```
fievelk, line 281: This logic seems overly complex to be in a function, and I think it would be a very good candidate for a separate class (e.g. `RequirementsValidator` or `RequirementsComplianceValidator`, or any better name).
Actually, the validation of each individual requirement could also deserve its own class instead of having so many conditionals. When requirements will increase this function will become almost impossible to expand and test.

We're using generic objects `obj1` and `obj2`, and I have the feeling that this is complicating the internal logic. Could you explain why this is implemented this way? Isn't it possible to distinguish between offers/bids? This would probably remove the need for some conditionals and `hasattr` checks.

src/d3a/models/market/two_sided.py
fievelk, line 319: This looks like a leftover from a docstring that was placed somewhere else. In any case, I think this might return `True` rather than `None`. What do you think?

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +277,70 @@ def match_recommendation(self, recommended_list):
...
+                                 _obj2.attributes.get("hashed_identity") != value)):
+                            assert isinstance(value, str)
+                            is_satisfied = False
+                            continue
+
+                        if is_satisfied:
+                            """Requirement dict is satisfied."""
+                            return
+                # If no requirement dict is satisfied
+                raise InvalidBidOfferPair
```
fievelk, line 321: If this fails, it means that the external Myco matcher has made some kind of mistake, right? It has returned a pair of bid/offer that it thinks is valid, but we know that it isn't. If that's the case, we might need to have some more information to allow debugging the issue.

What about adding some message here, to provide a better explanation? For example:
```suggestion
                raise InvalidBidOfferPair(f"Requirements are not satisfied for offer {offer} and bid {bid}")
```
Of course, the `{offer}` and `{bid}` parts might need to be changed. It should be something that allows us the Myco (or us) to understand what went wrong.

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -35,24 +35,32 @@ def publish_offers_bids(self, message):
         published data are of the following format
         market_offers_bids_list_mapping = {"market_id" : {"bids": [], "offers": [] }, }
         """
-        # TODO: message can contain filters
-        data = {"event": "offers_bids_response"}
+        response_data = {"event": "offers_bids_response"}
+        data = json.loads(message.get("data"))
+        filters = data.get("filters", {})
+        # IDs of markets the client is interested in
+        market_ids_in_interest = filters.get("markets", None)
```
fievelk, line 42: Nit: I find this name a bit difficult to interpret. Do you think that any of these would be a potential alternative?
- `relevant_market_ids`
- `considered_market_ids`
- `filtered_market_ids`
- `requested_market_ids`

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -35,24 +35,32 @@ def publish_offers_bids(self, message):
...
                 market_offers_bids_list_mapping[market.id] = {"bids": [], "offers": []}
                 bids, offers = market.open_bids_and_offers
                 market_offers_bids_list_mapping[market.id]["bids"].extend(
                     list(bid.serializable_dict() for bid in bids.values()))
                 market_offers_bids_list_mapping[market.id]["offers"].extend(
                     list(offer.serializable_dict() for offer in offers.values()))
-        data.update({
-            "market_offers_bids_list_mapping": market_offers_bids_list_mapping,
+        response_data.update({
+            "orders": market_offers_bids_list_mapping,
```
hannesdiedrich, line 59: I agree, that `market_offers_bids_list_mapping` could be rephrased, but `orders` is not very descriptive IMO. Is this a requirement or can we change it to something like: `offer_bid_lists_in_markets` ( I know, not better, but maybe somebody else has a better idea)?

src/d3a/models/myco_matcher/external_matcher.py
fievelk:
```suggestion
        market_ids_in_interest = filters.get("markets", [])
```
If you change the default to an empty list, you could simplify the conditional below:
```python
if area_uuid not in market_ids_in_interest:
```

src/d3a/models/myco_matcher/external_matcher.py
fievelk, line 59: I vaguely remember @spyrostz saying something about the "orders" keyword (I think he wasn't too convinced about using it). I might be wrong, but I'll just tag him in case I remember correctly :)

src/d3a/models/market/two_sided.py
```diff
@@ -277,6 +351,6 @@ def _convert_match_to_residual(match):
                 match = replace(match, bid=bid_trade.residual)
             return match
 
-        matchings[start_index:] = [_convert_match_to_residual(match)
-                                   for match in matchings[start_index:]]
-        return matchings
+        matches[start_index:] = [_convert_match_to_residual(match)
```
hannesdiedrich, line 354: I think we dont play with fire here, so please revert to `matchings` again. Thanks! :-D

tests/test_market.py
```diff
@@ -132,6 +133,59 @@ def market_matcher():
     return PayAsBidMatcher()
 
 
+class TestTwoSidedMarket:
+    """Class Responsible for testing two sided market's functionality.
+
+    TODO: bring all related two sided market tests from this module into this class.
+    """
+    def test_double_sided_validate_requirements_satisfied(self, market):
+        offer = Offer("id", now(), 2, 2, 'other', 2,
```
fievelk, line 142: Nit:
```suggestion
        offer = Offer("id", now(), 2, 2, "other", 2,
```

tests/test_market.py
```diff
@@ -132,6 +133,59 @@ def market_matcher():
...
+                      requirements=[{"preferred_trading_partner": ["bid_id2", ]}],
```
fievelk, line 143: Is there a reason why you prefer using the trailing comma in single-item lists?

tests/test_market.py
```diff
@@ -132,6 +133,59 @@ def market_matcher():
     return PayAsBidMatcher()
 
 
+class TestTwoSidedMarket:
+    """Class Responsible for testing two sided market's functionality.
+
+    TODO: bring all related two sided market tests from this module into this class.
```
hannesdiedrich, line 139: I think it is even time to create a separate module for this class even now, right?

tests/test_market.py
fievelk: The same applies to the rest of the PR (make sure to run the linter to catch these)

hannesdiedrich:
I have a general comment/question: AFAIS, currently, it is not possible to post a bid or offer with attributes or requirements for example via the external API (please point we to it, if I missed it).
Is this targeted in a later ticket?

src/d3a/models/market/two_sided.py
Hassan754: I also thought this is wrong, I was expecting a discussion on this here,  anw this is taken from [our wiki](https://gridsingularity.atlassian.net/wiki/spaces/D3A/pages/643497985/Two+Sided+Market) so I suppose it needs to be fixed

hannesdiedrich:
> I have a general comment/question: AFAIS, currently, it is not possible to post a bid or offer with attributes or requirements for example via the external API (please point we to it, if I missed it).
> Is this targeted in a later ticket?

forget it, I just saw the d3a-api-client PR. Thanks!

tests/test_market_redis_connection.py
Hassan754: Yes, the only logic inside this _clear_market function is a call to a deprecated `match_offers_bids` which doesn't carries implementation
I removed the `match_offers_bids` function and this wrapper one too 
[you can check the conversation](https://gridsingularity.slack.com/archives/C01G89NENQZ/p1624285347002000)

src/d3a/models/market/two_sided.py
hannesdiedrich: Thanks for putting my doubts into proper words! I second that.

hannesdiedrich:
> > I have a general comment/question: AFAIS, currently, it is not possible to post a bid or offer with attributes or requirements for example via the external API (please point we to it, if I missed it).
> > Is this targeted in a later ticket?
> 
> forget it, I just saw the d3a-api-client PR. Thanks!

Sorry, I just had another look. 
Wouldn't adding an `attribute` key to the bid sent by the d3a-api-client result in an error here:
https://github.com/gridsingularity/d3a/blob/4247c16b859616a88ab226e263037c0c41b45d4b/src/d3a/models/strategy/external_strategies/load.py#L146

Did you do an e2e test?

src/d3a/models/market/two_sided.py
Hassan754: when matching offers and bids, one of them can have more energy so we select the minimum energy value and set it in a match
The rest of energy are left to a residual offer | bid

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +277,70 @@ def match_recommendation(self, recommended_list):
...
+        :raises:
+            InvalidBidOfferPair: Bid offer pair failed the validation
+        """
+        def _validate_requirements_satisfied(_obj1: Union[Offer, Bid], _obj2: Union[Offer, Bid]):
+            if _obj1.requirements:
+                for requirement in _obj1.requirements:
+                    is_satisfied = True
+                    for key, value in requirement.items():
+                        if (key == "preferred_trading_partner" and
+                                not ((hasattr(_obj2, "buyer_origin_id") and
```
fievelk, line 293: Why is it necessary to check both `buyer_origin_id` and `buyer_id`? (Same for `seller_*`)

src/d3a/models/market/two_sided.py
Hassan754: I agree, I thought about this but then ignored it while implementing and focused on making things work, anw I'll create something that doesn't break the open for extension closed for modification principle

src/d3a/models/market/two_sided.py
Hassan754: I'm not interested in returning anything actually, this is just to exit the function without raising any exceptions (meaning that validation passed)
Anw this is not relevant anymore as I'll group the validation into a dedicated class

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +277,70 @@ def match_recommendation(self, recommended_list):
...
+                                      _obj2.buyer_origin_id in value) or
+                                     (hasattr(_obj2, "buyer_id") and
+                                      _obj2.buyer_id in value) or
+                                     (hasattr(_obj2, "seller_origin_id") and
+                                      _obj2.seller_origin_id in value) or
+                                     (hasattr(_obj2, "seller_id") and
+                                      _obj2.seller_id in value))):
+                            assert isinstance(value, list)
```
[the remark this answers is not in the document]
hannesdiedrich: I see, but this assertions should be called in any case. I mean also when only `key == "preferred_trading_partner"` right?

tests/test_market.py
fievelk, line 139: Is this a TODO for this PR?

fievelk (commented):
Left some comments. The main point that I think should be discussed together is the big requirements validation function in the two-sided market.

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: Actually this can break the logic 
I do need to check the if filtered_market_ids, in case the client did not pass it or it's empty, I don't want to do the filter (In this case nothing will be returned whereas this is supposed to return all the records)

tests/test_market_redis_connection.py
fievelk: Great, thanks!

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: I agree, I'm open to suggestions here

src/d3a/models/market/two_sided.py
Hassan754: We are referring to the plural of match IIUC so it should be matches not matchings (matching is the verb)

tests/test_market.py
fievelk: That would be good in my opinion too

Hassan754: I agree, but grouping all two sided market tests from this module can take long time and this is PR cannot wait as the stakeholders (research team) is regularly asking for updates and deadlines
should I create a new module and leave the TODO comment in this module ?

Hassan754:
> > > I have a general comment/question: AFAIS, currently, it is not possible to post a bid or offer with attributes or requirements for example via the external API (please point we to it, if I missed it).
> > > Is this targeted in a later ticket?
> > 
> > 
> > forget it, I just saw the d3a-api-client PR. Thanks!
> 
> Sorry, I just had another look.
> Wouldn't adding an `attribute` key to the bid sent by the d3a-api-client result in an error here:
> https://github.com/gridsingularity/d3a/blob/4247c16b859616a88ab226e263037c0c41b45d4b/src/d3a/models/strategy/external_strategies/load.py#L146
> 
> Did you do an e2e test?

Thanks for catching this !!
I'm currently investigating the failing ci tests in the api-client and doing some e2e tests (They are taking long time because I'm running the simulation in slow time)

tests/test_market.py
Hassan754: Nope I did not plan to move these tests into the class in the context of this PR, however I'll give it another look and may move them if the won't take long time

hannesdiedrich: IMO yes, you can proceed like this.

hannesdiedrich (commented):
Left some comments, happy to review again.

src/d3a/models/market/two_sided.py
Hassan754: By `preferred_trading_partner` we could mean the seller | buyer
These, in their turns can have their offers or bids forwarded from a market to another by the IAA, in this case the IAA is the buyer|seller and thus we can't rely on this field, but the origin buyer | origin seller which are the real contributors

fievelk: But one of the two fields should always point to the original seller, right? Couldn't we just always check that field?

src/d3a/models/myco_matcher/external_matcher.py
spyrostz: Thanks for the ping, I still stand by my statement that bids_offers would be a much better term than the dubious "orders" (not every person is familiar with the term 'order book')

Hassan754:
IMO a unit test for the MycoExternalMatcher is needed, I'm planning on adding it but not in the context of this PR, straight after pushing it if everyone agrees

src/d3a/models/market/market_validators.py
```diff
@@ -0,0 +1,106 @@
...
+    @classmethod
+    def is_satisfied(cls, offer: Offer, bid: Bid, requirement: Dict) -> bool:
+        bid_required_hashed_identity = requirement.get("hashed_identity", "")
+        assert isinstance(bid_required_hashed_identity, str), \
+            f"Invalid requirement {requirement}"
+        offer_hashed_identity = (
+                offer.attributes or {}).get("hashed_identity", None)
+        if bid_required_hashed_identity and offer_hashed_identity != bid_required_hashed_identity:
+            return False
+        return True
```
hannesdiedrich, line 62: Double negations are always harder to understand IMO. How about:
```suggestion
        return True if offer_hashed_identity == bid_required_hashed_identity else False
```

hannesdiedrich: Same for the other Requirement classes.

src/d3a/models/market/market_validators.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+# Supported offers/bids requirements
+# To add a new requirement create a class extending the Requirement abstract class and add it below
```
hannesdiedrich, line 66: ❤️

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +278,36 @@ def match_recommendation(self, recommended_list):
...
                 )
 
+    @staticmethod
+    def validate_requirements_satisfied(offer: Offer, bid: Bid) -> None:
+        """Validate if both trade parties satisfy each other's requirements.
+
+        :raises:
+            InvalidBidOfferPair: Bid offer pair failed the validation
+        """
+        if ((offer.requirements or bid.requirements) and
```
hannesdiedrich, line 290: You mean
```suggestion
        if ( not (offer.requirements or bid.requirements) or
```
Right?

src/d3a/models/market/two_sided.py
hannesdiedrich: You are right. First translation that popped up for `matches` was the small wooded thing that you light a fire with. Sorry.

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,178 @@
...
         self.area_uuid_markets_mapping = {}
-        self.markets_mapping = {}
+        self.markets_mapping = {}  # Dict[market_id: market] mapping
         self.recommendations = []
 
     def _setup_redis_connection(self):
         self.myco_ext_conn = ResettableCommunicator()
         self.myco_ext_conn.sub_to_multiple_channels(
-            {"external-myco/get-simulation-id": self.get_simulation_id,
+            {"external-myco/get-simulation-id": self.publish_simulation_id,
```
hannesdiedrich, line 32: IMO `get_simulation_id` is better even for the callback method (i was confused at first)

src/d3a/models/market/two_sided.py
Hassan754: Not really, actually I'm checking if at least the bid or offer have requirements, if True then we'll move to the second check which is to see if these requirements are satisfied
So if both offer and bid have no requirements, the statement will be False and it won't raise an exception

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: I'm trying to homogenize the callbacks, the ones that publish to redis are appended with `publish`

hannesdiedrich: I got that, but still. Lets wait for other opinions.

src/d3a/models/market/two_sided.py
hannesdiedrich: Yes, you are totally right!

tests/test_two_sided_market.py
```diff
@@ -0,0 +1,288 @@
...
+            # should raise an exception as hashed_identity requirement of offer isn"t satisfied
+            market.validate_requirements_satisfied(offer, bid)
+        offer.attributes["hashed_identity"] = "AreaX"
+        market.validate_requirements_satisfied(offer, bid)
+
+    @pytest.mark.parametrize(
+        "bid_energy, offer_energy, clearing_rate, selected_energy", [
+            (2, 3, 2, 2.5),
+            (3, 2, 2, 2.5),
+            (2, 2, 2 / 2 + 1, 2),
```
hannesdiedrich, line 69: Any reason to have this calculation here?
```suggestion
            (2, 2, 2 , 2),
```

hannesdiedrich: I see that this is copy paste code, Sorry. But i guess you can change it anyways, right?

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,183 @@
...
+        for area_id, market_slots in self.area_uuid_markets_mapping.items():
+            if filtered_area_ids and area_id not in filtered_area_ids:
+                # Client is uninterested in this Area -> skip
+                continue
+            for market_slot in market_slots:
+                # Cache the market (needed while matching)
+                self.markets_mapping[market_slot.id] = market_slot
+                bids_list, offers_list = self._get_bids_offers(market_slot, filters)
+                market_offers_bids_list_mapping[market_slot.id] = {
+                    "bids": bids_list, "offers": offers_list}
```
hannesdiedrich, line 57: I vote to rename `market_slots` into simple `markets` and `market_slot` into `market`.
It is kind of a best practice in the d3a code. `market_slot` usually only refers to the time of the market. Thanks!

src/d3a/models/myco_matcher/external_matcher.py
hannesdiedrich: I dont have any and vote to keep `market_offers_bids_list_mapping ` as you changed it back.

hannesdiedrich (commented):
Some cosmetic changes. LGTM in general.

src/d3a/models/area/__init__.py
```diff
@@ -374,26 +380,33 @@ def publish_market_cycle_to_external_clients(self):
...
+            (self.strategy.redis.aggregator.
+             consume_all_area_commands(self.uuid,
+                                       self.strategy.trigger_aggregator_commands))
 
     def tick(self):
+        """Tick event handler.
+
+        invokes aggregator commands consumer, publishes market clearing, updates events,
+         updates cached market's bids and offers in case of myco matching and matches
+         bid offer pairs otherwise.
```
faizan2590, line 398: Seems like out of indent.

src/d3a/models/market/market_validators.py
```diff
@@ -0,0 +1,105 @@
+"""Validators for d3a markets."""
```
faizan2590, line 1:
```suggestion
"""
Copyright 2018 Grid Singularity
This file is part of D3A.
This program is free software: you can redistribute it and/or modify
it under the terms of the GNU General Public License as published by
the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.
This program is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU General Public License for more details.
You should have received a copy of the GNU General Public License
along with this program.  If not, see <http://www.gnu.org/licenses/>.
"""

"""Validators for d3a markets."""
```

src/d3a/models/market/market_validators.py
```diff
@@ -0,0 +1,105 @@
...
+from d3a.models.market import Offer, Bid
+
+
+class Requirement(ABC):
+    """Define an offer/bid requirement with is_satisfied interface."""
+
+    @classmethod
+    @abstractmethod
+    def is_satisfied(cls, offer: Offer, bid: Bid, requirement: Dict) -> bool:
+        """Check whether a requirement is satisfied."""
```
faizan2590, line 15:
```suggestion
        """Check whether a requirement is satisfied."""
        pass
```
Correct me if I am wrong, as far as I know one needs to add `pass` for methods without any implementation.

src/d3a/models/market/market_validators.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+class PreferredPartnersRequirement(Requirement):
+    """Check if preferred_trading_partners requirement is satisfied for both bid and offer."""
+
+    @classmethod
+    def is_satisfied(cls, offer: Offer, bid: Bid, requirement: Dict) -> bool:
+        preferred_trading_partners = requirement.get("preferred_trading_partners", [])
+        assert isinstance(preferred_trading_partners, list),\
+            f"Invalid requirement {requirement}"
```
faizan2590, line 25:
```suggestion
            f"Invalid data type for preferred_trading_partners {requirement}"
```

src/d3a/models/market/market_validators.py
```diff
@@ -0,0 +1,105 @@
...
+class RequirementsSatisfiedChecker:
+    """Check if a list of bid/offer requirements are satisfied."""
+
+    @classmethod
+    def is_satisfied(cls, offer: Offer, bid: Bid) -> bool:
+        """Receive offer and bid and validate that at least 1 requirement is satisfied."""
+        offer_requirements = offer.requirements or []
+        bid_requirements = bid.requirements or []
+
+        # If a bid of offer has no requirements, consider it as satisfied.
```
faizan2590, line 87:
```suggestion
        # If bid and offer has no requirements, consider it as satisfied.
```

src/d3a/models/market/two_sided.py
```diff
@@ -19,21 +19,33 @@
...
 
 
 class TwoSidedMarket(OneSidedMarket):
+    """Extends One sided market class and adds support for bidding functionality.
+
+    A market type that allows producers to place energy offers to the markets
+    (exactly the same way as on the one-sided market case), but also allows the consumers
+    to place energy bids on their respective markets.
+    Contrary to the one sided market, where the offers are selected directly by the consumers,
+     the offers and bids are matched by the myco matcher.
```
faizan2590, line 47:
```suggestion
     the offers and bids are being matched via some matching algo.
```
Myco is an extension of two sided market, there are already 2 traditional TwoSidedMarket based matching algos (pay-as-bid/pay-as-clear)

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +280,36 @@ def match_recommendation(self, recommended_list):
...
+    def validate_requirements_satisfied(offer: Offer, bid: Bid) -> None:
+        """Validate if both trade parties satisfy each other's requirements.
+
+        :raises:
+            InvalidBidOfferPair: Bid offer pair failed the validation
+        """
+        if ((offer.requirements or bid.requirements) and
+                not RequirementsSatisfiedChecker.is_satisfied(offer, bid)):
+            # If no requirement dict is satisfied
+            raise InvalidBidOfferPair
```
faizan2590, line 293:
```suggestion
            raise InvalidBidOfferPair(f" OFFER: {offer} & BID: {bid} pair failed the validation")
```

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -372,14 +376,14 @@ def event_market_cycle(self):
...
+                market_info["device_info"] = self._device_info_dict
                 market_info["event"] = "market"
-                market_info['device_bill'] = self.device.stats.aggregated_stats["bills"] \
+                market_info["device_bill"] = self.device.stats.aggregated_stats["bills"] \
                     if "bills" in self.device.stats.aggregated_stats else None
                 market_info["area_uuid"] = self.device.uuid
                 market_info["last_market_maker_rate"] = \
                     get_market_maker_rate_from_config(self.area.current_market)
-                market_info['last_market_stats'] = \
+                market_info["last_market_stats"] = \
```
faizan2590, line 386: Please skip backslash used for line continuation from parts of code being touched by this PR.

src/d3a/models/market/market_validators.py
Hassan754: Adding docstring works as a place holder so no need for the pass statement

src/d3a/models/market/market_validators.py
```diff
@@ -1,5 +1,19 @@
...
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+Validators for d3a markets.
```
hannesdiedrich, line 15: IMO this can be removed. The module name tells that already. Also, the Licence string should not be modified. Thanks!

src/d3a/models/myco_matcher/__init__.py
```diff
@@ -9,19 +9,32 @@
...
-            self.match_algorithm = ExternalMatcher()
-        else:
-            raise WrongMarketTypeException("Wrong market type setting flag "
-                                           f"{ConstSettings.IAASettings.MARKET_TYPE}")
+    def __init__(self):
+        self.match_algorithm = None
+
+    def reconfigure(self):
+        """Reconfigure the myco matcher properties at runtime."""
+        self.match_algorithm = self.get_matcher_algorithm()
```
hannesdiedrich, line 17: IMO this is counter intuitive. `self.get_matcher_algorithm()` should also be called when the class is activated. I understand that it does not work when the class is initialised, but I would propose to stickt with the naming scheme that we have for the 2 stage initialisation on d3a and call this method `activate`. This will make the call on simulation.py more understandable IMO.
If we really need a `reconfigure` function in the future, then IMO, it should stay here and the `activate` method should be added. I know they would do the same for now, but maybe not in the future. What do you think?

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -36,10 +36,10 @@ def __init__(self, *args, **kwargs):
 
     @property
     def filtered_bids_next_market(self) -> List[Dict]:
-        """Get a representation of each of the device's bids from the next market."""
+        """Get a representation of each of the device"s bids from the next market."""
```
hannesdiedrich, line 39: Nit: replacing globally does not always work unfortunately :-)

hannesdiedrich (commented):
Left a few comments. In general looks good! Will review again, once the oder d3a PR is merged into here. Thanks!

src/d3a/models/myco_matcher/__init__.py
Hassan754: You are right, at the moment AFAIU the user cannot change the matching algorithm while a simulation is running, so this should be renamed to `activate`, and if later on we add the ability to change the algorithm we can add the reconfigure method

src/d3a/models/area/__init__.py
```diff
@@ -374,26 +380,33 @@ def publish_market_cycle_to_external_clients(self):
...
                 self.dispatcher.publish_market_clearing()
             else:
                 if is_external_matching_enabled():
-                    bid_offer_matcher.match_algorithm.area_uuid_markets_mapping.\
-                        update({self.uuid: self.all_markets})
+                    bid_offer_matcher.match_algorithm.update_area_uuid_markets_mapping(
+                        {self.uuid: self.all_markets})
                 else:
                     for market in self.all_markets:
                         bid_offer_pairs = bid_offer_matcher.calculate_recommendation(
```
fievelk, line 412: This comment is probably hardly addressed in this PR, since it's a bit broad. However, it is not a "Nit" comment, in the sense that I believe it is quite important for the future.

I think that there's something off with this part (conceptually).
We implemented a `MycoMatcher` class that should act as an interface for all different types of matching. This means that we should call the same methods for all different types of `MycoMatcher`.
However, we're not using it in that way: here we are differentiating between the `MycoMatcher`s whose `match_algorithm` is external, and the other `MycoMatcher`s. In fact, calling this line:
```python
bid_offer_matcher.match_algorithm.update_area_uuid_markets_mapping
```
only works for `ExternalMatcher` instances, and will break for `PayAsClearMatcher` or `PayAsBidMatcher`. I think this defeats the purpose of using interfaces (the `BaseMatcher`).

Conceptually, we should not directly call methods on the `match_algorithm` here. We should have some interface in `bid_offer_matcher` (the `MycoMatcher`) that deals with it, so that we can remove the `if is_external_matching_enabled()` check from here.

fievelk: To be a bit clearer, what I would expect is (as a first step) something like this:
```python
def tick(self):
    self._consume_commands_from_aggregator()

    if (ConstSettings.IAASettings.MARKET_TYPE != SpotMarketTypeEnum.ONE_SIDED.value
            and not self.strategy):
        if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
        else:
        	bid_offer_matcher.my_interface_method()
```

src/d3a/models/market/market_validators.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+from abc import ABC, abstractmethod
+from typing import Dict
+
+from d3a.models.market import Offer, Bid
+
+
+class Requirement(ABC):
+    """Define an offer/bid requirement with is_satisfied interface."""
```
fievelk, line 24: Nit: what about
```suggestion
    """Interface for offer and bid requirements."""
```

src/d3a/models/area/__init__.py
Hassan754: @fievelk I see your points, however I think there is a confusion here, we should distinguish between a matcher and a matching algorithm
I think this is the root issue, kindly note that there is no BaseMatcher anymore in the d3a, there is a BaseMatchingAlgorithm in the d3a-interface, this is the `interface` for all matching algorithms and they will all have the same interface `get_matches_recommendations`
Most probably we need to rename either `MycoMatcher` or `ExternalMatcher`, the matcher suffix should not be on both
Unfortunately we cannot homogenize anymore here because the external matcher is the one that requests offers and bids , we cannot explicitly call `match_recommendations` on it, he is the one to fire events unlike our default setup
Please note that the ExternalMatcher could make use of PayAsBidMatchingAlgorithm, PayAsClearMatchingAlgorithm, AttributedMatchingAlgorithm or any other client defined algorithm
I agree about the last point (my_interface_method), I'll move it to the MycoMatcher, and btw we should pass data as well so if external matching is not enabled we should match on the go.

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +280,40 @@ def match_recommendation(self, recommended_list):
                     recommended_list, index+1, trade, bid_trade
                 )
 
+    @staticmethod
+    def validate_requirements_satisfied(
```
fievelk, line 284: If this method is used only inside of this class, can we consider making it private?

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +280,40 @@ def match_recommendation(self, recommended_list):
...
+        """
+        if ((offer.requirements or bid.requirements) and
+                not RequirementsSatisfiedChecker.is_satisfied(
+                    offer, bid, clearing_rate, selected_energy)):
+            # If no requirement dict is satisfied
+            raise InvalidBidOfferPairException(
+                f"OFFER: {offer} & BID: {bid} requirements failed the validation.")
+
+    @classmethod
+    def validate_authentic_bid_offer_pair(
```
fievelk, line 300: Nit: what do we mean by "authentic" here? For some reason every time I find this method I have a brain freeze, so maybe there's a more descriptive term we could use?

fievelk: BTW, I would add the description of the arguments in the docstring of this method. I would specifically describe what "selected_energy" and "clearing_rate" are (I think we have discussed this in a previous PR).

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +280,40 @@ def match_recommendation(self, recommended_list):
...
+            cls, bid: Bid, offer: Offer, clearing_rate: float, selected_energy: float) -> None:
+        """Basic validation function for bid against an offer.
+
+        :raises:
```
fievelk, line 304:
```suggestion
        Raises:
```

src/d3a/models/market/two_sided.py
```diff
@@ -264,9 +280,40 @@ def match_recommendation(self, recommended_list):
...
+            InvalidBidOfferPairException: Bid offer pair failed the validation
+        """
+        if not (bid.energy >= selected_energy and
+                offer.energy >= selected_energy and
+                (bid.energy_rate + FLOATING_POINT_TOLERANCE) >= clearing_rate and
+                (bid.energy_rate + FLOATING_POINT_TOLERANCE) >= offer.energy_rate):
+            raise InvalidBidOfferPairException
+        cls.validate_requirements_satisfied(bid, offer, clearing_rate, selected_energy)
```
fievelk, line 312: The arguments are in the wrong order here: you're passing the bid instead of the offer.
```suggestion
        cls.validate_requirements_satisfied(offer, bid, clearing_rate, selected_energy)
```
To make it clearer, please use named arguments.

BTW, this was spotted by pylint. Please make sure you run it again on all files before asking for reviews, so that we can proceed faster on this kind of warnings :)

fievelk: Also: if this didn't fail in the tests, it probably means that there's not enough coverage. It would be great if you could add tests for this part.

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
 class ExternalMatcher(BaseMatcher):
+    """Class responsible for external bids / offers matching."""
     def __init__(self):
         super().__init__()
         self.simulation_id = d3a.constants.COLLABORATION_ID
         self.myco_ext_conn = None
-        self.channel_prefix = f"external-myco/{self.simulation_id}/"
+        self.channel_prefix = f"external-myco/{self.simulation_id}"
+        self.response_channel = f"{self.channel_prefix}/response"
+        self.events_channel = f"{self.response_channel}/events/"
```
fievelk, line 24: It seems like most of these variables are just internals and they should not be changed nor accessed from the outside. If this is the case, I would suggest to name them as "private" (e.g. `self._events_channel`) to make it clear.

src/d3a/models/market/two_sided.py
Hassan754: I'm not sure, this was here before I touch the code, I also think that it can be renamed to validate_valid_bid_offer_pair for example, what do you think ?

src/d3a/models/market/two_sided.py
Hassan754: I'm building the habit of doing that progressively, although sometimes I introduce a `small should not harm` changes without rerunning pylint 😄

Hassan754: and FYI I'm adapting in the other PR as I  thought it could make it easier for me and the reviewer

src/d3a/models/market/market_validators.py
Hassan754: I'm thinking about making the is_satisfied function accepts `arguments` key instead of (offer, bid,
            clearing_rate, selected_energy), what do you think?

src/d3a/models/myco_matcher/external_matcher.py
fievelk: Actually, I'm thinking that using `get-*` or `publish-*` in channel names is not the best thing. The reason is that you always have two sides of the communication that do opposite things (pub/sub) so if you use `get/publish/subscribe` in the channel name you're only covering one part of the story.
I would personally use something like `external-myco/simulation-id`

fievelk: The same applies to `post-recommendations`: only one side is posting them, while the other side is receiving them.

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
+            {"external-myco/get-simulation-id": self.publish_simulation_id,
+             f"{self.channel_prefix}/offers-bids/": self.publish_offers_bids,
+             f"{self.channel_prefix}/post-recommendations/": self.match_recommendations})
 
     def publish_offers_bids(self, message):
         """Publish open offers and bids.
 
-        published data are of the following format
-        market_offers_bids_list_mapping = {"market_id" : {"bids": [], "offers": [] }, }
+        published data are of the following format:
```
fievelk, line 40:
```suggestion
        Published data are of the following format:
```

src/d3a/models/market/two_sided.py
Hassan754: One can set the market of an asset as a preferred trading agent
Let's say your house_id house_ppp, you created an offer which was forwarded to the city higher market id : ppp_city , I can set my preferred trading agent to be `ppp_city` or from `house_ppp`, so both make sense

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
+            {"bids_offers": {`market_id` : {"bids": [], "offers": [] }, }}
         """
-        # TODO: message can contain filters
-        data = {"event": "offers_bids_response"}
+        response_data = {"event": "offers_bids_response"}
+        data = json.loads(message.get("data"))
+        filters = data.get("filters", {})
```
fievelk, line 45: If the `myco-api-client` sends a wrong request (e.g. it has a typo in the `"filters"` key) it won't receive any meaningful error notification. I think we should validate this data somehow on the external matcher side. Not necessarily in this PR, but I think it's quite important since the external matchers will be implemented from external agents.

fievelk: Also, I think it would be good to describe the structure of the expected `message` in the docstring.

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
         """Publish open offers and bids.
 
-        published data are of the following format
-        market_offers_bids_list_mapping = {"market_id" : {"bids": [], "offers": [] }, }
+        published data are of the following format:
+            {"bids_offers": {`market_id` : {"bids": [], "offers": [] }, }}
         """
-        # TODO: message can contain filters
-        data = {"event": "offers_bids_response"}
+        response_data = {"event": "offers_bids_response"}
```
fievelk, line 43: Here we are using strings to describe the events here, rather than enums as we usually do with market events. Is there a specific reason for this?

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
+        data = json.loads(message.get("data"))
+        filters = data.get("filters", {})
+        # IDs of markets (Areas) the client is interested in
+        filtered_market_ids = filters.get("markets", None)
```
fievelk, line 47:
```suggestion
        filtered_market_ids = filters.get("markets")
```

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
         market_offers_bids_list_mapping = {}
-        for area_uuid, markets in self.area_uuid_markets_mapping.items():
+        for area_id, markets in self.area_uuid_markets_mapping.items():
+            if filtered_market_ids and area_id not in filtered_market_ids:
```
fievelk, line 50: Regarding naming: reading `area_id not in filtered_market_ids` feels strange to me. I would expect to find market IDs in the `filtered_market_ids` variable.
Maybe we could use a name like `areas_to_market_ids`? 

On a different note: why are we using `_id` and not `_uuid`?

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
+        bids_list = list(bid.serializable_dict() for bid in bids.values())
+        filtered_offers_energy_type = filters.get("energy_type", None)
+        if filtered_offers_energy_type:
+            offers_list = list(
+                offer.serializable_dict() for offer in offers.values()
+                if offer.attributes and
+                offer.attributes.get("energy_type") == filtered_offers_energy_type)
+        else:
+            offers_list = list(
+                offer.serializable_dict() for offer in offers.values())
```
fievelk, line 142: Sorry if I'm getting a bit confused at this point.
1. The myco-api-client requests open bids and offers from d3a.
2. d3a answers with a list of open bids/offers

Now we want to add some filtering. However, shouldn't this filtering be done in the matching algorithm? In theory, `d3a` shouldn't know anything about filtering: it should just validate that the matches between bids and offer are valid. Right?

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
 
 class ExternalMatcher(BaseMatcher):
+    """Class responsible for external bids / offers matching."""
     def __init__(self):
         super().__init__()
         self.simulation_id = d3a.constants.COLLABORATION_ID
         self.myco_ext_conn = None
-        self.channel_prefix = f"external-myco/{self.simulation_id}/"
+        self.channel_prefix = f"external-myco/{self.simulation_id}"
+        self.response_channel = f"{self.channel_prefix}/response"
```
fievelk, line 23: At the moment we're structuring the channels as:
```
- external-myco/<simulation-id>/
  - offers-bids
  - response/
    - offers-bids

```

What would you think of a different structure, like this:
```
- external-myco/<simulation-id>/
  - offers-bids
    - response/
```
It would be more natural in my opinion and less "branched".

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
+        try:
+            validated_records = self._get_validated_bid_offer_match_list(recommendations)
             for market_id, records in validated_records.items():
                 market = self.markets_mapping.get(market_id)
-                if market.readonly:
-                    # The market has just finished
+                try:
+                    market.match_recommendation(records)
+                except (OfferNotFoundException, BidNotFoundException):
+                    # If the offer or bid have just been consumed
```
fievelk, line 82: When can this happen? It might be worth adding some more explanation to the comment. Also, we're losing any information about this scenario (since we're not logging the exception, but just continuing with the loop). This can be fine, but I'd just like to be sure that it's desired.

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
                     continue
-                market.match_recommendation(records)
-        self.myco_ext_conn.publish_json(channel, response_dict)
+        except Exception as ex:
```
fievelk, line 84: Do we have any way to expect what kind of exception we'll have? I'm asking because of the too broad exception.

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
+            response_data["status"] = "fail"
+            response_data["message"] = "Validation Error"
+            if not isinstance(ex, MycoValidationException):
+                logging.exception("Bid offer pair matching failed.")
```
fievelk, line 88: This message should help us to understand what went wrong. Is it possible to add some more information to help our future selves? Or would any information be useless at this point?

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
 
-    def get_simulation_id(self, message):
-        """Publish the simulation id to the Myco client."""
+    def publish_simulation_id(self, message):
+        """Publish the simulation id to the redis myco client.
+
+        At the moment the id of the simulations run by the cli is set as ""
+        however, this function guarantees that the myco is aware of the running collaboration id
+        regardless of the value set in d3a.
+        """
```
fievelk, line 99: Nit: I think the whitespace is not needed here

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,179 @@
...
     def publish_event_tick_myco(self):
         """Publish the tick event to the Myco client."""
 
-        channel = f"external-myco/{d3a.constants.COLLABORATION_ID}/response/events/"
         data = {"event": "tick"}
-        self.myco_ext_conn.publish_json(channel, data)
+        self.myco_ext_conn.publish_json(self.events_channel, data)
 
-    def publish_market_cycle_myco(self):
+    def publish_event_market_myco(self):
```
fievelk, line 109: What's a market event? Why has it been renamed from the previous "market_cycle"?

fievelk (commented):
Left some comments. It's quite a big PR so I hope I didn't miss important stuff.

src/d3a/models/area/__init__.py
fievelk: > Unfortunately we cannot homogenize anymore here because the external matcher is the one that requests offers and bids , we cannot explicitly call match_recommendations on it, he is the one to fire events unlike our default setup

I think this is exactly the issue here. The `bid_offer_matcher` (`MycoMatcher`) can contain components whose behaviour is incompatible (I'm talking in terms of interface). That's why you need that `if` in front of it: you can't call the same interface method because it won't work for all components. I think this points out the fact that something could be changed in the structure of the classes. In some cases the `MycoMatcher` is just contacting an API, while in some other cases it's actually computing recommendations (and this create an incoherent structure IMO).

src/d3a/models/market/two_sided.py
fievelk: Or `validate_bid_offer_match`? Not sure though, you definitely have a clearer view of the context.

src/d3a/models/market/market_validators.py
fievelk: I think that if we don't need it, it's better to explicitly name all the arguments for the moment. If they become too many, it means that we need to create a specific class for them (e.g. `BidArguments` or something).

src/d3a/models/area/__init__.py
Hassan754: @fievelk why are you considering the Myco Matcher as an interface ?
this is the class that is aware of the different matching algorithms and their different interfaces (the externalMatcher vs all other matching algorithms)
this can be solved when we (if ever) be in charge of sending the open offers/bids to the myco and wait for its recommendations, however I suggested this to Colin and he seemed positive but I still believe that this is to be thought more of.

src/d3a/models/market/two_sided.py
Hassan754: Ah my suggested name has valid twice, you are right

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: we are not enforcing filters atm and I see no point in doing this, this is why  I did not add any validation here.

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: Nope, an enum is better than having strings into our code, I had to group them since test1, I'll do it in the context of this PR

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: I absolutely have no idea why I renamed it to area_id, I faintly recall this was discussed somewhere but I'm not sure
I'll fix this

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: image-20210423-190639
The requirement states that myco can request and filter orders and then it would pass them to the matching algorithm, I can see some business goals behind this
anw this also favors the communication part, we only send the minimum size of needed data

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: I made it this way for the sake of this subscription `f"{self.redis_channels_prefix}/response/*": self._on_event_or_response`, anw no harm of adding a wildcard before the response, will do it

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: This can happen in several cases, for example

1. an offer or bid can be deleted
2. myco could send the same offer or bid in several recommendations, one could be matched so the next time we encounter this offer this error will be raised

This is AFAIS, there could be other cases too, I'll update the comment accordingly

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: Of course I can think of some exceptions that can be raised, however I left it as broad as this is in the early stages,  and matching can fail for whatever reason, I would log the error and gradually start excepting the different kinds of exceptions

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: atm this statement `if market is None or market.readonly`  is the only place where this exception is raised
anw it is too late to display something at this stage, so I will add some message when raising the exception

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: I was trying to homogenize the handlers names, publish_tick_myco, publish_finish_myco...
In the myco and api-client, they are aware of the market event, not the market cycle, so I named it here like that
However I suppose that it should be called the same as we call the event in d3a, so I will restore it

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: And the expected message structure is already described in the docstring,  it is not displayed by github as this was not changed, it was there since Test1 V2, anw I'll add the filters key.

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: this is indeed late here, however the exception traceback should be enough

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: Which whitespace?

src/d3a/models/area/__init__.py
```diff
@@ -225,9 +233,8 @@ def _update_descendants_strategy_prices(self):
                 self.strategy.event_activate_price()
             for child in self.children:
                 child._update_descendants_strategy_prices()
-        except Exception as e:
-            log.error(f"area._update_descendants_strategy_prices failed. Exception: {e}. "
-                      f"Traceback: {traceback.format_exc()}")
+        except Exception:
+            log.exception("area._update_descendants_strategy_prices failed.")
```
hannesdiedrich, line 237: Is the traceback also returned for `log.exception` or only the exception type?

src/d3a/models/market/two_sided.py
```diff
@@ -234,49 +254,85 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
...
+        if not (bid.energy >= selected_energy and
+                offer.energy >= selected_energy and
+                (bid.energy_rate + FLOATING_POINT_TOLERANCE) >= clearing_rate and
+                (bid.energy_rate + FLOATING_POINT_TOLERANCE) >= offer.energy_rate):
+            raise InvalidBidOfferPairException
+        cls._validate_requirements_satisfied(
+            bid=bid, offer=offer, clearing_rate=clearing_rate, selected_energy=selected_energy)
+
     @classmethod
     def _replace_offers_bids_with_residual_in_matching_list(
```
hannesdiedrich, line 324: A docstring could be helpful here

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,182 @@
...
-    BidOfferMatch, offer_or_bid_from_json_string)
+from d3a.models.market import Market
+
 
-from d3a.models.myco_matcher.base_matcher import BaseMatcher
+class ExternalMatcherEventsEnum(Enum):
+    OFFERS_BIDS_RESPONSE = "offers_bids_response"
+    MATCH = "match"
+    TICK = "tick"
+    MARKET = "market"
```
hannesdiedrich, line 20: I propose to be consistent.
```suggestion
    MARKET = "market_cycle"
```

hannesdiedrich (commented):
Left some minor comments, LGTM once addressed

src/d3a/models/area/__init__.py
Hassan754: Yes it will print the full exception traceback

src/d3a/models/market/two_sided.py
Hassan754: Actually this was here before and I did not pay enough attention to what it does
Now that you mentioned it, I see some room for refactoring/optimizations here and in the `match_recommendations`, but I prefer leaving this to be tackled in [this ticket](https://gridsingularity.atlassian.net/browse/D3ASIM-3532) to not break things and to have enough time to find the best implementation, I'll be sure I address this in the mentioned ticket right after this one is merged, of course if you agree

src/d3a/models/market/market_structures.py
```diff
@@ -92,18 +94,22 @@ def serializable_dict(self):
...
-        return self.id == other.id and \
-            self.price == other.price and \
-            self.original_offer_price == other.original_offer_price and \
-            self.energy == other.energy and \
-            self.seller == other.seller
+        return (self.id == other.id and
+                self.price == other.price and
+                self.original_offer_price == other.original_offer_price and
+                self.energy == other.energy and
+                self.seller == other.seller and
```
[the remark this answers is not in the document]
faizan2590: Reminder

src/d3a/models/market/market_structures.py
```diff
@@ -183,9 +191,23 @@ def serializable_dict(self):
...
 
+    def __hash__(self):
+        return hash(self.id)
+
+    def __eq__(self, other):
+        return (self.id == other.id and
+                self.price == other.price and
+                self.original_bid_price == other.original_bid_price and
+                self.energy == other.energy and
+                self.buyer == other.buyer and
```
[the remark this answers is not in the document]
faizan2590: Reminder

src/d3a/models/market/two_sided.py
```diff
@@ -234,49 +254,85 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
...
-            cls, matchings, start_index, offer_trade, bid_trade
+            cls, matches: List[Dict], start_index, offer_trade: Trade, bid_trade: Trade
```
faizan2590, line 325:
```suggestion
            cls, matches: List[Dict], start_index: int, offer_trade: Trade, bid_trade: Trade
```

src/d3a/models/market/two_sided.py
faizan2590: As suggested, following could be added even now
```
"""
If a trade resulted in a residual offer/bid, upcoming matching list needs to be replaced with residual offer/bid
:param matchings: Recomended list of offer/bid matches
:param start_index: Current index of matching
:param offer_trade: Trade info of the successful offer
:param bid_trade: Trade info of the successful bid
:return: The updated matching offer/bid pair list with existing offer/bid replaced with corresponding residual offer/bid
"""
```

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -1,140 +1,182 @@
...
+             f"{self._channel_prefix}/offers-bids/": self.publish_offers_bids,
+             f"{self._channel_prefix}/recommendations/": self.match_recommendations})
 
     def publish_offers_bids(self, message):
         """Publish open offers and bids.
 
-        published data are of the following format
-        market_offers_bids_list_mapping = {"market_id" : {"bids": [], "offers": [] }, }
+        Published data are of the following format:
+            {"bids_offers": {`market_id` : {"bids": [], "offers": [] }, filters: {}}}
```
faizan2590, line 47:
```suggestion
            {"bids_offers": {"market_id" : {"bids": [], "offers": [] }, "filters": {}}}
```

src/d3a/models/market/market_structures.py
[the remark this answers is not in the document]
Hassan754: https://github.com/gridsingularity/d3a/blob/0d73b307d98f502c868833d5fb29f6868276d6b7/src/d3a/models/market/market_structures.py#L113

src/d3a/models/area/__init__.py
```diff
@@ -374,37 +380,53 @@ def publish_market_cycle_to_external_clients(self):
...
+        elif (self.strategy
+              and getattr(self.strategy, "is_aggregator_controlled", False)):
+            (self.strategy.redis.aggregator.
+             consume_all_area_commands(self.uuid,
+                                       self.strategy.trigger_aggregator_commands))
 
     def tick(self):
+        """Tick event handler.
+
+        invokes aggregator commands consumer, publishes market clearing, updates events,
```
faizan2590, line 395:
```suggestion
        Invokes aggregator commands consumer, publishes market clearing, updates events,
```

faizan2590 (approved):
LGTM

spyrostz (dismissed):
LGTM

faizan2590 (approved):

src/d3a/models/market/__init__.py
```diff
@@ -189,9 +189,16 @@ def __repr__(self):  # pragma: no cover
...
+        if reverse_order:
+            # Sorted bids in descending order
+            return list(reversed(sorted(
+                offers_bids.values(),
+                key=lambda obj: obj.energy_rate)))
+        else:
+
+            return sorted(offers_bids.values(),
+                          key=lambda obj: obj.energy_rate,
+                          reverse=reverse_order)
```
hannesdiedrich: Shall we reuse `sort_list_of_dicts_by_attribute` here in oder to be DRY?

Hassan754: The data structure here is different (dict of lists as values) whereas the mentioned function deals with list of dicts
However this is possible, I can do it in the ticket I'm currently working on ( supporting 1-N bids/offers matching) if you agree

hannesdiedrich (approved):
One suggestion that can be changed in an upcoming PR. Please dont forget. Thanks!

PR #1177, opened 2021-07-02
Fix case when offer is None in Home Meter tick event

fievelk:
This PR corrects some `if` statements that were causing exceptions when no offers were present.


hannesdiedrich (approved):
LGTM

Hassan754 (approved):
LGTM

PR #1179, opened 2021-07-05
DD-44: updatet ws address of node

lana-shanghai:


faizan2590 (dismissed):

tests/test_two_sided_market.py
```diff
@@ -1,16 +1,17 @@
...
-from d3a_interface.dataclasses import BidOfferMatch
-from pendulum import now
-from math import isclose
-from unittest.mock import MagicMock, patch
 from d3a_interface.constants_limits import ConstSettings
+from d3a_interface.dataclasses import BidOfferMatch
 from d3a_interface.matching_algorithms import (
     PayAsBidMatchingAlgorithm, PayAsClearMatchingAlgorithm
 )
+from pendulum import now
```
faizan2590, line 12: Since the imports are ordered now and I noticed, I guess one of the two should be enough like here [some lines above ](https://github.com/gridsingularity/d3a/pull/1179/commits/27956385f36abc33d073cbefbe6a98006b0a9880#diff-e443008143292bdfa0c5f6d4d871a132e7d974f5078bc4493649dbfa5b2506e4R5), whole pendulum library is imported too.

faizan2590 (approved):
LGTM

tests/test_two_sided_market.py
lana-shanghai: Agree here

spyrostz: Pycharm (which I used to order the imports automatically) disagrees with both of you though :D . The way the third party imports are organized are first to use the imports that use the `import` syntax (using alphanumeric order) and then to order the imports that use the `from * import * ` syntax, again in alphanumeric order.

spyrostz (approved):

src/d3a/models/area/__init__.py
```diff
@@ -266,10 +266,9 @@ def set_events(self, event_list):
     def activate(self, bc=None, current_tick=None, simulation_id=None):
```
hannesdiedrich, line 266: `bc` is now only a flag to enable of disable the blockchain interface, right? Could you please use a boolean value as default value then? Thanks!

lana-shanghai: A second too late :) let me add to the ticket with the next blockchain updates

hannesdiedrich: i was too late. But all good, will change it in my next PR. Thanks!

PR #1180, opened 2021-07-06
Feature/D3ASIM-3356 export matching algorithms

Hassan754:
…3a-interface


src/d3a/models/market/__init__.py
```diff
@@ -187,18 +187,11 @@ def __repr__(self):  # pragma: no cover
...
-                obj.values(),
-                key=lambda b: b.energy_rate)))
-
-        else:
-            # Sorted bids in ascending order
-            return list(sorted(
-                obj.values(),
-                key=lambda b: b.energy_rate))
+    def sorting(offers_bids: Dict, reverse_order=False) -> List[Union[Bid, Offer]]:
+        """Sort a list of bids or offers by a energy_rate attribute."""
```
hannesdiedrich, line 191: Nit:
```suggestion
        """Sort a list of bids or offers by their energy_rate attribute."""
```

src/d3a/models/market/two_sided.py
```diff
@@ -253,14 +253,14 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
                                     seller_id=offer.seller_id)
         return bid_trade, trade
 
-    def match_recommendation(self, recommended_list):
+    def match_recommendations(self, recommended_list: List[Dict]):
```
hannesdiedrich, line 256: Hmm, isn't is rather
```suggestion
    def match_recommendations(self, recommended_list: List[BidOfferMatch]):
```
? 
In the d3a-interface in `_create_bid_offer_matches` indeed BidOfferMatch objects are appended.

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -74,10 +72,10 @@ def match_recommendations(self, message):
         recommendations = data.get("recommended_matches", [])
         try:
             validated_records = self._get_validated_bid_offer_match_list(recommendations)
-            for market_id, records in validated_records.items():
-                market = self.markets_mapping.get(market_id)
+            for record in validated_records:
+                market = self.markets_mapping.get(record["market_id"])
```
hannesdiedrich, line 76: Hmm, neither here nor in the other PR I do see the `self.markets_mapping` being populated. Can you help me please?
Also (this should have been commented on by me earlier, sorry):
Why not populating a `market: BifOfferPair` mapping in the frame of `_get_validated_bid_offer_match_list` as we are looping over the recommendations and getting the market objects there already?
This way we could do something like the following here:
``` python
for market, validated_records in self.market_record_mapping.items():
        try:
            market.match_recommendation(records)
        except (OfferNotFoundException, BidNotFoundException):
            # If the offer or bid have just been consumed
            continue
```

tests/test_market_redis_connection.py
```diff
@@ -137,7 +137,7 @@ def test_accept_offer_calls_market_method_and_publishes_response(self):
                       seller="trade_seller", buyer="trade_buyer")
         self.market.accept_offer = MagicMock(return_value=trade)
         self.subscriber._accept_offer(payload)
-        sleep(0.01)
+        sleep(.1)
```
hannesdiedrich, line 140: Can you please explain this change? Why was it needed to increase the timeout?
Nit:I know it is the same and there is no general rule against it, but I personally prefer writing it like a human would write it on paper: `0.1` Thanks!

hannesdiedrich (commented):
Left some comments and questions, will review again after addressed.

src/d3a/models/market/two_sided.py
Hassan754: It's True I am returning BidOfferMatch after calling the `serializable_dict()` function on it, so these are dicts
I'll update the type hints so it reflects this

src/d3a/models/area/__init__.py
```diff
@@ -404,20 +404,29 @@ def tick(self):
...
+    def _match_bids_offers(self) -> None:
+        """Match bids and offers for all markets."""
+        if is_external_matching_enabled():
+            # Update the open offer bids cache that the myco client will request
+            bid_offer_matcher.match_algorithm.update_area_uuid_markets_mapping(
+                {self.uuid: self.all_markets})
+            return
+        # If the external matching is not enabled, get and match bid offer pairs
+        for market in self.all_markets:
+            bid_offer_pairs = True
```
fievelk, line 420: Nit: this one seems a bit counterintuitive to me, because the content of `bid_offer_pairs` shouldn't generally be a boolean. I understand this is a sort of trick to get the `while` loop started, but in this case I would prefer having a more verbose `while True` loop. For example:
```python
while True:
    bid_offer_pairs = ...
    if not bid_offer_pairs:
        break
```

src/d3a/models/market/__init__.py
```diff
@@ -187,18 +187,11 @@ def __repr__(self):  # pragma: no cover
...
+        return list(sorted(offers_bids.values(),
```
fievelk, line 192: Why is this forced to list?

src/d3a/models/market/two_sided.py
fievelk: Nit: since we use types, maybe we don't need to specify that `_list`. But this is really not so important.
```suggestion
    def match_recommendations(self, recommendations: List[BidOfferMatch]):
```

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: Thanks a lot, this is indeed a better approach!
I will adapt the _get_validated_bid_offer_match_list function then

tests/test_market_redis_connection.py
Hassan754: Yes I changed the value because due to race condition and the fact the code is async, on my machine sometimes this fails.
I will change the format

src/d3a/models/area/__init__.py
Hassan754: you are right, the do while loop is missed in python indeed.
Thanks!

src/d3a/models/market/__init__.py
Hassan754: I thought the sorted function returns a sorted iterable object (like the filter function)
I checked it again and it seems the returned type is a list, I will remove the redundant parsing

src/d3a/models/market/two_sided.py
hannesdiedrich: > It's True I am returning BidOfferMatch after calling the `serializable_dict()` function on it, so these are dicts
> I'll update the type hints so it reflects this

Ok, i see now. Can you maybe explain why you did that? Isn't calling `record.market_id` as convenient as `record["market_id"]` . One of the advantages of the DataClass, right?

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: Actually I gave it another look and recalled that the markets_mapping is cached in publish_offers_bids function, this is before sending the open offers and bids to the myco

hannesdiedrich: ok, even better, why not populating the `market: BifOfferPair` there (converting the `markets_mapping` into such).
This way we can match per market here (as suggested).

Hassan754: We cannot build bid offer pairs before sending the offers and bids, this is myco-api-client role

src/d3a/models/market/two_sided.py
Hassan754: True, but I'm sending recommendations through redis or websocket from myco  to d3a, this is why we can't send Dataclass instances

PR #1184, opened 2021-07-21
 Update Trade Market Structure

Hassan754:
Rename the Trade market structure member offer to be offer_bid, as it can be either type | refactor accordingly


integration_tests/steps/balancing_market.py
```diff
@@ -168,15 +168,17 @@ def follow_device_registry_energy_rate_or(context, device1, device2):
...
-               for device in [house1, house2, context.simulation.area]
-               for market in device.past_balancing_markets
-               for trade in market.trades if trade.offer.energy < 0)
-
-    assert all(int(round(trade.offer.price / trade.offer.energy, 2)) in positive_energy_rate
-               for device in [house1, house2, context.simulation.area]
-               for market in device.past_balancing_markets
-               for trade in market.trades if trade.offer.energy > 0)
+    assert all(
+        int(round(trade.offer_bid.price / trade.offer_bid.energy, 2)) in negative_energy_rate
```
spyrostz, line 172: This is a good opportunity to use the `Offer/Bid.energy_rate` member and replace the division. This applies to also other occurrences in the PR

integration_tests/steps/two_sided_market.py
```diff
@@ -156,26 +156,26 @@ def energy_origin(context, producer, consumer):
...
-    house1_consumed_energy = sum(trade.offer.energy
+    house1_consumed_energy = sum(trade.offer_bid.energy
                                  for market in house1.past_markets
                                  for trade in market.trades)
     assert isclose(house1_consumed_energy, load_energy_per_day, rel_tol=1e-3)
 
     # Assert that all House 2 trades contain Offer objects
     house2 = next(c for c in context.simulation.area.children if c.name == "House 2")
-    assert all(isinstance(trade.offer, Offer)
+    assert all(isinstance(trade.offer_bid, Offer)
```
spyrostz, line 174: Now that all these occurrences were revealed, it would be nice if we got rid of control flow via type checking. How about adding two extra methods/properties in both Offer and Bid classes, named `is_offer` and `is_bid` that will return True / False respectively and will be used for these methods ? I think it would help a lot in the readability of the caller code, if you disagree though feel free to skip it because this is purely according to personal preference.

src/d3a/models/market/market_structures.py
```diff
@@ -296,13 +296,13 @@ def to_json_string(self):
     def serializable_dict(self):
         return {
             "type": "Trade",
-            "match_type": "Offer" if isinstance(self.offer, Offer) else "Bid",
+            "match_type": "Offer" if isinstance(self.offer_bid, Offer) else "Bid",
```
spyrostz, line 299: I think you can use the classname for this, and thus removing the need for the isinstance.

src/d3a/models/market/market_structures.py
```diff
@@ -296,13 +296,13 @@ def to_json_string(self):
...
             "id": self.id,
-            "offer_bid_id": self.offer.id,
+            "offer_bid_id": self.offer_bid.id,
             "residual_id": self.residual.id if self.residual is not None else None,
-            "energy": self.offer.energy,
-            "energy_rate": self.offer.energy_rate,
-            "price": self.offer.energy * self.offer.energy_rate,
+            "energy": self.offer_bid.energy,
+            "energy_rate": self.offer_bid.energy_rate,
+            "price": self.offer_bid.energy * self.offer_bid.energy_rate,
```
spyrostz, line 305: You can use just `self.offer_bid.price`

tests/test_strategy_pvpredefined.py
```diff
@@ -218,8 +218,8 @@ def testing_event_trade(area_test3, pv_test4):
     pv_test4.state._available_energy_kWh[area_test3.test_market.time_slot] = 1
     pv_test4.event_trade(market_id=area_test3.test_market.id,
                          trade=Trade(id='id', time='time',
-                                     offer=Offer(id='id', time=pendulum.now(), price=20,
-                                                 energy=1, seller='FakeArea'),
+                                     offer_bid=Offer(id='id', time=pendulum.now(), price=20,
```
spyrostz, line 221: Very minor, to replace ' with "

spyrostz (commented):
Some comments, LGTM once resolved though.

hannesdiedrich (commented):
LGTM once Spyros' comments are dealt with.
IMO, If there is no ticket with a specification behind the PR, there should be at least one sentence in the PR description for the rationale behind this. Even if it is only a link to a slack or confluence discussion. I kind of got the rationale behind this change from reading the diff, but a more verbose explanation than `Update Trade Market Structure` should be given.

integration_tests/steps/balancing_market.py
Hassan754: This has exposed a bug, atm in two_sided's accept_bid we do not update the energy_rate of the bid.
I have fixed it

integration_tests/steps/two_sided_market.py
Hassan754: What do you think about adding these interfaces to the Trade class ? maybe named is_offer_trade() and is_bid_trade()

spyrostz: Even better! As long as the check is self-contained in the Trade class and not done by the caller code it works for me, thanks!

hannesdiedrich (dismissed):
LGTM

src/d3a/models/market/market_structures.py
```diff
@@ -34,75 +35,97 @@ class Clearing:
...
 
-        offer_dict["type"] = "Offer"
-        offer_dict.pop('energy_rate', None)
+        return json.dumps(obj_dict, default=my_converter)
 
-        return json.dumps(offer_dict, default=my_converter)
+    def _to_csv(self) -> Tuple:
+        rate = round(self.energy_rate, 4)
+        return (rate, self.energy, self.price,
+                self.seller if isinstance(self, Offer) else self.buyer)
```
fievelk, line 78: I would personally duplicate this code and put it both in Offer and Bid. I think it would be a bit more appropriate than having this `if` here.

src/d3a/models/market/market_structures.py
```diff
@@ -34,75 +35,97 @@ class Clearing:
...
+                 energy: float, seller: str, original_offer_price: float = None,
+                 seller_origin: str = None, seller_origin_id: str = None,
+                 seller_id: str = None,
+                 attributes: Dict = None,
+                 requirements: List[Dict] = None):
+        super().__init__(id=id, time=time, price=price, energy=energy,
+                         attributes=attributes, requirements=requirements)
+        self.seller = seller
+        self.original_offer_price = original_offer_price
+        self.seller_origin = seller_origin
```
fievelk, line 103: Nit: since we're unifying the interface for Bids and Offers, I got to think that these are basically the same concepts:
- `self.seller` and `self.buyer`
- `self.seller_origin` and `self.buyer_origin`

So they could in theory be unified and put inside `BidOffer`. The difficult part is the naming, though. What do these represent for Bid/Offer objects? Some hypotheses:
- `self.seller` and `self.buyer` --> `self.parent_origin` or `parent_actor` or `parent_trader`
- `self.seller_origin` and `self.buyer_origin` --> `self.root_origin` or `root_actor` or `root_trader`

Maybe you can come up with something better. If you think that this only complicates things, no need to do anything.

src/d3a/models/market/market_structures.py
```diff
@@ -114,93 +137,62 @@ def __eq__(self, other):
...
-        return 'rate [ct./kWh]', 'energy [kWh]', 'price [ct.]', 'seller'
-
-    def _to_csv(self):
-        rate = round(self.energy_rate, 4)
-        return rate, self.energy, self.price, self.seller
+        return "rate [ct./kWh]", "energy [kWh]", "price [ct.]", "seller"
 
 
-def copy_offer(offer):
+def copy_offer(offer) -> Offer:
```
fievelk, line 143: What about renaming this method so that it will be the same for Offer and Bids? E.g.
```suggestion
def copy(offer) -> Offer:
# or
def create_duplicate(offer) -> Offer:
```

src/d3a/models/market/market_structures.py
```diff
@@ -34,75 +35,97 @@ class Clearing:
...
+    def __init__(self, id: str, time: datetime, price: float,
+                 energy: float, seller: str, original_offer_price: float = None,
+                 seller_origin: str = None, seller_origin_id: str = None,
+                 seller_id: str = None,
+                 attributes: Dict = None,
+                 requirements: List[Dict] = None):
+        super().__init__(id=id, time=time, price=price, energy=energy,
+                         attributes=attributes, requirements=requirements)
+        self.seller = seller
+        self.original_offer_price = original_offer_price
```
fievelk, line 102: As I suggested on slack, this could also be renamed to be 
```suggestion
        self.original_price = original_offer_price
```
and put into `BidsOffer`

src/d3a/models/market/market_structures.py
```diff
@@ -211,16 +203,16 @@ def __eq__(self, other):
                 self.requirements == other.requirements)
 
 
-def offer_or_bid_from_json_string(offer_or_bid, current_time=None):
+def offer_or_bid_from_json_string(offer_or_bid, current_time=None) -> Union[Offer, Bid]:
```
fievelk, line 206: I think that this would be a good candidate to be a class method for `BaseBidOffer`:
```python
BaseBidOffer.from_json()
```
(Also, no need to specify the `_string` part in `from_json_string`, I think).

src/d3a/models/market/market_structures.py
```diff
@@ -211,16 +203,16 @@ def __eq__(self, other):
...
     object_type = offer_bid_dict.pop("type")
-    if "price" not in offer_bid_dict:
-        offer_bid_dict["price"] = offer_bid_dict["energy_rate"] * offer_bid_dict["energy"]
+
+    offer_bid_dict.pop("energy_rate", None)
+
     if object_type == "Offer":
-        offer_bid_dict.pop('energy_rate', None)
-        offer_bid_dict['time'] = current_time
+        offer_bid_dict["time"] = current_time
```
fievelk, line 213: Nit: I have the feeling that this logic doesn't belong here. But I see that you just took it from the existing version, so no need to fix it here.

src/d3a/models/market/market_structures.py
```diff
@@ -232,15 +224,15 @@ class TradeBidOfferInfo:
     propagated_offer_rate: float
     trade_rate: float
 
-    def to_json_string(self):
+    def to_json_string(self) -> str:
         return json.dumps(asdict(self), default=my_converter)
 
     @classmethod
-    def len(cls):
+    def len(cls) -> int:
```
fievelk, line 231: Nit: where is this used?

src/d3a/models/market/market_structures.py
```diff
@@ -232,15 +224,15 @@ class TradeBidOfferInfo:
...
         return cls.len()
 
 
-def trade_bid_info_from_json_string(info_string):
+def trade_bid_info_from_json_string(info_string) -> TradeBidOfferInfo:
```
fievelk, line 235: Similar comment as above: I would see this as a class method for `TradeBidOfferInfo`:
```python
trade_bid_offer = TradeBidOfferInfo.from_json(json_str)
```

fievelk: Nit: On another side: why is this called `TradeBidOfferInfo` and not `TradeInfo`? Do you have any idea about the reason?

src/d3a/models/market/market_structures.py
```diff
@@ -263,46 +255,56 @@ class Trade:
...
 
     @classmethod
-    def _csv_fields(cls):
+    def _csv_fields(cls) -> Tuple:
         return (tuple(cls.__dataclass_fields__.keys())[1:2] + ("rate [ct./kWh]", "energy [kWh]") +
                 tuple(cls.__dataclass_fields__.keys())[3:5])
 
-    def _to_csv(self):
-        rate = round(self.offer.energy_rate, 4)
+    def _to_csv(self) -> Tuple:
```
fievelk, line 271: Nit: two points about this method:
1. the method name is a bit misleading. The output is not a CSV file, but rather a tuple that can be used for many purposes (one of those is to create a CSV file);
2. the method should be public, since it's being used outside of this class

I know this is not part of the PR, but I'm taking this opportunity that you gave us with the refactoring to note it.

src/d3a/models/market/market_structures.py
```diff
@@ -263,46 +255,56 @@ class Trade:
...
         if key_in_dict_and_not_none(trade_dict, "residual"):
             trade_dict["residual"] = trade_dict["residual"].to_json_string()
         if key_in_dict_and_not_none(trade_dict, "offer_bid_trade_info"):
             trade_dict["offer_bid_trade_info"] = (
                 trade_dict["offer_bid_trade_info"].to_json_string())
         return json.dumps(trade_dict, default=my_converter)
 
-    def serializable_dict(self):
+    @property
+    def is_bid_trade(self) -> bool:
```
fievelk, line 289: Nit: a possible alternative would just be
```suggestion
    def is_bid(self) -> bool:
```
Do you think this would be accurate? (Same for the offer part)

src/d3a/models/strategy/__init__.py
```diff
@@ -458,9 +458,9 @@ def event_market_cycle(self):
         self.event_responses = []
 
     def assert_if_trade_offer_price_is_too_low(self, market_id, trade):
-        if isinstance(trade.offer, Offer) and trade.offer.seller == self.owner.name:
-            offer = [o for o in self.offers.sold[market_id] if o.id == trade.offer.id][0]
-            assert trade.offer.energy_rate >= \
+        if trade.is_offer_trade and trade.offer_bid.seller == self.owner.name:
```
fievelk, line 461: Could you please add a type hint to the `trade` argument of this method? I suppose it is a `TradeBidOfferInfo` and not a `Trade` instance, right?

fievelk (commented):
Left some comments. Thanks for the refactoring!

Hassan754:
@fievelk Thanks for your comments, I agree with the majority of them
In order to not overcomplicate this PR and make the review process harder, I'm thinking about creating another PR to address them
In this case, do you think it makes sense to gradually introduce these changes into the master branch (Merging the changes introduced here then the ones in the next PR)?

fievelk:
> In order to not overcomplicate this PR and make the review process harder, I'm thinking about creating another PR to address them

@Hassan754 Since my comments were not major nor blocking, that's definitely fine for me.

Hassan754:
> > In order to not overcomplicate this PR and make the review process harder, I'm thinking about creating another PR to address them
> 
> @Hassan754 Since my comments were not major nor blocking, that's definitely fine for me.

Awesome, I'll move forward with this PR and keep it as a reference to address the comments

fievelk (approved):

fievelk:
@Hassan754 regarding your latest commit: https://github.com/gridsingularity/d3a/pull/1184/commits/10120fdf51268c6e140476ed9b490bba1bd0506d

This bug should have made tests fail. Since it didn't happen, I suppose that this part of code is not tested. It would be great if you could add tests for it and make sure they would have failed with the previous bug.

Hassan754:
> @Hassan754 regarding your latest commit: [10120fd](https://github.com/gridsingularity/d3a/commit/10120fdf51268c6e140476ed9b490bba1bd0506d)
> 
> This bug should have made tests fail. Since it didn't happen, I suppose that this part of code is not tested. It would be great if you could add tests for it and make sure they would have failed with the previous bug.

Actually they were failing, there is a dedicated unit test function that was testing this, it failed after resolving the merge conflicts with master

src/d3a/models/market/market_structures.py
hannesdiedrich: No, IMO this would be confusing. A trade can not be a bid. This is to check if the trade was issued when accepting a bid. Please leave the property name as is, thanks!

hannesdiedrich (approved):
LGTM

PR #1186, opened 2021-07-22
Added ability to run tests in random order. Fixed issues with failing…

spyrostz:
… tests that were revealed when executed randomly.


tests/test_area.py
```diff
@@ -55,6 +55,7 @@ def setUp(self):
         self.config.ticks_per_slot = int(self.config.slot_length.seconds /
                                          self.config.tick_length.seconds)
         self.config.start_date = today(tz=TIME_ZONE)
+        GlobalConfig.sim_duration = duration(days=1)
```
Hassan754, line 58: Why is this needed?, you have already refactored the below usages to use `self.config`

tests/test_market.py
```diff
@@ -45,6 +45,14 @@
 }
 
 
+@pytest.fixture(scope="function", autouse=True)
+def device_registry_auto_fixture():
+    DeviceRegistry.REGISTRY = device_registry_dict
+    ConstSettings.IAASettings.MARKET_TYPE = 1
```
Hassan754, line 51: If this test is dependent from the state of the system this should be set by default, is there something I'm missing ?

tests/test_read_user_profile.py
```diff
@@ -1,14 +1,21 @@
...
+import d3a.constants
 from d3a.d3a_core.util import d3a_path
 
 
 class TestReadUserProfile(unittest.TestCase):
 
+    def setUp(self):
+        GlobalConfig.sim_duration = duration(days=1)
+        GlobalConfig.slot_length = duration(minutes=15)
+
```
Hassan754, line 18: AFAIS we are resetting these values to their default, these most probably are altered in another test which is breaking this test
If this is True, it makes more sense if we tear down the altered value in the tests touching this variable
However this is minor and can be skipped, you have already made huge refactoring, Thanks!

tests/test_strategy_storage.py
```diff
@@ -142,6 +151,9 @@ def sorted_offers(self):
         ]
         return offers[self.count]
 
+    def get_bids(self):
```
Hassan754, line 154: I don't see this being used anywhere here, and this is not a property of Area class as well

Hassan754 (dismissed):
LGTM

tests/test_area.py
```diff
@@ -92,8 +93,9 @@ def test_markets_are_cycled_according_to_market_count(self):
             assert len(self.area.all_markets) == i
 
     def test_delete_past_markets_instead_of_last(self):
+        constants.D3A_TEST_RUN = False
```
fievelk, line 96: (Not directly related to the refactoring) I'm a bit confused by this: what is this constant used for? Shouldn't this tell us that we're running tests?

tests/test_read_user_profile.py
fievelk: If @Hassan754 's comment is correct, this might fix this specific test but other tests might still break. I would also suggest changing this in the other test (it should be `test_strategy_pvpredefined`).

tests/test_market.py
spyrostz: Another test changes the state of the system to two sided market and does not revert it to the original value. I unfortunately had a hard time and did not manage to find out the culprit test, still though this should be here IMO. Since these tests require one-sided market in order to pass, it should be explicitly configured

tests/test_strategy_commercial_producer.py
```diff
@@ -159,6 +159,7 @@ def test_event_market_cycle_does_not_create_balancing_offer_if_not_in_registry(
 
 def test_event_market_cycle_creates_balancing_offer_on_last_market_if_in_registry(
         commercial_test1, area_test1):
+    ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET = True
```
fievelk, line 162: This is also changing the global default value. Won't this affect all the tests that are run after this?

tests/test_area.py
spyrostz: self.config changes only the SimulationConfig instance unfortunately. GlobalConfig is a different object, which includes global default settings for the configuration parameters. This uses class variables deliberately in order to be a singleton-style object, with the typical singleton disadvantage: if the class member is mutated, then the mutation affects all instances of this class. That's why changes in this object persist throughout multiple test runs.

tests/test_read_user_profile.py
spyrostz: I agree with you that this would definitely be the best way to go! Similar to the one-sided market configuration comment  though, I did not find the culprit test unfortunately, therefore I had to deal with the effect and not the cause of the issue :(

tests/test_strategy_storage.py
spyrostz: Thanks for catching that! I added these fake methods for the FakeMarket class when the two-sided market was activated by mistake, because I thought that this was the problem. Thus they should not be needed anymore, changing these

tests/test_strategy_commercial_producer.py
fievelk: I think one possible option would be to patch this value, either with unittest's mock.patch or with pytest (maybe [`monkeypatch`](https://docs.pytest.org/en/latest/how-to/monkeypatch.html) would be a possibility for this).

spyrostz: I thought that I had a teardown method for this class, however you are right as far as I see it is missing! Adding it now and thanks for catching it

tests/test_area.py
spyrostz: This is a badly-named constant from the past, that used to have a more general behavior. The only reason for the existence of this variable at the moment though, is to control if the old markets are deleted once they move to the past on every market cycle, or if the old market object instances will remain in-memory forever (which is something that some tests - unit and integration - actually require). I think it will be much better to rename this variable to the slightly more verbose DELETE_PAST_MARKETS, what do you think ? If you agree I'll rename this in the context of this PR

fievelk: I agree. In general, I think production code shouldn't know anything about tests. In this case however, I don't see many quick alternatives!

tests/test_read_user_profile.py
fievelk: So this doesn't depend on `test_strategy_pvpredefined`? I thought that was the only usage of these two constants in our tests. If not, let's leave it alone for the moment :)

tests/test_strategy_infinite_bus.py
```diff
@@ -15,22 +15,35 @@
...
+
 TIME = pendulum.today(tz=TIME_ZONE).at(hour=10, minute=45, second=0)
 
 
+@pytest.fixture(scope="function", autouse=True)
+def auto_fixture():
+    ConstSettings.IAASettings.MARKET_TYPE = 1
+    ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET = True
+    GlobalConfig.sim_duration = pendulum.duration(days=1)
+    GlobalConfig.slot_length = pendulum.duration(minutes=15)
```
fievelk, line 42: Just to double-check: will these two constants be reverted to their default value once the tests are done?

tests/test_strategy_load_hours.py
```diff
@@ -41,6 +41,13 @@
 MIN_BUY_ENERGY = 50  # wh
 
 
+@pytest.fixture(scope="function", autouse=True)
+def auto_fixture():
+    GlobalConfig.sim_duration = duration(days=1)
+    GlobalConfig.slot_length = duration(minutes=15)
```
fievelk, line 47: As above: shouldn't we provide a teardown step after the `yield` to restore the default values?

tests/test_area.py
spyrostz: On second thought, I would like to also ask @hannesdiedrich to see if he recalls the reason why we named this variable this way. The reason for that is because I recalled that we used to call this variable KEEP_PAST_MARKETS, but we renamed it for some reason down the road, and I really do not want to go full circle without a good reason for this.

tests/test_strategy_infinite_bus.py
spyrostz: Deleted these calls completely because I found out the test that did not revert these variables in the first place, therefore these should not be needed anymore.

tests/test_strategy_load_hours.py
spyrostz: Similar to the former comment, found out the root cause and the setup method should not be needed for this module

tests/test_read_user_profile.py
spyrostz: It did not depend only on this but on a couple more tests (test_simulation and test_live_events). Once I fixed these teardown methods the test succeeded, therefore this should not be needed anymore.

tests/test_live_events.py
```diff
@@ -65,6 +66,9 @@ def setUp(self):
         self.area_grid = Area("Grid", children=[self.area_house1, self.area_house2],
                               config=self.config)
 
+    def tearDown(self) -> None:
+        GlobalConfig.sim_duration = duration(days=1)
```
fievelk, line 70: Now I see why people don't like singletons :O Basically, here we have to restore the initial value of `GlobalConfig.sim_duration` even though we never directly changed: it was changed somewhere inside one of the classes we're using, but this effect is not reversed at the end of the test. If we start having too many singletons (buried deep in the classes hierarchy) this would become a nightmare.

fievelk (approved):
In general LGTM. There are two related topics still open:
1. the naming of the `D3A_TEST_RUN` constant;
2. if we change the default values of `GlobalConfig` (or any others) in `d3a-interface`, the teardown functions of these tests will become incorrect and need to be changed again, and I suppose we will easily forget to do it. One possible solution would be to store the default values in their own constant and use them here. E.g.:
```python
# in d3a-interface
DEFAULT_SIMULATION_DURATION = duration(days=DURATION_D)  # For example
class GlobalConfig:
    self.sim_duration = DEFAULT_SIMULATION_DURATION

# In d3a tests
from d3a_interface.constants_limits import DEFAULT_SIMULATION_DURATION
def tearDown(self) -> None:
    GlobalConfig.sim_duration = DEFAULT_SIMULATION_DURATION
```

tests/test_live_events.py
spyrostz: That's one of the long list of reasons that singletons are more or less an antipattern indeed. In general, there are 2 cases for a class to be allowed to be a singleton: 
1. configuration shared across the application (this case)
2. database or network connections, since these are singletons by nature

As you can see, even though we have a typical usecase for a singleton here, it is not without problems, since the singleton has become more or less a set of global variables in the unit tests context. For the production code we do not have this problem because the GlobalConfig is only instantiated once when the simulation is started and never mutated again, still for the tests it is very annoying

spyrostz:
> In general LGTM. There are two related topics still open:
> 
> 1. the naming of the `D3A_TEST_RUN` constant;
> 2. if we change the default values of `GlobalConfig` (or any others) in `d3a-interface`, the teardown functions of these tests will become incorrect and need to be changed again, and I suppose we will easily forget to do it. One possible solution would be to store the default values in their own constant and use them here. E.g.:
> 
> ```python
> # in d3a-interface
> DEFAULT_SIMULATION_DURATION = duration(days=DURATION_D)  # For example
> class GlobalConfig:
>     self.sim_duration = DEFAULT_SIMULATION_DURATION
> 
> # In d3a tests
> from d3a_interface.constants_limits import DEFAULT_SIMULATION_DURATION
> def tearDown(self) -> None:
>     GlobalConfig.sim_duration = DEFAULT_SIMULATION_DURATION
> ```

For the first part, we'll have to wait a bit. For the second part, these variables are already in d3a-interface, therefore I took advantage of them and replaced the hardcoded values with these.

fievelk (approved):

tests/test_area.py
hannesdiedrich: You are right Spyros, we had this in the past. I couldn't recall why we changed that either, but had I quick look into the code. This constant is not only used to keep the past markets but also to keep all bids and offers in the BaseStrategy buffers in order to assess these in the tests later.  Additionally, have a look here `src/d3a/d3a_core/simulation.py:259` where we also use this constant to export the raw results to JSON files. IMO these usages qualifies the variable naming and not `KEEP_PAST_MARKETS`.

hannesdiedrich (dismissed):
LGTM

tests/test_area.py
fievelk: But still, I would expect a variable named `D3A_TEST_RUN` to always be `True` during tests (and to be set once and for all before all tests). That's why I think that maybe we could find a better name for it (that doesn't have to do with "tests").
Right now this is used to:
1. keep the past markets in memory
2. keep all bids and offers in memory
3. export the raw results to JSON files

I think that (3) is unrelated to the other two. Maybe we could split this variable into two (or event three) variables, like `KEEP_MARKETS_IN_MEMORY`, `KEEP_BIDS_OFFERS_IN_MEMORY`, `EXPORT_RESULTS_TO_JSON`. It's just an idea of course, but it might make the variable less confusing in my opinion.

spyrostz: Thanks Hannes for clarifying this! In this case I would be also hesitant to rename this, and I would not like to split it into many variables since the reason that this variable is used is one (exporting to json has the same root cause, that json exporting depends on past state). Also setting it globally for all tests is not really possible because some tests take advantage of the False value of this variable. How about renaming this variable to RETAIN_PAST_MARKET_STRATEGIES_STATE?

hannesdiedrich: I am also not in favour of having 3 different variables and making it more complex than it is. In order to fully consider all parts, the variable name should be `RETAIN_PAST_MARKET_STRATEGIES_STATE_EXPORT_RAW_JSON` which is obviously bad. So I vote for keeping the original name and maybe add a verbose comment in the constants module to explain what this constant does.

spyrostz: Export raw json is not really needed as part of the name IMO. The only reason why this variable is used in this feature is because some raw JSON files depend on the results from all markets, because they do not accumulate on every market slot, but they are calculated once at the end of the simulation, and expect all market info (trades, bids, offers) to be there for all past markets for the whole simulation.
Having said that, I agree that probably a verbose comment in the constants.py would be sufficient for our purposes.

fievelk: Not blocking of course, but I would still vote for changing the name, to be honest 😄 `D3A_TEST_RUN` is just saying something different and misleading in my opinion. I would be in favor of changing it to `RETAIN_PAST_MARKET_STRATEGIES_STATE` (or even `RETAIN_PAST_MARKETS_STATES` for brevity, if it makes sense).

src/d3a/constants.py
```diff
@@ -38,7 +38,13 @@
 
 SIMULATION_PAUSE_TIMEOUT = 600
 
-D3A_TEST_RUN = False
+# Controls whether the past markets are still being kept in the simulation memory for the duration
+# of the simulation. Helpful in the unit / integration tests, since some of these rely on the
+# markets remaining in-memory until the end of the simulation run.
+# Is also needed for the raw JSON files and the plots when running in CLI mode.
+# Also helpful when debugging, in order for the interpreter to have access to all markets that a
+# simulation has ran through.
```
hannesdiedrich, line 46: For completenes, can you please add, that the bids and offers of all markets are buffered in the BaseStrategy when set to True, too? Or do you think this is clear?

hannesdiedrich (dismissed):
One minor comment. Thanks for the additional unittests!
LGTM

src/d3a/constants.py
spyrostz: Makes sense, I'll add it too. Thanks!

fievelk (approved):

PR #1187, opened 2021-07-22
Feature/D3ASIM-3532: Support 1 to N and N to M bid/offer matching

Hassan754:


setup.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 
-d3a_interface_branch = os.environ.get("BRANCH", "master")
+d3a_interface_branch = os.environ.get("BRANCH", "feature/D3ASIM-3532")
```
fievelk, line 5: Reminder

src/d3a/models/market/two_sided.py
```diff
@@ -19,6 +19,7 @@
 from dataclasses import replace
 from logging import getLogger
 from math import isclose
+import itertools
```
fievelk, line 22: Nit: are these imports optimized?

src/d3a/models/market/two_sided.py
```diff
@@ -257,35 +257,57 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
...
-
-            if trade.residual is not None or bid_trade.residual is not None:
-                recommendations = self._replace_offers_bids_with_residual_in_matching_list(
-                    recommendations, index+1, trade, bid_trade
+            clearing_rate = recommended_pair["trade_rate"]
+            market_offers = [
+                self.offers.get(offer["id"]) for offer in recommended_pair["offers"]]
+            market_bids = [self.bids.get(bid["id"]) for bid in recommended_pair["bids"]]
+
+            if not all(list(market_offers)):
```
fievelk, line 269: Why is this being forced to `list`?

src/d3a/models/market/two_sided.py
```diff
@@ -257,35 +257,57 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
...
+                # If not all received offers exist in the market, skip the current recommendation
+                continue
+            if not all(list(market_bids)):
```
fievelk, line 272: Same as above: why do we use `list()` if `market_bids` is already a list?

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -7,14 +7,13 @@
...
-    InvalidBidOfferPairException, OfferNotFoundException,
-    BidNotFoundException, MycoValidationException)
+    InvalidBidOfferPairException, MycoValidationException)
 from d3a.d3a_core.redis_connections.redis_area_market_communicator import ResettableCommunicator
 from d3a.models.market import Market
 
 
 class ExternalMatcherEventsEnum(Enum):
-    OFFERS_BIDS_RESPONSE = "offers_bids_response"
+    MATCHING_DATA_RESPONSE = "matching_data_response"
```
hannesdiedrich, line 16: Can you please explain the reason for this renaming? AFAIS, there are still only lists of bids and offers sent. The new name suggests to me that there was already some matching done. Is this because you also return the `filters`? I understand that these filters were sent by the myco and only returned here convenience, right?

src/d3a/models/market/two_sided.py
```diff
@@ -257,9 +257,9 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
     def match_recommendations(
             self, recommendations: List[BidOfferMatch.serializable_dict]) -> None:
         """Match a list of bid/offer pairs, create trades and residual offers/bids."""
-        if recommendations is None:
-            return
-        for index, recommended_pair in enumerate(recommendations):
+
+        while recommendations:
```
hannesdiedrich, line 261: So I understand that you want to digest all recommendations here. But what if e.g. a residual i can not be traded? I guess this will be en endless loop then, right?

src/d3a/models/market/two_sided.py
```diff
@@ -261,30 +261,53 @@ def match_recommendations(
...
+                    propagated_offer_rate=market_offer.price / market_offer.energy,
+                    trade_rate=original_bid_rate)
+
+                bid_trade, offer_trade = self.accept_bid_offer_pair(
+                    market_bid, market_offer, clearing_rate,
+                    trade_bid_info, selected_energy)
+                if offer_trade.residual:
+                    market_offer = offer_trade.residual
+                else:
+                    market_offer = next(market_offers, None)
```
hannesdiedrich, line 299: I dont understand why this is done here. Isn't the `market_offer` overwritten in Line 282 again in the next iteration?

src/d3a/models/market/two_sided.py
```diff
@@ -305,19 +328,33 @@ def _validate_requirements_satisfied(
...
+        # Example List1: [A, B], List2: [C, D] -> combinations: [(A, C), (A, D), (B, C), (B, D)]
+        bids_offers_combinations = itertools.product(bids, offers)
+        if not (
+                bids_total_energy >= selected_energy and
+                offers_total_energy >= selected_energy
+                and all(
+                    (bid.energy_rate + FLOATING_POINT_TOLERANCE) >= clearing_rate for bid in bids)
+                and all(
+                    combination[0].energy_rate +
+                    FLOATING_POINT_TOLERANCE >= combination[1].energy_rate
```
hannesdiedrich, line 350: I understand that the 0th element is the bid and the 1st element is the offer, right? Can you please add this to your comment in line 341? Thanks!

hannesdiedrich (commented):
Left some comments and questions. Will review again after they are addressed.

src/d3a/models/market/two_sided.py
Hassan754: I have originally made them as iterable and changed this later on, I'll update this line

src/d3a/models/market/two_sided.py
Hassan754: https://github.com/gridsingularity/d3a/pull/1187#discussion_r676399548

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: The matching algorithms should receive bids/offers + atm some of them receive an additional member which is a timestamp, so in order for this to be scalable I renamed the variable, we could send even more data in the future

src/d3a/models/market/two_sided.py
Hassan754: In the first line I'm popping the object, `recommended_pair = recommendations.pop()`
So we're consuming the list same an a for loop, I suppose no endless loop should happen here

src/d3a/models/market/two_sided.py
Hassan754: No, the one being reassigned later is the market_bid

src/d3a/models/myco_matcher/external_matcher.py
hannesdiedrich: `some of them receive an additional member which is a timestamp` Can you please explain which timestamp and why only some of them have this member?
Even if some of them contain a timestamp, this is still not a good rationale for renaming this IMO.

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -7,14 +7,13 @@
...
 import d3a.constants
 from d3a.d3a_core.exceptions import (
-    InvalidBidOfferPairException, OfferNotFoundException,
-    BidNotFoundException, MycoValidationException)
+    InvalidBidOfferPairException, MycoValidationException)
 from d3a.d3a_core.redis_connections.redis_area_market_communicator import ResettableCommunicator
 from d3a.models.market import Market
 
 
 class ExternalMatcherEventsEnum(Enum):
```
fievelk, line 15: Nit: Not so relevant, but I think we can also avoid adding the `Enum` suffix to the class name.
```suggestion
class ExternalMatcherEvents(Enum):
```

src/d3a/models/market/two_sided.py
hannesdiedrich: OK, I misunderstood the purpose of `_replace_offers_bids_with_residual_in_recommendations_list`. I understood now.

src/d3a/models/market/two_sided.py
hannesdiedrich: Sorry, this does not answer my question IMO. Can you please explain why this is done here more verbosely? Thanks!

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -37,16 +36,16 @@ def _setup_redis_connection(self):
         self.myco_ext_conn = ResettableCommunicator()
         self.myco_ext_conn.sub_to_multiple_channels(
             {"external-myco/simulation-id/": self.publish_simulation_id,
-             f"{self._channel_prefix}/offers-bids/": self.publish_offers_bids,
+             f"{self._channel_prefix}/matching-data/": self.publish_matching_data,
```
fievelk, line 39: I don't see any PR that changes this on the `myco-api-client`. Is there something missing?

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -79,14 +78,10 @@ def match_recommendations(self, message):
...
         recommendations = data.get("recommended_matches", [])
         try:
-            validated_records = self._get_validated_bid_offer_match_list(recommendations)
-            for record in validated_records:
-                market = self.markets_mapping.get(record["market_id"])
-                try:
-                    market.match_recommendations([record])
-                except (OfferNotFoundException, BidNotFoundException):
-                    # If the offer or bid have already been matched or deleted.
-                    continue
```
fievelk, line 89: Why is this part not needed anymore? Where were the `BidNotFoundException` and `OfferNotFoundException` exceptions raised before?

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -145,38 +140,36 @@ def _get_bids_offers(market: Market, filters: Dict):
...
                 # The market is already finished or doesn't exist
                 raise MycoValidationException
 
             # Get the original bid and offer from the market
-            market_bid = market.bids.get(record.get("bid").get("id"), None)
-            market_offer = market.offers.get(record.get("offer").get("id"), None)
+            market_bids = [
+                market.bids.get(bid.get("id"), None) for bid in recommendation.get("bids")]
+            market_offers = [
+                market.offers.get(offer.get("id"), None) for offer in recommendation.get("offers")]
```
fievelk, line 157:
```suggestion
            market_bids = [
                market.bids.get(bid.get("id")) for bid in recommendation.get("bids")]
            market_offers = [
                market.offers.get(offer.get("id")) for offer in recommendation.get("offers")]
```

fievelk: Also, can `recommendation["bids"]` be `None`? In that case, this would fail. Same applies for `recommendation["offers"]`

src/d3a/models/market/two_sided.py
Hassan754: Sure
So this is how the algorithm works:

1. Receive a list of recommendations, each recommendation can have multiple bids against multiple offers
2. While this list is not empty:
3. pop the first recommendation `recommended_pair = recommendations.pop()`
4. Get the market bids/offers version of the ones in the recommendation
5. If any bid or offer id doesn't exist in the market, continue | move on to the next recommendation
6. validate_bid_offer_match is called to validate that these can be matched
7. Make the bids/offers of the current recommendation as iterable
8. Pick the first bid and the first offer using next()
8. In a while True loop, we match the formers together
9. If there is a bid residual, make the market_bid = bid_trade.residual, the same goes for the offer (Otherwise we will pick the next item in the iterable)
10. We will always be matching the market_bid and market_offer until one of two iterables are fully consumed, this will break the while True loop and we'll move on to the next recommendation

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: Yes, the myco PR is still not created

src/d3a/models/myco_matcher/external_matcher.py
```diff
@@ -145,38 +140,36 @@ def _get_bids_offers(market: Market, filters: Dict):
...
 
-            if not (market_bid and market_offer):
-                # Offer or Bid either don't belong to market or were already matched
+            if not (all(market_bids) and all(market_offers)):
```
fievelk, line 159: Why is `all()` needed here? An empty list evaluates to `False`, so this should also work:
```suggestion
            if not (market_bids and market_offers):
```

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: If the bid or offer are not found, we will skip the recommendation in the TwoSided `match_recommendations` function.
However, if at some point the bid/offer were gone after the check, the accept_offer and accept_bid functions will raise `OfferNotFoundException` and `BidNotFoundException` respectively

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: I'm checking here a different case, not if we have an empty list [] , I'm checking the case when we have [offer1, offer2, None, offer3]
If some offer doesn't exist we will have  None , which is the case I'm testing here

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: The pay as clear matching algorithm expects the matching data to be {"market": {"bids": [], "offers":[], "current_time: ""}}
So the very outer key is the market uuid with mapping to bids/offers and some additional data
so the matching_data makes more sense to me than bids_offers, especially after taking into consideration that this can be scalable and we could be sending more data in the future

src/d3a/models/myco_matcher/external_matcher.py
fievelk: And how does a `None` end up in `recommendation["offers"]` (or bids)?

Hassan754: we don't, we get them in the market bids/offers in this line `market_offers = [
                market.offers.get(offer.get("id")) for offer in recommendation.get("offers")]`
 market.offers.get(offer.get("id"))  could return None if the offer id doesn't exist in the market, and then I'm checking `if not (all(market_bids) and all(market_offers)):`

fievelk: Ah yes, I see it now. Thanks

fievelk: Another question: why do you use `offer.get("id")` rather than `offer["id"]`? In which case can we have an offer without ID?

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: @hannesdiedrich let me know what you think, if this is okay I will continue with the renaming and create the Myco PR

hannesdiedrich: Hmm, I am still not convinces but would like to hear other opinions here. If more people agree with you, please use your suggestion.

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: The data is received from a client side, there should be a validation class-function imo to make sure the format is supported, anw I will not add this in my current PR

fievelk: I suggest creating a Jira task for it then, since it looks quite an important requirement 👍

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: I have decided to leave this change to my next feature to minimize the number of changes that should be tested

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: A validation function should guarantee this won't happen https://gridsingularity.atlassian.net/browse/D3ASIM-3542

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: https://gridsingularity.atlassian.net/browse/D3ASIM-3542

src/d3a/models/myco_matcher/external_matcher.py
fievelk: In the description of the task you only mention that the validation function should contain the right keys. So it wouldn't necessarily check for `None` though (given the current description).

src/d3a/models/market/two_sided.py
spyrostz: Furthermore, as far as I see, the reason for the permutation here is in order to validate that all bids have a higher energy rate than all offers, right ? In this case, in order to avoid the product which can potentially be expensive for long lists of matches, why not simply checking that all offers are less than or equal to the clearing rate ? In combination with the check of the bids it will achieve the same result, assuming that the clearing rate is the same for all pairs.

src/d3a/models/market/two_sided.py
```diff
@@ -257,35 +255,57 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
...
+
+            market_offers = iter(market_offers)
+            market_bids = iter(market_bids)
+            market_offer = next(market_offers)
+            market_bid = next(market_bids)
+            while True:
+                original_bid_rate = market_bid.original_bid_price / market_bid.energy
+                trade_bid_info = TradeBidOfferInfo(
+                    original_bid_rate=original_bid_rate,
+                    propagated_bid_rate=market_bid.price / market_bid.energy,
```
spyrostz, line 286: Trivial, but you can use `energy_rate`. Same goes for 2 lines below for the propagated_offer_rate

spyrostz (commented):
One comment, once this and the other comments are resolved LGTM

src/d3a/models/market/two_sided.py
hannesdiedrich: Darn! I overlooked `while True` loop. I thought this is all in the `while recommendations` loop. Ok, Understood. Thanks for the explanation!

hannesdiedrich:
LGTM once https://github.com/gridsingularity/d3a/pull/1187#discussion_r676391585 is dealt with. Thanks!

hannesdiedrich (approved):
LGTM

src/d3a/models/myco_matcher/external_matcher.py
Hassan754: @fievelk I have updated the ticket description to better reflect the requirement

src/d3a/models/market/market_structures.py
```diff
@@ -54,6 +54,8 @@ class Offer:
     def __post_init__(self):
         self.id = str(self.id)
         self.energy_rate = self.price / self.energy
+        if self.original_offer_price is None:
```
fievelk, line 57: For future PRs: I think it would be good to rename `self.original_offer_price` and `self.original_bid_price` to `self.original_price`. This has some advantages:
1. the name would be a bit better, since you remove a repetition coming from the class name (`offer.original_price` vs `offer.original_offer_price`)
2. it would allow homogenizing `Offer` and `Bid` objects' behaviour (and maybe someday they'll share the same interface)

Hassan754: I totally agree, I also suggested this in a slack conversation , I may do it in a next PR

tox.ini
```diff
@@ -75,6 +75,8 @@ commands =
 	python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
     pip-sync requirements/tests.txt requirements/pandapower.txt
     pip install -e .
+	pip uninstall -y d3a-interface
```
fievelk, line 78: Nit: is this needed?

fievelk (approved):

src/d3a/models/market/two_sided.py
```diff
@@ -257,35 +255,57 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
...
+            market_offers = [
+                self.offers.get(offer["id"]) for offer in recommended_pair["offers"]]
+            market_bids = [self.bids.get(bid["id"]) for bid in recommended_pair["bids"]]
+
+            if not all(market_offers):
+                # If not all received offers exist in the market, skip the current recommendation
+                continue
+            if not all(market_bids):
+                # If not all received bids exist in the market, skip the current recommendation
+                continue
```
faizan2590, line 272:
```suggestion
            if not all(market_offers) or not all(market_bids):
                # If not all received offers OR bids exist in the market, skip the current recommendation
                continue
```

src/d3a/models/market/two_sided.py
```diff
@@ -257,35 +255,57 @@ def accept_bid_offer_pair(self, bid, offer, clearing_rate, trade_bid_info, selec
...
+                    market_bid = bid_trade.residual
+                else:
+                    market_bid = next(market_bids, None)
+                recommendations = (
+                    self._replace_offers_bids_with_residual_in_recommendations_list(
+                        recommendations, offer_trade, bid_trade)
                 )
+                if not (market_bid and market_offer):
+                    # If we reach the end of the offers/bids lists, break
+                    break
```
faizan2590, line 308:
```suggestion
```

leading to https://github.com/gridsingularity/d3a/pull/1187/files#r682691970

faizan2590 (approved):
LGTM, though left some enhancement comment.

PR #1192, opened 2021-07-28
Feature/d3 asim 3493 Renamed max_panel_power_W into capacity_kW and also added this power scaling to the predefined PV

hannesdiedrich:
Here I propose to look at each commit individually as I tried to distinguish the steps.

 For more information please refer to this comment in the corresponding ticket:
 https://gridsingularity.atlassian.net/browse/D3ASIM-3493?focusedCommentId=23277


integration_tests/steps/area_events.py
```diff
@@ -146,6 +146,7 @@ def load_consumes_less_than_between(context, energy, start_time, end_time):
             continue
         assert market.trades[0].seller == "IAA House 1" and \
             market.trades[0].buyer == "Grid Load"
+        print(energy, market.trades[0].offer.energy)
```
Hassan754, line 149: reminder

src/d3a/models/config.py
```diff
@@ -66,9 +66,9 @@ def __init__(self, sim_duration: duration, slot_length: duration, tick_length: d
         self.read_pv_user_profile(pv_user_profile)
         self.read_market_maker_rate(market_maker_rate)
 
-        max_panel_power_W = ConstSettings.PVSettings.MAX_PANEL_OUTPUT_W \
-            if max_panel_power_W is None else max_panel_power_W
-        self.max_panel_power_W = max_panel_power_W
+        capacity_kW = ConstSettings.PVSettings.DEFAULT_CAPACITY_KW \
+            if capacity_kW is None else capacity_kW
```
Hassan754, line 70:
```suggestion
        capacity_kW = capacity_kW or ConstSettings.PVSettings.DEFAULT_CAPACITY_KW
```

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -113,17 +116,24 @@ def _read_predefined_profile_for_pv(self):
             else:
                 self._power_profile_index = self.owner.config.cloud_coverage
         if self._power_profile_index == 0:  # 0:sunny
-            profile_path = pathlib.Path(d3a_path + '/resources/Solar_Curve_W_sunny.csv')
+            profile_path = (
+                pathlib.Path(d3a_path + '/resources/Solar_Curve_W_sunny_normalized.csv'))
```
Hassan754, line 120: Nit, single quotes

Hassan754 (approved):
LGTM

src/d3a/resources/Solar_Curve_W_cloudy_normalized.csv
```diff
@@ -0,0 +1,97 @@
+INTERVAL;POWER(W)
```
faizan2590, line 1: I was expecting it to be kW

faizan2590: Okay, how about using the pre-existing CSVs  in codebase?

hannesdiedrich: Sorry, don't understand. What do you mean by `pre-existing CSVs` ?
The scaling of the predefined load is a feature requested in the ticket.

faizan2590: I meant why not updating the old CSV itself `Solar_Curve_W_cloudy` rather than creating this one, as ATM, they both only been used in `src/d3a/models/strategy/predefined_pv.py`

hannesdiedrich: As described in the ticket, its discussion and in the documentation I provided here https://gridsingularity.atlassian.net/wiki/spaces/D3A/pages/3008004097/PV+Workflow+d3a-web+--+d3a+as+of+Jul+2021  We need a way of scaling the predefined PV with the `capacity_kW`. That is why we normalized the predefined profiles so that we only need to multiply `capacity_kW`.

faizan2590: Sorry, still haven't understood it
Screen Shot 2021-08-05 at 11 02 26
There your talking about both `capacity_kW` as well as `max_panel_powerW` whereas what I can see in `PVPredefinedStrategy`, no `max_panel_powerW` parameter exist now
Screen Shot 2021-08-05 at 11 04 30

hannesdiedrich: You looked at the d3a-web paragraph. But you are right, there was no max_panel_power before in the PredefinedPV. `capacity_kW` was added now to scale the predefined profiles (which again are normalized to be between 0 and 1).

hannesdiedrich: I just realise that the numbers never reach one. But it was decided that the exiting profiles should be divided by 250.

src/d3a/resources/Solar_Curve_W_sunny_normalized.csv
```diff
@@ -0,0 +1,97 @@
...
+11:45;0.8
+12:00;0.8
+12:15;0.7
+12:30;0.65
+12:45;0.75
+13:00;0.75
+13:15;0.6
+13:30;0.7
+13:45;0.65
+14:00;0.7
```
hannesdiedrich, line 58: @chrisd12 Please have a look at the weight profiles here which I derived from dividing the exiting predefined PV profiles by 250 (as you suggested in the ticket D3ASIM-3493). However, the maximum value was 200W in the sunny profile. So, If you scale a sunny profile with `capacity_kW=250`, you will only get a peak production of 200W for the sunny case (and less for the cloudy and partially cloudy cases). Do you see this as a problem?

src/d3a/models/power_flow/__init__.py
```diff
@@ -55,7 +55,7 @@ def _d3a_area_to_electrical_grid(self, area):
         elif isinstance(area.strategy, PVStrategy):
             area.pv_device = \
                 self.add_generation_device(area,
-                                           convert_unit_to_kilo(area.strategy.max_panel_power_W))
+                                           convert_unit_to_kilo(area.strategy.capacity_kW))
```
fievelk, line 58: If capacity is already in kWh, this shouldn't be necessary, right?

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -63,6 +65,7 @@ def __init__(
         :param fit_to_limit: Linear curve following initial_selling_rate & final_selling_rate
         :param update_interval: Interval after which PV will update its offer
         :param energy_rate_decrease_per_update: Slope of PV Offer change per update
+        :param capacity_kW: power rating of the predefined profiles
```
fievelk, line 68: Nit: it would be nice to change the docstring style to be Google-style

src/d3a/models/power_flow/__init__.py
hannesdiedrich: Wow, good catch! Thanks!

fievelk: Sure! It would be great to have some tests that would fail with the previous bug, if it's not too convoluted with the current structure.

setup.py
```diff
@@ -1,8 +1,7 @@
 from setuptools import find_packages, setup
-import os
 
 
-d3a_interface_branch = os.environ.get("BRANCH", "master")
+d3a_interface_branch = "feature/D3ASIM-3493"
```
fievelk, line 4: Reminder

fievelk:
`max_panel_power_W` is also still mentioned in the wiki docs: https://github.com/gridsingularity/d3a/blob/f6253efa3aba1255afa0652e02ea63a1b2f547ea/wiki/docs/general-settings.md#L162

src/d3a/models/strategy/predefined_pv.py
```diff
@@ -113,17 +116,24 @@ def _read_predefined_profile_for_pv(self):
...
+            profile_path = (
+                pathlib.Path(d3a_path + "/resources/Solar_Curve_W_cloudy_normalized.csv"))
         else:
             raise ValueError("Energy_profile has to be in [0,1,2,4]")
 
-        # Populate energy production forecast data
-        self.power_profile = read_arbitrary_profile(
-            InputProfileTypes.POWER, str(profile_path))
+        power_weight_profile = read_arbitrary_profile(
+            InputProfileTypes.IDENTITY, str(profile_path))
```
fievelk, line 131: Nit:
```suggestion
        power_weight_profile = read_arbitrary_profile(
            InputProfileTypes.IDENTITY, profile_path)
```
This check is already done here: https://github.com/gridsingularity/d3a-interface/blame/bfa8804079ec0edad2a6e5907fea526da3805eed/d3a_interface/read_user_profile.py#L232

src/d3a/models/strategy/pv.py
```diff
@@ -102,8 +102,8 @@ def area_reconfigure_event(self, **kwargs):
 
         if key_in_dict_and_not_none(kwargs, 'panel_count'):
             self.panel_count = kwargs['panel_count']
-        if key_in_dict_and_not_none(kwargs, 'max_panel_power_W'):
-            self.max_panel_power_W = kwargs['max_panel_power_W']
+        if key_in_dict_and_not_none(kwargs, 'capacity_kW'):
+            self.capacity_kW = kwargs['capacity_kW']
```
fievelk, line 106: Nit
```suggestion
        if key_in_dict_and_not_none(kwargs, "capacity_kW"):
            self.capacity_kW = kwargs["capacity_kW"]
```

src/d3a/models/strategy/pv.py
```diff
@@ -221,14 +221,15 @@ def gaussian_energy_forecast_kWh(self, time_in_minutes=0):
...
+            gauss_forecast = self.capacity_kW * math.exp(
                 # time/5 is needed because we only have one data set per 5 minutes
 
                 (- (((round(time_in_minutes / 5, 0)) - 147.2)
                     / 38.60) ** 2
                  )
             )
-        return round(convert_W_to_kWh(gauss_forecast, self.area.config.slot_length), 4)
+
+        return round(convert_kW_to_kWh(gauss_forecast, self.area.config.slot_length), 4)
```
fievelk, line 232: I think we're usually rounding to a shared precision:
```suggestion
from d3a.constants import DEFAULT_PRECISION
        # ...
        return round(convert_kW_to_kWh(gauss_forecast, self.area.config.slot_length), DEFAULT_PRECISION)
```

src/d3a/setup/area_events/cloud_coverage_event.py
```diff
@@ -30,15 +30,15 @@ def get_setup(config):
                 'House 1',
                 event_list=[ConfigEvents(12, {'cloud_coverage': 1})],
                 children=[
-                    Area('H1 PV', strategy=PVPredefinedStrategy(),
+                    Area('H1 PV', strategy=PVPredefinedStrategy(capacity_kW=0.25),
```
fievelk, line 33: Nit: if you happen to refactor, you could change from `'` to `"`. Same applies to the rest of the file/PR.

src/d3a/setup/d3a-settings.json
```diff
@@ -5,7 +5,7 @@
     "tick_length": "15s",
     "market_count": 1,
     "cloud_coverage": 0,
-    "start_date": "2021-03-25"
+    "start_date": "2021-07-28"
```
fievelk, line 8: Why is this needed?

src/d3a/setup/kpi/d3asim_3464_young_couple.py
```diff
@@ -44,7 +44,7 @@ def get_setup(config):
                                 fit_to_limit=True)),
                          ]),
                     Area("Community PV", strategy=PVStrategy(
-                        max_panel_power_W=4000, panel_count=1,
+                        capacity_kW=4000, panel_count=1,
```
fievelk, line 47: Shouldn't this be divided by 1000?

fievelk: It probably doesn't change much, unless some researcher has based their computations on this.

tests/test_live_events.py
```diff
@@ -38,7 +38,7 @@ def setUp(self):
         self.strategy_pv = PVStrategy(
             panel_count=3, initial_selling_rate=34, final_selling_rate=12,
             fit_to_limit=False, update_interval=6, energy_rate_decrease_per_update=4,
-            max_panel_power_W=432)
+            capacity_kW=432)
```
fievelk, line 41: Nit: probably unnecessary, but should it be converted to kW?

fievelk: The same applies to the rest of the file. It's probably not needed since all computations will be scaled up and remain consistent, I assume.

fievelk (commented):
One major comment about `convert_unit_to_kilo`. 
The rest is mostly about conversion to kW (which probably doesn't make any difference with the final test results).

hannesdiedrich:
> `max_panel_power_W` is also still mentioned in the wiki docs:
> 
> https://github.com/gridsingularity/d3a/blob/f6253efa3aba1255afa0652e02ea63a1b2f547ea/wiki/docs/general-settings.md#L162

@chrisd12  will deal with adapting the docs

src/d3a/setup/d3a-settings.json
hannesdiedrich: No, the date is only updated when you run `fab write_default_settings_file`. Does not matter really. This setup file is not used by users at all unfortunately AFAIK.

src/d3a/setup/kpi/d3asim_3464_young_couple.py
hannesdiedrich: Yeah, I skipped adapting all power values in all setup files and only adapted the ones that are used in the integrationtests which failed. There are a lot of setup files that are not used any more. 
Do you think I should take the time to adapt the power values in all of them?

tests/test_live_events.py
hannesdiedrich: https://github.com/gridsingularity/d3a/pull/1192#discussion_r684089286

src/d3a/setup/kpi/d3asim_3464_young_couple.py
fievelk: I think it's not necessary, as long as the values are internally coherent with each other. The only small issue I see is that 4000kW is an extremely high value for a solar panel (about 1000 times bigger than what you would expect :P )

src/d3a/resources/Solar_Curve_W_cloudy_normalized.csv
faizan2590: > which again are normalized to be between 0 and 1

Would be nice then to have validator function for validating that the provided CSVs/profiles always have a value between 0 and 1.

hannesdiedrich: IMO there is no need. You can not upload such a profile, we only support these three: Sunny, Cloudy, PartiallyCloudy

faizan2590: > IMO there is no need. You can not upload such a profile, we only support these three: Sunny, Cloudy, PartiallyCloudy

Okay thanks got it after looking into PR again, BUT still why are you still keeping the profile that were used earlier ?

- Solar_Curve_W_partial.csv
- Solar_Curve_W_sunny.csv
- Solar_Curve_W_cloudy.csv

hannesdiedrich: Yeah, I could remove them, but I thought that people might want to use them as example profiles for the PVUserProfileStrategy. If you like me to delete them, will do, but I don't see any harm in leaving them.

faizan2590: Okay, leave it then.

faizan2590 (dismissed):
LGTM

src/d3a/resources/Solar_Curve_W_sunny_normalized.csv
chrisd12: @hannesdiedrich I don't see it as a problem at all. Usually when a PV manufacturer state the capacity (e.g. 250W) it means the maximum power the panel can produce under perfect conditions (irradiance, angle, temperature) and these are very rarely met. So it is fine to have an output a little lower than the capacity.
I have tested a couple of simulation and it seems to be working. Just one thing, can you please rename the normalised csv files to kW (e.g. `Solar_Curve_W_cloudy_normalized` --> `Solar_Curve_kW_cloudy_normalized`)?

hannesdiedrich: @chrisd12  Ok, thanks a lot! Makes a lot of sense!
RWT the renaming: I dont agree and would even change it to: `Solar_Curve_cloudy_normalized` as the weights in there dont have any unit (I will also remove the unit form the table header). Right?

chrisd12: @hannesdiedrich The profile is multiplied by the capacity_kW. If capacity_kW=1 then the profile remained unchanged. If capacity_kW=2 then the profile values are multiplied by 2. Basically this file is a normalised profile but it is also a profile of a PV with a capacity of 1kW.

hannesdiedrich: @chrisd12 I am not convinced. Indeed, `it is also a profile of a PV with a capacity of 1kW`, but only if you multiply by `1 kW`. Sorry, but my physics professor would agree with me, these weights are unitless and you only make a power profile out of it, when you multiply `capacity_kW` . 
If you want to say that these are 1kW profiles, then we need to rename the scaling value to only `capacity_scale` or so.

src/d3a/setup/kpi/d3asim_3464_young_couple.py
hannesdiedrich: It was actually the only place where the power value was so high. SO I converted it to kW.

src/d3a/models/power_flow/__init__.py
hannesdiedrich: This is only the power flow calculation that is not used operationally ATM but is tested in two integrationtests. I removed the W->kW conversion but don't plan to add another test.

src/d3a/resources/Solar_Curve_W_sunny_normalized.csv
chrisd12: @hannesdiedrich ok then let's proceed **without** the unit. Please name it `Solar_Curve_cloudy_normalized`. The column name `Profile weigh` looks good to me

tox.ini
```diff
@@ -76,7 +76,7 @@ commands =
     pip-sync requirements/tests.txt requirements/pandapower.txt
     pip install -e .
 	pip uninstall -y d3a-interface
-	pip install git+https://github.com/gridsingularity/d3a-interface@{env:BRANCH:master}
+	pip install git+https://github.com/gridsingularity/d3a-interface@feature/D3ASIM-3493
```
Hassan754, line 79: reminder

src/d3a/models/strategy/pv.py
hannesdiedrich: AFAIR, It made some integrationtests fail which where not that easy to adapt. That is why i left this for a later change.

PR #1193, opened 2021-07-28
 External myco setup add extra assets

Hassan754:


hannesdiedrich:
@Hassan754 Here again, please explain why this was needed if there is not connected ticket. Thank!

Hassan754:
> @Hassan754 Here again, please explain why this was needed if there is not connected ticket. Thank!

This is not needed anymore, the research team wanted more assets to have more offers/bids, however they have them now on their machines and I'll leave this script to them then

PR #1196, opened 2021-07-30
D3ASIM-3536: simulate energy imbalance between production/consumption and forecast

fievelk:
The real amount of energy produced/consumed by a device is usually different from what the device anticipated to consume/produce. So far, our simulations only used this second type of energy (the forecasts). However, for the implementation of settlement markets, we also need track the "real" energy.

For simulations we don't have real energy readings. We therefore simulate the "real" readings by altering the energy forecasts by some amount.

This PR adds a variable in the state classes to track the amount of energy actually consumed/produced by the device.


src/d3a/models/state.py
```diff
@@ -56,17 +56,33 @@ def get_state(self) -> Dict:
 
     @abstractmethod
     def restore_state(self, state_dict: Dict):
-        pass
+        """Update the state of the device using the provided dictionary."""
 
     @abstractmethod
     def delete_past_state_values(self, time_slot):
```
hannesdiedrich, line 62: While you are on it, can you please add a type to `time_slot` here (i think it is pendulum.DateTime) ?
I find this very helpful. Thanks a lot!

src/d3a/models/state.py
```diff
@@ -56,17 +56,33 @@ def get_state(self) -> Dict:
...
 
-class ConsumptionState(StateInterface):
+class ProsumptionInterface(StateInterface, ABC):
+    """Interface with common methods/variables shared by consumption and production devices."""
+
+    def __init__(self):
+        # Actual energy consumed/produced by the device at specific market slots
+        self._real_energy: Dict[DateTime, float] = {}
+
+    def set_real_energy_kWh(self, energy_kWh: float, time_slot: DateTime) -> None:
```
hannesdiedrich, line 76: I read `-> None` a best practice? IMO it is not needed.

src/d3a/models/strategy/home_meter.py
```diff
@@ -189,6 +190,21 @@ def event_market_cycle(self):
...
+        """Set the (simulated) actual energy of the device in the previous market slot."""
+        if self.area.current_market:
+            self._set_real_energy_kWh(self.area.current_market.time_slot)
+
+    def _set_real_energy_kWh(self, time_slot: DateTime) -> None:
+        """Set the (simulated) actual energy of the device in a market slot."""
+        energy_forecast_kWh = self.state.get_energy_at_market_slot(time_slot)
+        simulated_real_energy = utils.alter_energy(energy_forecast_kWh)
+
+        # This value can be either positive or negative. This is different than with the other
```
hannesdiedrich, line 203: Please add again the sign convention to this comment (which sign means consumption and which production), thanks!

src/d3a/models/strategy/utils.py
```diff
@@ -0,0 +1,33 @@
...
+def alter_energy(
+        energy_kWh: float, max_deviation_pct: float = MAX_ENERGY_DEVIATION_PERCENT) -> float:
+    """
+    Compute a new energy amount that differs from the given one by (at most) a max percentage.
+    """
+    deviation_pct = random.uniform(-max_deviation_pct, max_deviation_pct)
+    # NOTE: the deviation can be either positive or negative
+    deviation_kWh = round((energy_kWh * deviation_pct) / 100.0, DEFAULT_PRECISION)
+    altered_energy = energy_kWh + deviation_kWh
+    # The sign of the new energy should never be flipped compared to the original
```
hannesdiedrich, line 30: But this is easily done when setting `MAX_ENERGY_DEVIATION_PERCENT>100` right?
- For prosumers this is expected and wanted, and it should only be limited for values>0 for load and PV, right?

tests/test_strategy_utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+from d3a.models.strategy import utils
+
+
+class TestStrategyUtils:
+    """Test utility functions for strategy modules."""
+
+    @staticmethod
+    @pytest.mark.parametrize("execution_number", range(10))
+    def test_alter_energy(execution_number):  # pylint: disable=W0613
```
hannesdiedrich, line 25: Once settled on my comment in `alter_energy`, please also add a check for the 0 boundary here. Thanks!

hannesdiedrich (commented):
Looks good in general. Left some comments. Thanks

src/d3a/models/state.py
fievelk: I would personally treat `None` as any other type.
We are using type hints just as docs, so right now this is not that relevant (even though I like documentation). But we could decide to use a static type checker like mypy to debug our codebase. As an example, it would find these:
```python
# my_test.py
class SomeClass:
    @staticmethod
    def do_something() -> None:
        return 12
```
From console
```sh
$ mypy my_test.py
my_test.py:4: error: No return value expected
```
Note that removing `-> None:` is not the same! In that case, the static checker will just not check the return type:
```python
# my_test.py
class SomeClass:
    @staticmethod
    def do_something():
        return 12
```
```sh
$ mypy my_test.py
Success: no issues found in 1 source file
```

hannesdiedrich: ok, makes sense. I have to get used to it then. BTW, this does not apply to `__init__` ?

src/d3a/models/strategy/utils.py
fievelk: > But this is easily done when setting MAX_ENERGY_DEVIATION_PERCENT>100 right?

You're right, that's the only way this can happen at the moment.
Unless `MAX_ENERGY_DEVIATION_PERCENT` is set to `> 100`, even prosumers would not flip their energy sign. If a HomeMeter produces `1 kWh`, the maximum deviation it could have is `+/- 0.5 KwH` at the moment (so it would not be possible for it to switch sign).

If we want to allow prosumers to switch their sign, we need to use a different approach to limit the maximum energy deviation than the one we discussed on Slack.

fievelk: > it should only be limited for values>0 for load and PV, right?

Sorry I think I didn't get this part. What should be limited? Do you mean `MAX_ENERGY_DEVIATION_PERCENT`?

Suppose `MAX_ENERGY_DEVIATION_PERCENT == -50`, this would not change the computations I think:
```python
deviation_pct = random.uniform(-max_deviation_pct, max_deviation_pct)
# becomes random.uniform(+50, -50)
```
This part is the only line where the constant is used, so if you have a `-` sign in front of it it will just flip the operands of `random.uniform`.

hannesdiedrich: IMO, we should skip this premature optimisation and not think to much about this approach as it will be replaced soonish by probably something more real (e.g. real energy measurements, even in simulations).
So how about  doing something like 
```python
altered_energy = min(energy_kWh + deviation_kWh, 0) if not is_prosumer else energy_kWh + deviation_kWh
```
We would need to have the flag `is_prosumer`. Not pretty, i know, but should work, right?

src/d3a/models/state.py
fievelk: I think it can apply to everything: I'm also not completely used to it but in theory you could use it everywhere. If you want, I can also add it to the `__init__`.

src/d3a/models/strategy/utils.py
fievelk: The thing is that `deviation_kWh` (in your example) will never be a value that could flip the sign of any device.
```python
min(energy_kWh + deviation_kWh, 0)
```
In this code, `energy_kWh + deviation_kWh` will never be `< 0`, because `deviation_kWh` will always be `< energy_kWh` (since it is a percentage). 

Of course, this is giving for granted that we do not set `MAX_ENERGY_DEVIATION_PERCENT > 100`.

src/d3a/models/state.py
hannesdiedrich: IMO dont add it at all. ;-) But yeah, leave it in `set_real_energy_kWh` but omit it in `__init__`, Thanks!

src/d3a/models/strategy/utils.py
hannesdiedrich: > The thing is that `deviation_kWh` (in your example) will never be a value that could flip the sign of any device.

Yes, it could, when the `is_prosumer` flag is set, it behaves like you proposal. I would set the `is_prosumer` flag to True only for PV and Load. (of course, a better naming could be chosen for this flag).

fievelk: I'm not sure I understand how it would flip the sign. Whatever the value of `is_prosumer` is, the energy deviation would always be minor than the original energy, because it is a percentage (minor than 100) of the original energy. So `energy_kWh + deviation_kWh` would always be `> 0`. Sorry if I'm still missing something!

hannesdiedrich: We are talking about different things I believe. 
I was assuming we would allow `MAX_ENERGY_DEVIATION_PERCENT>100%` and only wanted to make it work so that `altered_energy` is always grater than 0 for PV and Load.
AFAIU, you want to make also a flipping sign possible and more frequent for HomeMeter (and Storage later). IMO this is not needed ATM. We only need some deviations for the settlement-market POC. And if the user really want to have consumption although production was forecasted for a HomeMeter from time to time, they could set the MAX_ENERGY_DEVIATION_PERCENT higher than 100%. 
Does it make sense? Did I get it right?

fievelk: Ah ok. I'm assuming we would not allow `MAX_ENERGY_DEVIATION_PERCENT > 100` , that's why this code wouldn't need to check for the type of device.
I don't want to flip the sign for HomeMeter/Storage, and at the moment the code won't allow it (unless `MAX_ENERGY_DEVIATION_PERCENT` is > 100).

So I suppose what is needed here is more information about the requirements:
- Should we allow MAX_ENERGY_DEVIATION_PERCENT > 100% ?

If not, then flipping the sign will not be possible (so the code shouldn't need to be changed). If yes, then we should only allow this for certain devices (like HomeMeter and storage).

hannesdiedrich: I know that Colin updated the requirements a bit. I like to use the opportunity to correct myself.
Of course it should not be `min(energy_kWh + deviation_kWh, 0)` but `max(energy_kWh + deviation_kWh, 0)` in order to be always above 0. Sorry. I shot to soon. :-)

src/d3a/models/strategy/load_hours.py
```diff
@@ -169,8 +169,23 @@ def update_state(self):
...
         self._delete_past_state()
 
+    def set_real_energy_of_last_market(self):
+        """Set the (simulated) actual energy of the device in the previous market slot."""
+        if self.area.current_market:
+            self._set_real_energy_kWh(self.area.current_market.time_slot)
+
+    def _set_real_energy_kWh(self, time_slot: DateTime) -> None:
+        """Set the (simulated) actual energy consumed by the device in a market slot."""
+        energy_forecast_kWh = self.state.get_desired_energy_Wh(time_slot) * 1000
```
hannesdiedrich, line 184: Just while I am seeing this: 
```suggestion
        energy_forecast_kWh = self.state.get_desired_energy_Wh(time_slot) / 1000.
```

fievelk: Oh wow, thanks for noticing this!

fievelk: Why do you recommend using the `.`?

src/d3a/models/strategy/utils.py
```diff
@@ -0,0 +1,38 @@
...
+
+You should have received a copy of the GNU General Public License along with this program. If not,
+see <http://www.gnu.org/licenses/>.
+"""
+import numpy as np
+
+from d3a.constants import DEFAULT_PRECISION, RELATIVE_STD_FROM_FORECAST_ENERGY
+
+
+def alter_energy(
```
Hassan754, line 21: The name gives impression that this function will `slightly` change something (the energy), however it basically receives energy value and `compute` the relative std from it without changing it, so a name starting by compute/calculate or any other option seem better choices imo

Hassan754 (commented):
Left some comments, In general it LGTM, I'll review it again once all issues/discussions are settled

src/d3a/models/state.py
```diff
@@ -56,17 +56,33 @@ def get_state(self) -> Dict:
 
     @abstractmethod
     def restore_state(self, state_dict: Dict):
-        pass
+        """Update the state of the device using the provided dictionary."""
```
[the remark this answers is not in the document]
fievelk: When a simulation is restarted and a previous area state was stored (I think in the database), the `restore_state` method is used to update all the internal variables of the State instance using those db values. I'm not sure about the perfect name for this function.

src/d3a/models/state.py
```diff
@@ -56,17 +56,33 @@ def get_state(self) -> Dict:
...
         return self.__class__.__name__
 
 
-class ConsumptionState(StateInterface):
+class ProsumptionInterface(StateInterface, ABC):
+    """Interface with common methods/variables shared by consumption and production devices."""
+
+    def __init__(self):
+        # Actual energy consumed/produced by the device at specific market slots
+        self._real_energy: Dict[DateTime, float] = {}
```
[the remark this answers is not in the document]
fievelk: I think the original reason for this structure is that we want to discourage having something like this in the strategies:
```python
self.state.real_energy.get(time_slot, 0.0)
```
Also, have a look at `set_desired_energy()`: in that case the setter is doing more than just setting a single variable. Doing it with public variables would be cumbersome and confusing in strategies, so we provide getters/setters to do it and to provide an interface that could be unified in the future.

Moreover, the State classes were already implemented and I thought it was better to be coherent rather than mixing up styles.

src/d3a/models/strategy/utils.py
```diff
@@ -18,19 +18,21 @@
 from d3a.constants import DEFAULT_PRECISION, RELATIVE_STD_FROM_FORECAST_ENERGY
 
 
-def alter_energy(
-        energy_kWh: float, relative_std: float = RELATIVE_STD_FROM_FORECAST_ENERGY) -> float:
+def compute_altered_energy(
+        energy_kWh: float, relative_std: float = RELATIVE_STD_FROM_FORECAST_ENERGY,
+        random_generator: np.random.Generator = np.random.RandomState()) -> float:
```
hannesdiedrich, line 23: Hmm, not sure but this will not generate the same results for the same random seed (global Setting) , right?

fievelk: This will not interfere with the global random seed (the one we set with `np.random.seed()`). But at the same time, this will not inherit the global seed (so the energy values will be completely random). The only way I could think of to change this is to refactor the global random seed completely (maybe using a singleton). However, placing such a singleton in `singletons.py` causes circular imports.

src/d3a/models/strategy/load_hours.py
hannesdiedrich: Hmm, good question. This i burned into my brain. I remember having issues with this and data types in my studies. But this was python2. I just checked and even if `self.state.get_desired_energy_Wh(time_slot)` returns an int, the devision always returns a floating point number. So, all good and I learned something. Thanks a lot!

src/d3a/models/strategy/utils.py
hannesdiedrich: You could add it to the GlobalConfig (and consequently to the config.py), no circular imports should present then.
However, again, this is probably an overkill. I think it is worth to ask Colin in the ticket whether the user wants to have reproducible results ( I fear that he/she does).

fievelk: Another way to go would be to just store a constant for the seed, and then replace all usages of numpy's `random` to use `np.random.default_rng(seed=constants.SEED)`

hannesdiedrich: ok! This would be the discussed additional PR right? IMO it will be worth it (if reproducible results are wanted)

fievelk: > However, again, this is probably an overkill. I think it is worth to ask Colin in the ticket whether the user wants to have reproducible results ( I fear that he/she does).

Sure, I'll ask Colin and if that's the case I'll create another subtask for it then!

src/d3a/models/strategy/load_hours.py
hannesdiedrich: Interesting how experience in pain while debugging results into habits (I always add a point to the divisor). But from now on I can save "the effort".

fievelk: Haha yes, I remember it also from the past. When you suggested it I thought that it was necessary again and my nightmares were coming back! :P

src/d3a/models/strategy/utils.py
fievelk: Update: Colin suggested to add a constant seed for the moment. If necessary, we can add new functionalities later on to allow users to change this seed.

src/d3a/models/state.py
```diff
@@ -56,17 +56,33 @@ def get_state(self) -> Dict:
...
+
+    def set_real_energy_kWh(self, energy_kWh: float, time_slot: DateTime) -> None:
+        """Set the actual energy consumed/produced by the device in the given market slot."""
+        self._real_energy[time_slot] = energy_kWh
+
+    def get_real_energy_kWh(self, time_slot: DateTime) -> float:
```
hannesdiedrich, line 80: I just realise, that my current story (D3ASIM-3537) is indeed dependant on this one, because of your changes in the state.py here. Unfortunately, there is no consistent naming for the real energy, but I would vote to name it `energy_measurement` actually. Here in this story these are just simulated. Do you agree?

src/d3a/models/strategy/utils.py
hannesdiedrich: Why cant we use the seed we already have in the simulation config?

hannesdiedrich: It was discussed offline to use a constant and separate seed for the `compute_altered_energy`.

src/d3a/models/state.py
fievelk: I agree about the naming issue. The problem is that all the previous values we had were "forecasts" but their names make it look like they're real energy readings.
That's why I was looking for a way to make this new energy look different from the others (and adding `_real_` to the name was a way to achieve this).
I'm worried that calling it `energy_measurement` will not make it clear that this is different from any other energy we currently have. For example, suppose you read:
- `self.available_energy`
- `self.energy_measurement`
How can you know that these two are actually talking about two very different types of energy (forecast vs real)?

hannesdiedrich: Yeah. I agree. I think it is time to change these. If I correctly remember, this is on the table since forever. This will make also other coders help more to understand that there is a big difference between the forecast and the measurement. What do you think @faizan2590 ?

hannesdiedrich: My proposal is the following renaming that we could put into a separate story:
`ConsumptionState._energy_requirement_Wh` --> `ConsumptionState._energy_consumption_forecast_kWh`
(this will contain a unit change and also includes renaming the corresponding set and get methods)
The PV state is already aligned with this.
With this you could name the `_real_energy ` into `_energy_consumption_measurement_kWh` and `_energy_production_measurement_kWh`
What do you think? Also tagging @faizan2590 again.

hannesdiedrich: For this PR here I would be strongly for adding the unit to the variable, too (so to name if `self._real_energy_kWh`)

fievelk: > For this PR here I would be strongly for adding the unit to the variable, too (so to name if self._real_energy_kWh)

I agree, will do it now.
I also think that the renaming you suggest would work. I am thinking about some bigger refactoring for the energy values (to create a simpler interface shared by the states) but this is just an idea for the moment.

src/d3a/models/strategy/utils.py
```diff
@@ -0,0 +1,40 @@
...
+see <http://www.gnu.org/licenses/>.
+"""
+import numpy as np
+
+from d3a.constants import DEFAULT_PRECISION, RELATIVE_STD_FROM_FORECAST_ENERGY
+
+
+def compute_altered_energy(
+        energy_kWh: float, relative_std: float = RELATIVE_STD_FROM_FORECAST_ENERGY,
+        random_generator: np.random.Generator = np.random.default_rng(seed=12)) -> float:
```
hannesdiedrich, line 23: I have one more doubt here though: Every time this function is called we initiate a new RNG. Doesn't this increase the computational effort? for the whole simulation?

hannesdiedrich (commented):
One last comment. Once addressed, LGTM

src/d3a/models/strategy/utils.py
fievelk: On one side, initializing this kind of instance should be very light so I don't think it would represent an issue.
But most importantly:
> Python’s default arguments are evaluated once when the function is defined, not each time the function is called.
(https://docs.python-guide.org/writing/gotchas/#what-actually-happens)

You can see it with this example:
```python
class MyClass:
   pass

def my_function(foo=MyClass()):
   """Print the ID of the actual object in memory"""
   print(id(foo))

>>> my_function()
4491117472

>>> my_function()
4491117472

>>> my_function()
4491117472
```

src/d3a/models/state.py
faizan2590: > My proposal is the following renaming that we could put into a separate story:
ConsumptionState._energy_requirement_Wh --> ConsumptionState._energy_consumption_forecast_kWh
(this will contain a unit change and also includes renaming the corresponding set and get methods)
The PV state is already aligned with this.
With this you could name the _real_energy into _energy_consumption_measurement_kWh and _energy_production_measurement_kWh
What do you think? Also tagging @faizan2590 again.


Sorry @hannesdiedrich for the late response, but I totally agree with your proposal.

Hassan754:
I have a general question, AFAIS the d3a-api-client will be able to send both `forecast` and `measurement`.
So are we allowing the client to send his measurement + we compute measurement on our own in the d3a? 
If both options should be supported, doesn't it make sense if the compute altered energy reside somewhere in the d3a-interface where both can use it ? (Assuming the client still isn't sending real values and decided for some reason to also simulate these)

fievelk:
> I have a general question, AFAIS the d3a-api-client will be able to send both `forecast` and `measurement`.
> So are we allowing the client to send his measurement + we compute measurement on our own in the d3a?
> If both options should be supported, doesn't it make sense if the compute altered energy reside somewhere in the d3a-interface where both can use it ? (Assuming the client still isn't sending real values and decided for some reason to also simulate these)

This PR does not consider external strategies (so it doesn't consider data sent through the d3a-api-client). For that part, you can take a look at this PR: https://github.com/gridsingularity/d3a/pull/1199. As far as I understand, if the client API doesn't send any real-measurement data, the implementation provided here (in `alter_energy()`) should be used automatically. So there should be no need to explicitly send simulated values from the client. I'm tagging @hannesdiedrich for confirmation about this.

hannesdiedrich:
> alter_energy



> > I have a general question, AFAIS the d3a-api-client will be able to send both `forecast` and `measurement`.
> > So are we allowing the client to send his measurement + we compute measurement on our own in the d3a?
> > If both options should be supported, doesn't it make sense if the compute altered energy reside somewhere in the d3a-interface where both can use it ? (Assuming the client still isn't sending real values and decided for some reason to also simulate these)
> 
> This PR does not consider external strategies (so it doesn't consider data sent through the d3a-api-client). For that part, you can take a look at this PR: #1199. As far as I understand, if the client API doesn't send any real-measurement data, the implementation provided here (in `alter_energy()`) should be used automatically. So there should be no need to explicitly send simulated values from the client. I'm tagging @hannesdiedrich for confirmation about this.

Very good point. In my PR https://github.com/gridsingularity/d3a/pull/1199 I used the `set_energy_measurement_kWh ` but did not think about the `_set_energy_measurement_kWh` in the device implementations that use `compute_altered_energy `. IMO the external strategy should disable this behaviour and use the measurements  sent by the client.

src/d3a/models/state.py
```diff
@@ -146,7 +162,7 @@ def get_desired_energy_Wh(self, time_slot, default_value=0.0):
         return self._desired_energy_Wh.get(time_slot, default_value)
 
 
-class ProductionState(StateInterface):
+class ProductionState(ProsumptionInterface):
```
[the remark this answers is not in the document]
fievelk: Sure. `PVState` needs to have access to `set/get_energy_measurement_kWh()` methods. These methods are defined in `ProsumptionInterface` and not in `StateInterface`.

src/d3a/models/state.py
```diff
@@ -56,17 +56,33 @@ def get_state(self) -> Dict:
...
+        # Actual energy consumed/produced by the device at specific market slots
+        self._energy_measurement_kWh: Dict[DateTime, float] = {}
+
+    def set_energy_measurement_kWh(self, energy_kWh: float, time_slot: DateTime) -> None:
+        """Set the actual energy consumed/produced by the device in the given market slot."""
+        self._energy_measurement_kWh[time_slot] = energy_kWh
+
+    def get_energy_measurement_kWh(self, time_slot: DateTime) -> float:
+        """Get the actual energy consumed/produced by the device in the given market slot."""
+        return self._energy_measurement_kWh.get(time_slot, 0.0)
```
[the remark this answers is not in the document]
fievelk: This PR is just for simulations without external connection, but I see your point.
However, what should happen when a device doesn't send its measured energy? Don't we apply any penalty? Also, can this scenario happen?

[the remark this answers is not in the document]
faizan2590: >  what should happen when a device doesn't send its measured energy

Thats a matter of regulatory requirement that could be extremely strict or have some relaxation to simply ignore if the device have missed some beats due to network connectivity.

> Don't we apply any penalty?

If later is the case as described in a possible scenario above then the device penalty is simply ignored for some time duration threshold

> Also, can this scenario happen?

I see it quite like to happen quite often at least in the early implementation phase of such systems.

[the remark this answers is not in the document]
fievelk: It looks like this aspect has not been discussed in the requirements. I'll add a comment on the ticket.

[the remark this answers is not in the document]
fievelk: From the ticket conversation:
> A None value should be fine, or however a data point is represented if there’s no forecast in our current implementation.
> We can use that for penalties or whatever we decide later as the mechanism.

I decided to use `None` for the moment. Keep in mind that this method is currently not called anywhere: I implemented it because it will be needed later on for the settlement market implementation. In that moment, we need to pay attention to this `None` value (to avoid `TypeError` exceptions).

src/d3a/models/strategy/utils.py
```diff
@@ -0,0 +1,40 @@
...
+License, or(at your option) any later version.
+
+This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without
+even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+General Public License for more details.
+
+You should have received a copy of the GNU General Public License along with this program. If not,
+see <http://www.gnu.org/licenses/>.
+"""
+import numpy as np
```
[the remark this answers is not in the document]
fievelk: Done.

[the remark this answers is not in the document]
fievelk: I've checked the run times with and without numpy to be sure. If you're interested:
```python
from timeit import timeit


number_of_runs = 1000000
energy_kWh = 1000
std = (50 * energy_kWh) / 100

python_gauss = timeit(
    "random.gauss(mu=energy_kWh, sigma=std)",
    setup=f"import random; energy_kWh={energy_kWh}; std={std}",
    number=number_of_runs
)
numpy_normal = timeit(
    "np.random.normal(loc=energy_kWh, scale=std)",
    setup=f"import numpy as np; energy_kWh={energy_kWh}; std={std}",
    number=number_of_runs
)

print(f"Python random.gauss:\t{python_gauss}s")
print(f"NumPy random.normal:  \t{numpy_normal}s")

# Python random.gauss:	0.5805347780000005s
# NumPy random.normal:  	2.389011778000011s
```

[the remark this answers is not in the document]
hannesdiedrich: Hmm, we already install numpy and for a good reason, to not fall into the rabbit hole that we had once when we were fixing the random seed. Is using numpy really using more resources than the `random.gauss()` ?
But more importantly: Is `random.gauss()` always creating the same random numbers, I remind you on our long discussion [here](https://github.com/gridsingularity/d3a/pull/1196#discussion_r681732859)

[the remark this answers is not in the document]
hannesdiedrich: Sorry, I did not see your last comment, we must have written them in parrelel.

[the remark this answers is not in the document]
fievelk: @hannesdiedrich in order not to alter the global seed, I'm using `random.Random(12)`. This is the same I was doing earlier with numpy: we create a new instance of a random number generator and we alter the seed only on this instance (not on the global `random` module). So it should have the same effect as before and it will always generate the same random numbers.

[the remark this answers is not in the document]
hannesdiedrich: Ok, great! Thank you!

src/d3a/models/state.py
```diff
@@ -56,17 +56,33 @@ def get_state(self) -> Dict:
...
 
-class ConsumptionState(StateInterface):
+class ProsumptionInterface(StateInterface, ABC):
+    """Interface with common methods/variables shared by consumption and production devices."""
+
+    def __init__(self):
+        # Actual energy consumed/produced by the device at specific market slots
+        self._energy_measurement_kWh: Dict[DateTime, float] = {}
+
+    def set_energy_measurement_kWh(self, energy_kWh: float, time_slot: DateTime) -> None:
```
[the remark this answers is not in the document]
fievelk: I agree with you, but I'd rather leave this to the context of the more specific ticket that has been opened for it (3487).

hannesdiedrich (approved):
LGTM

faizan2590 (approved):
LGTM

PR #1197, opened 2021-08-04
Feature/D3ASIM-3542: Redesign Myco matchers to homogenize interfaces between Internal and External matchers | Add Input validation

Hassan754:


src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -145,6 +160,8 @@ def _get_validated_recommendations(
         """Return a validated list of BidOfferMatch instances."""
         validated_recommendations = []
         for recommendation in recommendations:
+            if not BidOfferMatch.is_valid_dict(recommended_pair):
+                raise MycoValidationException
```
Hassan754, line 164: This is the actual task

integration_tests/two_sided_pay_as_clear_market.feature
```diff
@@ -27,7 +27,9 @@ Feature: Two sided pay_as_clear market tests
      When we run the d3a simulation on console with two_sided_pay_as_clear.default_2a for 2 hrs
      Then there are files with offers, bids & market_clearing_rate for every area
 
+  # TODO: re-enable in context of bug/D3ASIM-3534
   @slow
+  @disabled
```
hannesdiedrich, line 32: Hmm, so now an additional test is failing? Do you have a clue about the reason already ?

src/d3a/d3a_core/singletons.py
```diff
@@ -16,7 +16,7 @@
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from d3a.d3a_core.global_objects import ExternalConnectionGlobalStatistics
-from d3a.models.myco_matcher import MycoMatcher
+from d3a.models.myco_matcher.myco_matcher import MycoMatcher
```
hannesdiedrich, line 19: You could eliminate this additional route with adding an `__all__ = []` statement in the `myco_matcher.__init__` right?

src/d3a/models/myco_matcher/myco_matcher.py
```diff
@@ -0,0 +1,55 @@
...
+"""
+from typing import Dict, Optional
+
+from d3a.d3a_core.util import is_external_matching_enabled
+from d3a.models.myco_matcher import MycoExternalMatcher, MycoInternalMatcher
+from d3a.models.myco_matcher.myco_matcher_interface import MycoMatcherInterface
+
+
+class MycoMatcher(MycoMatcherInterface):
+    """Interface for market matching, set the matching algorithm and expose recommendations."""
```
hannesdiedrich, line 26: Probably copy paste (same doc string is in the MycoInternalMatcher).

src/d3a/models/myco_matcher/myco_matcher.py
```diff
@@ -0,0 +1,55 @@
...
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+from typing import Dict, Optional
+
+from d3a.d3a_core.util import is_external_matching_enabled
+from d3a.models.myco_matcher import MycoExternalMatcher, MycoInternalMatcher
+from d3a.models.myco_matcher.myco_matcher_interface import MycoMatcherInterface
+
+
+class MycoMatcher(MycoMatcherInterface):
```
hannesdiedrich, line 25: I needed some time and re-reading your PR description to understand was this class does.
IMO a class name like the following would make it more clear right away. But I may be the only one. Open for other proposals too.
```suggestion
class MycoMatcherBase(MycoMatcherInterface):
```

src/d3a/d3a_core/simulation.py
```diff
@@ -374,10 +373,13 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
...
-                        external_global_statistics.is_it_time_for_external_tick(
-                            current_tick_in_slot)):
-                    bid_offer_matcher.match_algorithm.publish_event_tick_myco()
+                        is_external_matching_enabled()):
+                    # If External matching is enabled, limit the number of ticks dispatched.
+                    if external_global_statistics.is_it_time_for_external_tick(
+                            current_tick_in_slot):
+                        bid_offer_matcher.event_tick()
+                else:
+                    bid_offer_matcher.event_tick()
```
hannesdiedrich, line 382: How about putting this into the `MycoMatcher.event_tick` so only `bid_offer_matcher.event_tick()` is left in here?

src/d3a/models/area/__init__.py
```diff
@@ -403,29 +402,16 @@ def tick(self):
...
-                bid_offer_pairs = bid_offer_matcher.get_matches_recommendations(data)
-                if not bid_offer_pairs:
-                    break
-                market.match_recommendations(bid_offer_pairs)
+    def _update_myco_matcher(self) -> None:
+        """Update the markets cache that the myco matcher will request"""
+        bid_offer_matcher.matcher.update_area_uuid_markets_mapping(
+            area_uuid_markets_mapping={self.uuid: {"markets": self.all_markets, "current_time": self.now}})
+        if not is_external_matching_enabled():
+            bid_offer_matcher.matcher.match_recommendations()
```
hannesdiedrich, line 414: This implementation violates the single responsibilty principle. Please refactor into separate functions (or just move 
``` python
if not is_external_matching_enabled():
            bid_offer_matcher.matcher.match_recommendations()
```
out of here. Thanks!

src/d3a/models/myco_matcher/myco_matcher_interface.py
```diff
@@ -0,0 +1,41 @@
...
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
+
+from typing import Dict
+
+
+class MycoMatcherInterface:
```
hannesdiedrich, line 22: Any reason why this is not an ABC class?

src/d3a/models/myco_matcher/myco_external_matcher.py
hannesdiedrich: Minor, but how about being more specific in order to know already from the Exception which validation failed?
```suggestion
            if not BidOfferMatch.is_valid_dict(recommended_pair):
                raise MycoValidationException("BidOfferMatch is not valid")
```

hannesdiedrich (commented):
In general it looks like a great enhancement. 
Left some comments. Will review again once addressed.

src/d3a/d3a_core/singletons.py
Hassan754: This does make sense,
However in the `MycoMatcher` I do import the Internal and External Matchers, so if I put the `MycoMatcher` in the __init__ besides these two I will have a circular imports issue

src/d3a/models/myco_matcher/myco_matcher.py
Hassan754: IMO `Base` could mean that this is a base class that others will extend, which is not the case here as this is more like a `wrapper` for `MycoInternalMatcher` and `MycoExternalMatcher`
What do you think about `MycoMatchersWrapper`?

hannesdiedrich: It is indeed a wrapper, I was confused by the fact that all three classes inherit from `MycoMatcherInterface` . However, l would not vote for adding "Wrapper" to the name. Leave it as is or somebody else has a better idea.

src/d3a/d3a_core/singletons.py
hannesdiedrich: I see. leave it as is then.

integration_tests/two_sided_pay_as_clear_market.feature
hannesdiedrich: Ah yes. Thanks for trying the slow tests from time to time!

src/d3a/models/myco_matcher/myco_matcher_interface.py
Hassan754: Not really, It's just that I did not want to force the implementation of the methods defined here, but I can mark the class as abstract

src/d3a/models/myco_matcher/myco_matcher.py
fievelk: The structure looks like a sort of [Bridge Pattern](https://refactoring.guru/design-patterns/bridge/python/example). Maybe we could change the names of the implementation classes (e.g. `MycoExternalMatcherImplementation`) and not the name of this `MycoMatcher` class.

---

One of the differences with the bridge pattern is that in our case it's not currently possible to do something like this:
```python
matcher_implementation = MycoExternalMatcher()
MycoMatcher(implementation=matcher_implementation)
```
(That's because we want to instantiate this as a singleton before we know its implementation).

Hassan754: @fievelk Thanks for the addition!
I thought about conditionally passing the selected `implementation` to the Myco matcher in the singleton definition and as you correctly stated this can't work because this is called even before the configurations are updated
Honestly we're not interested in having only one MycoMatcher instance, we're only interested in the one in the `singletons` module .
Although it doesn't make sense to have 2 Myco Matchers, there is no harm if this happens, if we reassign this variable the python gc should remove it anyway and if we create an instance in other places it won't bother us (this is not like instantiating two different connections instances for example)
Anw the current implementation makes it easy to alter the matcher (implementation) or matcher's properties (matching algorithm) in the runtime, for example we have the `activate` method and we could introduce another `reactivate-update or any other naming`
One extra benefit from wrapping Internal and External matchers is that we have full control flow, it's true both have the same interface but the callbacks are a bit different (External matcher's match_recommendations is called upon the arrive of a redis message)
Another example is the `event_tick`, for internal matchers we always want to forward the tick whereas for the external matchers we only forward a percentage of the ticks.

Anyway the naming `Implementation` sounds good to me if you both agree

fievelk: > Honestly we're not interested in having only one MycoMatcher instance, we're only interested in the one in the singletons module .

I wasn't suggesting to change the `MycoMatcher` class to turn it into a singleton. I was suggesting to create a `SingletonBidOfferMatcher` class (or any name) that would wrap the `MycoMatcher` and could be used as a singleton. In this way, the `MycoMatcher` could be instantiated with the implementation parameter. Anyway, this is not so important here.

Regarding the different implementation of `MycoExternal/InternalMatcher`, I'm not completely sure I understood your point. The implementation is of course different, but they can be used interchangeably just by setting them as `MycoMatcher().matcher`, right? If not, something might not be quite right, because we should be able to call the same methods on `MycoMatcher().matcher` regardless of the object that this variable contains (without using `if`s).

Hassan754: Of course they both have the same interfaces `event_tick` `event_market_cycle` `event_finish` `match_recommendations`, it's just that as I mentioned the example of dispatching the tick event, we don't want to dispatch **All** ticks to the External matcher so we have a check here, the other exception is that we do not want to call the `match_recommendations` of the external matcher from within the area tick event (we call it once we receive a redis command)

fievelk: I personally think this could be improved: we should not differentiate based on the content of `MycoMatcher().matcher`, otherwise this defeats the point of using a common interface.
For example, these lines:
```python
if (self.simulation_config.external_connection_enabled and
        is_external_matching_enabled()):
    # If External matching is enabled, limit the number of ticks dispatched.
    if external_global_statistics.is_it_time_for_external_tick(
            current_tick_in_slot):
        bid_offer_matcher.event_tick()
else:
    bid_offer_matcher.event_tick()
```
IMO should become:
```python
bid_offer_matcher.event_tick()
```
Then you would apply the `if` statements inside the `event_tick()` implementation.

Same reasoning goes for the `match_recommendations()` call (I couldn't find the point you were mentioning).

fievelk: BTW, this is the exact same thing that Hannes suggested in his comment https://github.com/gridsingularity/d3a/pull/1197#pullrequestreview-722358911

src/d3a/models/myco_matcher/myco_matcher.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def event_tick(self, **kwargs) -> None:
+        """Handler for the tick event."""
+        is_it_time_for_external_tick = kwargs.pop("is_it_time_for_external_tick", False)
+        if is_external_matching_enabled():
+            # If External matching is enabled, limit the number of ticks dispatched.
+            if is_it_time_for_external_tick:
+                self.matcher.event_tick(**kwargs)
+        else:
+            self.matcher.event_tick(**kwargs)
```
fievelk, line 61:
```suggestion
    def event_tick(self, **kwargs) -> None:
        """Handler for the tick event."""
        self.matcher.event_tick(**kwargs)
```
Then in `MycoExternalMatcher`:
```python
from d3a.d3a_core.singletons import external_global_statistics
def event_tick(self, **kwargs):
        if not external_global_statistics.is_it_time_for_external_tick:
            return

        data = {"event": ExternalMatcherEventsEnum.TICK.value}
        self.myco_ext_conn.publish_json(self._events_channel, data)
```
1. You can import the singleton directly and avoid passing arguments around (in this case)
2. You don't need to check `if is_external_matching_enabled():` because, if that is `True`, you would have already changed `self.matcher` to be a `MycoExternalMatcher` instance. So you just need to change the content of `MycoExternalMatcher.event_tick()`.

src/d3a/models/myco_matcher/myco_matcher.py
Hassan754: @fievelk Hannes initially suggested moving this check from within the simulation.py (caller code)
So I moved it inside the MycoMatcher, one major difference from the Bridge to notice is that in the MycoMatcher I also extend the interface and forwarding messages inside the matchers after having my own control logic in there 
This way, the caller code doesn't call `bid_offer_matcher.matcher.event_tick()`, it is enough to do `bid_offer_matcher.event_tick()` .

hannesdiedrich: I took some time to understand the Bridge Pattern but I fail to see the connection to our "Problem" here. 
I would agree that selecting the Mycomatcher already at initiation of the MycoMatcher class and not only when calling the activate method (why do we need this two-stage init here anyway?).
Apart from that I am having trouble to follow the discussion here. Could maybe somebody write a short summary what the discussion solutions are? Thanks

src/d3a/models/myco_matcher/myco_matcher_interface.py
hannesdiedrich: OK, no. I see now that this is more like a fallback and the single point of implementation for the `update_area_uuid_markets_mapping` that is used in both matcher cases. Right?
Then leave it as is. Thanks!

src/d3a/d3a_core/simulation.py
```diff
@@ -371,16 +371,10 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
                         external_global_statistics.update()
 
                 self.area.tick_and_dispatch()
+                bid_offer_matcher.event_tick(
+                    is_it_time_for_external_tick=external_global_statistics.
+                    is_it_time_for_external_tick(current_tick_in_slot))
```
hannesdiedrich, line 376: So you need to pass `is_it_time_for_external_tick` ? `external_global_statistics` is a singleton and you could import it in MycoMatcher right? (or is there a circular import again ?)

hannesdiedrich (commented):
One comment. Once the discussion about bridge pattern is agreed upon, LGTM

src/d3a/d3a_core/simulation.py
Hassan754: Right, there was a circular import error, I could import it locally but still I will need to pass the `current_tick_in_slot` anw, so I believe passing the flag makes more sense

fievelk: Regarding circular imports: this might be another reason to replace the variable in `singletons.py` with an actual singleton class. We could put it in another module (since `singletons.py` doesn't give information about the type of objects it contains, besides them generically being singletons). At that point, we could just have something like:
```python
from d3a.models.myco_matcher.myco_matcher_singleton import MycoMatcherSingleton
matcher = MycoMatcherSingleton()
```
This would create a new `MycoMatcher` when it doesn't already exist, and it will return the existing one if it exists (singleton).

src/d3a/models/myco_matcher/myco_matcher.py
Hassan754: I have adapted to the suggestions and removed the MycoMatcherInterface from the Myco matcher, this way it only serves as a bridge without having any logic control flow, let me know what you think

Hassan754: @hannesdiedrich we have a two stage init here because the get_setup() methods of simulations set the SpotMarketType and MatchingAlgorithm at runtime, this happens after the initialization of the Myco Matcher so in this case the myco matcher will always be initialized with the default settings (Internal matching with PAB )
This is why there is an activate method that will be called after the settings are loaded

src/d3a/d3a_core/simulation.py
Hassan754: @fievelk this is basically the same, the `simulation.py` would still import both `MycoMatcher` and `external_global_statistics`, the `MycoMatcher` will still import `ExternalMatcher` that in its turn will import the `external_global_statistics` and we fall back in the same maze, however as I stated we would have to pass an argument in any case (current_tick_in_slot), so why not passing the flag` is_time_for_external_tick` instead and save imports?

src/d3a/models/myco_matcher/myco_matcher.py
fievelk: Bridge pattern:
> The abstraction will be able to delegate some (sometimes, most) of its calls to the implementations object. Since all implementations will have a common interface, they’d be interchangeable inside the abstraction.

In our case I considered the "abstraction" to be `MycoMatcher`, and the implementations are `MycoExternalMatcher` and `MycoInternalMatcher`.
The discussion was about the fact that our implementations are not interchangeable (we couldn't just call `bid_offer_matcher.matcher.event_tick()` without adding `if` statements or custom function parameters).

src/d3a/d3a_core/simulation.py
fievelk: I don't think that the situation you described would generate a circular import. If you have the time you could give it a try.

Hassan754: Okay I'll give it a try but won't design a singleton Myco matcher, I'll just move the instantiation to another module as I believe both will have the same impact

Hassan754: If this succeeded, I'll either make it singleton or permanently move the instantiation somewhere else

hannesdiedrich: > Right, there was a circular import error, I could import it locally but still I will need to pass the `current_tick_in_slot` anw, so I believe passing the flag makes more sense

Indeed, we need to pass this anyway. Didn't think about it. So. If you don't succeed with removing the circular import, just keep as you proposed, IMO

Hassan754: It works!
Great, so we have two options here

1. From the `simulations.py` pass the `current_tick_in_slot` and in the external matcher import `external_global_statistics` and check if it is time for external tick (Move the Myco matcher singleton as in the suggestion you provided)
2. Directly pass the is_time_for_external_tick from `simulation.py` to external matcher (Leave the currentsingletons)

src/d3a/models/myco_matcher/myco_matcher.py
hannesdiedrich: Thanks a lot for this summary! So if my comment https://github.com/gridsingularity/d3a/pull/1197#discussion_r683374174 would be addressed, the bridge pattern could be applied and we decide which implementation is used in simulation.init and not in the Abstraction class, right?

Hassan754: The bridge pattern is already implemented, in the Myco Matcher we have a property `matcher`, so in the future we can create more matchers and still make sure they can fit inside the MycoMatcher and still the interface won't break
```python
class NewMatcher(MycoMatcherInterface):
# new way of doing things
```
In the MycoMatcher:
```python
            return MycoExternalMatcher()
        elif conditionX:
            return MycoInternalMatcher()
       else:
           return NewMatcher() # This is it, the codebase would still be working exactly as expected
```

fievelk: To get closer to a real Bridge patter we would need to avoid calling `bid_offer_matcher.matcher`: the `.matcher` should be a private variable (e.g. `_matcher` or `_implementation`) that is not seen from the outside, and `MycoMatcher` should provide just one unique interface that interact internally with `_matcher`.

> Thanks a lot for this summary! So if my comment #1197 (comment) would be addressed, the bridge pattern could be applied and we decide which implementation is used in simulation.init and not in the Abstraction class, right?

@Hassan754 I think that @hannesdiedrich means that it would be better to just pass the matcher implementation to the `MycoMatcher` init:
```python
# simulation.py
matcher_implementation = MycoExternalMatcher() if is_external_matching_enabled() else MycoInternalMatcher()
bid_offer_matcher = MycoMatcher(implementation=matcher_implementation)
```
I would personally vote for this, because your current solution (adding new `else` statements inside the `MycoMatcher`) break the Open/Closed SOLID principle:
> write your code so that you will be able to add new functionality without changing the existing code

To be honest, the choice of which myco matcher implementation to use should be made by yet another class if we want to be stricter :D

Hassan754: Indeed the last piece of separating the abstractions from the implementation is by defining MycoMatcher's unique interface, however AFAIS there are some places in the code we're accessing matcher.matching_algorithm.state `src/d3a/d3a_core/export.py:219` so I decided to keep the `matcher` property public ( However I very much believe that this implementation should be changed to make the design complete, but I decided not to interfere with these -at least now- as Faizan is in the same territory fixing the PAC exports and we're way out of the context of the ticket)

For the second point, yes it is not feasible due to the `activate` and possibly the option to change the matchers/matching algorithm in the runtime (as Events for example)
What I tried to do is the `get_matcher` static_method as a workaround to evitate creating an extra class, but if u believe this is inevitable, I'll create it . Thanks again for bearing with me !

fievelk: > What I tried to do is the get_matcher static_method as a workaround to evitate creating an extra class, but if u believe this is inevitable, I'll create it . Thanks again for bearing with me !

I don't think that right now this is necessary. I mentioned it because in your example you were adding another `else` to this method, and in that case I think it would make sense to have some separate logic for it. No need to do it now.

src/d3a/d3a_core/simulation.py
fievelk: Both are fine for it. Since the current tick must unfortunately be passed in any case, I think this wouldn't make a big difference. Maybe the 2nd solution would be lighter, but do what you think is best!

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -101,30 +123,29 @@ def publish_simulation_id(self, message):
         channel = "external-myco/simulation-id/response/"
         self.myco_ext_conn.publish_json(channel, {"simulation_id": self.simulation_id})
 
-    def publish_event_tick_myco(self):
+    def event_tick(self, **kwargs):
         """Publish the tick event to the Myco client."""
-
+        is_it_time_for_external_tick = kwargs.pop("is_it_time_for_external_tick", False)
+        # If External matching is enabled, limit the number of ticks dispatched.
+        if is_it_time_for_external_tick:
```
hannesdiedrich, line 130: This should be :
```suggestion
        if not is_it_time_for_external_tick:
```
right?

hannesdiedrich (commented):
One important comment, once addressed, LGTM

src/d3a/models/myco_matcher/myco_external_matcher.py
Hassan754: Indeed!

hannesdiedrich: Actually, this shows, that there should be a test for that. Could you please add one? Thanks a lot!

tests/test_myco_external_matcher.py
```diff
@@ -52,28 +52,30 @@ def test_publish_simulation_id(self):
...
 
-    def test_publish_event_tick_myco(self):
+    def test_event_tick(self):
         data = {"event": "tick"}
-        self.matcher.publish_event_tick_myco()
+        self.matcher.event_tick()
         self.matcher.myco_ext_conn.publish_json.assert_called_once_with(
             self.events_channel, data)
+        self.matcher.event_tick(is_it_time_for_external_tick=False)
+        assert self.matcher.myco_ext_conn.publish_json.call_count == 1
```
hannesdiedrich, line 61: Wait, if is it not time for an external tick, `self.matcher.myco_ext_conn.publish_json.call_count` should be 0, right?

Hassan754: We are calling the event tick above and testing wether it was called (it should)
So at this point, when we call it again it won't invoke the redis publisher but the counter is still 1 (not 2)

src/d3a/setup/myco_setup/external_myco.py
```diff
@@ -16,40 +16,40 @@ def get_setup(config):
...
-                        hrs_of_day=list(range(0, 24)),
-                        final_buying_rate=40)),
-                    Area("H1 Storage1", strategy=StorageStrategy(initial_soc=100,
-                                                                 battery_capacity_kWh=20)),
-                    Area("H1 Storage2", strategy=StorageStrategy(initial_soc=100,
-                                                                 battery_capacity_kWh=20)),
-                    Area("H1 PV1", strategy=PVStrategy(panel_count=4)),
-                    Area("H1 PV2", strategy=PVStrategy(panel_count=4)),
+                    Area("H1 General Load1",
+                         strategy=LoadHoursExternalStrategy(avg_power_W=200,
```
hannesdiedrich, line 22: Can you please explain why you used the external strategies here? AFAIU, this is not needed to run the external myco matcher right?

Hassan754: True, this was just in order to also connect the d3a-api-client

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -520,12 +520,14 @@ def _bid_aggregator(self, arguments: Dict):
...
             assert self.can_bid_be_posted(self.next_market.time_slot, **arguments)
 
-            replace_existing = arguments.get("replace_existing", True)
+            replace_existing = arguments.pop("replace_existing", True)
             bid = self.post_bid(
                 self.next_market,
                 arguments["price"],
                 arguments["energy"],
-                replace_existing=replace_existing)
+                replace_existing=replace_existing,
```
hannesdiedrich, line 528: So, we apparently dont have an integrationtest set up in the d3a-api-client that tests the requirements and attributes feature right? Can you please add a story for that and put it into the sprint?

tests/test_myco_external_matcher.py
hannesdiedrich: I see. Please add a comment that explains that. Thanks!

src/d3a/setup/myco_setup/external_myco.py
hannesdiedrich: Ok. We can not rely on manual testing, so I really vote for https://github.com/gridsingularity/d3a/pull/1197#discussion_r685760708

hannesdiedrich (approved):
Please add integrationtests for the requirements and attributes feature (or add a story for that) in the d3a-api-client. LGTM

Hassan754:
> Please add integrationtests for the requirements and attributes feature (or add a story for that) in the d3a-api-client. LGTM

https://gridsingularity.atlassian.net/browse/D3ASIM-3554

src/d3a/models/strategy/external_strategies/storage.py
Hassan754: https://gridsingularity.atlassian.net/browse/D3ASIM-3554, Yes I will start with it just after merging this

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -51,28 +69,32 @@ def publish_offers_bids(self, message):
         # IDs of markets (Areas) the client is interested in
         filtered_areas_uuids = filters.get("markets")
         market_offers_bids_list_mapping = {}
-        for area_uuid, markets in self.area_uuid_markets_mapping.items():
+        for area_uuid, area_data in self.area_uuid_markets_mapping.items():
             if filtered_areas_uuids and area_uuid not in filtered_areas_uuids:
```
faizan2590, line 73:
```suggestion
            if area_uuid not in filtered_areas_uuids:
```

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -51,28 +69,32 @@ def publish_offers_bids(self, message):
...
                 # Client is uninterested in this Area -> skip
                 continue
-            for market in markets:
+            for market in area_data["markets"]:
                 # Cache the market (needed while matching)
                 self.markets_mapping[market.id] = market
```
faizan2590, line 78: I know its been there from the old PR but are you clearing this cache after some time or after every market cycle?

faizan2590:
I noticed now while reviewing this PR, AFAIS [recommendation](https://github.com/gridsingularity/d3a/pull/1197/files#diff-3364cb14ec351a94609946ecf9d589fff9518fa491004616fdceedab745eba90R106) already is a list, if yes then why being concatenated into list again while passing it to `match_recommendations`

src/d3a/models/myco_matcher/myco_external_matcher.py
Hassan754: This will break the functionality, if the client did not send filtered areas, `filtered_areas_uuids` will be None and then we can skip the condition
If I remove this check it will be if area_uuid in None -> Error

src/d3a/models/myco_matcher/myco_external_matcher.py
Hassan754: I am not deleting them and I am not sure about the best place to do this, what do you think about using a lru cache here with maxsize?

Hassan754:
> I noticed now while reviewing this PR, AFAIS [recommendation](https://github.com/gridsingularity/d3a/pull/1197/files#diff-3364cb14ec351a94609946ecf9d589fff9518fa491004616fdceedab745eba90R106) already is a list, if yes then why being concatenated into list again while passing it to `match_recommendations`

The `recommendations` is indeed a list, however the `recommendation` (single instance) is not .
the match_recommendations expects a list of recommendations, this is why when passing single instance I'm parsing it to a list

src/d3a/models/myco_matcher/myco_matcher_interface.py
```diff
@@ -0,0 +1,44 @@
...
+        self.area_uuid_markets_mapping: Dict[str, Dict] = {}
+
+    def activate(self):
+        """Method to be called upon the activation of MycoMatcher."""
+
+    def update_area_uuid_markets_mapping(self, area_uuid_markets_mapping: Dict[str, Dict]) -> None:
+        """Interface for updating the area_uuid_markets_mapping of Myco matchers."""
+        self.area_uuid_markets_mapping.update(area_uuid_markets_mapping)
+
+    def match_recommendations(self, **kwargs) -> None:
```
faizan2590, line 34:
```suggestion
    @abstractmethod
    def match_recommendations(self, **kwargs) -> None:
```

faizan2590 (approved):
Left some comments, LGTM once addressed/answered. Would be waiting for integration tests in upcoming PR for better understanding of the workflow and its validation.

src/d3a/models/myco_matcher/myco_matcher_interface.py
Hassan754: I did not want to force the children to implement the methods, I only wanted them to have this property so the interface is shared between them

src/d3a/models/myco_matcher/myco_external_matcher.py
Hassan754: I have tested clearing the cache in the memory cycle and it seems to be working, adapted
Thanks!

src/d3a/models/myco_matcher/myco_matcher_interface.py
faizan2590: That is actually the backbone of myco matcher. Isn't it? Can you imagine any myco_matcher without having its own `match_recommendations` implementation?

src/d3a/models/myco_matcher/myco_external_matcher.py
faizan2590: Ohkay, I thought it would least be an empty list.

src/d3a/models/myco_matcher/myco_matcher_interface.py
Hassan754: You're right, for this specific function it absolutely makes sense to have it abstract

faizan2590 (approved):
LGTM

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -175,6 +176,8 @@ def _get_validated_recommendations(
                     f"Market with id {recommendation.get('market_id')} doesn't exist.")
             if market.readonly:
                 # The market is already finished
+                # TODO: we're clearing markets cache after each market cycle so is this check
+                # TODO: .. really relevant? (All finished markets should be deleted already)
```
hannesdiedrich, line 180: IMO you could safely `pass` for the `market.readonly` case. Or is it important for the client to receive an exception in this case?

hannesdiedrich (approved):
LGTM

src/d3a/models/myco_matcher/myco_external_matcher.py
Hassan754: It's a requirement , if any recommendation is to be placed in a finished market, we cancel the whole matching 
So either match all or don't match at all

PR #1198, opened 2021-08-05
 Revert d3a-interface installation in setup.py

Hassan754:
I have recently adapted to a suggestion to remove the d3a-interface installation from setup.py as it is already covered in the requirements
However, there is a check here that skips the requirements starting with -e prefix, so the d3a-interface was being skipped and the deploys were failing
I have reverted the original setup in this PR


hannesdiedrich (dismissed):
LGTM

PR #1200, opened 2021-08-16
Feature/d3 asim 3537 Added ability to send energy measurements via external strategies

hannesdiedrich:


src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -444,38 +442,101 @@ def _set_energy_forecast(self, payload):
                 IncomingRequest("set_energy_forecast", arguments,
                                 energy_forecast_response_channel))
 
-    def _set_energy_forecast_impl(self, arguments, response_channel):
+    def _set_energy_measurement(self, payload: Dict) -> None:
```
fievelk, line 445: Nit: if we want to follow SOLID principles, we should probably think about refactoring the `ExternalMixin` class (and the other strategies). The `_set_energy_forecast/measurement` methods, for example, are defined in the `ExternalMixin` and inherited in all the external strategies, even though they're not used (for example in the `StorageExternalMixin`).

fievelk: Since this method is called from other classes, I think it should be a public method (without the leading underscore). The same applies to other methods in the same class, in case you are doing some refactoring.

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -444,38 +442,101 @@ def _set_energy_forecast(self, payload):
...
+        else:
+            self.pending_requests.append(
+                IncomingRequest("set_energy_measurement", arguments,
+                                energy_measurement_response_channel))
+
+    @staticmethod
+    def _validate_values_positive_in_profile(profile: Dict) -> None:
+        """Validate whether all values are positive in a profile."""
+        for time_str, energy in profile.items():
+            assert energy >= 0.0, f"Energy is not positive for time stamp {time_str}."
```
fievelk, line 471: Nit: if this is dealing with external input (e.g. user input or client) it would be better to raise an exception instead of using assertions.

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -444,38 +442,101 @@ def _set_energy_forecast(self, payload):
...
+        Digest command from buffer and perform set_energy_forecast.
+        Args:
+            arguments: Dictionary containing "energy_forecast" that is a dict containing a profile
+                key: time string, value: energy in kWh
+            response_channel: redis channel string where the response should be sent to
+        """
         try:
-            assert arguments["energy_forecast"] >= 0.0
-            self.energy_forecast_buffer_Wh = arguments["energy_forecast"]
+            self._validate_values_positive_in_profile(arguments["energy_forecast"])
```
fievelk, line 482: I have the feeling that this won't work for Home Meter devices (since the profile values can be both positive and negative).

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: > Nit: if we want to follow SOLID principles, we should probably think about refactoring the `ExternalMixin` class (and the other strategies). The `_set_energy_forecast/measurement` methods, for example, are defined in the `ExternalMixin` and inherited in all the external strategies, even though they're not used (for example in the `StorageExternalMixin`).

Yeah, Hassan proposed to create yet another base class, but I like your proposal. Will add these to the `ExternalMixin`. thanks

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: This function is called in a try except block which will deal with sending the error back to the client.
You prepended `Nit:`, so I guess that I could skip it right?

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: Indeed it wont work. We dont have a external HomeMeter yet. Shall we change it in the frame of https://gridsingularity.atlassian.net/browse/D3ASIM-3525 ?

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -444,38 +442,101 @@ def _set_energy_forecast(self, payload):
...
+                response_channel,
+                {"command": "set_energy_measurement", "status": "ready",
+                 "transaction_id": arguments.get("transaction_id", None)})
+        except Exception:
+            logging.exception(f"Error when handling _set_energy_measurement_impl "
+                              f"on area {self.device.name}. Arguments: {arguments}")
+            self.redis.publish_json(
+                response_channel,
+                {"command": "set_energy_measurement", "status": "error",
+                 "error_message": f"Error when handling _set_energy_measurement_impl "
```
fievelk, line 523: Nit:
```suggestion
            error_message = ("Error when handling _set_energy_measurement_impl "
                             f"on area {self.device.name}. Arguments: {arguments}")
            logging.exception(error_message)
            self.redis.publish_json(
                response_channel,
                {"command": "set_energy_measurement", "status": "error",
                 "error_message": error_message,
                 "transaction_id": arguments.get("transaction_id", None)})
```

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -42,17 +40,17 @@ def __init__(self, *args, **kwargs):
...
-                f'{self.channel_prefix}/list_bids': self._list_bids,
+                f"{self.channel_prefix}/bid": self._bid,
+                f"{self.channel_prefix}/delete_bid": self._delete_bid,
+                f"{self.channel_prefix}/list_bids": self._list_bids,
                 }
 
     @property
     def filtered_bids_next_market(self) -> List[Dict]:
-        """Get a representation of each of the device's bids from the next market."""
+        """Get a representation of each of the device"s bids from the next market."""
```
fievelk, line 50:
```suggestion
        """Get a representation of each of the device's bids from the next market."""
```

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -398,35 +396,54 @@ def event_tick(self):
         for req in self.pending_requests:
             if req.request_type == "set_energy_forecast":
                 self._set_energy_forecast_impl(req.arguments, req.response_channel)
+            elif req.request_type == "set_energy_measurement":
+                self._set_energy_measurement_impl(req.arguments, req.response_channel)
 
         self.pending_requests = deque(
             req for req in self.pending_requests
-            if req.request_type not in "set_energy_forecast")
```
fievelk, line 404: How was this working before? O.o

src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -398,35 +396,54 @@ def event_tick(self):
...
+    def _clear_energy_buffers(self):
+        """Clear forecast and measurement buffers."""
+        self.energy_forecast_buffer = {}
+        self.energy_measurement_buffer = {}
+
+    def update_energy_forecast(self) -> None:
+        """Set energy forecast for future markets."""
+        for slot_time, energy_kWh in self.energy_forecast_buffer.items():
+            if slot_time >= self.area.next_market.time_slot:
+                self.state.set_desired_energy(energy_kWh * 1000, slot_time, overwrite=True)
```
fievelk, line 437: I think that the inconsistency between kWh and Wh is becoming a bit difficult to track here. It might be because neither `set_desired_energy` nor `energy_forecast_buffer` have the suffix with their unit, and they're using different units.
I think it's about time we homogenize units across all d3a to only always use kWh.

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -374,52 +371,76 @@ def __init__(
...
         super().event_tick()
 
     def _incoming_commands_callback_selection(self, req):
+        """Map commands to callbacks for forecast and measurement reading."""
         if req.request_type == "set_energy_forecast":
             self._set_energy_forecast_impl(req.arguments, req.response_channel)
+        elif req.request_type == "set_energy_measurement":
+            self._set_energy_measurement_impl(req.arguments, req.response_channel)
         else:
             super()._incoming_commands_callback_selection(req)
```
fievelk, line 413: Couldn't this method be used in `event_tick()` to decide which callback to call?

src/d3a/models/strategy/home_meter.py
```diff
@@ -189,6 +190,20 @@ def event_market_cycle(self):
 
         self._delete_past_state()
 
+    def set_energy_measurement_of_last_market(self):
```
fievelk, line 193: If this method is not supposed to be called from the outside, do you think we could mark it as private?

fievelk: This was introduced in my PR, but I think it could be changed. Anyway, not major.

src/d3a/models/strategy/external_strategies/__init__.py
fievelk: Wait, what would you add to the `ExternalMixin`? My point was exactly that the `ExternalMixin` contains too many methods and that these methods are not being used in all the classes that inherit the `ExternalMixin` (and that's not good) :D

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: Indeed and this is targeted in https://gridsingularity.atlassian.net/browse/D3ASIM-3487.

src/d3a/models/strategy/external_strategies/pv.py
hannesdiedrich: It is used in the `event_tick()`, have a look into the `PVExternalMixin`.

src/d3a/models/strategy/home_meter.py
hannesdiedrich: sorry, forgot to remove it (when I changed to your branch as target).

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: Upsi, sorry, read too fast. Here we start a gain the never ending discussion about whether and how to handle the ExternalMixin. What would be your proposal of refactoring the ExternalMixin? Having one mixing for each device type? 
In any case, lets create a separate story for that. This settlement market is dragging too long already, we need to get it done soon. OK?

src/d3a/models/strategy/external_strategies/__init__.py
fievelk: TL;DR: This is not blocking so you can skip it. But still quite important and it would be good to keep it in mind for the future.

The try/except block could still stay where it is. Assertions are for errors that should never happen in production, they're meant for developers (so that they know they have a bug in their code) and could/should fail during local development with a bad crash. Also, assertions can be disabled globally with a specific python flag.
Exceptions (with try/except) are errors that are expected, they're meant for production code and should be caught to avoid crashes and provide users with meaningful error messages.
In this case, we're validating the client input, so we expect it to raise exceptions in some cases. It's a normal thing and it does not depend on a bug in the code. That's why I'm suggesting to use exceptions.

src/d3a/models/strategy/external_strategies/__init__.py
fievelk: That's an option. The reason why I pointed this out here is that the `ExternalMixin` is defining methods that are incompatible (or not used) with the classes that inherit from it. That's what I also meant in the comment above: https://github.com/gridsingularity/d3a/pull/1200#discussion_r690140977

src/d3a/models/strategy/external_strategies/pv.py
fievelk: Sorry, I meant in `PVForecastExternalStrategy`.

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: OK, will add an exception here.

src/d3a/models/strategy/external_strategies/pv.py
hannesdiedrich: Sorry, I dont follow. Can you please explain some more? Thanks!

src/d3a/models/strategy/external_strategies/load.py
hannesdiedrich: Haha, indeed. I dont understand either. Maybe this was the reason why [this](https://github.com/gridsingularity/d3a/blob/30538c352420bc0b1e5b8b6ae2a2c3c2125a6620/src/d3a/models/strategy/external_strategies/pv.py#L389)  hat to be added.

src/d3a/models/strategy/external_strategies/pv.py
fievelk: The `event_tick()` method in `PVForecastExternalStrategy` doesn't call `_incoming_commands_callback_selection()`, but it looks like it duplicates its code.

hannesdiedrich: `PVForecastExternalStrategy._incoming_commands_callback_selection` extends the usage of `PVExternalMixin._incoming_commands_callback_selection`. We need to also call `PVExternalMixin.event_tick` (and the `PVExternalMixin._incoming_commands_callback_selection`) in the `PVForecastExternalStrategy.event_tick`.
 The duplication you see was added by Spyros, which also added a comment where he mentions that this is needed to make it work for the MQTT subscriber. I am willing to add a ticket that looks into the need for this duplication (which is probably not needed any more because of the fix i added [here](https://github.com/gridsingularity/d3a/pull/1200#discussion_r690172774). What do you think?

src/d3a/models/strategy/home_meter.py
hannesdiedrich: sorry, again not read correctly. I'll make it private. Thanks!

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: Ok, let me try something.

hannesdiedrich: Hmmm, I tried, but I can only think of a way that will lead to subclass explosion (that we discussed in the workshop). This is mainly due to the fact that we have this special architecture of subclassing in the external strategies.
Injecting the measurement and forecast endpoints into e.g. the `PVForecastExternalStrategy` is not pretty, because we need lots of functionality of `ExternalMixin` in there. 
If you can not think of a better way, I propose to keep it like it is now and maybe deal with refactoring all classes and inheritance in a dedicated story. What do you think?

src/d3a/models/strategy/external_strategies/pv.py
fievelk: I think that the comment is referring to the repetition of that operation, rather than to the duplication of the code. What I meant is that those lines could maybe be replaced by
```python
self._incoming_commands_callback_selection(req)
```
Anyway, this is not essential and I just noticed I should have prepended it with "Nit".

src/d3a/models/strategy/external_strategies/__init__.py
fievelk: I agree. I think that a wider refactoring would be good, even though it might need some discussion (it's quite a big topic).

src/d3a/models/strategy/external_strategies/pv.py
hannesdiedrich: Sorry, still do not get it. We need to take care of the req.type's `set_energy_forecast` and `set_energy_measurement` here. for all other req.type's  the `super()._incoming_commands_callback_selection(req)` will take care of . Right?

src/d3a/models/strategy/external_strategies/__init__.py
fievelk: Definitely: working on the `ExternalMixin` deserves a separate story (and probably some conversation to agree on a new structure). I'll create a task on Jira.

src/d3a/models/strategy/external_strategies/pv.py
fievelk: Maybe we're getting confused :D 
In `PVForecastExternalStrategy` we have:
```python
def event_tick(self):
    # ...
    for req in self.pending_requests:
        if req.request_type == "set_energy_forecast":
            self._set_energy_forecast_impl(req.arguments, req.response_channel)
        elif req.request_type == "set_energy_measurement":
            self._set_energy_measurement_impl(req.arguments, req.response_channel)
    # ...
    super().event_tick()
```
But it could be simplified to be:
```python
def event_tick(self):
    # ...
    for req in self.pending_requests:
        self._incoming_commands_callback_selection(req)
    # ...
    super().event_tick()
```
Because `_incoming_commands_callback_selection(req)` is defined inside `PVForecastExternalStrategy` and it already checks `set_energy_forecast` and `set_energy_measurement`. I hope this makes sense.

hannesdiedrich: @fievelk Please have a look at [this inline comment](https://github.com/gridsingularity/d3a/pull/1200/files#diff-44fccfc06172fd26d724ae1e5828f14d8e0a6e4d9dfa259090365298a9ba1410L389). @spyrostz added the duplication for a reason, that he maybe can explain better. I fear that this was a fix for this bug: https://github.com/gridsingularity/d3a/pull/1200#discussion_r690172774

spyrostz: The main problem here is that the event_tick of the PVExternalMixin base class is checking the `self.connected` member and does not trigger the call to `_incoming_commands_callback_selection` if no user has registered / connected to the PV before (that's what the connected flag controls, it essentially checks if a user has registered to this device before accepting commands). Unfortunately this whole model breaks for the MQTT subscriber, where we need to send measurements directly to the device without registering first, using an admin account and authenticating only in d3a-web / client api service. This is what the comment describes, that specifically for this case we have to circumvent the normal flow and deliberately accept energy forecasts (and measurements IMO, since we are in the topic) without relying on the super().event_tick() call.

fievelk: @spyrostz Thanks for the explanation about the comment and this part of the code, which is now clearer to me 👍

Still, I don't think this is connected to what we were discussing earlier. The refactoring I proposed doesn't change the functionality of the code. The comment can stay there, the parsing will happen as before. It's just that we would re-use code (already defined in `PVForecastExternalStrategy._incoming_commands_callback_selection()`) instead of duplicating it. If this doesn't make sense, please ignore my suggestion: it might be that I'm not seeing something here and this was just a nit :)

spyrostz: @fievelk no worries at all, I did not read carefully enough your comment and did not figure out that this was your exact suggestion, thanks for clarifying it! 
There is one tricky cornercase with your suggestion: `_incoming_commands_callback_selection` also uses super() https://github.com/gridsingularity/d3a/blob/96a69bf0ba8029991a87a0e397906105514feecb/src/d3a/models/strategy/external_strategies/pv.py#L405. This means that if we call this function from the PVForecastExternalStrategy event_tick method, all the commands will be processed immediately (including offer, bid ...) which is not the desirable outcome here (we only want the forecast and measurement commands to circumvent the normal process, not every command). Of course, the `_incoming_commands_callback_selection` method could be refactored to avoid this (e.g. by moving the if blocks for the forecast and measurement to their own method and calling this method twice), I am not sure if it will make the code more complex/unreadable than it already is.

tests/test_external_strategies.py
```diff
@@ -0,0 +1,119 @@
...
+
+from d3a.models.area import Area
+from d3a.models.strategy.external_strategies.load import LoadForecastExternalStrategy
+from d3a.models.strategy.external_strategies.pv import PVForecastExternalStrategy
+from d3a_interface.constants_limits import GlobalConfig
+from pendulum import now, Duration
+from pytest import mark
+
+
+def ext_strategy_fixture(strategy):
```
fievelk, line 28: I'm a bit confused: pytest fixtures are defined using the [`@pytest.fixture`](https://docs.pytest.org/en/6.2.x/fixture.html) decorator, but I don't see it here. This looks like a normal function (so we could use a naming like `generate_strategy` or similar.

A more proper alternative might be using `@pytest.fixture` with the `indirect` flag (so that you could still pass the `strategy` argument):
- [Indirect parametrization](https://docs.pytest.org/en/latest/example/parametrize.html#indirect-parametrization)
- [Pass a parameter to a fixture function](https://stackoverflow.com/a/33879151/1916588)

tests/test_external_strategies.py
```diff
@@ -0,0 +1,119 @@
...
+    config = MagicMock()
+    config.ticks_per_slot = 90
+    GlobalConfig.end_date = GlobalConfig.start_date + Duration(days=1)
```
fievelk, line 31: I fear that this would change the value of `GlobalConfig.end_date` for all the tests that are run after this. The reason is that at the moment this is not a fixture (in pytest's jargon) and there is no teardown method.

tests/test_external_strategies.py
```diff
@@ -0,0 +1,119 @@
...
+    area = Area(name="forecast_pv", config=config, strategy=strategy,
+                external_connection_available=True)
+    parent = Area(name="parent_area", children=[area])
+    parent.activate()
+    area.activate()
+    strategy.connected = True
+    market = MagicMock()
+    market.time_slot = GlobalConfig.start_date
+    parent.get_future_market_from_id = lambda _: market
+    area.get_future_market_from_id = lambda _: market
```
fievelk, line 41: Nit: I think the common way to handle this in tests is to use mocks. For example:
```suggestion
    parent.get_future_market_from_id = MagicMock(return_value=market)
    area.get_future_market_from_id = MagicMock(return_value=market)
```

tests/test_external_strategies.py
```diff
@@ -0,0 +1,119 @@
...
+    return strategy
+
+
+class TestPVForecastExternalStrategy:
+
+    @mark.parametrize("strategy", [ext_strategy_fixture(LoadForecastExternalStrategy()),
+                                   ext_strategy_fixture(PVForecastExternalStrategy())])
+    def test_clear_energy_buffers(self, strategy):
```
fievelk, line 49: Nit: private methods are usually not considered for unit-tests (of course you can have exceptions). If you want more information about it, you can check this talk: https://youtu.be/URSWYvyc42M?t=947 . From this point, she talks about private methods and why not to test them.

tests/test_external_strategies.py
```diff
@@ -0,0 +1,119 @@
...
+        strategy._clear_energy_buffers()
+        assert strategy.energy_forecast_buffer == {}
+        assert strategy.energy_measurement_buffer == {}
+
+    @mark.parametrize("strategy", [ext_strategy_fixture(LoadForecastExternalStrategy()),
+                                   ext_strategy_fixture(PVForecastExternalStrategy())])
+    def test_update_energy_forecast_measurement_is_called_on_market_cycle(self, strategy):
+        strategy.update_energy_forecast = Mock()
+        strategy.update_energy_measurement = Mock()
+        strategy._clear_energy_buffers = Mock()
```
fievelk, line 61: Why is it necessary to mock this private method?

tests/test_external_strategies.py
```diff
@@ -0,0 +1,119 @@
...
+    def test_clear_energy_buffers(self, strategy):
+        strategy.energy_forecast_buffer = {now(): 1}
+        strategy.energy_measurement_buffer = {now(): 1}
+        strategy._clear_energy_buffers()
+        assert strategy.energy_forecast_buffer == {}
+        assert strategy.energy_measurement_buffer == {}
+
+    @mark.parametrize("strategy", [ext_strategy_fixture(LoadForecastExternalStrategy()),
+                                   ext_strategy_fixture(PVForecastExternalStrategy())])
+    def test_update_energy_forecast_measurement_is_called_on_market_cycle(self, strategy):
```
fievelk, line 58: Nit: this test works and it's good to have it. I was just a bit mislead by the test name, because here you're actually testing `event_market_cycle`, not `update_energy_forecast/measurement`. 
One popular way to name test methods is to follow a similar schema: 
> `[MethodUnderTest]_[Scenario]_[ExpectedResult]`

For example: `test_event_market_cycle_calls_energy_update_methods()`. This is good because it immediately tells you which method you are testing, and for what. It makes it much easier to find tests for one method by using `test_<method-name>` in your code.

[In reality, you might want to test the entire `event_market_cycle` method (not just these calls) and that's one of the reasons why it's recommended to keep methods very short.]

tests/test_external_strategies.py
```diff
@@ -0,0 +1,119 @@
...
+        strategy.update_energy_forecast = Mock()
+        strategy.update_energy_measurement = Mock()
+        strategy._clear_energy_buffers = Mock()
+        strategy.event_market_cycle()
+        strategy.update_energy_forecast.assert_called_once()
+        strategy.update_energy_measurement.assert_called_once()
+        strategy._clear_energy_buffers.assert_called_once()
+
+    def test_update_forecast_calls_set_available_energy(self, strategy=ext_strategy_fixture(
```
fievelk, line 67:
```suggestion
    def test_update_energy_forecast_calls_set_available_energy(self, strategy=ext_strategy_fixture(
```

fievelk: Nit: since you're also testing `set_available_energy.assert_not_called()`, the method name is saying something a bit different than its content. Minor anyway.

tests/test_external_strategies.py
```diff
@@ -0,0 +1,119 @@
...
+        strategy.state.set_available_energy.assert_called_once_with(energy,
+                                                                    time, overwrite=True)
+        # do not call set_available_energy for time_slots in the past
+        time = strategy.area.next_market.time_slot.subtract(minutes=15)
+        strategy.energy_forecast_buffer = {time: energy}
+        strategy.state.set_available_energy.reset_mock()
+        strategy.update_energy_forecast()
+        strategy.state.set_available_energy.assert_not_called()
+
+    def test_update_forecast_calls_set_desired_energy_for(self, strategy=ext_strategy_fixture(
```
fievelk, line 83:
```suggestion
    def test_update_energy_forecast_calls_set_desired_energy(self, strategy=ext_strategy_fixture(
```
I'm not sure what the `_for` stood for, maybe it was a leftover?

src/d3a/models/strategy/external_strategies/pv.py
fievelk: Thanks! I did not think about the fact that processing the other commands wasn't desirable in this case! Thanks for the patience :D

tests/test_external_strategies.py
hannesdiedrich: I am still not confident with the answer to the question: "When to mock and when not to". As a rool of thumb I saved into my brain: "if you want to call `assert_called_once()`, you need a mock object.
I just tested it without the above mocks and it works without, as you suggested. But why?

tests/test_external_mixin.py
```diff
@@ -395,3 +401,141 @@ def test_restore_state(self, strategy):
...
+    ])
+    def test_set_energy_forecast(self, strategy):
+        transaction_id = str(uuid.uuid4())
+        strategy._get_transaction_id = Mock(return_value=transaction_id)
+        arguments = {"transaction_id": transaction_id,
+                     "energy_forecast": {now().format(DATE_TIME_FORMAT): 1}}
+        payload = {"data": json.dumps(arguments)}
+        assert strategy.pending_requests == deque([])
+        strategy._set_energy_forecast(payload)
+        strategy._get_transaction_id.assert_called_with(payload)
```
fievelk, line 417: Nit: it is not necessary to check this (as it is a private method and we should know nothing about it from the "outside").

tests/test_external_strategies.py
hannesdiedrich: Thanks for sharing the link to the talk! 
I really can not connect it to this special case her though. Why wouldn't I test this unit (function) ?
I need to make sure that the buffers are cleared afterwards. Or how would you test this behaviour?
BTW; does the coverage calculation know about the rule "prefer not to test private methods"?

tests/test_external_mixin.py
```diff
@@ -395,3 +401,141 @@ def test_restore_state(self, strategy):
...
+        assert len(strategy.pending_requests) > 0
+        energy_forecast_response_channel = f"{strategy.channel_prefix}/" \
+                                           "response/set_energy_forecast"
+        assert (strategy.pending_requests ==
+                deque([IncomingRequest("set_energy_forecast", arguments,
+                                       energy_forecast_response_channel)]))
+
+        strategy.redis.publish_json = Mock()
```
fievelk, line 425: Here you start a different set of tests. If you don't want to put them into a separate method, it would be great if you could add a comment to state how they differ from the previous ones from the lines above (i.e. the payload is now empty).

tests/test_external_strategies.py
hannesdiedrich: Understood and agreed. Thanks! Will change it.

tests/test_external_mixin.py
hannesdiedrich: Do you mean `_set_energy_forecast` ? IMO we should test it indeed. It is the callback for setting the energy_forecast.

fievelk: I mean that it's not necessary to test the call to `_get_transaction_id()`

tests/test_external_strategies.py
hannesdiedrich: I was expecting this comment. The problem is that you can not pass pytest.fixtures into `@mark.parametrize`. So i just "mocked" the fixture.

tests/test_external_strategies.py
hannesdiedrich: Ah, i was overreading `private`. I will change it

tests/test_external_strategies.py
hannesdiedrich: I had this name before but I spend like 15 mins in order to comply to pylint. I could not find a way to put this function header into multiple lines.

fievelk: I think this should work, right?
```python
    def test_update_energy_forecast_calls_set_available_energy(
            self, strategy=ext_strategy_fixture(PVForecastExternalStrategy())):
        """Some docstring."""
        time = strategy.area.next_market.time_slot
        # ...
```

hannesdiedrich: yeah, I realised also by myself (see my latest commits).

tests/test_external_strategies.py
fievelk: Suppose you have a private method (`_clear_energy_buffers()` in this case). This private method will be called inside a public method (e.g. `event_market_cycle()`). We only want to test the _public_ side-effects and the outputs of our public method*.
Therefore, when testing `event_market_cycle()`, you will also check that `self.energy_forecast_buffer` and `self.energy_measurement_buffer` are empty. As you see, we are testing the public side-effects of `event_market_cycle()` and we don't care about which private methods are involved in the process (we only care about the final effect).

*Sidenote: ideally, a method should only do one thing: either return an output, or change the state of the object (not both things). It's called ["Command-Query separation" principle](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation).

fievelk: Regarding the coverage calculation: the important thing is that your tests "hit" (execute) the code of the private methods. If you don't mock them, their code will be executed and should be therefore included in the coverage.

tests/test_external_strategies.py
fievelk: I think this is what the second link is doing: it uses just one fixture (with a parameter which would be the Strategy in our case). Or am I missing something?

tests/test_external_strategies.py
hannesdiedrich: yes, this is copy and paste code, will adapt (or remove). Thanks (forgot it until now)

tests/test_external_strategies.py
hannesdiedrich: I tried it with 
```python
from pytest import mark, fixture


@fixture
def ext_strategy_fixture(strategy):
    config = Mock()
    config.slot_length = duration(minutes=15)
    config.tick_length = duration(seconds=15)
    config.ticks_per_slot = 90
    config.start_date = GlobalConfig.start_date
    config.grid_fee_type = ConstSettings.IAASettings.GRID_FEE_TYPE
    config.end_date = GlobalConfig.start_date + duration(days=1)
    config.market_count = 1
    area = Area(name="forecast_pv", config=config, strategy=strategy,
                external_connection_available=True)
    parent = Area(name="parent_area", children=[area], config=config)
    parent.activate()
    area.activate()
    strategy.connected = True
    market = MagicMock()
    market.time_slot = GlobalConfig.start_date
    return strategy


class TestPVForecastExternalStrategy:

    @mark.parametrize("ext_strategy_fixture", [LoadForecastExternalStrategy(),
                                               PVForecastExternalStrategy()], indirect=True)
    def test_event_market_cycle_calls_energy_update_methods(self, ext_strategy_fixture):
        ext_strategy_fixture.energy_forecast_buffer = {now(): 1}
        ext_strategy_fixture.energy_measurement_buffer = {now(): 1}
        ext_strategy_fixture.update_energy_forecast = Mock()
        ext_strategy_fixture.update_energy_measurement = Mock()
        ext_strategy_fixture.event_market_cycle()
        ext_strategy_fixture.update_energy_forecast.assert_called_once()
        ext_strategy_fixture.update_energy_measurement.assert_called_once()
        assert ext_strategy_fixture.energy_forecast_buffer == {}
        assert ext_strategy_fixture.energy_measurement_buffer == {}
```

But got 
```
test setup failed
file /Users/hannesd/d3a/d3a/tests/test_external_strategies.py, line 52
      @mark.parametrize("ext_strategy_fixture", [LoadForecastExternalStrategy(),
                                     PVForecastExternalStrategy()], indirect=["ext_strategy_fixture"])
      def test_event_market_cycle_calls_energy_update_methods(self, ext_strategy_fixture):
file /Users/hannesd/d3a/d3a/tests/test_external_strategies.py, line 29
  @fixture
  def ext_strategy_fixture(strategy):
E       fixture 'strategy' not found
>       available fixtures: cache, called, capfd, capfdbinary, caplog, capsys, capsysbinary, doctest_namespace, ext_strategy_fixture, monkeypatch, pytestconfig, record_property, record_testsuite_property, record_xml_attribute, recwarn, requests_mock, tmp_path, tmp_path_factory, tmpdir, tmpdir_factory
>       use 'pytest --fixtures [testpath]' for help on them.
```
I dont understand really (not even the stackoverflow post. Sorry. Can you help?

spyrostz: I think all you have to do is change this `def ext_strategy_fixture(strategy):` to this `def ext_strategy_fixture():` . Fixture function arguments should be either other fixtures, or one of the predefined fixtures from pytest (e.g. `cache`, `called` ... ). `strategy` is not a symbol in this file as far as I see.

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -374,52 +372,76 @@ def __init__(
                          update_interval=update_interval,
                          energy_rate_decrease_per_update=energy_rate_decrease_per_update,
                          use_market_maker_rate=use_market_maker_rate)
-        self.energy_forecast_buffer_Wh = energy_forecast_Wh
+
+        # Buffers for energy forecast and measurement values,
+        # that have been sent by the d3a-api-client in the duration of one market slot
+        self.energy_forecast_buffer = None  # Dict[DateTime, float]
+        self.energy_measurement_buffer = None  # Dict[DateTime, float]
```
faizan2590, line 379: Since in type hint it is expected to be a dict with key as DateTime and a float value, why not initialising it with an empty dict?

tests/test_external_mixin.py
```diff
@@ -377,3 +401,160 @@ def test_restore_state(self, strategy):
...
+                                                        "Available parameters: (energy_forecast).",
+                                                        "transaction_id": transaction_id})
+        assert len(strategy.pending_requests) == 0
+
+    @parameterized.expand([
+        [ext_strategy_fixture(LoadForecastExternalStrategy())],
+        [ext_strategy_fixture(PVForecastExternalStrategy())]
+    ])
+    def test_set_energy_measurement(self, strategy):
+        # test successful call of set_energy_measurement:
```
faizan2590, line 442:
```suggestion
        # test successful call of set_energy_measurement.
```
NIT

tests/test_external_strategies.py
```diff
@@ -0,0 +1,125 @@
...
+from d3a_interface.constants_limits import GlobalConfig
+from pendulum import now, duration
+from pytest import mark
+
+
+def ext_strategy_fixture(strategy):
+    config = Mock()
+    config.slot_length = duration(minutes=15)
+    config.tick_length = duration(seconds=15)
+    config.ticks_per_slot = 90
```
faizan2590, line 33: correct me if I am wrong, it should be 60

tests/test_external_strategies.py
```diff
@@ -0,0 +1,125 @@
...
+    config.start_date = GlobalConfig.start_date
+    config.grid_fee_type = ConstSettings.IAASettings.GRID_FEE_TYPE
+    config.end_date = GlobalConfig.start_date + duration(days=1)
+    config.market_count = 1
+    area = Area(name="forecast_pv", config=config, strategy=strategy,
+                external_connection_available=True)
+    parent = Area(name="parent_area", children=[area], config=config)
+    parent.activate()
+    area.activate()
```
faizan2590, line 42: Correct me if im wrong, simply triggering activate for parent area should trigger activate for child areas too.

tests/test_external_strategies.py
hannesdiedrich: > I think all you have to do is change this `def ext_strategy_fixture(strategy):` to this `def ext_strategy_fixture():` . Fixture function arguments should be either other fixtures, or one of the predefined fixtures from pytest (e.g. `cache`, `called` ... ). `strategy` is not a symbol in this file as far as I see.

But how do I pass the strategy in there?

tests/test_external_mixin.py
hannesdiedrich: Well, here I want to show that this is done in the following (until the next inline comment appears).

tests/test_external_strategies.py
hannesdiedrich: We only need a number here and independent of the value. I will change it anyway as you are right.

tests/test_external_strategies.py
spyrostz: I just noticed that you need to use the strategy inside the fixture. In order to do that, you have 2 options: 

1. construct the strategy class inside the fixture. Bad, because it does not allow parameterization of the test (as far as I saw you use the constructed strategy object as a test parameter, therefore this will not work easily)
2. do not use a module-scope fixture for this, but use a class method, that will be called right at the start of the test and will set up the strategy the way you want it. 
3. Create your own fixture decorator that would receive the output of the parameterized decorator and set up the already constructed strategy. More complex, and without much gain in readability of the test. 

I would go with option #2

tests/test_external_strategies.py
hannesdiedrich: Indeed.

tests/test_external_strategies.py
fievelk: @hannesdiedrich in order to pass and use the parameterized strategy with indirect parameterization, you need to use the `request` object. I used the snippet you posted above and just changed the first lines:
```python
@fixture
def ext_strategy_fixture(request):
    strategy = request.param
    # ...
```
It should work.

fievelk: You can find more information on the documentation:
- [`parametrize`](https://docs.pytest.org/en/latest/reference/reference.html?highlight=request#pytest.Metafunc.parametrize)

Specifically these lines regarding the `indirect` parameter:
> **indirect** (_Union[bool, Sequence[str]]_) – A list of arguments’ names (subset of argnames) or a boolean. If True the list contains all names from the argnames. Each argvalue corresponding to an argname in this list will be passed as request.param to its respective argname fixture function so that it can perform more expensive setups during the setup phase of a test rather than at collection time.

hannesdiedrich: Thanks a lot for pointing me to the correct point in the documentation. I need to learn to be more patient here and not to stop reading when I think I found the solution and then wonder why it does not work.
I adapted your proposal here https://github.com/gridsingularity/d3a/pull/1200/commits/87a5e4f1df6a16aa130f3e3dd1a4ae18cec980bb

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -444,38 +443,102 @@ def _set_energy_forecast(self, payload):
...
+        for time_str, energy in profile.items():
+            if energy < 0.0:
+                raise ValueError(f"Energy is not positive for time stamp {time_str}.")
+
+    def set_energy_forecast_impl(self, arguments: Dict, response_channel: str) -> None:
+        """
+        Digest command from buffer and perform set_energy_forecast.
+        Args:
+            arguments: Dictionary containing "energy_forecast" that is a dict containing a profile
+                key: time string, value: energy in kWh
```
spyrostz, line 480: We might be obliged to also adapt the device measurement subscriber to support this, I'll do it in the context of D3ASIM-3549

hannesdiedrich: yes, you are right, I wanted to mention this in the PR description, but forgot.

hannesdiedrich: > I'll do it in the context of D3ASIM-3549

Thanks a lot!

tests/test_external_strategies.py
```diff
@@ -129,3 +134,97 @@ def test_update_energy_measurement_doesnt_call_set_energy_measurement_kWh_for_fu
         ext_strategy_fixture.state.set_energy_measurement_kWh = Mock()
         ext_strategy_fixture.update_energy_measurement()
         ext_strategy_fixture.state.set_energy_measurement_kWh.assert_not_called()
+
+    @pytest.mark.parametrize("ext_strategy_fixture", [LoadHoursExternalStrategy(100),
+                                                      StorageExternalStrategy()],
+                             indirect=True)
+    def test_receive_bid_successful(self, ext_strategy_fixture):
```
fievelk, line 141: I'm not sure I understand one thing: these tests are collected under the class `TestPVForecastExternalStrategy`, but they're actually testing other strategies' methods. In this specific case they're not even testing the strategy method, but rather the mixin method.

I would highly recommend to have a one-to-one match between implementation modules and test modules. For example:
- `external_strategies/pv.py` --> `tests/external_strategies/test_pv.py`
- `external_strategies/load.py` --> `tests/external_strategies/test_load.py`
- - `external_strategies/__init__.py` --> `tests/external_strategies/test_init.py` (note: in the future I would refactor the `__init__.py` module and move the code outside of it).

Sorry for not noticing this earlier. Feel free to ignore the comment if you want to just move on with the PR. But I think it's something to keep in mind.

src/d3a/models/strategy/external_strategies/__init__.py
spyrostz: I agree that these classes desperately need refactoring (especially with the fact that these methods are not supported by all subclasses), however I think that this change is still compatible with the HomeMeter device external strategy. keep in mind that these are methods that can be overridden. Therefore all the HomeMeterExternalStrategy has to do in theory, is to override these methods and provide its own implementation. I promise to also try to refactor this in the context of D3ASIM-3549 (again), where I have started to make these 2 methods a bit more DRY.

tests/test_external_strategies.py
hannesdiedrich: Ok, I agree with the proposal to have a one-to-one relationship between implementation and tests. 
in this specific case, it would lead to code duplication (I would have to implement the `test_receive_bid_successful` in both storage and load. I could create a function that could be reused in a `tests/external_strategies/utils.py`module. what do you think?

src/d3a/models/strategy/external_strategies/storage.py
```diff
@@ -46,15 +46,15 @@ def event_activate(self, **kwargs):
...
-            f'{self.channel_prefix}/list_offers': self._list_offers,
-            f'{self.channel_prefix}/bid': self._bid,
-            f'{self.channel_prefix}/delete_bid': self._delete_bid,
-            f'{self.channel_prefix}/list_bids': self._list_bids,
+            f'{self.channel_prefix}/offer': self.offer,
+            f'{self.channel_prefix}/delete_offer': self.delete_offer,
+            f'{self.channel_prefix}/list_offers': self.list_offers,
+            f'{self.channel_prefix}/bid': self.bid,
+            f'{self.channel_prefix}/delete_bid': self.delete_bid,
+            f'{self.channel_prefix}/list_bids': self.list_bids,
```
spyrostz, line 54: Very minor nit, but if you can also change the ' with " would be great, thanks!

fievelk:
Nit: I would still recommend using the naming scheme we discussed. At least, the name of the method could start with `test_<method-name>` to make it easier to understand which unit we're testing.

When I read `test_receive_bid_successful()` I think we're testing a method called `receive_bid()`. I would suggest to rename this to something like `test_bid()`, `test_bid_with_correct_payload_succeeds()`, `test_bid_succeeds()`, `test_bid_is_received_successfully()` or similar. The important part IMO is the `test_<method-name>` part :)

tests/test_external_mixin.py
```diff
@@ -1,31 +1,75 @@
...
 import unittest
 import uuid
-import json
-from unittest.mock import MagicMock
-from parameterized import parameterized
-from pendulum import now, Duration
+from collections import deque
+from unittest.mock import MagicMock, Mock
+
+import d3a.models.strategy.external_strategies
```
spyrostz, line 24: One more nit, please optimize the imports since you are here.

tests/test_external_mixin.py
```diff
@@ -377,3 +421,146 @@ def test_restore_state(self, strategy):
         assert strategy._connected is False
         assert strategy._use_template_strategy is False
         strategy.state.restore_state.assert_called_once_with(state_dict)
+
+
+class TestForecastRelatedFeatures:
+    @pytest.mark.parametrize("ext_strategy_fixture", [LoadForecastExternalStrategy(),
+                                                      PVForecastExternalStrategy()], indirect=True)
+    def test_set_energy_forecast(self, ext_strategy_fixture):
```
spyrostz, line 429: Nit again, but this test could be split into 2 IMO: test_set_energy_forecast_command_is_accepted and test_energy_forecast_command_fails_wrong_arguments . In addition to that, the first test could also include the check that the publish_json method has been called (because even on success we still respond via Redis IIRC).

tests/test_external_mixin.py
```diff
@@ -377,3 +421,146 @@ def test_restore_state(self, strategy):
...
+        arguments = {"transaction_id": transaction_id,
+                     "energy_forecast": {now().format(DATE_TIME_FORMAT): 1}}
+        response_channel = "response_channel"
+        ext_strategy_fixture.set_energy_forecast_impl(arguments, response_channel)
+        ext_strategy_fixture.redis.publish_json.assert_called_once_with(
+            response_channel, {"command": "set_energy_forecast",
+                               "status": "ready",
+                               "transaction_id": arguments["transaction_id"]})
+
+        # test if error is sent when wrong time format is provided
```
spyrostz, line 501: I still think it is better if these are split into separate tests, nit again though

tests/test_external_mixin.py
```diff
@@ -377,3 +421,146 @@ def test_restore_state(self, strategy):
...
+        arguments = {"transaction_id": transaction_id,
+                     "energy_measurement": {now().format(DATE_TIME_FORMAT): 1}}
+        response_channel = "response_channel"
+        ext_strategy_fixture.set_energy_measurement_impl(arguments, response_channel)
+        ext_strategy_fixture.redis.publish_json.assert_called_once_with(
+            response_channel, {"command": "set_energy_measurement",
+                               "status": "ready",
+                               "transaction_id": arguments["transaction_id"]})
+
+        # test if error is sent when wrong time format is provided
```
spyrostz, line 542: Same as above

spyrostz (dismissed):
Some minor comments, LGTM in general

tests/test_external_strategies.py
fievelk: You read my mind about code duplication :D 
One of the reasons for this is that the `bid()` method is duplicated. In this specific case, I would actually vote for keeping the duplication in the tests too, because we're not testing the same method (even though it is identical!).

spyrostz (dismissed):
LGTM

fievelk:
In the last commit you refactored tests into separate modules: that's great and I think it really improves the structure. However, I think it's necessary to rename the modules using the `test_` prefix. As I wrote in my comment above:

> - `external_strategies/pv.py` --> `tests/external_strategies/test_pv.py`
> - `external_strategies/load.py` --> `tests/external_strategies/test_load.py`

In fact, I believe pytest is currently not running these tests because of this.

hannesdiedrich:
> In the last commit you refactored tests into separate modules: that's great and I think it really improves the structure. However, I think it's necessary to rename the modules using the `test_` prefix. As I wrote in my comment above:
> 
> > * `external_strategies/pv.py` --> `tests/external_strategies/test_pv.py`
> > * `external_strategies/load.py` --> `tests/external_strategies/test_load.py`
> 
> In fact, I believe pytest is currently not running these tests because of this. You can check test discovery conventions here:
> 
> * https://docs.pytest.org/en/6.2.x/goodpractices.html#conventions-for-python-test-discovery
> 
> (I also think we don't need the `__init__.py` files in these packages, but I didn't check).

You are right. I adapted to this in my last commit.

fievelk (approved):

PR #1203, opened 2021-08-18
Removed mix of import/from statements for the same library (pendulum)…

spyrostz:
…. Fixed ordering on imports on multiple test files.

### DIFF
--- a/tests/test_area.py
+++ b/tests/test_area.py
@@ -15,26 +15,27 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-from pendulum import duration, today
+import unittest
 from collections import OrderedDict
 from unittest.mock import MagicMock
-import unittest
+
+from d3a_interface.constants_limits import ConstSettings, GlobalConfig
 from parameterized import parameterized
+from pendulum import duration, today
+
+from d3a import constants
+from d3a.constants import TIME_ZONE
+from d3a.d3a_core.device_registry import DeviceRegistry
 from d3a.events.event_structures import AreaEvent, MarketEvent
 from d3a.models.area import Area, check_area_name_exists_in_parent_area
+from d3a.models.area.event_dispatcher import AreaDispatcher
 from d3a.models.area.events import Events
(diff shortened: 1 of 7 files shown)


faizan2590 (dismissed):
LGTM

tests/test_area.py
```diff
@@ -15,26 +15,27 @@
...
 from collections import OrderedDict
 from unittest.mock import MagicMock
-import unittest
+
+from d3a_interface.constants_limits import ConstSettings, GlobalConfig
 from parameterized import parameterized
+from pendulum import duration, today
+
+from d3a import constants
+from d3a.constants import TIME_ZONE
```
fievelk, line 27: Based on the conversation we had, these two imports should be unified into one style, right?

tests/test_external_mixin.py
```diff
@@ -1,20 +1,22 @@
+import json
 import unittest
 import uuid
-import json
 from unittest.mock import MagicMock
+
+from d3a_interface.constants_limits import ConstSettings
+from d3a_interface.constants_limits import GlobalConfig
```
fievelk, line 7:
```suggestion
from d3a_interface.constants_limits import ConstSettings, GlobalConfig
```

tests/test_external_mixin.py
```diff
@@ -1,20 +1,22 @@
...
 from parameterized import parameterized
 from pendulum import now, Duration
+
+import d3a.models.strategy.external_strategies
+from d3a.constants import DATE_TIME_FORMAT
+from d3a.d3a_core.singletons import external_global_statistics
```
fievelk, line 13: I have the feeling that these imports are not optimized. Can it be?

fievelk (approved):
Just two smaller comments but nothing major.

tests/test_external_mixin.py
spyrostz: Thanks, you were right they were not optimized. Will do that too, thanks again!

tests/test_area.py
spyrostz: Indeed, will fix that, thanks!

fievelk (approved):

hannesdiedrich (approved):
LGTM

PR #1207, opened 2021-08-19
D3ASIM-3549: Explicitly read attributes and requirements when receivi…

spyrostz:
…ng a bid from the aggregator, instead of expanding all the dict arguments that arrived with the request. This will allow the code to ignore new arguments that arrive via the API client that do not fit to the post_bid method definition.

This change is needed in order for the integration tests of the api client to pass. Without this change the argument "type" (that denotes the command type) is also passed with the other request arguments, and later it is expanded in the post_bid method (which does not support this argument, thus raising an exception)


src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -284,7 +284,9 @@ def _bid_aggregator(self, arguments):
                 arguments["price"],
                 arguments["energy"],
                 replace_existing=replace_existing,
-                **arguments)
+                attributes=arguments.get("attributes", None),
+                requirements=arguments.get("requirements", None)
```
faizan2590, line 288:
```suggestion
                attributes=arguments.get("attributes"),
                requirements=arguments.get("requirements")
```

faizan2590 (approved):
LGTM

src/d3a/models/strategy/external_strategies/load.py
spyrostz: Indeed, will fix it in all occurrences, thanks!

hannesdiedrich (commented):
I am missing the callback mapping for the `set_energy_forecast` and `set_energy_measurement` in ÈxternalMixin.trigger_aggregator_commands`.
I don't think it works now. Can you please make sure by adding these commands to the integrationtest on the api client side? Thanks!

spyrostz:
> I am missing the callback mapping for the `set_energy_forecast` and `set_energy_measurement` in ÈxternalMixin.trigger_aggregator_commands`.
> I don't think it works now. Can you please make sure by adding these commands to the integrationtest on the api client side? Thanks!

My bad. I was under the impression that these commands would be added in the context of this PR https://github.com/gridsingularity/d3a/pull/1200/files (I have left a corresponding comment in the d3a-api-client PR). I'll add it in this PR then, thanks!

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -419,112 +423,90 @@ def _reject_all_pending_requests(self):
 
     def set_energy_forecast(self, payload: Dict) -> None:
         """Callback for set_energy_forecast redis endpoint."""
+        self.set_device_energy_data(payload, "set_energy_forecast", "energy_forecast")
+
+    def set_energy_measurement(self, payload: Dict) -> None:
+        """Callback for set_energy_measurement redis endpoint."""
+        self.set_device_energy_data(payload, "set_energy_measurement", "energy_measurement")
+
+    def set_device_energy_data(self, payload: Dict, command_type: str,
```
hannesdiedrich, line 432: I like this, but this will make the unittests fail (that I recently added).
I dont know how we should continue here. Do you want to adapt to my branch once again or wait until I pushed my changes to master

spyrostz: Hmmm CI did not complain, that's why I was rest assured that my changes did not break the test. Seems like this is not the case. 

At any case don't worry about it, I'll fix the tests too. Thanks for pinging me

PR #1211, opened 2021-08-24
Feature/d3 asim 3549

spyrostz:
Added dedicated aggregator commands for set energy forecast and measurement, since the individual commands will not work for the aggregator case.


hannesdiedrich (dismissed):
LGTM

src/d3a/setup/strategy_tests/external_devices.py
```diff
@@ -19,7 +19,8 @@
 from d3a.models.strategy.storage import StorageStrategy
 from d3a.models.strategy.load_hours import LoadHoursStrategy
 from d3a.models.strategy.external_strategies.pv import PVExternalStrategy
-from d3a.models.strategy.external_strategies.load import LoadHoursExternalStrategy
+from d3a.models.strategy.external_strategies.load import (LoadHoursExternalStrategy,
+                                                          LoadForecastExternalStrategy)
```
fievelk, line 23: Nit: imports can be optimized

tests/strategies/external/test_init.py
```diff
@@ -583,3 +585,38 @@ def test_set_energy_measurement_impl_fails_for_negative_energy(self, ext_strateg
...
                                "transaction_id": arguments["transaction_id"],
                                "error_message": error_message})
+
+    @pytest.mark.parametrize(
+        "ext_strategy, command_name",
+        itertools.product(
+            [LoadForecastExternalStrategy(), PVForecastExternalStrategy()],
+            ["set_energy_forecast", "set_energy_measurement"]
+        )
+    )
```
fievelk, line 595: Pytest now supports cartesian product out of the box (by stacking decorators):
```suggestion
    @pytest.mark.parametrize(
        "ext_strategy", [LoadForecastExternalStrategy(), PVForecastExternalStrategy()])
    @pytest.mark.parametrize("command_name", ["set_energy_forecast", "set_energy_measurement"])
```

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -486,7 +486,30 @@ def _set_device_energy_data_impl(self, arguments: Dict, response_channel: str,
                 response_channel,
                 {"command": command_type, "status": "error",
                  "error_message": error_message,
-                 "transaction_id": arguments.get("transaction_id", None)})
+                 "transaction_id": arguments.get("transaction_id")})
+
+    def _set_energy_forecast_aggregator(self, arguments):
```
fievelk, line 491: As discussed on Slack, I think we could consider the callbacks called by redis as public methods. They are basically exposed outside the class, and they are the ones whose side-effects deserve to be tested.
My suggestion is to make these methods public and test them (instead of directly testing `_set_device_energy_data_aggregator`, which is private).

fievelk:
One main comment regarding public methods and testing.

src/d3a/models/strategy/external_strategies/__init__.py
spyrostz: These are actually not Redis callbacks. These are methods that are called whenever we receive an aggregator batch command. Each batch command consists of multiple individual commands, and for each individual command one of the _set_*_aggregator method is called. The only thing that the class receives from the outside is the batch request, and these methods will never be called outside of the scope of the class. That's why it is important to keep them private, in order to not give the wrong idea to someone that he/she can call them from outside of the class. 

Having said all that, since this method incorporates logic that is not covered by other tests, it is better to test it independently. We can check it by sending a full batch command to the trigger_aggregator_commands method, however this will implicitly test the private method and not explicitly. If you would prefer it that way I can adapt the test though, let me know.

tests/strategies/external/test_init.py
spyrostz: Did not know that, thanks!

src/d3a/setup/strategy_tests/external_devices.py
spyrostz: Not according to Pycharm, these are already optimized.

src/d3a/models/strategy/external_strategies/__init__.py
fievelk: TL;DR: Thanks for the clear explanation! In this case keeping the method private is a deliberate design choice and it makes sense to keep it as it is.

More detailed answer below :) 
As far as I know, private methods should not be directly tested (of course there will always be exceptions) and it's better to test them indirectly. In this case, the `trigger_aggregator_commands` method seems to be the closer public method that can be tested; given the many `if` statements inside of it, this could easily become a pain. I personally usually see these situations as a flag for refactoring; for example, the `if`s could be delegated to a mapper class, and the individual aggregator methods like `_set_energy_forecast_aggregator` could be public methods of a separate class - or they could even be one class each - not reachable by the end-user.

src/d3a/setup/strategy_tests/external_devices.py
fievelk: Weird...pylint complains about them!

fievelk: > pylint: warning
> C0411 - third party import "from d3a_interface.constants_limits import ConstSettings" should be placed before "from d3a.models.area import Area" (wrong-import-order)

spyrostz: Ah sorry! I misread the filename, you are of course right, fixing it.

src/d3a/models/strategy/external_strategies/__init__.py
spyrostz: Great, I'll use the trigger_aggregator_commands for testing then and leave this for the pending external strategies refactoring that we have planned. Thanks!

fievelk (approved):

PR #1213, opened 2021-08-26
D3ASIM-3538: Added settlement market handling to area markets.

hannesdiedrich:
This is the expected lifecycle of normal past and settlement markets:
```
---- Current slot 2021-08-30T00:00 -----
Markets: ['2021-08-30T00:00']
Past Market: []
Settlement Markets: []
Past Settlement Markets: []

---- Current slot 2021-08-30T00:00 -----
Markets: ['2021-08-30T00:00']
Past Market: []
Settlement Markets: []
Past Settlement Markets: []

---- Current slot 2021-08-30T00:15 -----
Markets: ['2021-08-30T00:15']
Past Market: ['2021-08-30T00:00']
Settlement Markets: ['2021-08-30T00:00']
Past Settlement Markets: []

---- Current slot 2021-08-30T00:30 -----
Markets: ['2021-08-30T00:30']
Past Market: ['2021-08-30T00:00', '2021-08-30T00:15']
Settlement Markets: ['2021-08-30T00:00', '2021-08-30T00:15']
Past Settlement Markets: []

---- Current slot 2021-08-30T00:45 -----
Markets: ['2021-08-30T00:45']
Past Market: ['2021-08-30T00:00', '2021-08-30T00:15', '2021-08-30T00:30']
Settlement Markets: ['2021-08-30T00:00', '2021-08-30T00:15', '2021-08-30T00:30']
Past Settlement Markets: []

---- Current slot 2021-08-30T01:00 -----
Markets: ['2021-08-30T01:00']
Past Market: ['2021-08-30T00:00', '2021-08-30T00:15', '2021-08-30T00:30', '2021-08-30T00:45']
Settlement Markets: ['2021-08-30T00:00', '2021-08-30T00:15', '2021-08-30T00:30', '2021-08-30T00:45']
Past Settlement Markets: []

---- Current slot 2021-08-30T01:15 -----
Markets: ['2021-08-30T01:15']
Past Market: ['2021-08-30T00:15', '2021-08-30T00:30', '2021-08-30T00:45', '2021-08-30T01:00']
Settlement Markets: ['2021-08-30T00:15', '2021-08-30T00:30', '2021-08-30T00:45', '2021-08-30T01:00']
Past Settlement Markets: ['2021-08-30T00:00']

---- Current slot 2021-08-30T01:30 -----
Markets: ['2021-08-30T01:30']
Past Market: ['2021-08-30T00:30', '2021-08-30T00:45', '2021-08-30T01:00', '2021-08-30T01:15']
Settlement Markets: ['2021-08-30T00:30', '2021-08-30T00:45', '2021-08-30T01:00', '2021-08-30T01:15']
Past Settlement Markets: ['2021-08-30T00:15']
```

### DIFF
--- a/src/d3a/models/area/market_rotators.py
+++ b/src/d3a/models/area/market_rotators.py
@@ -53,6 +53,8 @@ class DefaultMarketRotator(BaseRotator):
             return False
         else:
             if ConstSettings.GeneralSettings.ENABLE_SETTLEMENT_MARKETS:
+                # if the settlement markets are enabled, the same amount as the active
+                # settlement markets has to be kept in the past_market buffer
                 return (time_slot < current_time_slot.subtract(
                     hours=ConstSettings.GeneralSettings.MAX_AGE_SETTLEMENT_MARKET_HOURS))
             else:


src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
+        """Move settlement markets to self.past_settlement_markets."""
+        move_settlement_market_slots = [
+            time_slot for time_slot in self.settlement_markets.keys()
+            if self._is_it_time_to_delete_past_market(current_time, time_slot)]
+        for time_slot in move_settlement_market_slots:
+            self.past_settlement_markets[time_slot] = self.settlement_markets.pop(time_slot)
+            self.log.debug(
+                f"Moving {self.past_settlement_markets[time_slot]} to past.")
+
+    def _delete_past_markets(self, current_time: DateTime) -> None:
```
Hassan754, line 129: So we have the following similar methods:

1. _delete_past_markets
2. _delete_past_balancing_markets
3. _delete_past_settlement_markets

The first two have the exact same implementation (but each one of them refer to a different set of markets), whereas the third one checks the `_is_it_time_to_delete_past_settlement_market` instead of the `_is_it_time_to_delete_past_market`

Can we refactor these three to have only one method responsible for deleting the past markets?
we may need to pass the relevant markets to this new method (either markets, balancing markets or settlement markets).
+ for the third case we may need to update the `_is_it_time_to_delete_past_market` to cover the case of the settlement markets.

Otherwise the code is linearly scaling up for each new market type that we will add, let me know what you think about that.

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
+                self.log.debug(
+                    f"Moving {self.past_balancing_markets[time_slot]} to past.")
+
+    def _move_settlement_markets_to_past_settlement(self, current_time: DateTime) -> None:
+        """Move settlement markets to self.past_settlement_markets."""
+        move_settlement_market_slots = [
+            time_slot for time_slot in self.settlement_markets.keys()
+            if self._is_it_time_to_delete_past_market(current_time, time_slot)]
+        for time_slot in move_settlement_market_slots:
+            self.past_settlement_markets[time_slot] = self.settlement_markets.pop(time_slot)
```
Hassan754, line 125: we should only have one editable settlement market (the latest), and the rest should be readonly, right ?

tests/area/test_markets.py
```diff
@@ -0,0 +1,132 @@
...
+        area_fixture.current_tick = slot_length_in_ticks * 2
+        area_fixture.cycle_markets()
+        assert len(area_fixture.past_markets) == 1
+        assert len(area_fixture.all_markets) == 5
+
+    @patch("d3a.constants.RETAIN_PAST_MARKET_STRATEGIES_STATE", True)
+    def test_market_rotation_is_successful_keep_past_markets(self, area_fixture):
+        area_fixture.activate()
+        assert len(area_fixture.all_markets) == 5
+        slot_length_in_ticks = area_fixture.config.slot_length / area_fixture.config.tick_length
```
Hassan754, line 70: IMO this is more understandable
```suggestion
        ticks_per_slot = area_fixture.config.slot_length / area_fixture.config.tick_length
```

Hassan754 (commented):

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
+            self.log.debug(
+                f"Moving {self.past_settlement_markets[time_slot]} to past.")
```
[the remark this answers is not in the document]
hannesdiedrich: This does not work, because the following error would be returned: `RuntimeError: dictionary changed size during iteration`

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
+            if self._is_it_time_to_delete_past_settlement_market(current_time, time_slot)]
+        for time_slot in delete_settlement_market_slots:
+            self._delete_market_and_all_attributes(self.past_settlement_markets, time_slot)
+
+    @staticmethod
+    def _delete_market_and_all_attributes(market_buffer: Dict, time_slot: DateTime) -> None:
+        """Delete market and all its attributes from the provided market_buffer."""
+        if ConstSettings.GeneralSettings.EVENT_DISPATCHING_VIA_REDIS:
+            market_buffer[time_slot].redis_api.stop()
+        del market_buffer[time_slot].offers
```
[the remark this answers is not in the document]
hannesdiedrich: This is the result of a long search for a memory leak some months ago. We found out that the garbage collector does not delete all attributes correctly. One possible explanation is that these attributes are also referenced between each other and on other places. Once there is a reference to an object , the GC will not delete it from memory.
So, this is an additional safety net that I would like to keep even. I hope it makes sense.

src/d3a/models/area/markets.py
hannesdiedrich: I am aware of the code duplication here and wanted to leave a comment about this, but forgot.
I decided to keep this because of [Pierpaolos earlier comment](https://github.com/gridsingularity/d3a/pull/1213#discussion_r696477792).
I also kept this way to have a better readability and remembered the workshop where we quoted Sandi Metz
`Duplication is far cheaper than the wrong abstraction`. This part of the code will hopefully not be touched anytime in the future (it wasn't for a long time until now). So IMO, we can live with the duplication for better readability and debugging. Happy to discuss this and hear other opinions.

src/d3a/models/area/markets.py
hannesdiedrich: No, the participants should be able to place bids and offers in all of the settlement markets as along as they are not in `past_settlement_markets.`

tests/area/test_markets.py
```diff
@@ -0,0 +1,132 @@
...
+        area_fixture.current_tick = slot_length_in_ticks
+        area_fixture.cycle_markets()
+        assert len(area_fixture.past_markets) == 1
+        assert len(area_fixture.all_markets) == 5
+
+        area_fixture.current_tick = slot_length_in_ticks * 2
+        area_fixture.cycle_markets()
+        assert len(area_fixture.past_markets) == 1
```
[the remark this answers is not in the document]
hannesdiedrich: yes, but we only keep one past market in the memory (as long as `RETAIN_PAST_MARKET_STRATEGIES_STATE = False`)

tests/area/test_markets.py
```diff
@@ -0,0 +1,85 @@
...
+        child = Area(name="test_market_area", config=config, strategy=StorageStrategy())
+        area = Area(name="parent_area", children=[child], config=config)
+        yield area
+
+        DeviceRegistry.REGISTRY = {}
+
+    def test_cycle_markets(self, area_fixture):
+        ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET = True
+        area_fixture.activate()
+        assert len(area_fixture.all_markets) == 5
```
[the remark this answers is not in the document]
hannesdiedrich: Because the market count is set to 5:
https://github.com/gridsingularity/d3a/pull/1213/files/967e4d58532940a78e334daf10e935b8af11a33f#diff-1a3117c0a30c4f92672a17fce97e08afb57e80744eb491247394ac5e43fed03fR45

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
+    def all_future_spot_markets(self) -> List:
+        """Return all future markets in a list."""
+        return list(self.markets.values())
 
-    @property
-    def all_spot_markets(self):
-        return list(self.markets.values()) + list(self.past_markets.values())
+    def _update_indexed_future_markets(self) -> None:
+        """Update the indexed_future_markets mapping."""
+        self.indexed_future_markets = {m.id: m for m in self.all_future_spot_markets}
```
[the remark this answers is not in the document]
hannesdiedrich: I would postpone this discussion until we implement the new behaviour of the futures market. For now I think the time_slot dependant buffering of the markets, balancing markets and settlement markets is more handy. 
But maybe you have a better argument to change this to ids? How does anybody else think about it?

src/d3a/models/area/markets.py
Hassan754: Of course duplication is cheaper than wrong abstraction, but accurate abstraction is far better than duplication as well :P so if we can agree on the abstraction it would be great, this way you can also reduce the number of code lines here + the unittests functions

src/d3a/d3a_core/util.py
```diff
@@ -472,3 +471,12 @@ def is_external_matching_enabled():
     """
     return (ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE ==
             BidOfferMatchAlgoEnum.EXTERNAL.value)
+
+
+def is_time_slot_in_past_markets(time_slot: DateTime, current_time_slot: DateTime):
+    """Checks if the time_slot should be in the area.past_markets."""
```
fievelk, line 477: Nit: I think this function might be refactored a bit for clarity. Given the method name, if the function returns `True` I would expect the `time_slot` to be _already_ in `area.past_markets`. Instead, we are actually checking whether `time_slot` _should be_ in `area.past_markets` (but it's probably not there yet).

Maybe a different naming would help (e.g. `is_market_slot_expired` or similar). What do you think?

fievelk: Still regarding this function: I think it would be cool to have classes for it. I am thinking about an interface like the following (as an example):
```python
expiration_timer = SettlementMarketExpirationTimer() if ... else DefaultExpirationTimer()
# Then in the strategy you would have
state = LoadState(market_expiration_timer=expiration_timer)
```
Just an idea, not necessary for this PR.

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
-import d3a.constants
+from d3a.models.market import GridFee, Market
+from d3a.models.market.balancing import BalancingMarket
+from d3a.models.market.one_sided import OneSidedMarket
+from d3a.models.market.two_sided import TwoSidedMarket
+from d3a_interface.constants_limits import ConstSettings, GlobalConfig
+from pendulum import DateTime
+
+if TYPE_CHECKING:
+    from d3a.models.area import Area
```
fievelk, line 31: I think this is not needed if you use `"Area"` as a string, as you're doing below.

src/d3a/d3a_core/util.py
spyrostz: I would vote for `is_timeslot_before_spot_and_settlement_markets`  but it is overly verbose

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
         self.markets = OrderedDict()  # type: Dict[DateTime, Market]
         self.balancing_markets = OrderedDict()  # type: Dict[DateTime, BalancingMarket]
         # Past markets
         self.past_markets = OrderedDict()  # type: Dict[DateTime, Market]
         self.past_balancing_markets = OrderedDict()  # type: Dict[DateTime, BalancingMarket]
-        self._indexed_future_markets = {}
+        # Settlement markets
+        self.settlement_markets = OrderedDict()  # type: Dict[DateTime, Market]
+        self.past_settlement_markets = OrderedDict()  # type: Dict[DateTime, Market]
+        self.indexed_future_markets = {}
```
fievelk, line 48: What's the `indexed_` prefix stating?
For example, is there a difference between `self.past_settlement_markets` and `self.indexed_future_markets`? If not, I would suggest to just name this variable as the others:
```suggestion
        self.future_markets = {}
```
(And maybe even using `OrderedDict()` as with the other ones.

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
 
     @property
-    def indexed_future_markets(self):
-        return self._indexed_future_markets
+    def all_future_spot_markets(self) -> List:
+        """Return all future markets in a list."""
+        return list(self.markets.values())
```
fievelk, line 53: I don't think converting to `list` is needed here (it probably only adds some overhead).

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
-        self._indexed_future_markets = {}
+        # Settlement markets
+        self.settlement_markets = OrderedDict()  # type: Dict[DateTime, Market]
+        self.past_settlement_markets = OrderedDict()  # type: Dict[DateTime, Market]
+        self.indexed_future_markets = {}
 
     @property
-    def indexed_future_markets(self):
-        return self._indexed_future_markets
+    def all_future_spot_markets(self) -> List:
```
fievelk, line 51: Nit:
```suggestion
    def future_spot_markets(self) -> List:
```
Given the name, I would assume that it will return "all" the future spot markets in any case.

src/d3a/models/area/markets.py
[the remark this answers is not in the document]
spyrostz: Timeslot indexing is necessary IMO, and much more helpful than id-indexing. The only reason that it exists here is for performance improvement, because the operation of searching a market with an id is quite frequent therefore there was a significant performance gain from adding the indexed version. I agree with Hannes though that once we add the futures market this will become obsolete, therefore we should remove it then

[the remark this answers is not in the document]
fievelk: I would say that the important thing is to keep these markets variables consistent with each other. If we do so, we'll be able to unify their implementation in the future much more easily.

src/d3a/models/area/markets.py
fievelk: I would personally refactor it. However, if you think that refactoring these methods is not worth it that's fine. Still, I think it would be better to have them all in one place (e.g. all in `util.py`). That's because they're doing the same thing, and if we decide to refactor them in the future it will be useful to have them all in one place at least.

src/d3a/models/area/markets.py
```diff
@@ -101,37 +175,52 @@ def select_market_class(is_spot_market):
...
-        self._indexed_future_markets = {
-            m.id: m for m in self.all_future_spot_markets
-        }
+        self._update_indexed_future_markets()
         return changed
+
+    def create_settlement_market(self, time_slot: DateTime, area: "Area") -> None:
+        """Create a new settlement market."""
+        self.settlement_markets[time_slot] = (
+            self._create_market(market_class=self._select_market_class(is_spot_market=True),
```
spyrostz, line 202: I have changed it in my PR, however IMO the settlement market is always double-sided by definition

spyrostz (dismissed):

src/d3a/models/area/markets.py
spyrostz: Indexed future markets is used to denote that it is an id-indexed dict. Reason why we added this was a performance improvement because the operation of searching a market by id was quite frequent in the futures markets and maintaining an indexed dict had a significant performance improvement. 
Hannes already answered in another comment that will be dealt with the new implementation of the futures market, which will migrate multiple futures markets to one, thus rendering this indexing obsolete.

src/d3a/models/area/markets.py
spyrostz: I would agree to refactor them, I would not agree to move these to util.py. This inherently belongs to the market class, and taking it out of the way because it looks boilerplateish is not the right way IMO

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
+        self._move_markets_to_past(current_time)
+        self._delete_past_markets(current_time)
+
+        if ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET:
+            self._move_balancing_markets_to_past_balancing(current_time)
+            self._delete_past_balancing_markets(current_time)
+
+        if ConstSettings.GeneralSettings.ENABLE_SETTLEMENT_MARKETS:
+            self._move_settlement_markets_to_past_settlement(current_time)
+            self._delete_past_settlement_markets(current_time)
```
fievelk, line 93: Nit: this seems to be a good place for refactoring. All these blocks do two operations: "move to past" and "delete" markets. So this looks like a good place to implement an interface with 3 different implementations. E.g.: `MarketsRotator`, `BalancingMarketsRotator`, `SettlementMarketsRotator` etc.:
```python
# In the init
self.markets_rotator = MarketsRotatorMapper.get_market_rotator()
# And later on
self.markets_rotator.rotate()
```
Just an idea.

src/d3a/models/area/markets.py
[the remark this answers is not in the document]
fievelk: I think it would be worth adding your explanation to the docstring!

src/d3a/models/state.py
```diff
@@ -72,14 +72,52 @@ class ProsumptionInterface(StateInterface, ABC):
     def __init__(self):
         # Actual energy consumed/produced by the device at specific market slots
         self._energy_measurement_kWh: Dict[DateTime, float] = {}
+        self._unsettled_deviation_kWh: Dict[DateTime, float] = {}
+        self._forecast_measurement_deviation_kWh: Dict[DateTime, float] = {}
+
+    @abstractmethod
+    def _calculate_unsettled_energy_kWh(self, measured_energy_kWh, time_slot):
```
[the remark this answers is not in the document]
fievelk: Since this method is overridden by children classes, it should probably become public.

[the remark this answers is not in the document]
fievelk: Also, `LoadState` and `PVState` are mostly kept for backwards-compatibility reasons. So I have the feeling that `_calculate_unsettled_energy_kWh` can actually be overridden directly within `ConsumptionState` and `ProductionState`, and maybe even declared as abstract as Hannes was suggesting. What do you think?

src/d3a/models/state.py
```diff
@@ -72,14 +73,50 @@ class ProsumptionInterface(StateInterface, ABC):
...
     def get_energy_measurement_kWh(self, time_slot: DateTime) -> float:
         """Get the actual energy consumed/produced by the device in the given market slot."""
-        return self._energy_measurement_kWh.get(time_slot, None)
+        return self._energy_measurement_kWh.get(time_slot)
+
+    def get_forecast_measurement_deviation_kWh(self, time_slot: DateTime) -> float:
+        """Get the energy deviation of forecasted energy from measurement by the device in
+        the given market slot. Negative value means that the deviation is beneficial to the
+        grid (and can be posted as an offer), positive value means that the deviation is
+        detrimental to the grid (and can be posted as a bid)"""
```
fievelk, line 99: Nit: in order to make this compliant with multi-line docstrings ([PEP 257](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings)), this should be refactored.

src/d3a/models/state.py
```diff
@@ -72,14 +73,50 @@ class ProsumptionInterface(StateInterface, ABC):
...
+        return self._forecast_measurement_deviation_kWh.get(time_slot)
+
+    def should_post_bid(self, time_slot: DateTime) -> bool:
```
fievelk, line 102: The `should_post_bid` and `should_post_offer` methods should not belong to the `ProsumptionInterface`. As stated by the docstring, `ProsumptionInterface` is an
> Interface with common methods/variables shared by consumption and production devices.

The `should_post_offer` can't be used by consumption devices, and `should_post_bid` can't be used by production devices, so they would rather belong to `ConsumptionState` and `ProductionState`.

src/d3a/models/state.py
```diff
@@ -72,14 +73,50 @@ class ProsumptionInterface(StateInterface, ABC):
...
+
+    def should_post_offer(self, time_slot: DateTime) -> bool:
+        return self._forecast_measurement_deviation_kWh.get(time_slot) < 0.0
+
+    def get_unsettled_deviation_kWh(self, time_slot: DateTime) -> float:
+        """Get the unsettled energy deviation of forecasted energy from measurement by the device
+        in the given market slot."""
+        return self._unsettled_deviation_kWh.get(time_slot)
+
+    def decrement_unsettled_deviation(
```
fievelk, line 113: I'm not sure whether this can be used by both production and consumption devices. If so, disregard this comment. If not, it should be moved into the respective class (either `ProductionState` or `ConsumptionState`).

src/d3a/models/state.py
```diff
@@ -72,14 +73,50 @@ class ProsumptionInterface(StateInterface, ABC):
...
+            self, purchased_energy_kWh: float, time_slot: DateTime) -> None:
+        """Decrease the device unsettled energy in a specific market slot."""
+        self._unsettled_deviation_kWh[time_slot] -= purchased_energy_kWh
+        assert self._unsettled_deviation_kWh[time_slot] >= -FLOATING_POINT_TOLERANCE, (
```
fievelk, line 117: I suggest to make this potentially more flexible, in case we decide to start using +/- values for consumption/production (as we discussed some months ago):
```suggestion
        assert abs(self._unsettled_deviation_kWh[time_slot]) >= FLOATING_POINT_TOLERANCE, (
```

src/d3a/models/state.py
```diff
@@ -72,14 +73,50 @@ class ProsumptionInterface(StateInterface, ABC):
...
+            f"Unsettled energy deviation fell below zero "
```
fievelk, line 118:
```suggestion
            "Unsettled energy deviation fell below zero "
```

src/d3a/models/state.py
```diff
@@ -237,6 +274,18 @@ class PVState(ProductionState):
     Completely inherits ProductionState, but we keep this class for backward compatibility.
     """
 
+    def _calculate_unsettled_energy_kWh(self, measured_energy_kWh, time_slot):
+        """
+        Returns negative values for overproduction (offer will be placed on the settlement market)
+        and positive values for underproduction (bid will be placed on the settlement market)
+        :param measured_energy_kWh: Measured energy that the PV produced
+        :param time_slot: time slot of the measured energy
+        :return: Deviation between forecasted and measured energy
```
fievelk, line 283: Nit: please use Google-style docstrings instead of reStructuredText-style.

tests/area/test_markets.py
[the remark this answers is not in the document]
fievelk: IMO, this is worth a comment :)

src/d3a/d3a_core/util.py
hannesdiedrich: > Still regarding this function: I think it would be cool to have classes for it. I am thinking about an interface like the following (as an example):
> 
> ```python
> expiration_timer = SettlementMarketExpirationTimer() if ... else DefaultExpirationTimer()
> # Then in the strategy you would have
> self.state = LoadState(market_expiration_timer=expiration_timer)
> # And in the State methods
> if self.expiration_timer.is_expired(time_slot):
>     do_something()
> ```
> 
> This would unify the state interface for all kinds of markets (for the expiration checks, at least).
> Just an idea, not necessary for this PR.

This looks like a lot of Boilerplate code for a simple comparison. Not sure if the proposed `SettlementMarketExpirationTimer ` will be used in the future.

tox.ini
```diff
@@ -76,7 +76,7 @@ commands =
     pip-sync requirements/tests.txt requirements/pandapower.txt
     pip install -e .
 	pip uninstall -y d3a-interface
-	pip install git+https://github.com/gridsingularity/d3a-interface@{env:BRANCH:master}
+	pip install git+https://github.com/gridsingularity/d3a-interface@feature/D3ASIM-3538
```
fievelk, line 79: Reminder

fievelk:
I left some comments and questions.

General nits:
- some backslashes `\` that can be refactored;
- there are some missing docstrings (e.g. in the `state.py` module)
- some files have not-optimized imports

src/d3a/d3a_core/util.py
fievelk: `is_time_slot_in_past_markets` occurs 8 times in 3 files. I think that in both cases we would have boilerplate code that would benefit from further refactoring :)

src/d3a/models/area/markets.py
hannesdiedrich: But it does return all future markets in any case. Or did I miss something?

fievelk: Yes, what I mean is that the shorter name `future_spot_markets` means the same, and it's shorter. I think the `all_` part is not needed.

src/d3a/models/state.py
spyrostz: This check is due to the floating point precision (or lack thereof) of Python. For instance, assuming that the value of the _unsettled_deviation_kWh[time_slot] == 0.25 , and purchased_energy_kWh == 0.25 . In some cases, due to floating point precision errors, this operation `self._unsettled_deviation_kWh[time_slot] -= purchased_energy_kWh` might lead to a minuscule negative value (e.g. 1e-18). In this case the >= 0 assertion would fail, even though all these operations were correct. Therefore we are obliged to include the minuscule negative values in our assertion, thus the - is really needed here

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,158 @@
...
-                del past_markets[pm].traded_energy
-                del past_markets[pm]
+                self.past_markets[time_slot] = market
+                self.log.debug(
+                    f"Moving {self.past_markets[time_slot]} to past.")
+
+    def _move_balancing_markets_to_past_balancing(self, current_time: DateTime) -> None:
+        """Move balancing markets to self.past_balancing_markets."""
+        # the conversion to list is needed in order to omit 'mutated during iteration' error
+        for time_slot in list(self.balancing_markets.keys()):
```
faizan2590, line 111:
```suggestion
        for time_slot in self.balancing_markets.keys():
```

src/d3a/models/area/markets.py
hannesdiedrich: Actually no, flake will complain if I dont do that.

src/d3a/models/area/markets.py
hannesdiedrich: I did that here:
https://github.com/gridsingularity/d3a/pull/1213/commits/121e754b98467c6bde7f54d67df6c27bbc5526ca
Unfortunately, also the slot markets depend on the settlement market setting (need to keep more past markets in memory in such a case), so the MarketRotator is not free of settlementmarket logic. 
But still I like the change. Let me know if you agree, Thanks!

src/d3a/models/area/markets.py
hannesdiedrich: You are right, will remove it.

src/d3a/models/area/markets.py
hannesdiedrich: Actually, lets remove this property totally, as it is used only once inside the same class

src/d3a/models/area/markets.py
fievelk: @faizan2590 the comment above explains why the `list` conversion is needed in this case.

src/d3a/models/area/markets.py
fievelk: Weird, I didn't know that. Thanks!

src/d3a/models/area/markets.py
```diff
@@ -101,37 +142,51 @@ def select_market_class(is_spot_market):
...
-                    name=area.name,
-                    in_sim_duration=is_timeslot_in_simulation_duration(area.config, timeframe)
-                )
-
-                area.dispatcher.create_area_agents(is_spot_market, market)
-                markets[timeframe] = market
+            time_slot = current_time + offset
+            if time_slot not in markets:
+                markets[time_slot] = self._create_market(market_class, time_slot, area,
+                                                         is_spot_market)
```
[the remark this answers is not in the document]
faizan2590: Recommended earlier when settlement market was following the [same approach](https://github.com/gridsingularity/d3a/pull/1213/files/276d61a9e104af169938c31f3e2c013d31a2cd06#r696581832)

src/d3a/models/state.py
```diff
@@ -237,6 +274,18 @@ class PVState(ProductionState):
     Completely inherits ProductionState, but we keep this class for backward compatibility.
     """
 
+    def _calculate_unsettled_energy_kWh(self, measured_energy_kWh, time_slot):
```
faizan2590, line 277:
```suggestion
    def _calculate_unsettled_energy_kWh(self, measured_energy_kWh: float, time_slot: DateTime) -> float:
```

src/d3a/models/state.py
```diff
@@ -249,6 +298,19 @@ def total_energy_demanded_Wh(self):
     def get_desired_energy(self, time_slot):
         return self._desired_energy_Wh[time_slot]
 
+    def _calculate_unsettled_energy_kWh(self, measured_energy_kWh, time_slot):
+        """
+        Returns negative values for underconsumption (offer will be placed on the settlement
+        market) and positive values for overconsumption (bid will be placed on the settlement
+        market)
+        :param measured_energy_kWh: Measured energy that the PV produced
```
faizan2590, line 306:
```suggestion
        :param measured_energy_kWh: Measured energy that the load consumed
```

src/d3a/models/state.py
```diff
@@ -249,6 +298,19 @@ def total_energy_demanded_Wh(self):
...
+        :param time_slot: time slot of the measured energy
+        :return: Deviation between forecasted and measured energy
+        """
+        traded_energy_kWh = (self.get_desired_energy_Wh(time_slot) -
+                             self.get_energy_requirement_Wh(time_slot)) * 1000.0
```
faizan2590, line 311:
```suggestion
                             self.get_energy_requirement_Wh(time_slot)) / 1000.0
```
I think division is required to convert smaller unit to a bigger one.

tests/area/test_area.py
```diff
@@ -103,16 +103,16 @@ def test_delete_past_markets_instead_of_last(self):
         self.area.cycle_markets(False, False, False)
         assert len(self.area.past_markets) == 0
 
-        current_time = today(tz=constants.TIME_ZONE).add(hours=1)
+        current_time = today(tz=constants.TIME_ZONE).add(minutes=self.config.slot_length.minutes)
```
faizan2590, line 106:
```suggestion
        current_time = today(tz=constants.TIME_ZONE).add(minutes=self.config.slot_length.total_minutes())
```

src/d3a/models/area/markets.py
faizan2590: Oh, thanks for pointing out, missed that.

src/d3a/models/state.py
spyrostz: Nice catch, you are correct! Also adding unit tests to cover this case, as far as I see the state classes are all lacking. Thanks!

tests/area/test_markets.py
```diff
@@ -0,0 +1,132 @@
...
+            area_fixture.config.start_date: [5, 1, 1, 0],
+            area_fixture.config.start_date + 1 * area_fixture.config.slot_length: [5, 2, 2, 0],
+            area_fixture.config.start_date + 2 * area_fixture.config.slot_length: [5, 3, 3, 0],
+            area_fixture.config.start_date + 3 * area_fixture.config.slot_length: [5, 4, 4, 0],
+            area_fixture.config.start_date + 4 * area_fixture.config.slot_length: [5, 4, 4, 1],
+            area_fixture.config.start_date + 5 * area_fixture.config.slot_length: [5, 4, 4, 1],
+        }
+
+        area_fixture.activate()
+        slot_length_in_ticks = area_fixture.config.slot_length / area_fixture.config.tick_length
```
faizan2590, line 120:
```suggestion
        total_ticks_per_slot = area_fixture.config.slot_length / area_fixture.config.tick_length
```
NIT: Sounds clearer to me what it is

hannesdiedrich: I response to [Hassans comment](https://github.com/gridsingularity/d3a/pull/1213#discussion_r698360099) I renamed it to `ticks_per_slot` OK?

src/d3a/d3a_core/util.py
hannesdiedrich: :-) Indeed, I would like to skip this for the sake of time though.

src/d3a/models/state.py
spyrostz: Don't worry I was not blaming Faizan :D I was mostly thinking out loud here, and as long as it did not introduce circular imports then it is easy to add, and actually adapted it already. So all good

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,175 @@
...
+        self.settlement_markets: Dict[DateTime, Market] = OrderedDict()
+        # Past markets:
+        self.past_markets:  Dict[DateTime, Market] = OrderedDict()
+        self.past_balancing_markets:  Dict[DateTime, BalancingMarket] = OrderedDict()
+        self.past_settlement_markets: Dict[DateTime, Market] = OrderedDict()
+        self.indexed_future_markets = {}
+
+        self.slot_market_rotator = RotatorInterface()
+        self.balancing_market_rotator = RotatorInterface()
+        self.settlement_market_rotator = RotatorInterface()
```
fievelk, line 159: Interfaces are usually just for inheritance, and they're not used directly. This probably be the same:
```suggestion
        self.slot_market_rotator = None
        self.balancing_market_rotator = None
        self.settlement_market_rotator = None
```

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,175 @@
...
+        del market_buffer[time_slot].traded_energy
+        del market_buffer[time_slot]
+
+
+class SlotMarketRotator(MarketRotatorBase):
+    """Deal with market rotation of spot markets."""
+
+
+class BalancingMarketRotator(MarketRotatorBase):
+    """Deal with market rotation of balancing markets."""
```
fievelk: I would suggest removing these classes and directly use `MarketRotatorBase` (maybe even calling it `DefaultMarketRotator`).

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,175 @@
...
+    def _is_it_time_to_delete_past_market(current_time_slot: DateTime,
+                                          time_slot: DateTime) -> bool:
+        """Check if the past market for time_slot is ready to be deleted."""
+
+        if constants.RETAIN_PAST_MARKET_STRATEGIES_STATE:
+            return False
+        else:
+            if ConstSettings.GeneralSettings.ENABLE_SETTLEMENT_MARKETS:
+                return (time_slot < current_time_slot.subtract(
+                    hours=ConstSettings.GeneralSettings.MAX_AGE_SETTLEMENT_MARKET_HOURS))
```
fievelk, line 67: Since we have a specialized rotator for each market type, these `if` statements can be implemented as methods of those classes. E.g.:
```python
class MarketRotatorBase:
    # ...
    def is_it_time_to_delete_past_market(current_time_slot: DateTime, time_slot: DateTime) -> bool:
        if constants.RETAIN_PAST_MARKET_STRATEGIES_STATE:
            return False

        return time_slot < current_time_slot.subtract(
            minutes=GlobalConfig.slot_length.total_minutes())

class SettlementMarketRotator(MarketRotatorBase):
    # ...
    def is_it_time_to_delete_past_market(current_time_slot: DateTime, time_slot: DateTime) -> bool:
        if constants.RETAIN_PAST_MARKET_STRATEGIES_STATE:
            return False

        if ConstSettings.GeneralSettings.ENABLE_SETTLEMENT_MARKETS:
            return (time_slot < current_time_slot.subtract(
                hours=ConstSettings.GeneralSettings.MAX_AGE_SETTLEMENT_MARKET_HOURS))
```
There's some duplication in my example that could probably be further avoided.

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,175 @@
...
-                del past_markets[pm].offer_history
-                del past_markets[pm].notification_listeners
-                del past_markets[pm].bids
-                del past_markets[pm].bid_history
-                del past_markets[pm].traded_energy
-                del past_markets[pm]
+        # Children trade in `markets`
+        self.markets:  Dict[DateTime, Market] = OrderedDict()
+        self.balancing_markets:  Dict[DateTime, BalancingMarket] = OrderedDict()
+        self.settlement_markets: Dict[DateTime, Market] = OrderedDict()
```
faizan2590, line 150:
```suggestion
        self.settlement_markets: Dict[DateTime, TwoSidedMarket] = OrderedDict()
```
Since settlement_markets to be double sided

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,175 @@
...
+
+    def _is_it_time_to_rotate_market(self, current_time_slot: DateTime,
+                                     time_slot: DateTime) -> bool:
+        """Check if it is time to move market for time_slot into past markets."""
+        return time_slot < current_time_slot
+
+    def _move_markets_to_past(self, markets: Dict, past_markets: Dict,
+                              current_time_slot: DateTime) -> None:
+        """Move markets to self.past_markets."""
+        move_market_slots = [
```
fievelk, line 80: Nit:
```suggestion
        expired_market_slots = [
        # or
        market_slots_to_be_cycled = [
```

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,175 @@
...
+            if self._is_it_time_to_rotate_market(current_time_slot, time_slot)]
+        for time_slot in move_market_slots:
+            market = markets.pop(time_slot)
+            market.readonly = True
+            past_markets[time_slot] = market
+            log.debug("Moving %s to past.", past_markets[time_slot])
+
+    def _delete_past_markets(self, market_dict: Dict, current_time_slot: DateTime) -> None:
+        """Delete the unneeded markets from self.past_markets."""
+        delete_market_slots = [
```
fievelk, line 91: Nit
```suggestion
        market_slots_to_be_deleted = [
```

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,175 @@
...
+        del market_buffer[time_slot].traded_energy
+        del market_buffer[time_slot]
+
+
+class SpotMarketRotator(MarketRotatorBase):
+    """Deal with market rotation of spot markets."""
+
+
+class BalancingMarketRotator(MarketRotatorBase):
+    """Deal with market rotation of balancing markets."""
```
fievelk, line 122: I would suggest removing these classes and directly use `MarketRotatorBase` (maybe even calling it `DefaultMarketRotator`).

src/d3a/models/area/markets.py
```diff
@@ -15,84 +15,175 @@
...
+        self.markets:  Dict[DateTime, Market] = OrderedDict()
+        self.balancing_markets:  Dict[DateTime, BalancingMarket] = OrderedDict()
+        self.settlement_markets: Dict[DateTime, Market] = OrderedDict()
+        # Past markets:
+        self.past_markets:  Dict[DateTime, Market] = OrderedDict()
+        self.past_balancing_markets:  Dict[DateTime, BalancingMarket] = OrderedDict()
+        self.past_settlement_markets: Dict[DateTime, Market] = OrderedDict()
+        self.indexed_future_markets = {}
+
+        self.slot_market_rotator = RotatorInterface()
```
fievelk, line 157:
```suggestion
        self.spot_market_rotator = RotatorInterface()
```

fievelk:
Left some more comments. I leave it up to you to decide which ones are nits and which are more important. IMO, the location of `should_post_bid` and `should_post_offer` was also relevant, but decide based on the time constraints. I'll leave a LGTM.

fievelk (dismissed):

src/d3a/models/area/markets.py
hannesdiedrich: @fievelk we need to keep more past markets in the memory in case the settlement markets are enabled. This is one reason I decided to go for the code dublication in my first proposal.

src/d3a/models/area/markets.py
hannesdiedrich: @fievelk Please have a look at the docstring of `activate_market_rotators`, or I will rather contact you directly to explain why I did it like this

src/d3a/models/state.py
spyrostz: I thought I had answered here, sorry. What you say makes sense only for the spot market actually, and not for the settlement market. All assets can post both bids and offers on the settlement market, according to the deviation from the forecasted energy (for instance, if a load consumed less than it bought, then it has a deviation on the negative side, which then can sell in the settlement market. Same for a PV that overproduces)
One improvement could be to rename these to can_post_settlement_bid and can_post_settlement_offer, which I will add in the context of 3568 in order to not block Hannes' PR

spyrostz:
> Left some more comments. I leave it up to you to decide which ones are nits and which are more important. IMO, the location of `should_post_bid` and `should_post_offer` was also relevant, but decide based on the time constraints. I'll leave a LGTM to speed up the process.

@fievelk answered here https://github.com/gridsingularity/d3a/pull/1213#discussion_r699377133

src/d3a/models/state.py
```diff
@@ -237,6 +336,19 @@ class PVState(ProductionState):
...
+        """
+        Returns negative values for overproduction (offer will be placed on the settlement market)
+        and positive values for underproduction (bid will be placed on the settlement market)
+        :param measured_energy_kWh: Measured energy that the PV produced
+        :param time_slot: time slot of the measured energy
+        :return: Deviation between forecasted and measured energy
+        """
+        traded_energy_kWh = (self.get_energy_production_forecast_kWh(time_slot) -
+                             self.get_available_energy_kWh(time_slot))
+        return traded_energy_kWh - measured_energy_kWh
```
faizan2590, line 350:
```suggestion
        return self.get_energy_production_forecast_kWh(time_slot) - measured_energy_kWh
```
Shouldn't it be the difference of forecasted and measured energy?

hannesdiedrich: No, you need to substract the traded energy, too. Please have a look at the docstring.

spyrostz: Not really. This measures the deviation of the measurements from the traded energy. The way it goes: 
spot market -> takes forecast as input, and decides production / consumption based on traded energy. This is the energy that the asset has to produce / consume.
settlement market -> the measurements indicate a deviation of production/ consumption from the energy that the asset decided to sell / buy. Therefore it uses this deviation in order to palce bids and offers to the settlement market.

src/d3a/models/state.py
fievelk: Thanks @spyrostz! I have the feeling that this complicates a bit the understanding of the states on an intuitive level: it feels a bit counter-intuitive to have bid-related methods on PVs (and offer-related methods on Loads). The re-naming that you suggest would definitely help to make the difference more visible. Thanks again.

src/d3a/models/state.py
```diff
@@ -237,6 +336,19 @@ class PVState(ProductionState):
...
     """
 
+    def _calculate_unsettled_energy_kWh(
+            self, measured_energy_kWh: float, time_slot: DateTime) -> float:
+        """
+        Returns negative values for overproduction (offer will be placed on the settlement market)
+        and positive values for underproduction (bid will be placed on the settlement market)
+        :param measured_energy_kWh: Measured energy that the PV produced
+        :param time_slot: time slot of the measured energy
+        :return: Deviation between forecasted and measured energy
```
faizan2590, line 346:
```suggestion
        :return: Deviation between energy traded in spot market traded and measured energy
```

src/d3a/models/state.py
faizan2590: > No, you need to substract the traded energy, too. Please have a look at the docstring

Left a [comment](https://github.com/gridsingularity/d3a/pull/1213/files?file-filters%5B%5D=.py#r699407475) to update docstring accordingly.

src/d3a/models/area/markets.py
fievelk: I think my suggestion above shouldn't change what you described, right? `_delete_past_markets()` will call `self. _is_it_time_to_delete_past_market()`. At that point, if you're inside a `SettlementMarketRotator` instance you will keep more past markets in memory, otherwise if you're in a `MarketRotatorBase` instance you'll keep the "normal" amount (based on the method implementation).

src/d3a/models/state.py
hannesdiedrich: @spyrostz please add this in the frame of 3568. Thanks!

src/d3a/models/area/markets.py
hannesdiedrich: I meant that we also need to keep more SPOT past markets in the memory if `ConstSettings.GeneralSettings.ENABLE_SETTLEMENT_MARKET`. So, the `MarketRotatorBase. is_it_time_to_delete_past_market` should return 
```python
            return (time_slot < current_time_slot.subtract(
                hours=ConstSettings.GeneralSettings.MAX_AGE_SETTLEMENT_MARKET_HOURS))
```
in this case.

fievelk: Interesting, I hadn't understood that. Thanks for the explanation. Maybe a comment would help, if you have time to add it.

fievelk (dismissed):

faizan2590 (dismissed):
LGTM to avoid minor comments adaptation as this feature been a priority.

src/d3a/models/area/markets.py
hannesdiedrich: Added it https://github.com/gridsingularity/d3a/pull/1213/commits/983054e368a590c99a8e6e6bd37f5213200b3ed3
Will press push after the tests have passed.

fievelk: Great, thanks!

PR #1214, opened 2021-08-27
Feature/d3 asim 3568

spyrostz:
Implementation of the settlement market for the template strategies. Limited only to the load and PV strategies at the moment. Includes test coverage, and also helper methods for rejecting settlement offers / bids that are violating the energy deviation limit.

### DIFF
--- a/tests/area/test_markets.py
+++ b/tests/area/test_markets.py
@@ -20,11 +20,13 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 from unittest.mock import Mock, patch
 
 import pytest
+from d3a_interface.constants_limits import ConstSettings, TIME_ZONE
+from pendulum import duration, today
+
+import d3a
 from d3a.d3a_core.device_registry import DeviceRegistry
 from d3a.models.area import Area
 from d3a.models.strategy.storage import StorageStrategy
-from d3a_interface.constants_limits import ConstSettings, TIME_ZONE
-from pendulum import duration, today
 
 
 class TestMarketRotation:
@@ -45,13 +47,19 @@ class TestMarketRotation:
         config.market_count = 5
         child = Area(name="test_market_area", config=config, strategy=StorageStrategy())
         area = Area(name="parent_area", children=[child], config=config)
+        ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET = False
+        ConstSettings.GeneralSettings.ENABLE_SETTLEMENT_MARKETS = False
+        d3a.constants.RETAIN_PAST_MARKET_STRATEGIES_STATE = False
+
         yield area
 
         DeviceRegistry.REGISTRY = original_registry
         ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET = False
         ConstSettings.GeneralSettings.ENABLE_SETTLEMENT_MARKETS = False
+        d3a.constants.RETAIN_PAST_MARKET_STRATEGIES_STATE = False
 
     def test_market_rotation_is_successful(self, area_fixture):
+        ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET = True
         area_fixture.activate()
         assert len(area_fixture.all_markets) == 5
         ticks_per_slot = area_fixture.config.slot_length / area_fixture.config.tick_length


src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+                 initial_selling_rate, final_selling_rate):
+        super().__init__()
+
+        self._update_interval = SettlementTemplateStrategiesConstants.UPDATE_INTERVAL_MIN
+        self.bid_updater = SettlementTemplateStrategyBidUpdater(
+                initial_rate=initial_buying_rate,
+                final_rate=final_buying_rate,
+                fit_to_limit=True,
+                energy_rate_change_per_update=None,
+                update_interval=duration(minutes=self._update_interval),
```
faizan2590, line 77: Couldn't find `_update_interval ` instance variable part of `SettlementMarketStrategy ` or `SettlementMarketStrategyInterface `, where is it coming from?

spyrostz: Look at line 71:

`SettlementTemplateStrategiesConstants.UPDATE_INTERVAL_MIN`

faizan2590: Thanks got overlooked by me.

src/d3a/models/strategy/__init__.py
```diff
@@ -330,6 +330,33 @@ def post_offer(self, market, replace_existing=True, **offer_kwargs) -> Offer:
...
+        if market is not None:
+            return market
+        # Then check settlement markets
+        markets = [m for m in self.area.settlement_markets.values() if m.id == market_id]
+        if not markets:
+            return None
+        return markets[0]
+
+    def are_offers_posted(self, market_id):
+        """Checks if any bids have been posted in the market slot with the given ID."""
```
faizan2590, line 357:
```suggestion
        """Checks if any offers have been posted in the market slot with the given ID."""
```

src/d3a/models/strategy/__init__.py
```diff
@@ -330,6 +330,33 @@ def post_offer(self, market, replace_existing=True, **offer_kwargs) -> Offer:
 
         return offer
 
+    def post_first_offer(self, market, energy_kWh, initial_energy_rate):
+        if not all(offer.seller != self.owner.name for offer in market.get_offers().values()):
+            self.owner.log.warning("There is already another offer posted on the market, therefore"
+                                   " do not repost another first offer.")
+            return None
```
faizan2590, line 337:
```suggestion
            return
```

Any specific advantage having an explicit None as return?

spyrostz: Not really, apart from matching the post_first_bid method which more or less does the same. Will change both methods to only return.

src/d3a/models/strategy/__init__.py
spyrostz: Thanks, changing this

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+                        market, energy_deviation_kWh,
+                        self.offer_updater.initial_rate[market.time_slot]
+                    )
+                except MarketException:
+                    pass
+
+    def event_tick(self, strategy):
+        self.bid_updater.increment_update_counter_all_markets(strategy)
+        self.offer_updater.increment_update_counter_all_markets(strategy)
+        """Update posted settlement bids and offers on market tick"""
```
hannesdiedrich, line 115:
```suggestion
        """Update posted settlement bids and offers on market tick."""
        self.bid_updater.increment_update_counter_all_markets(strategy)
        self.offer_updater.increment_update_counter_all_markets(strategy)
        
```

src/d3a/models/strategy/__init__.py
```diff
@@ -330,6 +330,33 @@ def post_offer(self, market, replace_existing=True, **offer_kwargs) -> Offer:
 
         return offer
 
+    def post_first_offer(self, market, energy_kWh, initial_energy_rate):
+        if not all(offer.seller != self.owner.name for offer in market.get_offers().values()):
```
faizan2590, line 334:
```suggestion
        if any(offer.seller != self.owner.name for offer in market.get_offers().values()):
```
Wouldn't it be simpler in order to avoid reverse of logic check?

src/d3a/models/strategy/update_frequency.py
```diff
@@ -59,8 +59,12 @@ def delete_past_state_values(self, current_market_time_slot):
             self.energy_rate_change_per_update.pop(market_slot, None)
             self.update_counter.pop(market_slot, None)
 
+    @staticmethod
+    def get_all_markets(area):
+        return area.all_markets
```
hannesdiedrich, line 64: Nit: Is this really needed? Isn't `for market in area.all_markets:` even shorter?

src/d3a/models/strategy/__init__.py
```diff
@@ -469,6 +490,15 @@ def can_offer_be_posted(
         return self.offers.can_offer_be_posted(
             offer_energy, offer_price, available_energy, market, replace_existing=replace_existing)
 
+    def can_settlement_offer_be_posted(self, offer_energy, offer_price, market,
+                                       replace_existing=False):
+        if not self.state.should_post_offer(market.time_slot):
```
faizan2590, line 495: As consensus is to add docstring for each newly added methods.

tests/strategies/settlement/test_settlement_strategy.py
```diff
@@ -0,0 +1,120 @@
...
+                seller_origin_id=self.area_mock.uuid, seller_id=self.area_mock.uuid
+            )
+
+    @pytest.mark.parametrize(
+        "strategy_fixture", [LoadHoursStrategy(100), PVStrategy()])
+    @pytest.mark.parametrize("should_post_bid", [True, False])
+    @pytest.mark.parametrize("should_post_offer", [True, False])
+    def test_event_tick_updates_bids_and_offers(
+            self, strategy_fixture, should_post_bid, should_post_offer):
+        self._setup_strategy_fixture(strategy_fixture, should_post_bid, should_post_offer)
```
hannesdiedrich, line 75: Interesting way of looping over different cases! Will keep this in mind for future. Thanks!

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+            self.offer_updater.update(market, strategy)
+
+    def _get_settlement_market_by_id(self, area, market_id):
+        markets = [m for m in area.settlement_markets.values() if m.id == market_id]
+        if not markets:
+            return None
+        return markets[0]
+
+    def event_bid_traded(self, strategy, market_id, bid_trade):
+        """Updates the unsettled deviation with the traded energy from the market"""
```
hannesdiedrich, line 127: Probably copy paste remnant, also in `event_trade` right?

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
+from d3a_interface.constants_limits import ConstSettings
```
hannesdiedrich, line 1: Please don't forget the licence string (or are we not caring about this any more because we will not divide the d3a further?)

hannesdiedrich: Also, I am hesitant to request doc_strings in this module. IMO all class and method names are self-explanatory. So, feel free to skip it, but if you could bother, please add them (also in other parts of the new code). Thanks!

src/d3a/models/strategy/__init__.py
```diff
@@ -588,6 +617,14 @@ def can_bid_be_posted(
 
         return total_posted_energy <= required_energy_kWh and bid_price >= 0.0
 
+    def can_settlement_bid_be_posted(self, bid_energy, bid_price, market, replace_existing=False):
```
faizan2590, line 620: Can't find its usage, also of `can_settlement_offer_be_posted `

hannesdiedrich, line 620: I dont see this method being used somewhere. Can you maybe point me to it? (also `can_settlement_offer_be_posted`)

src/d3a/constants.py
```diff
@@ -56,3 +56,12 @@
 CN_PROFILE_EXPANSION_DAYS = 7
 
 RUN_IN_REALTIME = False
+
+
+class SettlementTemplateStrategiesConstants:
```
hannesdiedrich, line 61: Potentially, this could be part of ConstSettings, right?

hannesdiedrich (commented):
Left some rather minor comments. Apart from that it looks great.

tests/strategies/settlement/test_settlement_strategy.py
```diff
@@ -0,0 +1,120 @@
...
+
+import pytest
+from d3a_interface.constants_limits import ConstSettings
+from pendulum import datetime, duration
+
+from d3a.models.market.market_structures import Bid, Offer, Trade
+from d3a.models.market.two_sided import TwoSidedMarket
+from d3a.models.strategy.load_hours import LoadHoursStrategy
+from d3a.models.strategy.pv import PVStrategy
+from d3a.models.strategy.settlement_strategies import SettlementMarketStrategy
```
faizan2590, line 12:
```suggestion
import pytest
from d3a.models.market.market_structures import Bid, Offer, Trade
from d3a.models.market.two_sided import TwoSidedMarket
from d3a.models.strategy.load_hours import LoadHoursStrategy
from d3a.models.strategy.pv import PVStrategy
from d3a.models.strategy.settlement_strategies import SettlementMarketStrategy
from d3a_interface.constants_limits import ConstSettings
from pendulum import datetime, duration
```

NIT: A recommendation of an `optimise imports`

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+        initial_selling_rate: float = None, final_selling_rate: float = None
+):
+    if initial_selling_rate is None:
+        initial_selling_rate = SettlementTemplateStrategiesConstants.INITIAL_SELLING_RATE
+    if final_selling_rate is None:
+        final_selling_rate = SettlementTemplateStrategiesConstants.FINAL_SELLING_RATE
+    if initial_buying_rate is None:
+        initial_buying_rate = SettlementTemplateStrategiesConstants.INITIAL_BUYING_RATE
+    if final_buying_rate is None:
+        final_buying_rate = SettlementTemplateStrategiesConstants.FINAL_BUYING_RATE
```
faizan2590, line 21:
```suggestion
        initial_buying_rate: float = SettlementTemplateStrategiesConstants.INITIAL_BUYING_RATE, 
        final_buying_rate: float = SettlementTemplateStrategiesConstants.FINAL_BUYING_RATE,
        initial_selling_rate: float = SettlementTemplateStrategiesConstants.INITIAL_SELLING_RATE, 
        final_selling_rate: float = SettlementTemplateStrategiesConstants.FINAL_SELLING_RATE):
```

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+            return None
+        return markets[0]
+
+    def event_bid_traded(self, strategy, market_id, bid_trade):
+        """Updates the unsettled deviation with the traded energy from the market"""
+        market = self._get_settlement_market_by_id(strategy.area, market_id)
+        if not market:
+            return
+        if bid_trade.offer_bid.buyer == strategy.owner.name:
+            strategy.state.decrement_unsettled_deviation(
```
faizan2590, line 132:
```suggestion
            strategy.state.increment_unsettled_deviation(
```
Assuming the settlement market's bid is following the same approach as normal unsold BID's update rate.

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+    def __init__(self, *args, **kwargs):
+        pass
+
+    def event_market_cycle(self, strategy):
+        pass
+
+    def event_tick(self, strategy):
+        pass
+
+    def event_bid_traded(self, strategy, market_id, bid_trade):
```
faizan2590, line 57:
```suggestion
    def event_bid_traded(self, strategy: BaseStrategy, market_id: str, bid_trade: Trade):
```

As recommended in other [PR comments](https://github.com/gridsingularity/d3a-web/pull/1410#discussion_r696600945) to add type hint

src/d3a/models/strategy/__init__.py
spyrostz: Good idea, will apply it to both post_first_bid and post_first_offer, thanks!

src/d3a/models/strategy/__init__.py
spyrostz: Adding docstring

src/d3a/models/strategy/__init__.py
spyrostz: Will be used in this ticket https://gridsingularity.atlassian.net/browse/D3ASIM-3567 , but was easier to be introduced by this one.

src/d3a/models/strategy/settlement_strategies/__init__.py
spyrostz: Fixing

src/d3a/models/strategy/update_frequency.py
spyrostz: It is needed in order to be able to override this method for the settlement strategy. Have a look at the SettlementTemplateStrategyBidUpdater and SettlementTemplateStrategyOfferUpdater classes, these inherit from TemplateStrategyBid/OfferUpdater classes and override only this method, in order to apply the price increase / decrease on bids / offers of settlement markets instead of the spot markets.

tests/strategies/settlement/test_settlement_strategy.py
spyrostz: Kudos to @fievelk that pointed out some PRs ago that pytest.mark.parametrize applies cartesian product to arguments that are added one after the other!

src/d3a/models/strategy/settlement_strategies/__init__.py
spyrostz: Actually this is precise as far as I see. This comment explains that the unsettled energy of the state will be updated (essentially the traded energy will be decremented from the unsettled energy) only when the event_trade and event_bid_traded methods are called.

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+    def event_tick(self, strategy):
+        self.bid_updater.increment_update_counter_all_markets(strategy)
+        self.offer_updater.increment_update_counter_all_markets(strategy)
+        """Update posted settlement bids and offers on market tick"""
+        for market in strategy.area.settlement_markets.values():
+            self.bid_updater.update(market, strategy)
+            self.offer_updater.update(market, strategy)
+
+    def _get_settlement_market_by_id(self, area, market_id):
+        markets = [m for m in area.settlement_markets.values() if m.id == market_id]
```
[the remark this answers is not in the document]
spyrostz: I tend to disagree with the statement that it has better performance. List comprehensions are typically way faster than filters, especially for large number of elements (one example can be found here https://blog.finxter.com/python-lists-filter-vs-list-comprehension-which-is-faster/)

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+    def event_bid_traded(self, strategy, market_id, bid_trade):
+        """Updates the unsettled deviation with the traded energy from the market"""
+        market = self._get_settlement_market_by_id(strategy.area, market_id)
+        if not market:
+            return
+        if bid_trade.offer_bid.buyer == strategy.owner.name:
+            strategy.state.decrement_unsettled_deviation(
+                bid_trade.offer_bid.energy, market.time_slot)
+
+    def event_trade(self, strategy, market_id, trade):
```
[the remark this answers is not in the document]
spyrostz: Maybe in another story, definitely out of scope for this

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+
+    def event_tick(self, strategy):
+        self.bid_updater.increment_update_counter_all_markets(strategy)
+        self.offer_updater.increment_update_counter_all_markets(strategy)
+        """Update posted settlement bids and offers on market tick"""
+        for market in strategy.area.settlement_markets.values():
+            self.bid_updater.update(market, strategy)
+            self.offer_updater.update(market, strategy)
+
+    def _get_settlement_market_by_id(self, area, market_id):
```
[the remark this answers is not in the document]
spyrostz: Fixing, thanks!

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+class SettlementTemplateStrategyOfferUpdater(TemplateStrategyOfferUpdater):
+
+    @staticmethod
+    def get_all_markets(area):
+        return area.settlement_markets.values()
+
+
+class SettlementMarketStrategyInterface:
```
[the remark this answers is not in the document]
spyrostz: An abstract class cannot be constructed unfortunately, and in this case we need a class that will create an object that does nothing.

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
+from d3a_interface.constants_limits import ConstSettings
+from pendulum import duration
+
+from d3a.constants import SettlementTemplateStrategiesConstants
+from d3a.d3a_core.exceptions import MarketException
+from d3a.models.strategy.update_frequency import TemplateStrategyBidUpdater, \
+    TemplateStrategyOfferUpdater
+
+
+def settlement_market_strategy_factory(
```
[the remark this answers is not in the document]
spyrostz: I do not really understand the advantage of having a factory class vs the factory method that is implemented here. inline methods for validation are not really needed here, since the selection of the class is defined by a single constant (ENABLE_SETTLEMENT_MARKETS) and does not require extensive validation which could potentially justify changing the factory method to a factory class. Furthermore, if the need appears at some point in the future it can easily be converted to a class, but for now it is overengineering.

src/d3a/models/strategy/settlement_strategies/__init__.py
spyrostz: This is actually tricky. Unsettled deviation is an absolute value for both consumption and production (bid and offer). Therefore in both cases the value has to be decremented till it reaches zero. Reason why it was modelled this way, to avoid issues with sign convention like the one you mention

src/d3a/models/strategy/settlement_strategies/__init__.py
spyrostz: Will do so , thanks!

src/d3a/models/strategy/update_frequency.py
hannesdiedrich: Ahhh. thanks! Now I understand.

src/d3a/models/strategy/settlement_strategies/__init__.py
spyrostz: Let me do both just for consistency actually.

src/d3a/models/strategy/__init__.py
spyrostz: https://github.com/gridsingularity/d3a/pull/1214#discussion_r698711127 , TLDR will be used in the followup story

src/d3a/constants.py
spyrostz: That's what I was also in-between. Research / UI have not yet decided on the way that the settlement markets are configured for template strategies. If the settlement market rates can be configured by the UI, you are right that this should be in the ConstSettings. If this is not the case though, and the rates of the template strategies will only be configurable in d3a, then the right place is not the d3a-interface but the constants of d3a, that's why I placed them here. 
I can move it now, however I am not sure if at the very end it will be useful. If you believe that it is better to be future-proof and move this to d3a-interface please ping me and I'll do it, thanks!

tests/strategies/settlement/test_settlement_strategy.py
spyrostz: This is contrary to PEP8 though, and we have decided to follow this (pendulum and d3a-interface should precede d3a. Also lines should separate the 3 blocks of imports. Actually, the original commit is fine as is)

src/d3a/models/strategy/settlement_strategies/__init__.py
spyrostz: Fixing

src/d3a/models/strategy/settlement_strategies/__init__.py
```diff
@@ -0,0 +1,142 @@
+from d3a_interface.constants_limits import ConstSettings
+from pendulum import duration
```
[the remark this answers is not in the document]
spyrostz: 1. Will rename this to settlement/strategy.py
2. Not really. d3a_interface should precede pendulum since they are both third-party imports, and d3a_interface precedes pendulum in alphanumeric order
3. Already added from previous comments

src/d3a/models/strategy/settlement_strategies/__init__.py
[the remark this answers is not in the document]
Hassan754: Interesting! I used to believe the opposite (I read about this topic long ago but it had opposite opinion)

tests/strategies/settlement/test_settlement_strategy.py
faizan2590: https://user-images.githubusercontent.com/33482427/131500822-0705a05f-8eb2-4228-bf97-979107219df7.mov

Okay, may be PyCharm's default optimise import needs to be customised then.

spyrostz: Hmm, that's very strange, and it does not happen for me. I think the problem is not Pycharm per se, but that for some reason it recognises the d3a package as a third-party package, while in reality is the main package of the project. Can it be that you have multiple repos (e.g. d3a, d3a-web, d3a-interface) on the same Pycharm project ? Because in this case I could foresee such a behavior. Let's continue to discuss this on Slack.

src/d3a/constants.py
hannesdiedrich: Convinced! Leave it please. Thanks!

src/d3a/models/strategy/settlement/strategy.py
```diff
@@ -0,0 +1,199 @@
...
+        return markets[0]
+
+    def event_bid_traded(self, strategy: BaseStrategy, market_id: str, bid_trade: Trade) -> None:
+        """
+        Updates the unsettled deviation with the traded energy from the market
+        :param strategy: Strategy object of the asset
+        :param market_id: Id of the market that the trade took place
+        :param bid_trade: Trade object
+        :return: None
+        """
```
hannesdiedrich, line 151: Nit: We decided to go with the [google style doctring](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html)
```suggestion
        """
        Update the unsettled deviation with the traded energy from the market.
        Args:
            strategy: Strategy object of the asset
            market_id: Id of the market that the trade took place
            bid_trade: Trade obje
        Returns: None
        """
```

hannesdiedrich (dismissed):
Nit comment about the doctring format,
LGTM

src/d3a/models/strategy/settlement/strategy.py
spyrostz: Thank you, my Pycharm messed up the settings again. Will change it to google style

src/d3a/models/strategy/settlement/strategy.py
```diff
@@ -0,0 +1,215 @@
...
+        self.bid_updater.increment_update_counter_all_markets(strategy)
+        self.offer_updater.increment_update_counter_all_markets(strategy)
+        for market in strategy.area.settlement_markets.values():
+            self.bid_updater.update(market, strategy)
+            self.offer_updater.update(market, strategy)
+
+    def _get_settlement_market_by_id(
+            self, strategy: BaseStrategy, market_id: str) -> Optional["Market"]:
+        markets = [m for m in strategy.area.settlement_markets.values() if m.id == market_id]
+        if not markets:
```
faizan2590, line 147:
```suggestion
        if not markets or len(markets) > 1:
```

As there shouldn't be more than one market with an id

tests/strategies/settlement/test_settlement_strategy.py
```diff
@@ -0,0 +1,126 @@
...
+        self.settlement_strategy = SettlementMarketStrategy(10, 50, 50, 20)
+        self.time_slot = today(tz=TIME_ZONE).at(hour=12, minute=0, second=0)
+        self.market_mock = MagicMock(spec=TwoSidedMarket)
+        self.market_mock.time_slot = self.time_slot
+        self.market_mock.id = str(uuid.uuid4())
+        self.test_bid = Bid("123", self.time_slot, 10, 1, buyer="test_name")
+        self.test_offer = Offer("234", self.time_slot, 50, 1, seller="test_name")
+        self.market_mock.bid = MagicMock(return_value=self.test_bid)
+        self.market_mock.offer = MagicMock(return_value=self.test_offer)
+        self.market_mock.bids = {self.time_slot: self.test_bid}
```
faizan2590, line 29:
```suggestion
        self.market_mock.bids = {self.test_bid.id: self.test_bid}
```

As far as I recall

tests/strategies/settlement/test_settlement_strategy.py
faizan2590: > Can it be that you have multiple repos (e.g. d3a, d3a-web, d3a-interface) on the same Pycharm project ?

Not really, they are in different PyCharm projects. Anyways, marking this comment as resolved.

faizan2590 (approved):
LGTM, can avoid adaptation/addressing of minor comments from my side due to time constraints.

src/d3a/models/strategy/settlement/strategy.py
spyrostz: I am changing this to an assertion instead. The suggestion will hid the problem by returning None, while we should be aware that the settlement market list is corrupt in this case.

tests/strategies/settlement/test_settlement_strategy.py
spyrostz: Thanks, fixing it

PR #1217, opened 2021-09-02
Feature/d3 asim 3567

spyrostz:
Added ability for external strategies to support the timeslot command argument. This is selecting the market which the command should apply. If the argument is missing, the spot market will be used. If the timeslot is in the past, then the settlement market that matches the provided timeslot will be used for the commands.


src/d3a/models/strategy/external_strategies/load.py
```diff
@@ -45,13 +45,19 @@ def channel_dict(self):
...
+            market: Market object that will read the bids from
+
+        Returns: List of bids for the strategy asset
+
+        """
 
         return [
             {"id": bid.id, "price": bid.price, "energy": bid.energy}
-            for _, bid in self.next_market.get_bids().items()
+            for _, bid in market.get_bids().items()
```
faizan2590, line 60:
```suggestion
            for bid in market.get_bids().values()
```

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -212,6 +214,20 @@ def _device_info_aggregator(self, arguments):
...
+        return self._get_market_from_timeslot(arguments["timeslot"])
+
+    def _get_market_from_timeslot(self, timeslot: DateTime):
+        market = self.market_area.get_market(timeslot)
+        if market:
+            return market
+        market = self.market_area.get_settlement_market(timeslot)
+        if market:
+            return market
+        return None
```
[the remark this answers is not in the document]
spyrostz: Spot markets and settlement markets have always different time slots. Spot market is always close to the present and future, settlement markets are always in the past

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -217,16 +218,18 @@ def next_market(self):
...
     def _get_market_from_timeslot(self, timeslot: DateTime):
         market = self.market_area.get_market(timeslot)
         if market:
             return market
         market = self.market_area.get_settlement_market(timeslot)
-        if market:
-            return market
-        return None
+        if not market:
+            raise Exception(f"Timeslot {timeslot} is not currently in the spot, future or "
```
fievelk, line 230: I would suggest to use a specific exception, maybe one of the ones we implemented here: https://github.com/gridsingularity/d3a/blob/d6926537004f1ef411e5988ee7e8c2025957bed0/src/d3a/d3a_core/exceptions.py

faizan2590 (approved):

src/d3a/models/strategy/external_strategies/__init__.py
hannesdiedrich: @fievelk Lets discuss this topic again next time we are in the office ok? I am still not convinced by this rule, I know it is in PEP8, but isn't the message of the exception already specific enough? Why creating a special Exception for this specific case? 
Again, no need to answer now. You can pitch me this topic in person. :-D

tests/strategies/external/test_load.py
```diff
@@ -15,21 +15,25 @@
...
 def ext_load_fixture():
-    return create_areas_markets_for_strategy_fixture(LoadHoursExternalStrategy(100))
+    from d3a_interface.constants_limits import ConstSettings
+    ConstSettings.IAASettings.MARKET_TYPE = 2
+    yield create_areas_markets_for_strategy_fixture(LoadHoursExternalStrategy(100))
+    ConstSettings.IAASettings.MARKET_TYPE = 1
 
 
-class TestPVForecastExternalStrategy:
+class TestLoadForecastExternalStrategy:
```
hannesdiedrich, line 36: 😅

hannesdiedrich (approved):
LGTM

src/d3a/models/strategy/external_strategies/__init__.py
fievelk: Sure! I'll practice my public-speaking skills 😄

spyrostz: There is one advantage in subclassing and using specific exception types: grouping by type, and catching exceptions of a specific type and handling them differently than other exceptions. Unfortunately here the exception does not really help much because the exception handler catches all exception types and propagates them to the response of the API command. That's why I am also not really sold that a specific exception type is needed here. Having said all of the above, looking forward to discussing this in person, the discussion will definitely be interesting for me!

PR #1219, opened 2021-09-02
Feature/d3 asim 3539 Added export of results fro settlement markets

hannesdiedrich:
For now it was decided to only export bids/offers/trades and general statistics (e.g. avg.-price) in csv files locally for the settlement markets.
I also added the sending of the bids/offers/trades for the settlement market in the core_stats_dict to the DB.<
We can maybe decide whether we should save the DB space for that and disable this for now.


src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -176,3 +146,74 @@ def _specific_row(self, slot, market):
...
+        data = self.export_data_factory(area, past_market_type)
+        if area.slug not in out_dict:
+            out_dict[area.slug] = dict((label, []) for label in data.labels)
+        for row in data.rows():
+            for ii, label in enumerate(data.labels):
+                out_dict[area.slug][label].append(row[ii])
+
+    def _populate_plots_stats_for_supply_demand_curve(self, area: Area) -> None:
+        """Populate the statistics for supply_demand_curve."""
+        if ConstSettings.IAASettings.MARKET_TYPE == 3:
```
faizan2590, line 195:
```suggestion
        if ConstSettings.IAASettings.MARKET_TYPE == 2 and ConstSettings.IAASettings.BID_OFFER_MATCH_TYPE == 2:
```

faizan2590 (commented):
Left some comments, will give another look once addressed/answered and the added integration_test passes.

src/d3a/d3a_core/export.py
```diff
@@ -16,33 +16,35 @@
...
 import shutil
-import json
-import operator
-from typing import Dict
 from collections import namedtuple
 from copy import deepcopy
 from functools import reduce  # forward compatibility for Python 3
+from typing import Dict, Tuple, List, Mapping
 
+import d3a.constants
```
fievelk, line 30: Nit: something seems off with the order of these imports. Could you please try to optimize them again?

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -15,92 +15,59 @@
...
-                self.clearing[area.slug] = {}
-            if market.time_slot not in self.cumulative_offers[area.slug]:
-                self.cumulative_offers[area.slug][market.time_slot] = {}
-                self.cumulative_bids[area.slug][market.time_slot] = {}
-                self.clearing[area.slug][market.time_slot] = {}
-            self.cumulative_offers[area.slug][market.time_slot] = market.state.cumulative_offers
-            self.cumulative_bids[area.slug][market.time_slot] = market.state.cumulative_bids
-            self.clearing[area.slug][market.time_slot] = \
-                bid_offer_matcher.matcher.match_algorithm.state.clearing
+past_market_type_file_suffix_dict = {
```
fievelk, line 37:
```suggestion
PAST_MARKET_TYPE_FILE_SUFFIX_DICT = {
```
```suggestion
past_market_type_file_suffix_dict = {
```

src/d3a/d3a_core/export.py
```diff
@@ -223,81 +286,61 @@ def _export_area_clearing_rate(self, area, directory, file_suffix, is_first):
...
-            past_markets = area.past_balancing_markets
-        else:
-            file_path = self._file_path(directory, "{}-trades".format(area.slug))
-            labels = ("slot",) + Trade._csv_fields()
-            past_markets = area.past_markets
-
+    def _export_trade_csv_files(self, past_markets: List, directory: dir, filename: str,
+                                labels: Tuple = Trade.csv_fields(),
+                                is_first: bool = False) -> None:
+        """ Export files containing individual trades  (*-trades.csv  files)."""
```
fievelk, line 309: Nit:
```suggestion
        """Export files containing individual trades (*-trades.csv files)."""
```

src/d3a/d3a_core/export.py
```diff
@@ -169,48 +174,106 @@ def move_root_plot_folder(self):
...
+                                                market_member="bid_history",
+                                                directory=directory,
+                                                filename=f"{area.slug}-bids",
+                                                labels=("slot",) + Bid.csv_fields(),
+                                                is_first=is_first)
+
+    def _export_settlement_markets_stats(self, area: Area, directory: dir, is_first: bool) -> None:
+        """Export bids, offers, trades, statistics csv-files for all settlement markets."""
+        if not ConstSettings.SettlementMarketSettings.ENABLE_SETTLEMENT_MARKETS:
+            return
```
fievelk, line 204: Nit: why is this needed? If the settlement markets were not enabled, `area.past_settlement_markets.values()` should be empty, right? If we only take this into consideration, we can avoid adding the dependency from this global setting and make the code a bit more flexible. 

Related to this: if the values of some market slots are not available, should their files (e.g. the trades CSV files) still be generated? If not, we can refactor the `_export_*_csv_files` functions to first check if there is available data. If not, they can just `return` immediately.

src/d3a/d3a_core/export.py
```diff
@@ -169,48 +174,106 @@ def move_root_plot_folder(self):
...
+            market_member="bid_history",
+            directory=directory,
+            filename=f"{area.slug}-settlement-bids",
+            labels=("slot",) + Bid.csv_fields(),
+            is_first=is_first)
+
+    def _export_balancing_markets_stats(self, area: Area, directory: dir, is_first: bool) -> None:
+        """Export bids, offers, trades, statistics csv-files for all balancing markets."""
+        if not ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET:
+            return
```
fievelk, line 231: Same comment as here: https://github.com/gridsingularity/d3a/pull/1219/files#r705155641

src/d3a/d3a_core/export.py
```diff
@@ -169,48 +174,106 @@ def move_root_plot_folder(self):
...
+
+    def _export_settlement_markets_stats(self, area: Area, directory: dir, is_first: bool) -> None:
+        """Export bids, offers, trades, statistics csv-files for all settlement markets."""
+        if not ConstSettings.SettlementMarketSettings.ENABLE_SETTLEMENT_MARKETS:
+            return
+        self._export_area_stats_csv_file(area, directory, PastDataEnum.SETTLEMENT_MARKET, is_first)
+        if not area.children:
+            return
+        self._export_trade_csv_files(
+            past_markets=list(area.past_settlement_markets.values()),
```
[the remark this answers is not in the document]
fievelk: The code is basically the same for all `_export_*_markets_stats()` methods. However, you pass the markets in different ways to the export CSV methods:
```python
# spot markets
past_markets=area.past_markets
# balancing markets
past_markets=area.past_balancing_markets
# settlement markets
past_markets=list(area.past_settlement_markets.values())  # <-- different!
```
As far as I see in `AreaMarkets`, there is no difference in the type of structure of these variables: https://github.com/gridsingularity/d3a/blob/0368635e2bb26db3d5fadb64efff0d6e20ea190d/src/d3a/models/area/markets.py#L45-L47
So why can't we just use the following?
```python
past_markets=area.past_settlement_markets
```

src/d3a/d3a_core/export.py
```diff
@@ -309,26 +352,28 @@ def plot_device_stats(self, area: Area, node_address_list: list):
                 address_list = new_node_address_list + [child.name]
                 self._plot_device_stats(address_list, child.strategy)
 
+    @staticmethod
+    def _get_from_dict(data_dict: Dict, map_list: List) -> Mapping:
+        """Get data from dict by following a map_list."""
```
fievelk, line 357: Nit: I think the docstring might be further improved by providing an example.
```suggestion
        """Get nested data from a dict by following the keys provided in a list.
        
        >>> dictionary = {"Parent 1": {"Child 1": "value 1"}, "Parent 2": {"Child 2": "value 2"}}
        >>> keys_path = ["Parent 2", "Child 2"]
        >>> _get_from_dict(dictionary, keys_path)
        >>> 'value 2'
        """
```
(This is the [doctest syntax](https://docs.python.org/3/library/doctest.html))

src/d3a/d3a_core/sim_results/endpoint_buffer.py
```diff
@@ -131,26 +145,46 @@ def generate_json_report(self):
...
+    def _read_settlement_markets_stats_to_dict(self, area: "Area") -> Dict[str, Dict]:
+        """Loop over all settlement markets and return market_stats for each in a dict."""
+        stats_dict = {}
+        for time_slot, market in area.settlement_markets.items():
+            stats_dict[time_slot.format(DATE_TIME_FORMAT)] = (
+                self._read_market_stats_to_dict(market))
+        return stats_dict
+
+    @staticmethod
+    def _read_market_stats_to_dict(market: "Market") -> Dict:
```
fievelk, line 157: I have a question: for spot markets, we only use `_read_market_stats_to_dict` to get the stats of one single market (`area.current_market`). For settlement markets, instead, we cycle through all the `area.settlement_markets`.
Could you please clarify why this is like this? What about the other past spot markets?

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -15,92 +15,59 @@
...
-
-    def __call__(self, area):
-        self._populate_area_children_data(area)
-
-    def _populate_area_children_data(self, area):
-        if area.children:
-            for child in area.children:
-                self._populate_area_children_data(child)
-        self.update_plot_stats(area)
+class PastDataEnum(Enum):
```
fievelk, line 30: I think the name of this class could be improved a bit: "past data" doesn't say much about the meaning of this class. What the class contains are actually the types of available markets that we have in the entire exchange.
I would vote for naming this class something like `AVAILABLE_MARKET_TYPES` (if it's not already used) and maybe using it in other parts of the code in the future (not just here in the export modules).

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -15,92 +15,59 @@
...
+}
 
-    def update_plot_stats(self, area):
-        self._get_stats_from_market_data(self.plot_stats, area, False)
-        if ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET:
-            self._get_stats_from_market_data(self.plot_balancing_stats, area, True)
-        if ConstSettings.GeneralSettings.EXPORT_SUPPLY_DEMAND_PLOTS:
-            self._populate_plots_stats_for_supply_demand_curve(area)
 
+class ExportDataBase(ABC):
```
fievelk, line 44: Class names should possibly be nouns, not verbs. So I would suggest this renaming:
- `BaseDataExporter`
- `UpperLevelDataExporter`
- `BalancingDataExporter`
- `LeafDataExporter`

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -15,92 +15,59 @@
...
+    @abstractmethod
+    def labels(self) -> List:
+        """Labels for csv headers."""
 
-    @staticmethod
-    def create(area):
-        return ExportUpperLevelData(area) if len(area.children) > 0 else ExportLeafData(area)
+    @abstractmethod
+    def rows(self) -> List:
+        """Return rows containing the offer, bids, trades data."""
```
fievelk, line 52: You're defining these as abstract methods, but you're implementing them as abstract properties. Pylint also detects this and is giving you a warning:
> pylint: warning
> W0236 - Method 'labels' was expected to be 'method', found it instead as 'property' (invalid-overridden-method)
So I would recommend to either implement these as methods everywhere, os to implement them as abstract properties in the ABC class:
```python
@property
@abstractmethod
def labels(self) -> List:
    # ...

File: src/d3a/d3a_core/sim_results/file_export_endpoints.py [hunk 13]
```diff
@@ -112,44 +79,47 @@ def _row(self, slot, market):
...
-            return ['desired energy [kWh]', 'deficit [kWh]']
+            return ["desired energy [kWh]", "deficit [kWh]"]
         elif isinstance(self.area.strategy, PVStrategy):
-            return ['produced [kWh]', 'not sold [kWh]']
+            return ["produced [kWh]", "not sold [kWh]"]
         return []
 
     def rows(self):
-        return [self._row(m.time_slot, m) for m in self.area.parent.past_markets]
+        return [self._row(m.time_slot, m) for m in self.past_markets]
```
fievelk, line 122:
```suggestion
        return [self._row(market.time_slot, market) for market in self.past_markets]
```

src/d3a/d3a_core/sim_results/file_export_endpoints.py
```diff
@@ -176,3 +146,74 @@ def _specific_row(self, slot, market):
...
+    @staticmethod
+    def export_data_factory(area: Area, past_market_type: PastDataEnum) -> ExportDataBase:
+        """Decide which data acquisition class to use."""
+        if past_market_type == PastDataEnum.SPOT_MARKET:
+            return ExportUpperLevelData(area.past_markets) \
+                if len(area.children) > 0 else ExportLeafData(area, area.past_markets)
+        if past_market_type == PastDataEnum.BALANCING_MARKET:
+            return ExportBalancingData(area.past_balancing_markets)
+        if past_market_type == PastDataEnum.SETTLEMENT_MARKET:
+            past_markets = list(area.past_settlement_markets.values())
```
fievelk, line 179:
```suggestion
            past_markets = area.past_settlement_markets.values()
```

src/d3a/models/strategy/settlement/strategy.py
```diff
@@ -125,31 +125,40 @@ def event_market_cycle(self, strategy: BaseStrategy) -> None:
...
+        self.bid_updater.increment_update_counter_all_markets(strategy)
+        self.offer_updater.increment_update_counter_all_markets(strategy)
 
-    def event_tick(self, strategy: BaseStrategy) -> None:
+    def event_tick(self, strategy: BidEnabledStrategy) -> None:
         """
-        Update posted settlement bids and offers on market tick
+        Update posted settlement bids and offers on market tick.
+        Order matters here:
+            - FIRST: the bis and offers need to be updated (update())
```
fievelk, line 136: Nit
```suggestion
            - FIRST: the bids and offers need to be updated (update())
```

src/d3a/models/strategy/settlement/strategy.py
```diff
@@ -125,31 +125,40 @@ def event_market_cycle(self, strategy: BaseStrategy) -> None:
...
 
-    def _get_settlement_market_by_id(
-            self, strategy: BaseStrategy, market_id: str) -> Optional["Market"]:
+        self.bid_updater.increment_update_counter_all_markets(strategy)
+        self.offer_updater.increment_update_counter_all_markets(strategy)
+
+    @staticmethod
+    def _get_settlement_market_by_id(strategy: BidEnabledStrategy,
+                                     market_id: str) -> Optional["Market"]:
         markets = [m for m in strategy.area.settlement_markets.values() if m.id == market_id]
```
fievelk, line 154: Nit
```suggestion
        markets = [market_slot for market_slot in strategy.area.settlement_markets.values() if market_slot.id == market_id]
```
I know this is more verbose, but we discussed about the need to refactor our code to reduce ambiguity between markets, market slots, etc.

src/d3a/setup/settlement_market/default_2a_settlement.py
```diff
@@ -0,0 +1,76 @@
+"""
```
fievelk, line 1: Why is the file called `default_2a_settlement.py`? What's `2a`?

fievelk:
General comments:
- Please add unit-tests for the export classes (`ExportDataBase` and its children)
- Some methods became public, so they also became perfect candidates to be unit-tested (e.g. `to_csv` and `csv_fields`)
- Some important changes are not currently unit-tested: 
    - https://github.com/gridsingularity/d3a/pull/1219/files?file-filters%5B%5D=.py#diff-1f9e31cab3c95b71b905a269f0f618a2688ce0b70db49a0788881574acada0bcR341-R342
    - https://github.com/gridsingularity/d3a/pull/1219/files?file-filters%5B%5D=.py#diff-0dd20e06e48cecfc18cfafc4c6e393cd8ff85c51bdc73ebbce55417410a2079bR129-R130
    - https://github.com/gridsingularity/d3a/pull/1219/files?file-filters%5B%5D=.py#diff-0dd20e06e48cecfc18cfafc4c6e393cd8ff85c51bdc73ebbce55417410a2079bR148-R149
    - https://github.com/gridsingularity/d3a/pull/1219/files?file-filters%5B%5D=.py#diff-0a1a500bb749e6236cc4a29fdf8d57f58e5d47528a15d72413afad9164c0ff35R608-R609

Nits:
- The `ExportAndPlot` class is almost 800 LOC long. I believe that the `export.py` module (and the class inside of it) deserves to be split and tested thoroughly. I would personally vote for creating a ticket for this. 
- As Faizan said, there's quite a lot of redundant code in the export-related modules. We can address the issue some other time, but in that case I suggest creating a ticket for it (same as the previous point).
- What is the `is_first` parameter used for? If it's just used to avoid repeating CSV headers, this could be replaced by just checking if the file exists/is empty.
- Please refactor `'` into `"`
- Please replace `\` with `()` and - possibly - hanging indentation
- Some imports seem not to be optimized yet
- What is `src/d3a/setup/settlement_market/__init__.py` for? I think in python3 `__init__.py` files are not needed to identify packages.

src/d3a/setup/settlement_market/default_2a_settlement.py
hannesdiedrich: The default_2a setup file is the setup we alway us for sanity checking, is has been there since the very beginning of d3a and if you have a look, there are multiple adaptions of this setup for different external settings like market type. I thought to stick to the practice but to append the special setting. Will rename it to something more explanatory.

src/d3a/d3a_core/export.py
```diff
@@ -169,48 +174,106 @@ def move_root_plot_folder(self):
             shutil.move(os.path.join(old_dir, si), self.plot_dir)
         shutil.rmtree(old_dir)
 
-    def _export_area_with_children(self, area: Area, directory: dir, is_first: bool = False):
+    def _export_spot_markets_stats(self, area: Area, directory: dir, is_first: bool) -> None:
```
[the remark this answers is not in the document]
hannesdiedrich: I played around a lot with this in order to reduce the code repetition. However, I would have to add if/elif/else blocks to the potential universal function that would make it really ugly. Also, for the balancing markets we currently do not export the bids.

src/d3a/d3a_core/export.py
hannesdiedrich: you are right, there shouldn't be any past settlement markets if the settlement markets are disabled. However, we use the global setting in this module anyway. Is there a performance benefit when checking for whether a dictionary is empty in comparison to checking a global flag?

WRT your second question: CSV files should be generated even if there are no trades or bids or offers in the markets. Which is a pro argument to go with the global flag actually.

In general, this module is really not that flexible. This is really old code and I tried to make it a bit more flexible actually. Due to time contraints I wasn't able to refactor the whole module. This is worth a separate story IMO.

src/d3a/d3a_core/export.py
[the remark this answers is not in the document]
hannesdiedrich: This is because `area.past_settlement_markets -> Dict` and ´area.past_markets -> List´ . Have a look into the are class. I discussed this briefly with Spyros and we decided to leave it as this as we use he dictionary behaviour (time_slots are available) in other parts of the code for the settlement markets.

src/d3a/d3a_core/export.py
hannesdiedrich: My answer there also applies here.

src/d3a/d3a_core/sim_results/file_export_endpoints.py
hannesdiedrich: good catch. This. was due to copy and paste.

src/d3a/models/strategy/settlement/strategy.py
hannesdiedrich: I link to this discussion here https://github.com/gridsingularity/d3a-interface/pull/246#discussion_r705945794

src/d3a/setup/settlement_market/default_2a_settlement.py
fievelk: Maybe we could rename the `default_2a` file to something else (and later use different suffixes as you did here). I think the main issue is the base name we already had, not your adaptation in the context of this PR.

src/d3a/d3a_core/export.py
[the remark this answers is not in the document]
fievelk: Given your conversation, I think that these could actually be classes rather than functions. There would be a common interface (no universal function) and no need to use `if` statements, since each class would have its own implementation. Maybe this could be a way to proceed if you think it's suitable and you want to spend some time on it.

src/d3a/d3a_core/export.py
fievelk: > you are right, there shouldn't be any past settlement markets if the settlement markets are disabled. However, we use the global setting in this module anyway. Is there a performance benefit when checking for whether a dictionary is empty in comparison to checking a global flag?
I don't think there's any real performance benefit. There would be a code flexibility benefit though, because we would not need to rely on the global setting to use this function. It's always good to have one less dependency (less error-prone, less refactorings, minor complexity for unit-tests, no teardown issues for unit-tests in this case, etc.).

fievelk: If the CSV files should be generated even without trades, then the flag might still be a good way to proceed anyway, as you said.

hannesdiedrich:
> General comments:
> 
> * Please add unit-tests for the export classes (`ExportDataBase` and its children)
> * Some methods became public, so they also became perfect candidates to be unit-tested (e.g. `to_csv` and `csv_fields`)
> * Some important changes are not currently unit-tested:
>   
>   * https://github.com/gridsingularity/d3a/pull/1219/files?file-filters%5B%5D=.py#diff-1f9e31cab3c95b71b905a269f0f618a2688ce0b70db49a0788881574acada0bcR341-R342
>   * https://github.com/gridsingularity/d3a/pull/1219/files?file-filters%5B%5D=.py#diff-0dd20e06e48cecfc18cfafc4c6e393cd8ff85c51bdc73ebbce55417410a2079bR129-R130
>   * https://github.com/gridsingularity/d3a/pull/1219/files?file-filters%5B%5D=.py#diff-0dd20e06e48cecfc18cfafc4c6e393cd8ff85c51bdc73ebbce55417410a2079bR148-R149
>   * https://github.com/gridsingularity/d3a/pull/1219/files?file-filters%5B%5D=.py#diff-0a1a500bb749e6236cc4a29fdf8d57f58e5d47528a15d72413afad9164c0ff35R608-R609
> 
> Nits:
> 
> * The `ExportAndPlot` class is almost 800 LOC long. I believe that the `export.py` module (and the class inside of it) deserves to be split and tested thoroughly. I would personally vote for creating a ticket for this.
> * As Faizan said, there's quite a lot of redundant code in the export-related modules. We can address the issue some other time, but in that case I suggest creating a ticket for it (same as the previous point).
> * What is the `is_first` parameter used for? If it's just used to avoid repeating CSV headers, this could be replaced by just checking if the file exists/is empty.
> * Please refactor `'` into `"`
> * Please replace `\` with `()` and - possibly - hanging indentation
> * Some imports seem not to be optimized yet
> * What is `src/d3a/setup/settlement_market/__init__.py` for? I think in python3 `__init__.py` files are not needed to identify packages.

Thanks a lot for your comments,
I added the refactoring story for the export.py : https://gridsingularity.atlassian.net/browse/D3ASIM-3600
and will add the unittests for the important changes in the settlement strategies. I added the unittest of the ExportDataBase to the above mentioned story because this is not crucial code (the local export is hardly used any more). I hope you agree with that.
Of course will add all the other refactorings you suggested. Thanks for the tip with the __init__.py. I wasn't aware of this.

src/d3a/d3a_core/export.py
faizan2590: https://user-images.githubusercontent.com/33482427/132870684-113678a9-7629-4450-a3f4-3e5ebbbbcec3.mov

Tried, the ordered stayed as it is now

src/d3a/d3a_core/export.py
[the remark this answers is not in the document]
fievelk: Thanks. I think it would be good to refactor either one or the other (in order to always have the same return type). Maybe another time.

src/d3a/d3a_core/export.py
fievelk: Could you please check your settings? I'm not sure what the issue might be, but the order you have is not correct: imports from `d3a_interface` (which is an external library) should be above the ones from `d3a`.

image

faizan2590: Screenshot 2021-09-10 at 18 47 54


I think I have the same settings as you

src/d3a/d3a_core/export.py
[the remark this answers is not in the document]
faizan2590: To be dealt with as part of refactoring story: https://gridsingularity.atlassian.net/browse/D3ASIM-3600

tox.ini
```diff
@@ -76,7 +76,7 @@ commands =
     pip-sync requirements/tests.txt requirements/pandapower.txt
     pip install -e .
 	pip uninstall -y d3a-interface
-	pip install git+https://github.com/gridsingularity/d3a-interface@{env:BRANCH:master}
+	pip install git+https://github.com/gridsingularity/d3a-interface@feature/D3ASIM-3539
```
faizan2590, line 79: reminder to revert before merging

src/d3a/d3a_core/export.py
faizan2590: > If the CSV files should be generated even without trades, then the flag might still be a good way to proceed anyway, as you said.

Thinking from testing perspective, its better to not have an empty CSVs, else from the testing side one need to parse those CSVs too to validate

src/d3a/d3a_core/sim_results/endpoint_buffer.py
faizan2590: IMO, settlement market should too send the results  of `last_past_settlement_market`.

Tagging @spyrostz in case something related was discussed already that I am not aware of

faizan2590:
> What is src/d3a/setup/settlement_market/__init__.py for? I think in python3 __init__.py files are not needed to identify packages.

@fievelk I just checked it and its indeed needed, and one of the unittest `test_validate_all_setup_scenarios_are_available` failed too

fievelk:
> > What is src/d3a/setup/settlement_market/**init**.py for? I think in python3 **init**.py files are not needed to identify packages.
> 
> @fievelk I just checked it and its indeed needed, and one of the unittest `test_validate_all_setup_scenarios_are_available` failed too

Interesting. It would be nice to understand why exactly this happens in our case. Anyway, I found this resource that helps understanding when `__init__.py` files are needed and when/how they become a "trap":
- [Traps for the Unwary in Python’s Import System ](http://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html)

src/d3a/models/market/market_structures.py
```diff
@@ -408,3 +409,17 @@ def parse_event_and_parameters_from_json_string(payload) -> Tuple:
...
+
+
+class AvailableMarketTypes(Enum):
+    """Enum for deciding which markets to access."""
+    SPOT_MARKET = 1
+    BALANCING_MARKET = 2
+    SETTLEMENT_MARKET = 3
+
+
+market_type_file_suffix_dict = {
```
fievelk, line 421: Nit: same as here: https://github.com/gridsingularity/d3a/pull/1219#discussion_r705127605

src/d3a/d3a_core/export.py
fievelk: @faizan2590 this is what Hannes answered:
> WRT your second question: CSV files should be generated even if there are no trades or bids or offers in the markets. Which is a pro argument to go with the global flag actually.

I don't know exactly why we want to generate empty CSV files. I would personally be up for not generating them as you said, but if that's a requirement then we should keep them.

src/d3a/d3a_core/sim_results/endpoint_buffer.py
spyrostz: I think that last_past_settlement_market should be more appropriate here. The only reason I can think of to send the open settlement markets to the results service is to report the progress of the settlement market gradually, but I do not think that this is the intent here.

src/d3a/d3a_core/export.py
```diff
@@ -380,13 +425,11 @@ def plot_all_unmatched_loads(self):
...
         Plot unmatched loads of all loads in the configuration into one plot
         """
-        unmatched_key = 'deficit [kWh]'
-        data = list()
         root_name = self.area.slug
-        title = 'Unmatched Loads for all devices in {}'.format(root_name)
-        xtitle = 'Time'
-        ytitle = 'Energy (kWh)'
-        barmode = 'stack'
+        plot_desc = PlotDescription(data=[], barmode="stack",
```
fievelk, line 429: One nit and one more important comment.
Nit:
```suggestion
        plot_description = PlotDescription(data=[], barmode="stack",
```
These lines seem not to be tested. Hannes said that the exporter should be deprecated; if this will happen soon, then you can ignore it. If it won't happen soon, then I'd suggest to write tests.

src/d3a/d3a_core/export.py
spyrostz: The case is evident if you check the behavior of the balancing market. Under normal circumstances, when the balancing market is disabled, the csv files for the balancing market are not exported. Give it a try and run the d3a with e.g. default_2a, you will see that no e.g. balancing-offer.csv files will be generated. 

Hannes' case is the same, that he does not want to generate the csv files at all if the settlement market is disabled by configuration. What he also tried to explain and the confusion started, was that there is a case where the settlement markets are enabled, but they do not have any offers / bids / trades due to misconfiguration of the trading agents. In this case, we DO want the csv files to be generated and be empty in order to indicate that settlement markets are enabled but do not have any trades / offers / bids. All these checks can be done only by checking the settlement markets dict, I agree, but I think it is much easier for the reader of the code to see the constant here.

tox.ini
fievelk: Reminder

fievelk:
I left some comments.
Besides that:
- please refactor `\` into `()`
- please optimize imports

Regarding imports: maybe your PyCharm (@faizan2590) is declaring d3a-interface as some kind of special dependency? That might explain why it's being imported after d3a (while it should come before it).

src/d3a/d3a_core/export.py
faizan2590: https://github.com/gridsingularity/d3a/pull/1219/commits/6c441c8932252514d0f7e5bbcb0672ec2f390395

faizan2590:
> > > What is src/d3a/setup/settlement_market/**init**.py for? I think in python3 **init**.py files are not needed to identify packages.
> > 
> > 
> > @fievelk I just checked it and its indeed needed, and one of the unittest `test_validate_all_setup_scenarios_are_available` failed too
> 
> Interesting. It would be nice to understand why exactly this happens in our case. Anyway, I found this resource that helps understanding when `__init__.py` files are needed and when/how they become a "trap":
> 
> * [Traps for the Unwary in Python’s Import System ](http://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html)

Would avoid to go deep into it now, but will definitely keep an eye to find the answer

faizan2590:
> I left some comments.
> Besides that:
> 
> * please refactor `\` into `()`

Done


> * please optimize imports
> 
> Regarding imports: maybe your PyCharm (@faizan2590) is declaring d3a-interface as some kind of special dependency? That might explain why it's being imported after d3a (while it should come before it).

Don't, will reinstall pycharm later to double check, but for now explicitly moved d3a_interface before d3a

src/d3a/models/market/market_structures.py
```diff
@@ -408,3 +409,17 @@ def parse_event_and_parameters_from_json_string(payload) -> Tuple:
         kwargs["bid_trade"] = trade_from_json_string(kwargs["bid_trade"])
     event_type = MarketEvent(data["event_type"])
     return event_type, kwargs
+
+
+class AvailableMarketTypes(Enum):
```
spyrostz, line 414: I have a similar enum in feature/D3ASIM-3589, but don't worry about it. I'll delete one of the 2 once both PRs are merged

spyrostz (approved):
LGTM

fievelk (approved):

PR #1221, opened 2021-09-03
Feature/d3 asim 3589

spyrostz:
Added event dispatcher for the settlement markets. Adapted MycoMatcher to handle also settlement markets. Adapted the market classes to have dedicated debug info and a settlement market type.


src/d3a/models/market/two_sided.py
```diff
@@ -52,20 +52,30 @@ def __init__(self, time_slot=None, bc=None, notification_listener=None, readonly
         super().__init__(time_slot, bc, notification_listener, readonly, grid_fee_type,
                          grid_fees, name, in_sim_duration=in_sim_duration)
 
+    @property
+    def _class_name(self):
+        return "TwoSidedPayAsBid"
```
faizan2590, line 57:
```suggestion
        return "TwoSidedMarket"
```

Not guaranteed to always be pay-as-bid at this point.

src/d3a/models/area/__init__.py
```diff
@@ -349,8 +349,9 @@ def cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fal
             self._update_descendants_strategy_prices()
             self.should_update_child_strategies = False
 
+        from d3a.models.area.market_rotators import MarketClassType
```
faizan2590, line 352: NIT: To avoid lazy imports

src/d3a/models/area/event_dispatcher.py
```diff
@@ -46,6 +49,7 @@ class AreaDispatcher:
     def __init__(self, area):
         self._inter_area_agents = {}  # type: Dict[DateTime, Dict[str, OneSidedAgent]]
         self._balancing_agents = {}  # type: Dict[DateTime, Dict[str, BalancingAgent]]
+        self._settlement_agents = {}  # type: Dict[DateTime, Dict[str, SettlementAgent]]
```
faizan2590, line 52:
```suggestion
        self._settlement_agents: Dict[DateTime, Dict[str, SettlementAgent]] = {}
```

src/d3a/models/area/market_rotators.py
```diff
@@ -15,16 +15,24 @@
...
 
+from d3a import constants
+
 log = getLogger(__name__)
 
 
+class MarketClassType(Enum):
+    SPOT = 0
+    BALANCING = 1
+    SETTLEMENT = 2
```
faizan2590, line 33: Might better belongs to `d3a/src/d3a/models/market/__init__.py`

src/d3a/models/market/one_sided.py
```diff
@@ -47,15 +47,25 @@ def __init__(self, time_slot=None, bc=None, notification_listener=None,
...
+                sum(o.energy for o in self.offers.values()),
+                sum(o.price for o in self.offers.values()),
+                len(self.trades),
+                self.accumulated_trade_energy,
+                self.accumulated_trade_price
+            )
+
+    @property
+    def _class_name(self):
+        return "OneSided"
```
faizan2590, line 64:
```suggestion
        return "OneSidedMarket"
```

src/d3a/models/market/two_sided.py
```diff
@@ -373,3 +387,14 @@ def replace_recommendations_with_residuals(recommendation: Dict):
...
+
+class SettlementMarket(TwoSidedMarket):
+
+    @property
+    def _class_name(self):
+        return "Settlement"
+
+    @property
+    def _debug_log_market_type_identifier(self):
+        return "[SETTLEMENT]"
```
faizan2590, line 400: How about creating a module `src/d3a/models/market/settlement_market.py` and moving it there. Although now now `SettlementMarket ` is only a wrapper of `TwoSidedMarket `

src/d3a/models/market/two_sided.py
```diff
@@ -373,3 +387,14 @@ def replace_recommendations_with_residuals(recommendation: Dict):
             recommendations = [replace_recommendations_with_residuals(recommendation)
                                for recommendation in recommendations]
         return recommendations
+
+
+class SettlementMarket(TwoSidedMarket):
+
+    @property
+    def _class_name(self):
+        return "Settlement"
```
faizan2590, line 396:
```suggestion
        return "SettlementMarket"
```

src/d3a/models/market/two_sided.py
spyrostz: Works for me, will do so

src/d3a/models/area/market_rotators.py
spyrostz: Will move it thanks

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -274,8 +279,9 @@ def _delete_offer_aggregator(self, arguments):
 
     def _list_offers_aggregator(self, arguments):
         try:
+            market = self._get_market_from_command_argument(arguments)
             filtered_offers = [{"id": v.id, "price": v.price, "energy": v.energy}
-                               for _, v in self.next_market.get_offers().items()
+                               for _, v in market.get_offers().items()
```
faizan2590, line 284:
```suggestion
                               for v in market.get_offers().values()
```

tests/strategies/external/test_load.py
```diff
@@ -43,3 +47,45 @@ def test_list_bids_succeeds(self, ext_load_fixture):
...
+            }
+        )
+        assert_bid_offer_aggregator_commands_return_value(return_value, False)
+
+    def test_delete_bid_aggregator(self, ext_load_fixture):
+        bid = ext_load_fixture.post_bid(ext_load_fixture.next_market, 200.0, 1.0)
+
+        return_value = ext_load_fixture.trigger_aggregator_commands(
+            {
+                "type": "delete_bid",
```
faizan2590, line 69: I think `time_slot` being part of it too . Right?

spyrostz: These are generic tests for the aggregator commands, I am not testing the time_slot functionality here. This is tested by test_get_market_from_cmd_arg_returns_settlement_market , test_get_market_from_cmd_arg_returns_spot_market tests

tests/strategies/settlement/test_settlement_strategy.py
```diff
@@ -81,22 +81,24 @@ def test_event_tick_updates_bids_and_offers(
...
         strategy_fixture.area.config = Mock()
         strategy_fixture.area.config.ticks_per_slot = 60
         strategy_fixture.area.config.tick_length = duration(seconds=15)
         self.market_mock.bid.reset_mock()
         self.market_mock.offer.reset_mock()
 
+        strategy_fixture.area.current_tick = 19
+        self.settlement_strategy.event_tick(strategy_fixture)
+        strategy_fixture.area.current_tick = 20
         self.settlement_strategy.event_tick(strategy_fixture)
```
faizan2590, line 93: Sorry, didn't get the intuition of calling `event_tick ` twice

src/d3a/setup/settlement_market/settlement_market.py
```diff
@@ -0,0 +1,77 @@
...
+
+from d3a.models.area import Area
+from d3a.models.strategy.external_strategies.load import (LoadHoursExternalStrategy,
+                                                          LoadForecastExternalStrategy)
+from d3a.models.strategy.external_strategies.pv import PVExternalStrategy
+from d3a.models.strategy.load_hours import LoadHoursStrategy
+from d3a.models.strategy.storage import StorageStrategy
+
+
+def get_setup(config):
```
faizan2590, line 29: Is there any integration tests expected to use this setup file?

tests/strategies/settlement/test_settlement_strategy.py
spyrostz: This is related to the change in the SettlementMarketStrategy.event_tick method. You will notice that the order of the offer/bid update and increment/update counter has been reversed (now the offer is updated before the counter is incremented). This is mandatory to have it in that order because AFAIR the  UpdateFrequencyMixin class works that way (Load and PV strategies follow the same logic, first they update their bids and offers, then they increment the counter of the UpdateFrequencyMixin). This has the effect that if the event_tick is only called once, then the bids and the offers of the strategies will not be updated (because the update counter will not have incremented yet). Therefore a second tick is really required in order for the bids and the offers to be updated with a new energy rate.

src/d3a/setup/settlement_market/settlement_market.py
spyrostz: For now this works mostly as an example, but an integration test will follow soon

src/d3a/models/market/one_sided.py
```diff
@@ -47,15 +47,25 @@ def __init__(self, time_slot=None, bc=None, notification_listener=None,
...
-        return "<OneSidedMarket{} offers: {} (E: {} kWh V: {}) trades: {} (E: {} kWh, V: {})>"\
-            .format(" {}".format(self.time_slot_str),
-                    len(self.offers),
-                    sum(o.energy for o in self.offers.values()),
-                    sum(o.price for o in self.offers.values()),
-                    len(self.trades),
-                    self.accumulated_trade_energy,
-                    self.accumulated_trade_price
-                    )
+        return "<{}{} offers: {} (E: {} kWh V: {}) trades: {} (E: {} kWh, V: {})>"\
```
hannesdiedrich, line 50: Thanks a lot for refactoring this. I as also about to.

hannesdiedrich (approved):
LGTM

src/d3a/models/myco_matcher/myco_internal_matcher.py
```diff
@@ -58,18 +58,21 @@ def get_matching_algorithm():
...
-                    data = {
-                        market.id:
-                            {"bids": [bid.serializable_dict() for bid in bids.values()],
-                             "offers": [offer.serializable_dict() for offer in offers.values()],
-                             "current_time": area_data["current_time"]}}
-                    bid_offer_pairs = self._get_matches_recommendations(data)
-                    if not bid_offer_pairs:
-                        break
-                    market.match_recommendations(bid_offer_pairs)
+            for market_type in ["markets", "settlement_markets"]:
```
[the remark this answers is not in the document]
spyrostz: No, the settlement market will not be handled by Mycos ATM, and I am not sure if it ever will.

src/d3a/models/area/event_dispatcher.py
```diff
@@ -144,10 +150,47 @@ def create_agent_object(owner, higher_market, lower_market, is_spot_market):
...
+        elif market_type == MarketClassType.SETTLEMENT:
+            return SettlementAgent(**agent_constructor_arguments)
+        elif market_type == MarketClassType.BALANCING:
             return BalancingAgent(**agent_constructor_arguments)
+        else:
+            assert False, f"Market type not supported {market_type}"
 
-    def create_area_agents(self, is_spot_market, market):
+    @staticmethod
+    def _get_agents_for_market_type(dispatcher_object, market_type: MarketClassType):
```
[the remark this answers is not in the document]
spyrostz: This unfortunately is not feasible. This is a method of the AreaDispatcher class, therefore the AreaDispatcher class name is not visible from inside the class, thus the type hint will fail.

src/d3a/models/area/event_dispatcher.py
```diff
@@ -123,14 +129,14 @@ def event_listener(self, event_type: Union[MarketEvent, AreaEvent], **kwargs):
             self.area.strategy.event_on_disabled_area()
 
     @staticmethod
-    def create_agent_object(owner, higher_market, lower_market, is_spot_market):
+    def create_agent_object(owner, higher_market, lower_market, market_type):
```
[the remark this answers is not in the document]
spyrostz: First of all, double quotes does not remove the need to import the class of the type hint. Even in this case we need to do `from d3a.models.markets import Market` in order for python interpreter to not complain. As a followup to that, if I import the Area class in this module I will end up with a circular import error. I can avoid it by using this https://stackoverflow.com/a/39757388, but conditional imports only for type hints is very ugly IMO and IIRC we have discussed to not do it and simply not add type hints in these cases. TLDR, I will not add type hint for the owner argument

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -347,12 +348,16 @@ def _publish_trade_event(self, trade, is_bid_trade):
 
     def event_bid_traded(self, market_id, bid_trade):
         super().event_bid_traded(market_id=market_id, bid_trade=bid_trade)
-        if self.connected or self.redis.aggregator.is_controlling_device(self.device.uuid):
+        if (self.connected or
+                (self.redis.aggregator is not None
+                 and self.redis.aggregator.is_controlling_device(self.device.uuid))):
```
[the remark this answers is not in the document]
spyrostz: There is already such a property, `is_aggregator_controlled`. I'll reuse it

src/d3a/models/area/event_dispatcher.py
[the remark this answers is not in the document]
fievelk: I think using strings works in this case, as described in [PEP 484 (Forward references)](https://www.python.org/dev/peps/pep-0484/#forward-references)

src/d3a/models/area/event_dispatcher.py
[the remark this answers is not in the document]
fievelk: I haven't tested this, but [PEP 484 (Forward references)](https://www.python.org/dev/peps/pep-0484/#forward-references) describes a way to overcome circular imports:
> A common use for forward references is when e.g. Django models are needed in the signatures. Typically, each model is in a separate file, and has methods taking arguments whose type involves other models. Because of the way circular imports work in Python, it is often not possible to import all the needed models directly [...]
The solution is to switch to module-only imports and reference the models by their `_module_._class_ name`.

So for example:
```python
from models import b
class A(Model):
    def foo(self, b: 'b.B'):
        pass
```

src/d3a/models/area/event_dispatcher.py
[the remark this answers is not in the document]
fievelk: Another method would be to use:
```python
from __future__ import annotations
```
(Source: [How do I type hint a method with the type of the enclosing class?](https://stackoverflow.com/a/33533514/1916588))

src/d3a/models/area/event_dispatcher.py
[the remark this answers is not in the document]
spyrostz: I also do not like this solution actually, mainly due to the fact that module-based imports can be very broad, and also because they can lead to long/verbose types (e.g. one type hint could be d3a.models.area.Area which is overly verbose). More specifically in this example, this would not solve our problem because the Area class is on the d3a.models.area.__init__ module. Therefore when one does `from d3a.models import area` , the Area class will also be imported thus causing the circular import, despite the fact that area.Area would be used. 
The correct way to solve this is not by trying to cheat the Python interpreter, but by really refactoring the code in order to avoid such occurrences in the first place (by moving classes around files in a way that circular imports are avoided). However this is something that should be added as a separate task and not as part of this ticket.

src/d3a/models/area/event_dispatcher.py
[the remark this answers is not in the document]
spyrostz: Indeed the forward reference worked here, fixing this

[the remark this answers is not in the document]
Hassan754: I'm not sure about this but I think that using a string typehint will raise a warning in pylint if the reference is not imported, this way the IDEs won't recognize the type, but we can live with that atm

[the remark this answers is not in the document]
spyrostz: The forward reference (double quotes) works for me, already tried it with pylint and does not complain.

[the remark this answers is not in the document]
faizan2590: This raises another question of why not changing this method from static_method to instance_method ?

faizan2590 (commented):
Left one comment to a [thread](https://github.com/gridsingularity/d3a/pull/1221/files#r707225471), will LGTM once addressed/resolved.

src/d3a/models/area/event_dispatcher.py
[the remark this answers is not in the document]
spyrostz: This is not called only for `self`, but also with `area.parent.event_dispatcher` as an argument, in order to connect 2 IAAs. Therefore it is not possible to change it to an instance method which implies that the argument will only be `self`

faizan2590 (dismissed):
LGTM

src/d3a/d3a_core/redis_connections/redis_area_market_communicator.py
```diff
@@ -97,11 +97,15 @@ def publish_json(self, channel, data):
...
     def __init__(self, is_enabled):
         self.is_enabled = is_enabled
+        self.aggregator = None
         if self.is_enabled:
             super().__init__()
             self.channel_callback_dict = {}
             self.aggregator = AggregatorHandler(self.redis_db)
 
+    def is_aggregator_controlling_device(self, device_uuid):
+        return self.aggregator and self.aggregator.is_controlling_device(device_uuid)
```
fievelk, line 107: It looks like this method is never called. Is it a leftover?

src/d3a/d3a_core/util.py
```diff
@@ -171,6 +171,10 @@ def make_ba_name(owner):
     return f"BA {owner.name}"
 
 
+def make_sa_name(owner):
```
fievelk, line 174: I suppose that SA stands for "Settlement Agent", right?
I would suggest moving this method inside the `SettlementAgent` class, as that's the only place where it is currently used.

src/d3a/models/area/__init__.py
```diff
@@ -350,8 +351,11 @@ def cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fal
             self._update_descendants_strategy_prices()
             self.should_update_child_strategies = False
 
+        # TODO: Refactor and port the future, spot, settlement and balancing market creation to
+        # AreaMarkets class, in order to create all necessary markets with one call.
+
```
fievelk, line 356: Nit: do we already have a ticket for this? Or should we create one?

src/d3a/models/area/markets.py
```diff
@@ -78,29 +79,44 @@ def rotate_markets(self, current_time: DateTime) -> None:
...
+        elif market_type == AvailableMarketTypes.SETTLEMENT:
+            return self.settlement_markets
+        elif market_type == AvailableMarketTypes.BALANCING:
+            return self.balancing_markets
+        else:
+            assert False, f"Market type not supported {market_type}"
 
     def create_future_markets(self, current_time: DateTime,
-                              is_spot_market: bool, area: "Area") -> bool:
+                              market_type: AvailableMarketTypes, area: "Area") -> bool:
```
fievelk, line 108: I really like this change 👍

src/d3a/d3a_core/util.py
spyrostz: Not a good idea, especially for the future when we add more integration tests for the settlement market. Check out the usages of the make_ba_name and make_iaa_name, they are also used in the context of integration tests.

src/d3a/d3a_core/redis_connections/redis_area_market_communicator.py
spyrostz: Remnant after the merge from master, removing

src/d3a/models/area/__init__.py
spyrostz: I think we had one, but it is probably buried in Outdated. Creating one quickly

src/d3a/models/area/__init__.py
```diff
@@ -414,13 +419,18 @@ def _update_myco_matcher(self) -> None:
         """Update the markets cache that the myco matcher will request"""
         bid_offer_matcher.update_area_uuid_markets_mapping(
             area_uuid_markets_mapping={
-                self.uuid: {"markets": self.all_markets, "current_time": self.now}})
+                self.uuid: {"markets": self.all_markets,
+                            "settlement_markets": list(self.settlement_markets.values()),
```
fievelk, line 423: Is there a specific reason to use `list` on this? The following should work as well:
```suggestion
                            "settlement_markets": self.settlement_markets.values(),
```

spyrostz: Should work fine even without the casting to list, removing

src/d3a/models/area/event_dispatcher.py
```diff
@@ -73,34 +82,33 @@ def broadcast_balancing_market_cycle(self, **kwargs):
...
-        # Broadcast to children in random order to ensure fairness
-        for child in sorted(self.area.children, key=lambda _: random()):
-            child.dispatcher.event_listener(event_type, **kwargs)
-        # Also broadcast to IAAs. Again in random order
-        for time_slot, agents in self._inter_area_agents.items():
-            if time_slot not in self.area._markets.markets:
+    def _broadcast_notification_to_agents_of_market_type(
+            self, market_type: AvailableMarketTypes, event_type: AreaEvent, **kwargs):
+        for time_slot, agents in self._get_agents_for_market_type(self, market_type).items():
+            if time_slot not in self.area._markets.get_market_instances_from_class_type(
```
fievelk, line 88: Here we're accessing a private property of a different class (`area._markets`). We do the same in several other modules, so I think that `area._markets` wants to be made public.

src/d3a/models/area/event_dispatcher.py
```diff
@@ -73,34 +82,33 @@ def broadcast_balancing_market_cycle(self, **kwargs):
...
-                continue
 
-            if not self.area.events.is_connected:
-                break
-            for area_name in sorted(agents, key=lambda _: random()):
-                agents[area_name].event_listener(event_type, **kwargs)
+    def _broadcast_notification(self, event_type: Union[MarketEvent, AreaEvent], **kwargs):
+        if not self.area.events.is_enabled and \
+           event_type not in [AreaEvent.ACTIVATE, AreaEvent.MARKET_CYCLE]:
+            return
```
fievelk, line 101: Nit:
```suggestion
        if not self.area.events.is_enabled and event_type not in [
                AreaEvent.ACTIVATE, AreaEvent.MARKET_CYCLE]:
            return
```

src/d3a/models/area/event_dispatcher.py
```diff
@@ -73,34 +82,33 @@ def broadcast_balancing_market_cycle(self, **kwargs):
...
+
+        # Broadcast to children in random order to ensure fairness
+        for child in sorted(self.area.children, key=lambda _: random()):
+            child.dispatcher.event_listener(event_type, **kwargs)
+        self._broadcast_notification_to_agents_of_market_type(
+            AvailableMarketTypes.SPOT, event_type, **kwargs)
+        self._broadcast_notification_to_agents_of_market_type(
+            AvailableMarketTypes.BALANCING, event_type, **kwargs)
+        self._broadcast_notification_to_agents_of_market_type(
+            AvailableMarketTypes.SETTLEMENT, event_type, **kwargs)
```
fievelk, line 111: Nit: in this class we have many `if` methods to basically distinguish between SPOT, BALANCING and SETTLEMENT market types (like [here](https://github.com/gridsingularity/d3a/blob/ea445ce2d33609c69a7dcd577f6978113475b137/src/d3a/models/area/event_dispatcher.py#L165-L174) and [here](https://github.com/gridsingularity/d3a/blob/ea445ce2d33609c69a7dcd577f6978113475b137/src/d3a/models/area/event_dispatcher.py#L144-L163)). 
I think that this could be refactored into separate classes (each specialized on one market). No need to do it in the context of this PR, but it might be good to keep it in mind and maybe open a refactoring ticket.

spyrostz: I would leave this for when we reimplement the event dispatcher to not forward the bids and the offers from the higher to the lower market, since this is going to be implemented in the near future (this is the last task of the degrees of freedom market)

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -153,7 +153,7 @@ def deactivate(self):
         if self.is_aggregator_controlled:
             deactivate_msg = {'event': 'finish'}
             self.aggregator.add_batch_finished_event(self.area.uuid, deactivate_msg)
-        else:
+        elif self.redis_com.is_enabled:
```
fievelk, line 156: Nit:
```suggestion
        elif self.redis_communicator.is_enabled:
```
(Just for improved clarity)

src/d3a/models/market/one_sided.py
```diff
@@ -49,15 +49,25 @@ def __init__(self, time_slot=None, bc=None, notification_listener=None,
...
+                sum(o.energy for o in self.offers.values()),
+                sum(o.price for o in self.offers.values()),
+                len(self.trades),
+                self.accumulated_trade_energy,
+                self.accumulated_trade_price
+            )
+
+    @property
+    def _class_name(self):
+        return "OneSidedMarket"
```
fievelk, line 66: Nit: this could also be `self.__class__.__name__` (in case we want to make it dynamic)

src/d3a/models/market/one_sided.py
```diff
@@ -291,7 +304,8 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *, energy: in
 
         if already_tracked is False:
             self._update_stats_after_trade(trade, offer)
-            log.info(f"[TRADE] [{self.name}] [{self.time_slot_str}] {trade}")
+            log.info(f"{self._debug_log_market_type_identifier}[TRADE] "
```
fievelk, line 307: Do you confirm that the level of this is actually `info`? I'm asking because of the `_debug_log_market_type_identifier` variable.

src/d3a/models/area/event_dispatcher.py
spyrostz: I would rather do this in isolation after this PR is merged, because this touches many parts of the code and is not related with the PR itself

src/d3a/models/area/event_dispatcher.py
spyrostz: Fixing

src/d3a/models/myco_matcher/myco_internal_matcher.py
```diff
@@ -58,18 +58,21 @@ def get_matching_algorithm():
     def match_recommendations(self, **kwargs):
         """Request trade recommendations and match them in the relevant market."""
         for area_uuid, area_data in self.area_uuid_markets_mapping.items():
```
fievelk, line 60:
```suggestion
        for area_data in self.area_uuid_markets_mapping.values():
```

src/d3a/models/area/redis_external_market_connection.py
spyrostz: Fixing, will make it private too

src/d3a/models/market/one_sided.py
spyrostz: It is info and it is important to stay that way. Otherwise you will not be able to see the trades in the default d3a log level

src/d3a/models/market/one_sided.py
spyrostz: Will do and also remove this from the classes that inherit from this

src/d3a/models/myco_matcher/myco_internal_matcher.py
spyrostz: Will fix this, however keep in mind that this might change again during this task https://gridsingularity.atlassian.net/browse/D3ASIM-3593

fievelk:
General:
- Some parts of the new code are not unit-tested. Codecov is highlighting most of them, in case you want to have a look. 
  I personally believe that private methods should not be tested in general; however, I also have the feeling that some of these methods would actually be good candidates to become public methods of smaller classes (and therefore be tested) maybe in the context of a refactoring PR.

Nits:
- Please replace `\` with `()`
- Some missing docstrings for new classes and public methods

spyrostz:
> General:
> 
>     * Some parts of the new code are not unit-tested. Codecov is highlighting most of them, in case you want to have a look.
>       I personally believe that private methods should not be tested in general; however, I also have the feeling that some of these methods would actually be good candidates to become public methods of smaller classes (and therefore be tested) maybe in the context of a refactoring PR.
> 
> 
> Nits:
> 
>     * Please replace `\` with `()`
> 
>     * Some missing docstrings for new classes and public methods

Regarding the \ , I found only one occurrence in the diff, which already existed there. Did you mean just that ? 
Also regarding the docstrings, really the only new classes are the SettlementMarket and SettlementAgent, both of which inherit from the TwoSided market and agent respectively without any new functionality apart from debug logs. IMO these do not really need docstrings at all. If you insist I will add these, however it is not helpful to my eyes. 

Regarding test coverage, I'll check it out to see exactly the places where github complains, however I doubt that this is due to this PR. E.g. one occurrence I saw from a first glance is the `match_recommendations` method, which was already there in the first place (all I did was to include the settlement markets in the matching) therefore won't be able to do much. I'll give it a closer look and let you know.

fievelk:
> Regarding the \ , I found only one occurrence in the diff, which already existed there. Did you mean just that ?

Yes, I wrote a general comment as that's what we decided to do in the retrospective (for `'`, `\` and similar). Since it was just one case, I could have also just commented the line (but I didn't know yet while reviewing) :)

> Also regarding the docstrings, really the only new classes are the SettlementMarket and SettlementAgent, both of which inherit from the TwoSided market and agent respectively without any new functionality apart from debug logs. IMO these do not really need docstrings at all. If you insist I will add these, however it is not helpful to my eyes.

I wouldn't insist regarding this specific case, as the PR has already been thoroughly reviewed. But I would insist on it as a general process regarding PRs, because I think that docstrings are good to have, even when they seem redundant (as they are also used to generate automatic documentation); also, code without pylint warnings looks better to me 😄 

> Regarding test coverage, I'll check it out to see exactly the places where github complains, however I doubt that this is due to this PR. E.g. one occurrence I saw from a first glance is the `match_recommendations` method, which was already there in the first place (all I did was to include the settlement markets in the matching) therefore won't be able to do much. I'll give it a closer look and let you know.

This is actually a broader topic. I think that it would be nice for us to test methods even when they have not been implemented in our current PR. This is because we have a lot of existing untested code, and a sensible way to cover it is to gradually add tests for it when we modify it. Adding a line to an untested method without testing it would decrement our coverage even more.

spyrostz:
> > Regarding the \ , I found only one occurrence in the diff, which already existed there. Did you mean just that ?
> 
> Yes, I wrote a general comment as that's what we decided to do in the retrospective (for `'`, `\` and similar). Since it was just one case, I could have also just commented the line (but I didn't know yet while reviewing) :)
> 
> > Also regarding the docstrings, really the only new classes are the SettlementMarket and SettlementAgent, both of which inherit from the TwoSided market and agent respectively without any new functionality apart from debug logs. IMO these do not really need docstrings at all. If you insist I will add these, however it is not helpful to my eyes.
> 
> I wouldn't insist regarding this specific case, as the PR has already been thoroughly reviewed. But I would insist on it as a general process regarding PRs, because I think that docstrings are good to have, even when they seem redundant (as they are also used to generate automatic documentation); also, code without pylint warnings looks better to me smile
> 

Ok, since it helps you I'll add a couple of lines for each. However IMO we need to rethink this policy (I am not sure even if we have one for this). We have discussed it in the past too, but docstrings are not always good to have, even if automatic documentation is generated, because they are prone to diverge from the actual code. And especially for self-documenting classes that are almost dummy, I doubt that the comment serves more functionality other than satisfying pylint to be honest (speaking of, we can disable this in pylintrc, and I would vote in favor to do so)

> > Regarding test coverage, I'll check it out to see exactly the places where github complains, however I doubt that this is due to this PR. E.g. one occurrence I saw from a first glance is the `match_recommendations` method, which was already there in the first place (all I did was to include the settlement markets in the matching) therefore won't be able to do much. I'll give it a closer look and let you know.
> 
> This is actually a broader topic. I think that it would be nice for us to test methods even when they have not been implemented in our current PR. This is because we have a lot of existing untested code, and a sensible way to cover it is to gradually add tests for it when we modify it. Adding a line to an untested method without testing it would decrement our coverage even more.

It actually depends on the PR if you ask me. For this specific PR, where it touches almost all the tricky parts of the code (MycoInternalMatcher, AreaMarkets, EventDispatcher are some of these, that do not have dedicated tests but are tested through other classes' tests) one cannot really expect that all these classes should be unit tested in one PR, especially since there are integration tests in the PR that already cover the basic functionality.

fievelk:
> Ok, since it helps you I'll add a couple of lines for each. However IMO we need to rethink this policy (I am not sure even if we have one for this). We have discussed it in the past too, but docstrings are not always good to have, even if automatic documentation is generated, because they are prone to diverge from the actual code. And especially for self-documenting classes that are almost dummy, I doubt that the comment serves more functionality other than satisfying pylint to be honest (speaking of, we can disable this in pylintrc, and I would vote in favor to do so)

I agree with you on the fact that in some cases docstrings are not so necessary. However, I think this is the minority of the cases, and I'd rather have redundant docstrings enforced by pylint than not having important ones because we disabled the check. If a class is trivial, you can skip reading its docstring while you write new code. If a class/method is complex, you want to have a good docstring for it (and it will save you a lot of time). Of course, having confusing/outdated docstrings can be worse, but that's part of the tasks of a PR review IMO.

> It actually depends on the PR if you ask me. For this specific PR, where it touches almost all the tricky parts of the code (MycoInternalMatcher, AreaMarkets, EventDispatcher are some of these, that do not have dedicated tests but are tested through other classes' tests) one cannot really expect that this PR has to add dedicated unit tests for all these, especially since there are integration tests in the PR that already cover the basic functionality. Keep in mind that this PR is not a refactoring task, but is dealing with a task/feature that needs to be delivered in the short term.

Definitely: adding all unit-tests in one go for all these classes would not be feasible. As we discussed in the retrospective, all parts of the implementation process (including tests and conversation with stakeholders) should be considered when determining the complexity of new tasks. For settlement markets this is more complicated, and that's also why I added the comment as a "nit" that can be disregarded.

fievelk (dismissed):

src/d3a/d3a_core/util.py
fievelk: Thanks @spyrostz. Your answer got me thinking so I ask here, even though the PR has been merged. I have a doubt: when should these integration tests fail? For example, suppose I change the implementation of `make_ba_name` to be something like:
```python
def make_ba_name(owner):
    return "Some wrong string or value"
```
Should end-to-end tests fail?
I personally think they should, because one "side" is expecting to receive `f"BA {owner.name}"`, but it's receiving something else. If this is the case, then it would be good to directly use the expected string in integration tests, without calling `make_ba_name()`. Using the existing implementation is tempting, because we wouldn't have to manually write the string; but it actually will not allow us to be alerted if the implementation is incorrect.

What is your opinion on this?

PR #1223, opened 2021-09-07
D3ASIM-3564: Adding a callback method to return a simulation uuid and…

faizan2590:
… area name map

### DIFF
--- a/src/d3a/d3a_core/redis_connections/redis_communication.py
+++ b/src/d3a/d3a_core/redis_connections/redis_communication.py
@@ -24,7 +24,9 @@ from redis import StrictRedis
 from redis.exceptions import ConnectionError
 from rq import get_current_job
 from rq.exceptions import NoSuchJobError
+from typing import Dict, TYPE_CHECKING
 
+import d3a.constants
 from d3a_interface.results_validator import results_validator  # NOQA
 from d3a_interface.constants_limits import HeartBeat
 from d3a_interface.utils import RepeatingTimer
@@ -41,20 +43,26 @@ RESULTS_CHANNEL = "d3a-results"
 ZIP_RESULTS_CHANNEL = "d3a-zip-results"
 ZIP_RESULTS_KEY = "d3a-zip-results-key/"
 
+if TYPE_CHECKING:
+    from d3a.models.area import Area
+
 
 class RedisSimulationCommunication:
     def __init__(self, simulation, simulation_id, live_events):
-        if simulation_id is None:
-            return
         self._live_events = live_events
         self._simulation_id = simulation_id
         self._simulation = simulation
         self._sub_callback_dict = {
-            self._simulation_id + "/stop": self._stop_callback,
-            self._simulation_id + "/pause": self._pause_callback,
-            self._simulation_id + "/resume": self._resume_callback,
-            self._simulation_id + "/live-event": self._live_event_callback,
-            self._simulation_id + "/bulk-live-event": self._bulk_live_event_callback}
+            f"{d3a.constants.COLLABORATION_ID}/area-map/": self._area_map_callback
+        }
+        if simulation_id is not None:
+            self._sub_callback_dict.update({
+                self._simulation_id + "/stop": self._stop_callback,
+                self._simulation_id + "/pause": self._pause_callback,
+                self._simulation_id + "/resume": self._resume_callback,
+                self._simulation_id + "/live-event": self._live_event_callback,
+                self._simulation_id + "/bulk-live-event": self._bulk_live_event_callback,
+            })
         self.result_channel = RESULTS_CHANNEL
 
         try:
@@ -96,6 +104,21 @@ class RedisSimulationCommunication:
                     "error_message": f"Error when handling simulation {command_type}."
                 })
         self.publish_json(response_channel, response_json)
+
+    def _area_map_callback(self, payload: Dict) -> None:
+        """Triggers the calculation of area uuid and name map and publish it
+        back to a redis response channel"""
+        area_map = self._area_uuid_name_map_wrapper(self._simulation.area)
+        response_channel = f"external-myco/{d3a.constants.COLLABORATION_ID}/area-map/response/"
+        response_dict = {"area_mapp": area_map, "event": "area_map"}
+        self.publish_json(response_channel, response_dict)
+
+    def _area_uuid_name_map_wrapper(self, area: "Area", area_map: Dict = {}) -> Dict:
+        """Recursive method to populate area uuid and name map for area object"""
+        area_map[area.uuid] = area.name
+        for child in area.children:
+            area_map = self._area_uuid_name_map_wrapper(child, area_map)
+        return area_map
 
     def _stop_callback(self, payload):
         response = json.loads(payload["data"])


src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -97,6 +105,21 @@ def _generate_redis_response(self, response, simulation_id, is_successful, comma
                 })
         self.publish_json(response_channel, response_json)
 
+    def _area_map_callback(self, payload: Dict) -> None:
```
spyrostz, line 108: I think it can be a classmethod

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -41,20 +43,26 @@
 ZIP_RESULTS_CHANNEL = "d3a-zip-results"
 ZIP_RESULTS_KEY = "d3a-zip-results-key/"
 
+if TYPE_CHECKING:
+    from d3a.models.area import Area
+
 
 class RedisSimulationCommunication:
     def __init__(self, simulation, simulation_id, live_events):
-        if simulation_id is None:
```
spyrostz, line 47: Can you explain a bit why this is needed to be removed ? I am asking because if the simulation_id is None then there is indeed no need to instantiate all members of the constructor.

spyrostz (approved):
2 comments, once resolved / answered LGTM

src/d3a/d3a_core/redis_connections/redis_communication.py
faizan2590: If `simulation_id  is None` I am indeed not instantiating all elements of the constructor, give ~5 lines below `if simulation_id is not None:`  a look

Hassan754: I think Spyros meant that the original behavior (returning None if the simulation id is None) makes more sense
If the simulation id is None, imo we should raise an exception here

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -97,6 +105,22 @@ def _generate_redis_response(self, response, simulation_id, is_successful, comma
                 })
         self.publish_json(response_channel, response_json)
 
+    def _area_map_callback(self, payload: Dict) -> None:
+        """Triggers the calculation of area uuid and name map and publish it
```
Hassan754, line 109:
```suggestion
        """Trigger the calculation of area uuid and name mapping and publish it
```

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -97,6 +105,22 @@ def _generate_redis_response(self, response, simulation_id, is_successful, comma
...
+        back to a redis response channel"""
+        area_map = self._area_uuid_name_map_wrapper(self._simulation.area)
```
Hassan754, line 111: Nit, this could be renamed to `area_mapping` here and in the rest of occurrences 
```suggestion
        area_mapping = self._area_uuid_name_map_wrapper(self._simulation.area)
```

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -97,6 +105,22 @@ def _generate_redis_response(self, response, simulation_id, is_successful, comma
...
+        response_channel = f"external-myco/{d3a.constants.CONFIGURATION_ID}/area-map/response/"
+        response_dict = {"area_map": area_map, "event": "area_map"}
+        self.publish_json(response_channel, response_dict)
+
+    @classmethod
+    def _area_uuid_name_map_wrapper(cls, area: "Area", area_map: Dict = {}) -> Dict:
+        """Recursive method to populate area uuid and name map for area object"""
```
Hassan754, line 118: Better avoid using mutable default arguments
```suggestion
    def _area_uuid_name_map_wrapper(cls, area: "Area", area_mapping: Optional[dict] = None) -> Dict:
        """Recursive method to populate area uuid and name map for area object"""
        area_mapping = area_mapping or {}
```

Hassan754 (approved):
LGTM, it would be great if you can add rename `map` to `mapping` wherever used and also add some unit tests for the `_area_uuid_name_map_wrapper` and `_area_map_callback` in case there is a redis communication tester class

src/d3a/d3a_core/redis_connections/redis_communication.py
faizan2590: Before `RedisSimulationCommunication` was returning None for the case of local run when there is no need of `stop/pause/resume/live-event & bulk-live-event` . Now the same `RedisSimulationCommunication` is used for local run too via adding a new callback to channel `area-map` for local run. Thats why I have now skipped an early return of this class constructor.

faizan2590:
>Add some unit tests for the `_area_uuid_name_map_wrapper` and `_area_map_callback` in case there is a redis communication tester class

Unfortunately there was no redis communication tester class, considering testing output of area_mapper, its covered in [integration test under myco-matcher](https://github.com/gridsingularity/myco-api-client/blob/feature/D3ASIM-3564/integration_tests/test_redis_myco_matcher.py#L49).

src/d3a/d3a_core/redis_connections/redis_communication.py
Hassan754: How is it possible that the `simulation_id` is None and still you're getting the value of `d3a.constants.CONFIGURATION_ID`, isn't it also None in this case ?

faizan2590: `d3a.constants.CONFIGURATION_ID` is an empty string for local run.

Hassan754: IMO we should always rely on this `d3a.constants.CONFIGURATION_ID` variable then, and skip passing the `simulation_id`
And we should also remove the `if simulation_id is not None` check, because it doesn't make sense to only have an `area_mapping_callback` in case there is no simulation_id, if we're to interact with the simulation anw let's have all endpoints working (stop, pause, resume ...)

faizan2590: `simulation_id ` being part of `RedisSimulationCommunication`being there earlier, but I agree and would incline to remove it now

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -41,20 +43,26 @@
...
+    from d3a.models.area import Area
+
 
 class RedisSimulationCommunication:
-    def __init__(self, simulation, simulation_id, live_events):
-        if simulation_id is None:
-            return
+    def __init__(self, simulation, live_events):
         self._live_events = live_events
-        self._simulation_id = simulation_id
```
Hassan754, line 50: I suggest keeping this line `self._simulation_id = d3a.constants.CONFIGURATION_ID`
So you could revert the rest of the lines

Hassan754 (commented):
1 Last comment

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -41,20 +43,27 @@
...
+        self._simulation_id = d3a.constants.CONFIGURATION_ID
         self._simulation = simulation
         self._sub_callback_dict = {
-            self._simulation_id + "/stop": self._stop_callback,
-            self._simulation_id + "/pause": self._pause_callback,
-            self._simulation_id + "/resume": self._resume_callback,
-            self._simulation_id + "/live-event": self._live_event_callback,
-            self._simulation_id + "/bulk-live-event": self._bulk_live_event_callback}
+            f"{self._simulation_id}/area-map/": self._area_map_callback
```
Hassan754, line 56: Why is this allowed to be here even if there is no simulation_id there, whereas other endpoints are shut down?

faizan2590: Rest of the endpoints are shutdown for local run when `self._simulation_id` is an empty string and not needed as were designed to only facilitate `stop/pause/resume/live_events` for web-based simulation run , and not for local run. Contrary to that `area-map` is explicitly for local run. Hope its answers your confusion?

Hassan754: So where is the code responsible for pausing/stopping simulations run through cli? why isn't it the same for them?

faizan2590: On second thought I think all could be subscribed irrespective of local or web based run and irrespective of it being used or not.

Tagging @spyrostz in case he sees any performance degradation in doing so.

faizan2590: > So where is the code responsible for pausing/stopping simulations run through cli? 

[Here](https://github.com/gridsingularity/d3a/blob/master/src/d3a/d3a_core/simulation.py#L441)


> why isn't it the same for them?

It's been there even before we joined, as you can see from commit history.  Answering *why* , may be its not handy for user to publish a message to redis for local run `to stop/pause/resume` simulation rather than a simple keypress

spyrostz: The CLI code is not really relevant to the Redis channels, because as @faizan2590 mentioned Redis is not used in this flow. In general I agree with the latter observation of Faizan, that these channels are probably better to remain active. They are not used by anyone now, however does not hurt performance-wise to have them, and also might prove useful in the future in case we add such functionalities in the Redis version of the d3a-api-client

src/d3a/d3a_core/redis_connections/redis_communication.py
```diff
@@ -160,8 +185,9 @@ def _handle_redis_job_metadata(self):
             job = get_current_job()
             job.refresh()
             if job.meta.get("terminated"):
-                log.error(f"Redis job {self._simulation_id} received a stop message via the "
-                          f"job.terminated metadata by d3a-web. Stopping the simulation.")
+                log.error(f"Redis job {self._simulation_id} received a stop "
+                          f"message via the job.terminated metadata by d3a-web. "
+                          f"Stopping the simulation.")
```
Hassan754, line 190: Nit
```suggestion
                          "message via the job.terminated metadata by d3a-web. "
                          "Stopping the simulation.")
```

Hassan754 (approved):

src/d3a/d3a_core/redis_connections/redis_communication.py
spyrostz: This approach is not going to work, unless we also modify the d3a-web (which I would not agree). simulation_id is the redis job id that is different for each simulation run, even for the same configuration. configuration_id is assigned whenever a configuration/collaboration/CN is created, and is constant no matter how many times one relaunches the same configuration. Therefore the simulation_id in the Redis channels names is really needed and will lead to a bug if configuration_id is used instead (the simulation won't be able to be stopped from the UI). 

Please be careful when adapting Redis channels in the future, because there are multiple services that might rely on them and one does not have the liberty to change at will without reviewing other services !

Hassan754 (approved):
LGTM

PR #1226, opened 2021-09-14
Feature/d3 asim 3599

spyrostz:
The main root cause of the error was that there were times when offer / bid lists with `[None]` values were passed to the validator. This caused the validator to raise the exception described here https://gridsingularity.atlassian.net/browse/D3ASIM-3571 . It was fixed by correcting the if clause that controlled when to use the validator. 
The exception described in D3ASIM-3599 was also a byproduct of the aforementioned error, however in order to protect ourselves from such issues in the future I added a deepcopy to the offers and bids when read from the market. That way, even if the market mutates its offers and bids dicts in parallel, the dict returned by get_offers and get_bids will be possible to be used in loops without raising an exception.


Hassan754 (dismissed):
Looking at the ticket description and comments, it looks like adding the area uuid to bids/offers is not dealt with in this PR, is this meant to be?
Other than that, LGTM

spyrostz:
> Looking at the ticket description and comments, it looks like adding the area uuid to bids/offers is not dealt with in this PR, is this meant to be?
> Other than that, LGTM

Thanks for pointing it out, the branch name is wrong here. I have updated at least the PR name and description to depict the correct ticket (https://gridsingularity.atlassian.net/browse/D3ASIM-3599)

fievelk:
Just one comment: I would suggest adding the explanation you provided above as part of the docstring for the `get_offers()` and `get_bids()` methods. The reason why we're using `deepcopy` won't be immediately clear to our future selves (and to external contributors) so I think it would be good to state it in the docstrings.

> [...] in order to protect ourselves from such issues in the future I added a deepcopy to the offers and bids when read from the market. That way, even if the market mutates its offers and bids dicts in parallel, the dict returned by get_offers and get_bids will be possible to be used in loops without raising an exception.

spyrostz:
> Just one comment: I would suggest adding the explanation you provided above as part of the docstring for the `get_offers()` and `get_bids()` methods. The reason why we're using `deepcopy` won't be immediately clear to our future selves (and to external contributors) so I think it would be good to state it in the docstrings, to avoid reverting it.
> 
> > [...] in order to protect ourselves from such issues in the future I added a deepcopy to the offers and bids when read from the market. That way, even if the market mutates its offers and bids dicts in parallel, the dict returned by get_offers and get_bids will be possible to be used in loops without raising an exception.

Agreed, just added a docstring to fix it

fievelk (approved):

PR #1229, opened 2021-09-25
D3ASIM-3569: Report validation status of myco external matcher recommendations

Hassan754:


src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -162,45 +177,22 @@ def _get_bids_offers(market: TwoSidedMarket, filters: Dict):
...
-                continue
-            try:
-                market.validate_bid_offer_match(
-                    market_bids,
-                    market_offers,
-                    recommendation.get("trade_rate"),
-                    recommendation.get("selected_energy")
-                )
-            except InvalidBidOfferPairException:
-                continue
```
Hassan754, line 201: These were removed because they are duplicates available in the match_recommendations function of the Two Sided market
These are non blockers, they do not stop the matching process, they just omit the recommendation

src/d3a/models/market/two_sided.py
```diff
@@ -349,15 +349,20 @@ def validate_bid_offer_match(
...
-                        for offer in offers))):
-            raise InvalidBidOfferPairException
+        if selected_energy > bids_total_energy:
+            raise InvalidBidOfferPairException(
+                f"Energy traded {selected_energy} is higher than bids energy {bids_total_energy}.")
+        if selected_energy > offers_total_energy:
+            raise InvalidBidOfferPairException(
+                f"Energy traded {selected_energy} is higher than offers energy"
+                f" {offers_total_energy}.")
+        if any((bid.energy_rate + FLOATING_POINT_TOLERANCE) < clearing_rate for bid in bids):
```
spyrostz, line 359: If we really want the requirements matching to act as pay as bid, it should be this way IMO:

```suggestion
        if any(not isclose(bid.energy_rate + FLOATING_POINT_TOLERANCE, clearing_rate) for bid in bids):
```

spyrostz (commented):
One important clarification pending, but LGTM once this is fixed or discussed.

src/d3a/models/market/two_sided.py
Hassan754: In my understanding the match can be valid if the clearing rate is less than or equal to the bid's rate and greater than or equal to the offer's rate

Hassan754: @candrews42

spyrostz: Since this is a generic validator, and not trying to match the 2-sided PAB market, then I agree, you can leave it as is.

andreabertolini1995 (dismissed):
LGTM

andreabertolini1995 (approved):
LGTM

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -162,45 +171,69 @@ def _get_bids_offers(market: TwoSidedMarket, filters: Dict):
...
+            except Exception as exception:
+                if isinstance(exception, cls.BLOCKING_EXCEPTIONS):
+                    response["status"] = "fail"
+                    response[
+                        "message"] = f"Validation Error, matching will be skipped: {exception}"
+                    break
+                else:
+                    response["recommendations"].append(
+                        {**recommendation, "status": "fail", "message": str(exception)})
+                    continue
```
hannesdiedrich, line 238: Nice! Perfect example for motivation of not using simple `Exception`s, right @fievelk ? This is what you meant in Lisbon, right?

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -87,28 +86,38 @@ def publish_offers_bids(self, message):
         channel = f"{self._channel_prefix}/offers-bids/response/"
         self.myco_ext_conn.publish_json(channel, response_data)
 
-    def match_recommendations(self, message: Optional[str] = None) -> None:
+    def match_recommendations(self, message: Optional[dict] = None) -> None:
```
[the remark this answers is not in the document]
Hassan754: Because it can also be None. so I marked it as Optional

src/d3a/models/myco_matcher/myco_external_matcher.py
Hassan754: Update, I have removed the broad `except Exception` to not report **any** exception to the client

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -162,45 +171,69 @@ def _get_bids_offers(market: TwoSidedMarket, filters: Dict):
...
-            recommendation["bids"] = [bid.serializable_dict() for bid in market_bids]
-            recommendation["offers"] = [offer.serializable_dict() for offer in market_offers]
-            validated_recommendations.append(recommendation)
-        return validated_recommendations
+            except Exception as exception:
+                if isinstance(exception, cls.BLOCKING_EXCEPTIONS):
+                    response["status"] = "fail"
+                    response[
+                        "message"] = f"Validation Error, matching will be skipped: {exception}"
+                    break
```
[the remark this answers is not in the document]
Hassan754: As agreed with Spyros and Colin, these blocking exceptions should stop the whole matching process, so no need to further loop over the rest of the recommendations

src/d3a/models/myco_matcher/myco_external_matcher.py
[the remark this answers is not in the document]
fievelk: @hannesdiedrich if you're interested, you can find more information here: https://docs.python.org/3/library/typing.html#typing.Optional

src/d3a/models/myco_matcher/myco_external_matcher.py
fievelk: @hannesdiedrich this can be one of such cases where you want to have a distinction between different types of exceptions, yes!

src/d3a/models/myco_matcher/myco_external_matcher.py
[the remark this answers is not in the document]
hannesdiedrich: Thanks! had a look and am convinced. Can you use the `Dict` from `typing` instead of the build in `dict` though? Thanks!

src/d3a/models/myco_matcher/myco_external_matcher.py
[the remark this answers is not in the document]
hannesdiedrich: Interesting, just for me to understand (not in order to start a long discussion again, which I missed):
If I the myco matcher sends 10 recommendations and the second fails, the first one will be matched but not the other (probably valid) recommendations 3-10 ?

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -162,45 +171,69 @@ def _get_bids_offers(market: TwoSidedMarket, filters: Dict):
...
                 )
-            except InvalidBidOfferPairException:
-                continue
-
-            recommendation["bids"] = [bid.serializable_dict() for bid in market_bids]
-            recommendation["offers"] = [offer.serializable_dict() for offer in market_offers]
-            validated_recommendations.append(recommendation)
-        return validated_recommendations
+            except (MycoValidationException, InvalidBidOfferPairException) as exception:
+                if isinstance(exception, cls.BLOCKING_EXCEPTIONS):
```
hannesdiedrich, line 230: Can you maybe add an inline comment that explains why you handle the `MycoValidationException` differently than the `InvalidBidOfferPairException`  here?

src/d3a/models/myco_matcher/myco_external_matcher.py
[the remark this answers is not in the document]
Hassan754: I'm validating all recommendations first and then if no blocker exception is raised I start matching the validated records, take a look at the `match_recommendations` method:

```python
response_data.update(
            MycoExternalMatcherValidator.validate_and_report(self, recommendations)) # Here I'm sending all recommendations to validation
        if response_data["status"] != "success": # if this fail, there is a blocking exception raised
            self.myco_ext_conn.publish_json(channel, response_data)
            return
        for recommendation in response_data["recommendations"]:
            try:
                if recommendation["status"] != "success":
                    # skip the failed recommendations (non blocking ones)
                    continue
                market = self.markets_mapping.get(recommendation["market_id"])
                market.match_recommendations([recommendation]) # Here I'm doing the matching

```

src/d3a/models/myco_matcher/myco_external_matcher.py
Hassan754: I'm not dealing with `MycoValidationException` other than `InvalidBidOfferPairException`
I'm only dealing with the `BLOCKING_EXCEPTIONS` differently, so if `InvalidBidOfferPairException` is added to the `BLOCKING_EXCEPTIONS` it will be dealt with in the same way

src/d3a/models/myco_matcher/myco_external_matcher.py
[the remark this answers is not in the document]
hannesdiedrich: As far as my github diff tells, this has not been addressed yet.

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -162,45 +172,69 @@ def _get_bids_offers(market: TwoSidedMarket, filters: Dict):
...
+            # If not all offers bids exist in the market, skip the current recommendation
+            raise InvalidBidOfferPairException(
+                "Not all bids and offers exist in the market.")
+
+    @classmethod
+    def _validate(cls, matcher: MycoExternalMatcher, recommendation: Dict):
+        """Call corresponding validation methods."""
+        cls.validate_valid_dict(matcher, recommendation)
+        cls.validate_market_exists(matcher, recommendation)
+        cls.validate_orders_exist_in_market(matcher, recommendation)
```
faizan2590, line 217: `validate_valid_dict `, `validate_market_exists ` & `validate_orders_exist_in_market ` can be changed to private method as called within the same class.

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -162,45 +172,69 @@ def _get_bids_offers(market: TwoSidedMarket, filters: Dict):
...
+
+    @classmethod
+    def validate_and_report(cls, matcher: MycoExternalMatcher, recommendations: Dict) -> Dict:
```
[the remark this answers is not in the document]
Hassan754: I see your concern, This can be done indeed, however I see that these two classes are tightly coupled (This validator class will only work with the External Matcher)
So I propose either keeping them both in the same module, or creating a package (I also think that this should be avoided now but can be done if we planned to expand this external matcher functionality to add more classes)
In any case, separating them into 2 modules may cause some circular imports issues (The matcher imports the validator and the validator imports the matcher for type hints, this can be silenced though)
So it's your call, thanks!

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -87,28 +86,38 @@ def publish_offers_bids(self, message):
...
+        if response_data["status"] != "success":
+            self.myco_ext_conn.publish_json(channel, response_data)
+            return
+        for recommendation in response_data["recommendations"]:
+            try:
+                if recommendation["status"] != "success":
+                    continue
                 market = self.markets_mapping.get(recommendation["market_id"])
+                recommendation.pop("status")
                 market.match_recommendations([recommendation])
```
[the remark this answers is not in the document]
Hassan754: Ah okay, I couldn't do that because each recommendation can be targeted to a different market
So you can see the line `market = self.markets_mapping.get(recommendation["market_id"])`, I'm getting the market of each recommendation and posting the recommendation in this market
consider this example:
```python
recommendations = [{"market_id": "market1" ...}, {"market_id": "market2" ...}]
for recommendation in recommendations:
    market = self.markets_mapping.get(recommendation["market_id"]) # will return market1 in first iteration and market2 in the second
    market.match_recommendations([recommendation])
```

PR #1230, opened 2021-09-27
Adding callback for DSO to get energy deviances.

faizan2590:


src/d3a/models/area/__init__.py
```diff
@@ -580,6 +580,27 @@ def past_settlement_markets(self) -> Dict:
     def get_settlement_market(self, timeslot):
         return self._markets.settlement_markets.get(timeslot)
 
+    def _accumulate_energy_deviances_of_children(self, area, total_energy_deviance_kWh):
+        for child in area.children:
+            print(f"child: {child.name} | STRATEGY: {child.strategy}")
```
hannesdiedrich, line 585: Please remove

hannesdiedrich (commented):
Left an important question.
As always: Please add docstrings and tests.

src/d3a/models/area/__init__.py
```diff
@@ -580,6 +580,27 @@ def past_settlement_markets(self) -> Dict:
...
+                last_past_settlement_market = (
+                    self.last_past_settlement_market[0]
+                    if self.last_past_settlement_market is not None
+                    else None)
+                if last_past_settlement_market is not None:
+                    total_energy_deviance_kWh += (
+                        child.strategy.state.get_forecast_measurement_deviation_kWh(
+                            self.last_past_settlement_market[0]))
+            if child.children:
+                self._accumulate_energy_deviances_of_children(child, total_energy_deviance_kWh)
```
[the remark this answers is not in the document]
faizan2590: IMO, the DSO should be aware of expected deviation of energy in the grid. Let's say the DSO's point of reference is at Substation node. He may not be able to know  expected deviation just by looking only at its children(directly connected areas) where mostly Houses are the one directly connected to a substation but the actual deviation is caused by devices inside those houses.

src/d3a/models/area/__init__.py
```diff
@@ -143,6 +145,7 @@ def __init__(self, name: str = None, children: List["Area"] = None,
             if external_connection_available and self.strategy is None else None
         self.should_update_child_strategies = False
         self.external_connection_available = external_connection_available
+        self.total_energy_deviance_kWh: Dict[DateTime, float] = {}
```
spyrostz, line 148: I think that a much much better place for all these changes (total_energy_deviance member + the method `_calculate_energy_deviances` ) is the AreaStats class. That way the Area class will not bloat more than it already has. You can also handle the updates in the energy deviance without changing the Area class, by adding it in the `AreaStats.update_area_market_stats` method (it is already called every market cycle in order to update other statistics).

spyrostz (commented):
As a general comment, a unit test for the `total_energy_deviance_kWh` would be great. 

Besides that, the functionality LGTM and ready to be merged once the comments are addressed.

src/d3a/models/area/__init__.py
faizan2590: > I think that a much much better place for all these changes (total_energy_deviance member + the method _calculate_energy_deviances ) is the AreaStats class.

Done, https://github.com/gridsingularity/d3a/pull/1230/commits/9a79f2871e5898b88bfb21096e0e1f7695b8d11a

> You can also handle the updates in the energy deviance without changing the Area class, by adding it in the AreaStats.update_area_market_stats method (it is already called every market cycle in order to update other statistics).

Unfortunately, it need to be called for both cases when area has children or itself is a device. For later case due to [early return](https://github.com/gridsingularity/d3a/blob/master/src/d3a/models/area/__init__.py#L332) for cases when area is a device, and for other non-device area where market creation is done from top-to-bottom where concatinating into `update_area_market_stats` left me child with no deviance calculation due to market creation from top to bottom, so I also have to `calculate_energy_deviances ` before the end of `cycle_market`

src/d3a/models/area/stats.py
```diff
@@ -157,3 +167,29 @@ def _aggregate_exported_imported_energy(self):
             self.imported_traded_energy_kwh[self.current_market.time_slot] = 0.
         if self.current_market.time_slot not in self.exported_traded_energy_kwh:
             self.exported_traded_energy_kwh[self.current_market.time_slot] = 0.
+
+    def calculate_energy_deviances(self):
+        """
+        If area is a device - Get its forecated deviance
+        Else accumulate energy deviances of connected children
+        """
```
spyrostz, line 175: Nit, but would be great to convert this to the Google docstring format that we have decided to conform to.

PR #1231, opened 2021-09-28
D3ASIM-3596: Update myco external client on the market progression

Hassan754:


src/d3a/d3a_core/simulation.py
```diff
@@ -371,12 +373,11 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
 
                 self.area.tick_and_dispatch()
                 self.area.update_area_current_tick()
-                if self.simulation_config.external_connection_enabled:
```
spyrostz, line 374: The if statement is missing now, are you sure it is still not needed ? Reason I am asking is because the if statement is still there for event_market_cycle

spyrostz (commented):
One comment, will LGTM once resolved / discussed

src/d3a/d3a_core/simulation.py
Hassan754: Thanks for spotting this!
No need for the if clause in both calls as the bid_offer_matcher has the exact interface for both internal and external matcher
The if in the market_cycle event was left due to the merge conflict

spyrostz (approved):
LGTM

faizan2590 (approved):
LGTM

PR #1234, opened 2021-09-29
D3ASIM-3421: Conditionally select the appropriate psycopg2 package ac…

spyrostz:
…cording to whether Pypy or cPython is the current interpreter.

Default Postgres driver implementation in Python (psycopg2) cannot be installed via pip on cPython (to be more precise, cannot be compiled, since it is a C library with Python wrappers). In order for the Postgres driver to be installed in Pypy, psycopg2cffi package needs to be installed (which is the same C driver using cffi bindings). 

The separation on the installation path needs to be done in setup.py in order to cover all cases (docker version is using Pypy, locally cPython is used but for debugging purposes Pypy should also be supported)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,3 +1,5 @@
+import platform
+
 from setuptools import find_packages, setup
 
 d3a_interface_branch = "master"
@@ -15,6 +17,11 @@ except OSError:
 
 with open("README.rst", "r") as readme:
     README = readme.read()
+
+if platform.python_implementation() == "PyPy":
+    REQUIREMENTS.append("psycopg2cffi==2.9.0")
+else:
+    REQUIREMENTS.append("psycopg2==2.9.1")
 
 # *IMPORTANT*: Don't manually change the version here. Use the 'bumpversion' utility.
 VERSION = "1.1.0"


Hassan754 (approved):
LGTM

faizan2590 (approved):

fievelk:
@spyrostz this has been approved quickly while I was writing :)

I think there is a specific requirements format that allows defining a package depending on the platform. The relevant PEP is [PEP 508](https://www.python.org/dev/peps/pep-0508)

From a first read, I think that something like this might work:
```python
psycopg2cffi==2.9.0; platform_python_implementation == 'PyPy'
psycopg2==2.9.1; platform_python_implementation == 'CPython'
```

spyrostz:
> @spyrostz this has been approved quickly while I was writing :)
> 
> I think there is a specific requirements format that allows defining a package depending on the platform. The relevant PEP is [PEP 508](https://www.python.org/dev/peps/pep-0508).
> 
> From a first read, I think that something like this in `base.txt` might work:
> 
> ```python
> psycopg2cffi==2.9.0; platform_python_implementation == 'PyPy'
> psycopg2==2.9.1; platform_python_implementation == 'CPython'
> ```

It is unfortunately more complex than that. The platform_python_implementation field, if added in the files base.in, dev.in ... will not be evaluated in package installation time, but during pip-compile time. This has the effect that the generated files will include the package according to the interpreter that pip-compile uses. This is not the case for us though, since we normally use cPython (therefore if I run `fab compile`, only psycopg2 will be included in the generated requirement files) but we use the same generated files both for cPython and Pypy. That's the reason why we are obliged to do this distinction to a file that is evaluated during installation time (setup.py).

fievelk:
> @fievelk It is unfortunately more complex than that. The platform_python_implementation field, if added in the files base.in, dev.in ... will not be evaluated in package installation time, but during pip-compile time. This has the effect that the generated files will include the package according to the interpreter that pip-compile uses. This is not the case for us though, since we normally use cPython (therefore if I run `fab compile`, only psycopg2 will be included in the generated requirement files) but we use the same generated files both for cPython and Pypy. That's the reason why we are obliged to do this distinction to a file that is evaluated during installation time (setup.py).

@spyrostz Hmm that is trickier, yes. It is possible to use the "environment markers" defined in PEP 508 in the `base.in` file. However, pep-tools docs warn about this:

> As the resulting requirements.txt can differ for each environment, users must execute pip-compile on each Python environment separately to generate a requirements.txt valid for each said environment. The same requirements.in can be used as the source file for all environments, using PEP 508 environment markers as needed, the same way it would be done for regular pip cross-environment usage.
>
> If the generated requirements.txt remains exactly the same for all Python environments, then it can be used across Python environments safely. But users should be careful as any package update can introduce environment-dependant dependencies, making any newly generated requirements.txt environment-dependant too. As a general rule, it's advised that users should still always execute pip-compile on each targeted Python environment to avoid issues.

A possible fix would be to run `pip-compile` inside the pypy dockerfile, but we would lose some control over the content of requirement files (and it might become more difficult to debug issues).

spyrostz:
> A possible fix would be to run `pip-compile` inside the pypy dockerfile, but we would lose some control over the content of requirement files (and it might become more difficult to debug issues).

I think it is even worse, because it will always recompile and update the package versions on every deploy (essentially every time the container is built) and not whenever we want to. 

There is also the option of creating a separate requirements file only for docker / Pypy and use this only for the dockerfile. This option has the disadvantage that it complicates the requirements file even more than it currently is (currently we have base, dev .... after that we would have base, cpython_db, pypy_db, dev ... ) that's why I would not recommend to go this way. 

Frankly, handling the dependencies conditionally in the setup.py still seems the best option in my eyes.

fievelk:
> > A possible fix would be to run `pip-compile` inside the pypy dockerfile, but we would lose some control over the content of requirement files (and it might become more difficult to debug issues).
> 
> I think it is even worse, because it will always recompile and update the package versions on every deploy (essentially every time the container is built) and not whenever we want to.
> 
> There is also the option of creating a separate requirements file only for docker / Pypy and use this only for the dockerfile. This option has the disadvantage that it complicates the requirements file even more than it currently is (currently we have base, dev .... after that we would have base, cpython_db, pypy_db, dev ... ) that's why I would not recommend to go this way.
> 
> Frankly, handling the dependencies conditionally in the setup.py still seems the best option in my eyes.

I agree with you, it looks like this solution is the best one right now. It might change if we'll have more custom dependencies in the future, but no need to think about it now.

PR #1237, opened 2021-10-04
D3ASIM-3593: Add time_slot in the matching recommendations and myco Published data

Hassan754:


src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -60,7 +62,7 @@ def publish_offers_bids(self, message):
         """Publish open offers and bids.
 
         Published data are of the following format:
-            {"bids_offers": {"market_id" : {"bids": [], "offers": [] }, filters: {}}}
+        {"bids_offers": {`market_id` : {`time_slot`: {"bids": [], "offers": [] }, filters: {}}}}
```
hannesdiedrich, line 65:
```suggestion
        {"bids_offers": {'market_id' : {'time_slot': {"bids": [], "offers": [] }, filters: {}}}}
```

tests/test_two_sided_market.py
```diff
@@ -437,7 +450,8 @@ def test_match_recommendations(self, market):
         recommendations = [
             BidOfferMatch(
                 bids=[bid.serializable_dict()], offers=[offer.serializable_dict()],
-                trade_rate=2, selected_energy=1, market_id=market.id).serializable_dict()
+                trade_rate=2, selected_energy=1, market_id=market.id,
+                time_slot="time_slot").serializable_dict()
```
hannesdiedrich, line 454: Nit: For my taste it would be good to have a near-real timestamp here for the tests in order to show the code reader how a "normal" time stamp looks like:
```suggestion
                time_slot="20211006T12:00").serializable_dict()
```

hannesdiedrich (commented):
Some comments, only one major. Will have another look once addressed.

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -74,10 +76,11 @@ def publish_offers_bids(self, message):
                 continue
             for market in area_data["markets"]:
                 # Cache the market (needed while matching)
-                self.markets_mapping[market.id] = market
+                self.area_markets_mapping[f"{area_uuid}-{market.time_slot_str}"] = market
```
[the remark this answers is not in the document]
hannesdiedrich: I know, but I thought that we could at least internally stick with the market_id. But I get now, that it will be complicated when matching the recommendations later.

PR #1239, opened 2021-10-05
Stripped out timezone information from the datapoints read from the D…

spyrostz:
…B, in order to sync with the naive timezones in d3a. Enabled constant to read the profiles from the database. Added check for whether the profiles need to be read or rotated for the predefined load strategy.

### DIFF
--- a/src/d3a/d3a_core/user_profile_handler.py
+++ b/src/d3a/d3a_core/user_profile_handler.py
@@ -20,13 +20,11 @@ import uuid
 from datetime import datetime, timezone
 from typing import Dict
 
-import pendulum
 import pytz
 from d3a_interface.constants_limits import GlobalConfig, TIME_ZONE
 from d3a_interface.read_user_profile import read_arbitrary_profile, InputProfileTypes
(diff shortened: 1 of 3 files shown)


src/d3a/d3a_core/user_profile_handler.py
```diff
@@ -16,13 +16,12 @@
...
 """
 import os
-from datetime import datetime
-from typing import Dict
 import uuid
+from datetime import datetime, timezone
+from typing import Dict
 
-import d3a.constants
+import pendulum
```
hannesdiedrich, line 23: minor, and please leave it as it is a crucial PR:
Please homogenise the imports (there are already occurrences with `from pendulum import ...`

hannesdiedrich (approved):
LGTM

fievelk (approved):

src/d3a/d3a_core/user_profile_handler.py
spyrostz: no problem I added it now. It is more or less one more CI run after all

PR #1241, opened 2021-10-06
Bug/d3 asim 3624

spyrostz:


src/d3a/d3a_core/user_profile_handler.py
```diff
@@ -93,7 +93,8 @@ def get_first_week_from_profile(self, profile_uuid, current_timestamp) -> dict:
                  values as dict values.
 
         """
-        profile_uuid = uuid.UUID(profile_uuid)
+        if not isinstance(profile_uuid, uuid.UUID):
+            profile_uuid = uuid.UUID(profile_uuid)
```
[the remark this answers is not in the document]
spyrostz: Actually pony expects uuids as a type when trying to do a select statement with a uuid column. str does not work unfortunately. Therefore we are obliged to convert the string to uuid before we perform the query.

src/d3a/d3a_core/user_profile_handler.py
```diff
@@ -161,9 +162,16 @@ def _buffer_all_profiles(self, current_timestamp: DateTime):
...
+            self._user_profiles[profile_uuid] = {
+                self.strip_timezone_and_create_pendulum_instance_from_datetime(
+                    data_point.time): data_point.value
+                for data_point in query_ret_val if data_point.profile_uuid == profile_uuid
+            }
+
+        for profile_uuid in self._user_profiles:
+            if not self._user_profiles[profile_uuid]:
+                self._user_profiles[profile_uuid] = self.get_first_week_from_profile(
+                    profile_uuid, current_timestamp)
```
hannesdiedrich, line 174: So I get that this is for the case when the time range (start_date, end_date) is not included in the DB, right? Shall we save the effort performing this step in the `_read_new_datapoints_from_buffer_or_rotate_profile` or shall we keep it as a safety net (because it is not that expensive) . Just double checking your rationale. Thanks

src/d3a/models/strategy/predefined_load.py
```diff
@@ -81,20 +81,25 @@ def __init__(self, daily_load_profile=None,
                          use_market_maker_rate=use_market_maker_rate)
 
         self.profile_uuid = daily_load_profile_uuid
-        self.load_profile = None
+        self._load_profile_W = None
+        self._load_profile_kWh = {}
 
-        if should_read_profile_from_db(daily_load_profile):
+        if should_read_profile_from_db(daily_load_profile_uuid):
```
hannesdiedrich, line 87: Thanks a lot for catching this!

src/d3a/d3a_core/user_profile_handler.py
spyrostz: Exactly, and if you do not mind I would really like to leave this as an extra safety. At the current state of the code the second database call in read_new_datapoints_from_buffer_or_update_profile was not called during my tests, but it does not hurt to have and can save us from hours of debugging

src/d3a/d3a_core/user_profile_handler.py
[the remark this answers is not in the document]
hannesdiedrich: Weird that this was not appearing earlier.

hannesdiedrich (approved):
LGTM

PR #1251, opened 2021-10-12
Bug/D3ASIM-3601: Move external recommendations clearing call to the internal code

Hassan754:


src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -50,13 +52,14 @@ def __init__(self):
 
         # Dict[area_id-time_slot_str: market] mapping
         self.area_markets_mapping: Dict[str, TwoSidedMarket] = {}
+        self.recommendations = []
```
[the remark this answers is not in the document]
fievelk:
```suggestion
        self._recommendations = []
```

src/d3a/models/myco_matcher/myco_external_matcher.py
```diff
@@ -142,10 +147,12 @@ def publish_simulation_id(self, message):
...
         """Publish the tick event to the Myco client."""
-        is_it_time_for_external_tick = kwargs.pop("is_it_time_for_external_tick", True)
-        # If External matching is enabled, limit the number of ticks dispatched.
-        if not is_it_time_for_external_tick:
+        if (kwargs.get("current_tick_in_slot")
+                and not global_objects.external_global_stats.is_it_time_for_external_tick(
+                kwargs.pop("current_tick_in_slot"))):
+            # If External matching is enabled, limit the number of ticks dispatched.
             return
+        sleep(1)
```
fievelk, line 155: What is this `sleep` used for?

Hassan754: This is part of the temporary fix/workaround
In the case of the external matching, the `match_recommendations` will be called after allowing myco to send its recommendations
so this slight and adjustable sleep is waiting for the external script to submit recommendations in order to match them in the same tick
Tested using the original bug ticket files

src/d3a/models/area/__init__.py
```diff
@@ -411,10 +411,19 @@ def tick(self):
                 self.dispatcher.publish_market_clearing()
             else:
                 self._update_myco_matcher()
+                bid_offer_matcher.event_tick(
+                    **self._progress_info, current_tick_in_slot=self.current_tick_in_slot)
```
spyrostz, line 415: Does it really need to be moved here ? The way it is at the moment, the event_tick method will be called on every tick for every market area in the simulation. The MycoMatcher implementation though assumes that it is called once per tick for all markets AFAIR. I think that leaving this part of the code in simulation.py as it originally was would be fine. 

Similar comment exists for the sleep method, this should also not be required.

src/d3a/models/myco_matcher/myco_external_matcher.py
Hassan754: @spyrostz It's kind of working, however there are several orders that become invalid at the end of the tick (before clearing in the next tick), even if we do not allow d3a-api-client to replace orders before matching, the update frequency is replacing them anyway

PR #1252, opened 2021-10-13
 Start the external connection communicator if there are defined patterns

Hassan754:
Kudos to @hannesdiedrich
When we enable the global external connection, we subscribe to several patterns (mostly aggregator patterns) but we do not subscribe to any channels
This was restricting the external connection from being listening to external requests (Like create aggregator)
This was hidden because when we enable the external connection to any Area, we subscribe to some extra channels, and once we have channels the `start_communication` function was working properly


hannesdiedrich (dismissed):
LGTM

spyrostz (approved):
LGTM

src/d3a/d3a_core/redis_connections/redis_area_market_communicator.py
```diff
@@ -116,7 +121,7 @@ def sub_to_multiple_channels(self, channel_callback_dict):
     def start_communication(self):
         if not self.is_enabled:
             return
-        if not self.pubsub.channels:
+        if not self.is_subscribed_to_channels_patterns:
```
fievelk, line 124: Redis pubsub already provides the same functionality: https://github.com/redis/redis-py/blob/5cd878de3eca3322d6719822d4999595f943b218/redis/client.py#L1205-L1208

I'd suggest removing the `is_subscribed_to_channels_patterns` property and use the one provided by the library.

spyrostz: Thanks! Did not know that pubsub already provided this, awesome

tests/test_d3a_core/test_redis_area_market_communicator.py
```diff
@@ -0,0 +1,83 @@
...
+
+        communicator = ExternalConnectionCommunicator(is_enabled=True)
+        assert communicator.is_enabled is True
+        assert communicator.aggregator is not None
+        assert hasattr(communicator, "redis_db") is True
+        assert hasattr(communicator, "channel_callback_dict") is True
+
+    def test_sub_to_channel(self):
+        def callback():
+            return
```
fievelk, line 47: Nit: no need to define a function here. `callback` can be anything, as long as it's passed to `pubsub.subscribe`. I would personally probably use a mock object for example:
```suggestion
        callback = Mock(name="some-callback-function")
```

tests/test_d3a_core/test_redis_area_market_communicator.py
```diff
@@ -0,0 +1,83 @@
...
+        communicator = ExternalConnectionCommunicator(is_enabled=True)
+        communicator.sub_to_multiple_channels({"channel": callback})
+        communicator.pubsub.subscribe.assert_called_once_with(**{"channel": callback})
+
+    def test_start_communication(self):
+        communicator = ExternalConnectionCommunicator(is_enabled=True)
+        communicator.pubsub.subscribed = False
+        communicator.start_communication()
+        assert communicator.pubsub.run_in_thread.called is False
```
fievelk, line 63: The `mock` library now supports `assert_not_called()`:
```suggestion
        communicator.pubsub.run_in_thread.assert_not_called()
```

tests/test_d3a_core/test_redis_area_market_communicator.py
```diff
@@ -0,0 +1,83 @@
...
+        communicator = ExternalConnectionCommunicator(is_enabled=True)
+        assert communicator.is_enabled is True
+        assert communicator.aggregator is not None
+        assert hasattr(communicator, "redis_db") is True
+        assert hasattr(communicator, "channel_callback_dict") is True
+
+    def test_sub_to_channel(self):
+        def callback():
+            return
+        communicator = ExternalConnectionCommunicator(is_enabled=True)
```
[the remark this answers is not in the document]
fievelk: You can actually set `is_enabled` dynamically by using fixtures + indirect parametrization: https://stackoverflow.com/a/33879151/1916588

tests/test_d3a_core/test_redis_area_market_communicator.py
```diff
@@ -0,0 +1,83 @@
...
+
+    def test_start_communication(self):
+        communicator = ExternalConnectionCommunicator(is_enabled=True)
+        communicator.pubsub.subscribed = False
+        communicator.start_communication()
+        assert communicator.pubsub.run_in_thread.called is False
+
+        communicator.pubsub.subscribed = True
+        communicator.start_communication()
+        assert communicator.pubsub.run_in_thread.called is True
```
fievelk, line 67:
```suggestion
        communicator.pubsub.run_in_thread.assert_called_once()
```

Hassan754:
> One more general comment: please add docstrings in tests!

I recall that we agreed on not being pushy about writing docstrings, I find this most relevant in tests
Each individual unit test in the module has the name test_<name of method it's testing>, so I find it it pretty straightforward taking into consideration the `basic` assertions executed inside the tests

PR #1257, opened 2021-10-19
Feature/d3 asim 3631 New future markets handling

hannesdiedrich:


src/d3a/models/market/one_sided.py
```diff
@@ -101,7 +101,8 @@ def get_offers(self) -> Dict:
     def offer(self, price: float, energy: float, seller: str, seller_origin,
               offer_id=None, original_price=None, dispatch_event=True,
               adapt_price_with_fees=True, add_to_history=True, seller_origin_id=None,
-              seller_id=None, attributes: Dict = None, requirements: List[Dict] = None) -> Offer:
+              seller_id=None, attributes: Dict = None, requirements: List[Dict] = None,
+              time_slot: DateTime = None) -> Offer:
```
Hassan754, line 105: Nit, the same for the other arguments here and in other changed functions
```suggestion
              time_slot: Optional[DateTime] = None) -> Offer:
```

src/d3a/models/market/one_sided.py
```diff
@@ -117,9 +118,11 @@ def offer(self, price: float, energy: float, seller: str, seller_origin,
 
         if offer_id is None:
             offer_id = self.bc_interface.create_new_offer(energy, price, seller)
-        offer = Offer(offer_id, self.now, price, energy, seller, original_price,
+        now = self.now if self.time_slot else time_slot
```
Hassan754, line 121: I have the feeling that we actually want the opposite here, right?
If this is true, this is relevant to the other usages below
```suggestion
        now = time_slot if time_slot else self.now
```

Another nit, it would be great if we rename `now` to `time`

src/d3a/models/market/one_sided.py
```diff
@@ -168,7 +171,9 @@ def _calculate_original_prices(cls, offer):
 
     def split_offer(self, original_offer, energy, orig_offer_price):
 
+        time_slot = original_offer.time_slot
```
Hassan754, line 174: Nit, no need to declare a variable here, we can use `original_offer.time_slot` while passing the time_slot below

The same applies to the other assignment in the TwoSided market

src/d3a/models/market/two_sided.py
```diff
@@ -20,18 +20,19 @@
...
 from d3a.d3a_core.exceptions import (BidNotFoundException, InvalidBid,
                                      InvalidBidOfferPairException, InvalidTrade, MarketException)
 from d3a.d3a_core.util import short_offer_bid_log_str
 from d3a.events.event_structures import MarketEvent
 from d3a.models.market import lock_market_action
 from d3a.models.market.one_sided import OneSidedMarket
+from d3a_interface.constants_limits import ConstSettings
+from d3a_interface.data_classes import Bid, Offer, Trade, TradeBidOfferInfo, BidOfferMatch
+from d3a_interface.matching_algorithms.requirements_validators import RequirementsSatisfiedChecker
+from pendulum import DateTime
```
Hassan754, line 35: These should be above the d3a's imports, they were good before

src/d3a/models/myco_matcher/myco_internal_matcher.py
```diff
@@ -73,8 +73,29 @@ def match_recommendations(self, **kwargs):
...
                         market.match_recommendations(bid_offer_pairs)
+
+        self._match_recommendations_future_markets()
+
         self.area_uuid_markets_mapping = {}
 
+    def _match_recommendations_future_markets(self):
+        """Loop over all future markets and match bids and offers."""
+        for area_data in self.area_uuid_markets_mapping.values():
+            future_markets = area_data["future_markets"]
```
Hassan754, line 84: Is this supposed to be a single or multiple markets ? 
The name gives the impression that we have multiple markets despite that it looks like we're dealing with it as a single object

src/d3a/models/market/future.py
```diff
@@ -0,0 +1,217 @@
...
+    def __init__(self, bc=None, notification_listener=None, readonly=False,
+                 grid_fee_type=ConstSettings.IAASettings.GRID_FEE_TYPE,
+                 grid_fees=None, name=None):
+        super().__init__(time_slot=None, bc=bc, notification_listener=notification_listener,
+                         readonly=readonly, grid_fee_type=grid_fee_type,
+                         grid_fees=grid_fees, name=name, in_sim_duration=True)
+
+        self.slot_bid_mapping = {}  # type: Dict[DateTime, List[Bid]]
+        self.slot_offer_mapping = {}  # type: Dict[DateTime, List[Offer]]
+        self.slot_trade_mapping = {}  # type: Dict[DateTime, List[Trade]]
```
Hassan754, line 51:
```suggestion
        self.slot_bids_mapping: Dict[DateTime, List[Bid]] = {}
        self.slot_offers_mapping: Dict[DateTime, List[Offer]] = {}
        self.slot_trades_mapping: Dict[DateTime, List[Trade]] = {}
```

src/d3a/models/market/future.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+class FutureMarketException(Exception):
+    """Exception specific to the Future markets."""
+
+
+class FutureMarkets(TwoSidedMarket):
+    """Class responsible for future markets."""
+
+    def __init__(self, bc=None, notification_listener=None, readonly=False,
```
Hassan754, line 42: It would be great if you can add type hints here

src/d3a/models/market/future.py
```diff
@@ -0,0 +1,217 @@
...
+from pendulum import DateTime, duration
+
+log = getLogger(__name__)
+
+
+class FutureMarketException(Exception):
+    """Exception specific to the Future markets."""
+
+
+class FutureMarkets(TwoSidedMarket):
```
Hassan754, line 39: Can you explain why we name it in plural form?

src/d3a/models/market/future.py
```diff
@@ -0,0 +1,217 @@
...
+from typing import Dict, List, Union
+
+from d3a.d3a_core.exceptions import (BidNotFoundException, MarketReadOnlyException,
+                                     OfferNotFoundException)
+from d3a.events.event_structures import MarketEvent
+from d3a.models.market import lock_market_action
+from d3a.models.market.two_sided import TwoSidedMarket
+from d3a_interface.constants_limits import ConstSettings, GlobalConfig
+from d3a_interface.data_classes import Bid, Offer, Trade, BaseBidOffer
+from pendulum import DateTime, duration
```
Hassan754, line 30: The imports are not optimized

src/d3a/models/market/future.py
Hassan754: Scratch that, I understood now the reason and rational behind this implementation

PR #1258, opened 2021-10-22
D3ASIM-3630: Track devices energy requirement and availability for fu…

faizan2590:
…ture market

### DIFF
--- a/src/d3a/models/state.py
+++ b/src/d3a/models/state.py
@@ -21,10 +21,10 @@ from enum import Enum
 from math import isclose
 from typing import Dict
 
-from d3a_interface.constants_limits import ConstSettings, GlobalConfig
+from d3a_interface.constants_limits import ConstSettings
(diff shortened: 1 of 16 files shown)


src/d3a/models/state.py
```diff
@@ -50,6 +50,9 @@
 class StateInterface(ABC):
     """Interface containing methods that need to be defined by each State class."""
 
+    def __init__(self):
+        self._spot_market_time_slot: DateTime = None
```
Hassan754, line 54: I still vote for not adding any attribute to the state interface, we can still stick to the old function headers where we pass the market slot

Hassan754 (commented):
Left some comments, 1 important

src/d3a/models/state.py
faizan2590: Here I would like to tag @spyrostz as I added it after discussing with him.

PR #1261, opened 2021-10-22
Feature/d3 asim 3632 Added future IAAs

hannesdiedrich:


src/d3a/models/strategy/area_agents/future_engine.py
```diff
@@ -0,0 +1,73 @@
...
+            self._delete_trade_residual_buffers(buffer, current_market_slot)
+
+        delete_bids = []
+        for bid_info in self.forwarded_bids.values():
+            if bid_info.target_bid.time_slot <= current_market_slot:
+                delete_bids.append(bid_info.target_bid)
+                delete_bids.append(bid_info.source_bid)
+
+        for bid in delete_bids:
+            self._delete_forwarded_bid_entries(bid)
```
Hassan754, line 65: I don't believe there is a need for storing the ids and reiterate over them, same for the offers part
```suggestion
            if bid_info.target_bid.time_slot <= current_market_slot:
                self._delete_forwarded_bid_entries(bid_info.target_bid))
                self._delete_forwarded_bid_entries(bid_info.source_bid))

```

src/d3a/setup/kpi/d3asim_2103_nopv1.py
```diff
@@ -13,8 +13,8 @@ def get_setup(config):
 
     ConstSettings.GeneralSettings.DEFAULT_UPDATE_INTERVAL = 1
     ConstSettings.IAASettings.MARKET_TYPE = 1
-    ConstSettings.IAASettings.MIN_OFFER_AGE = 0
-    ConstSettings.IAASettings.MIN_BID_AGE = 0
+    # ConstSettings.IAASettings.MIN_OFFER_AGE = 0
+    # ConstSettings.IAASettings.MIN_BID_AGE = 0
```
Hassan754, line 17: why are these commented out?

tests/area/test_event_dispatcher.py
```diff
@@ -0,0 +1,253 @@
...
+    config.slot_length = duration(minutes=15)
+    config.tick_length = duration(seconds=15)
+    config.ticks_per_slot = 60
+    config.start_date = GlobalConfig.start_date
+    config.grid_fee_type = ConstSettings.IAASettings.GRID_FEE_TYPE
+    config.end_date = GlobalConfig.start_date + duration(days=1)
+
+    area = Area("name", config=config)
+    area.children = [Area("child1"), Area("child2")]
+    area.parent = Area("parent")
```
Hassan754, line 71: I suggest creating an area fixture instead of the AreaDispatcher and reuse it here
you can build up the `AreaDispatcher` wherever you need later on, and I also don't have a problem if you decide to keep the `AreaDispatcher` and create a separate area fixture

Hassan754 (commented):
Left some comments

integration_tests/kpi.feature
```diff
@@ -53,8 +53,8 @@ Feature: KPI integration tests
      Given we have a scenario named kpi.d3asim_2103_nopv1
      And d3a is installed
      When we run the d3a simulation on console with kpi.d3asim_2103_nopv1 for 24 hrs (30, 30)
-     Then self_sufficiency of {'Grid': 1.0, 'Community': 0.666, 'House 1': 0.0, 'House 2': 0.5, 'House 3': 1.0} are correctly reported
-     And self_consumption of {'Grid': 1.0, 'Community': 0.8, 'House 1': None, 'House 2': 1.0, 'House 3': 0.5} are correctly reported
+     Then self_sufficiency of {'Grid': 1.0, 'Community': 0.833, 'House 1': 0.0, 'House 2': 0.5, 'House 3': 1.0} are correctly reported
```
faizan2590, line 56: Whats the reason for change of KPI value in integration test?

hannesdiedrich: This was after a discussion with @spyrostz . We decided to move the expected outcome of the integrationtests again back to the original rationale behind this test: Test wether the Community used all its energy and sells the rest to the DSO (the same as in the corresponding 2PAB integrationtest).

tests/area/test_event_dispatcher.py
hannesdiedrich: I'll do the latter actually.

src/d3a/setup/kpi/d3asim_2103_nopv1.py
hannesdiedrich: This is related to [this](https://github.com/gridsingularity/d3a/pull/1261#discussion_r742843421) comment. Having a bid and offer age of 0 is a bit extreme and was probably added in order to bend the behaviour of the setup so that the expected test results are met. We used the default values now (which is 2). I will remove the comments though.

src/d3a/models/strategy/area_agents/future_engine.py
hannesdiedrich: No, unfortunately not, i will result in an "dictionary changed size during iteration" error.

src/d3a/models/strategy/area_agents/future_engine.py
```diff
@@ -0,0 +1,73 @@
...
+        return f"<FutureEngine [{self.owner.name}] {self.name} " \
+               f"{self.markets.source.time_slot:%H:%M}>"
+
+    @staticmethod
+    def _delete_trade_residual_buffers(residual_buffer: Dict,
+                                       current_market_slot: DateTime) -> None:
+        delete_order_ids = []
+        for order in residual_buffer.values():
+            if order.time_slot <= current_market_slot:
+                delete_order_ids.append(order.id)
```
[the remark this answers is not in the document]
hannesdiedrich: yes, same as https://github.com/gridsingularity/d3a/pull/1261/files/7f50801cca399e6c2d92e261de02a41ab571a25e#r742849167

src/d3a/models/area/event_dispatcher.py
```diff
@@ -159,12 +171,23 @@ def broadcast_notification(
...
+                AvailableMarketTypes.SPOT, event_type, **kwargs)
+        if isinstance(event_type, AreaEvent) or self.area.is_market_balancing(market_id):
+            self._broadcast_notification_to_area_and_child_agents(
+                AvailableMarketTypes.BALANCING, event_type, **kwargs)
+        if isinstance(event_type, AreaEvent) or self.area.is_market_settlement(market_id):
+            self._broadcast_notification_to_area_and_child_agents(
+                AvailableMarketTypes.SETTLEMENT, event_type, **kwargs)
+        if isinstance(event_type, AreaEvent) or self.area.is_market_future(market_id):
+            self._broadcast_notification_to_area_and_child_agents(
+                AvailableMarketTypes.FUTURE, event_type, **kwargs)
```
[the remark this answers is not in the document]
hannesdiedrich: `_broadcast_notification_to_area_and_child_agents` receives two types of events: `MarketEvent, AreaEvent`. For `AreaEvent`s (TICK, MARKET_CYCLE, ..), we want to dispatch it to all of the connected areas independent of the type. AFAIU, this is only temporary, because ideally, we have AreaEvents for each market type (like the `BALANCING_MARKET_CYCLE`) . Because the area may want to react differently if a FutureMarket TICK event is received rather then a normal TICK is received. Shall we add a story for that @spyrostz ?

In order to answer your question, the second condition is for the `MarketEvents` like `OFFER`, `BID`, `BID_TRADED` etc. These should only be dispachted to the corresponding IAA for the market type. For example, a Settlement IAA is not interested, if an OFFER is posted in the spot market. 
Does it make sense? BTW, I took a lot of time to come up with a more DRY version of this, but could not find one.

src/d3a/models/area/__init__.py
```diff
@@ -618,3 +620,19 @@ def update_config(self, **kwargs):
             self.strategy.read_config_event()
         for child in self.children:
             child.update_config(**kwargs)
+
+    def is_market_spot(self, market_id: str) -> bool:
```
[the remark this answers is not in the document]
hannesdiedrich: Yeah, I was refactoring these a lot forth and back. If you have a look at the caller of this method [here](https://github.com/gridsingularity/d3a/pull/1261/files/7f50801cca399e6c2d92e261de02a41ab571a25e#diff-f66774415440ea8a12f186f18c94dc58a2175b62ed94c9ab09ce8298b3eef863R179) would agree that `self.area.is_market_spot` is less confusing than `self.area.is_spot_market` ? Because in the latter version you could read "Area is spot market" which does not make sense IMO.
Maybe you or others have a better proposal to name this properly?

src/d3a/models/area/event_dispatcher.py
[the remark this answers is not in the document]
spyrostz: @hannesdiedrich feel free to add a story for that, however if we decide to move forward to organic grid structure (where we will not have area dispatcher but a central entity that keeps track of the grid structure and dispatches events to different markets or strategies - if we implement that soon enough then this ticket will not really be useful). Having said all the above, since I am not really certain how far in the future will we implement that, feel free to open a story to improve the situation in the short term

src/d3a/models/area/event_dispatcher.py
```diff
@@ -222,6 +245,10 @@ def _create_agent_object(owner: "Area", higher_market: Market,
             return SettlementAgent(**agent_constructor_arguments)
         if market_type == AvailableMarketTypes.BALANCING:
             return BalancingAgent(**agent_constructor_arguments)
+        if market_type == AvailableMarketTypes.FUTURE:
+            return FutureAgent(**agent_constructor_arguments,
+                               min_bid_age=ConstSettings.IAASettings.MIN_BID_AGE)
```
spyrostz, line 250: It should not be needed here

src/d3a/models/area/event_dispatcher.py
```diff
@@ -236,6 +263,32 @@ def _get_agents_for_market_type(
...
+            return False
+        if self.area.strategy:
+            return False
+        if not self.area.parent.events.is_connected:
+            return False
+        if not self.area.children:
+            return False
+        return True
+
+    def create_area_agents_for_future_markets(self, market: Market) -> None:
```
spyrostz, line 278: I think this method is not needed and can be incorporated in the create_area_agents method. I will try to tackle it in the context of feature/D3ASIM-3633

src/d3a/models/strategy/__init__.py
```diff
@@ -646,14 +646,16 @@ def update_offer_rates(self, market: "OneSidedMarket", updated_rate: float) -> N
             return
 
         for offer, iterated_market_id in self.offers.open.items():
+            updated_price = round(offer.energy * updated_rate, 10)
+            if offer.price == updated_price:
```
spyrostz, line 650: Minor, but since we are talking about floating points, and we know how well does Python handle floating point precision (not), I would feel much safer if you used math.isclose here. Thanks!

src/d3a/models/strategy/area_agents/two_sided_agent.py
```diff
@@ -15,35 +15,41 @@
...
         super().__init__(owner=owner,
                          higher_market=higher_market, lower_market=lower_market,
                          min_offer_age=min_offer_age, do_create_engine=False)
 
     def usable_bid(self, bid):
-        """Prevent IAAEngines from trading their counterpart's bids"""
+        """Prevent IAAEngines from trading their counterpart's bids."""
         return all(bid.id not in engine.forwarded_bids.keys() for engine in self.engines)
 
+    # pylint: disable=unused-argument
```
spyrostz, line 47: I am frankly surprised that this error is raised in some cases and not others (for instance, in other cases pylint can detect that the base class is BaseStrategy that implements the same methods with the same argument names, therefore it can detect that you do it that way here in order to adhere to the base class interface). Nevertheless nothing to do here unfortunately .

spyrostz (approved):
Some nits, LGTM otherwise

src/d3a/models/area/event_dispatcher.py
hannesdiedrich: Hmm, I was thinking about this a lot, too. But then you would need a check for the market_type in `create_area_agents` which dies not make it more pretty. This method is already called from a future market specific method (`area.markets.activate_future_markets`). So, IMO it is good like it is here.

src/d3a/models/area/__init__.py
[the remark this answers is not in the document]
Hassan754: It really seems off to have this attribute/property in the Area class, the two names don't make such sense

I would vote for getting rid of these functions and only have 1 method that returns the type of market (spot, future, settlement ...)

spyrostz (dismissed):
LGTM

src/d3a/models/area/event_dispatcher.py
```diff
@@ -107,17 +113,23 @@ def broadcast_balancing_market_cycle(self, **kwargs) -> None:
...
-        agent_dict = self._get_agents_for_market_type(agent_area.dispatcher, market_type)
-        for time_slot, agent in agent_dict.items():
-            if time_slot not in agent_area.get_market_instances_from_class_type(
-                    market_type):
-                # exclude past IAAs
-                continue
 
-            agent.event_listener(event_type, **kwargs)
+        if market_type == AvailableMarketTypes.FUTURE:
+            if self.future_agent:
```
faizan2590, line 118:
```suggestion
        if market_type == AvailableMarketTypes.FUTURE and self.future_agent:
```

src/d3a/models/strategy/area_agents/future_agent.py
```diff
@@ -0,0 +1,63 @@
...
+        ]
+        self.name = make_sa_name(self.owner)
+
+    def _clean_up_engine_buffers(self) -> None:
+        if not self.owner.current_market:
+            return
+        for engine in self.engines:
+            engine.clean_up_order_buffers(self.owner.current_market.time_slot)
+
+    def delete_engines(self) -> None:
```
faizan2590, line 57:
```suggestion
    def _delete_engines(self) -> None:
```

faizan2590 (commented):
Left some minor comments, other than that LGTM

src/d3a/models/strategy/__init__.py
```diff
@@ -646,14 +647,16 @@ def update_offer_rates(self, market: "OneSidedMarket", updated_rate: float) -> N
             return
 
         for offer, iterated_market_id in self.offers.open.items():
+            updated_price = round(offer.energy * updated_rate, 10)
```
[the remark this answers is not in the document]
hannesdiedrich: This would change the behaviour and would make the updated price differences more granular as the precision in `limit_float_precision` is 8. I vote for keeping it like this.

src/d3a/models/strategy/area_agents/future_agent.py
hannesdiedrich: This method is called by the dispatcher and needs to be public. https://github.com/gridsingularity/d3a/blob/52b48f1b2ccb5e1aaf9cbeb5f53b4cec703594b9/src/d3a/models/area/event_dispatcher.py#L335

src/d3a/models/area/event_dispatcher.py
hannesdiedrich: Unfortunately, this has to be like this, otherwise a future market event would be posted to the other IAAs. I had changed this maybe 5 times, because I saw the refactoring opportunity, but tests always failed. I will leave an inline comment in order to not get refactored in a future PR.

spyrostz (approved):
LGTM

integration_tests/savings_kpi.feature
```diff
@@ -8,5 +8,5 @@ Feature: Savings KPI integration tests
   Examples: Cases
     | scenario                          | setup                               | saving_percentage |
     | Young Couple House                | kpi.d3asim_3464_young_couple        |     16.6667      |
-    | Family 2 children with PV         | kpi.d3asim_3464_family_with_pv      |     43.537       |
+    | Family 2 children with PV         | kpi.d3asim_3464_family_with_pv      |     28.427       |
```
spyrostz, line 11: I do not have a problem per se by this, however I am curious as to why it happened. As far as I understand, the current MIN_OFFER_AGE value for this test is still 0, therefore it should be able to broadcast immediately every update in the energy rate, therefore the deleting + reposting of the PV offer on every tick (the behavior that the commit solved) should in theory have exactly the same effect as reposting the event only during update_intervals. I would recommend to leave it as is, and open a ticket in order to further debug this behavior.

spyrostz (dismissed):
LGTM

src/d3a/models/strategy/__init__.py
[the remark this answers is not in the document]
faizan2590: I would prefer to keep the precision level to global, unless in this specific case where changing precision level from 10 to the global one (8) revealed any adverse affect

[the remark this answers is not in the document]
hannesdiedrich: I ran all tests locally and they are passing even with precision level 8.
I will change it now but fear some tickets from researchers.

tests/market/test_future.py
```diff
@@ -194,7 +184,8 @@ def test_accept_offer(future_market):
         """Test if trade is added to trade buffers when accept_offer is called."""
         first_future_market = next(iter(future_market.slot_bid_mapping))
         offer = future_market.offer(1, 1, "seller", "seller_origin", time_slot=first_future_market)
-        trade = future_market.accept_offer(offer, "buyer")
+        trade = future_market.accept_offer(offer, "buyer",
+                                           trade_bid_info=TradeBidOfferInfo(1, 1, 1, 1, 1))
```
Hassan754, line 188: Why did you pass the `trade_bid_info` ?

faizan2590 (approved):
LGTM

tests/area/test_event_dispatcher.py
```diff
@@ -0,0 +1,255 @@
...
+         BalancingAgent]
+    ])
+    def test_create_area_agents_creates_correct_objects(self, market_type: AvailableMarketTypes,
+                                                        spot_market_type: SpotMarketTypeEnum,
+                                                        market_class: Market,
+                                                        expected_agent_type: InterAreaAgent,
+                                                        area_dispatcher):
+        """Test if create_area_agents creates correct objects in the agent dicts."""
+        original_matching_type = ConstSettings.IAASettings.MARKET_TYPE
+        ConstSettings.IAASettings.MARKET_TYPE = spot_market_type
```
Hassan754, line 128: Why is this needed, I couldn't find the reason behind it.
Reason for asking this is to check whether we can only pass either the `spot_market_type` or the `market_class`, not both

tests/market/test_future.py
hannesdiedrich: I can not remember. I think the test was failing.I just checked, it is indeed not needed.

tests/area/test_event_dispatcher.py
hannesdiedrich: `AreaDispatcher._create_agent_object` depends on this global setting

src/gsy_e/models/area/__init__.py
```diff
@@ -373,6 +373,8 @@ def cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fal
         else:
             changed_balancing_market = None
 
+        self._markets.update_area_market_id_lists()
```
Hassan754, line 376: I have the feeling that we may end up out of sync between the markets ids lists and the actual markets in this _markets handler
I think a better solution would be to call this `update_area_market_id_lists` at the end of each `create_new_*_market` method call, this will also allow us to make `update_area_market_id_lists` private

src/gsy_e/models/area/__init__.py
```diff
@@ -618,3 +620,19 @@ def update_config(self, **kwargs):
             self.strategy.read_config_event()
         for child in self.children:
             child.update_config(**kwargs)
+
+    def is_market_spot(self, market_id: str) -> bool:
```
Hassan754, line 624: Nit
AFAIS the `is_market_spot`, `is_market_settlement`, `is_market_balancing` and `is_market_future` are only used in the `broadcast_notification` method.
I prefer to refrain from adding all of these, imo we can get rid of them and create 1 single method that returns the `type` of the passed market_id

src/gsy_e/models/area/event_dispatcher.py
```diff
@@ -108,17 +114,25 @@ def broadcast_balancing_market_cycle(self, **kwargs) -> None:
...
-                    market_type):
-                # exclude past IAAs
-                continue
 
-            agent.event_listener(event_type, **kwargs)
+        if market_type == AvailableMarketTypes.FUTURE:
+            # First check for the market_type and then check if an future agent exists, otherwise
+            # the future event would be posted on non-future agents.
+            if self.future_agent:
+                self.future_agent.event_listener(event_type, **kwargs)
```
Hassan754, line 122:
```suggestion
        if market_type == AvailableMarketTypes.FUTURE and self.future_agent:
            # First check for the market_type and then check if an future agent exists, otherwise
            # the future event would be posted on non-future agents.
            self.future_agent.event_listener(event_type, **kwargs)
```

src/gsy_e/models/strategy/area_agents/future_agent.py
```diff
@@ -0,0 +1,63 @@
...
+                         do_create_engine=False)
+        self.engines = [
+            FutureEngine("High -> Low", higher_market, lower_market,
+                         min_offer_age, min_bid_age, self),
+            FutureEngine("Low -> High", lower_market, higher_market,
+                         min_offer_age, min_bid_age, self),
+        ]
+        self.name = make_sa_name(self.owner)
+
+    def _clean_up_engine_buffers(self) -> None:
```
Hassan754, line 51: IMO this doesn't deserve its own method, it's only used once inside the `delete_engines` (It belongs there AFAIS)

src/gsy_e/models/strategy/area_agents/balancing_agent.py
```diff
@@ -55,14 +58,15 @@ def event_bid_traded(self, *, market_id, bid_trade):
...
-        market = self._get_market_from_market_id(market_id)
+        market = self.get_market_from_market_id(market_id)
         if market is None:
             return
 
         self._calculate_and_buy_balancing_energy(market, bid_trade)
         super().event_bid_traded(market_id=market_id, bid_trade=bid_trade)
 
     def _calculate_and_buy_balancing_energy(self, market, trade):
+        print("_calculate_and_buy_balancing_energy")
```
Hassan754, line 69: reminder

src/gsy_e/models/strategy/area_agents/balancing_agent.py
```diff
@@ -98,6 +102,7 @@ def _balancing_trade(self, offer, target_energy):
         buyer = make_ba_name(self.owner) \
             if make_ba_name(self.owner) != offer.seller \
             else f"{self.owner.name} Reserve"
+        print("_balancing_trade", abs(offer.energy), abs(target_energy))
```
Hassan754, line 105: reminder

src/gsy_e/models/strategy/area_agents/two_sided_agent.py
```diff
@@ -15,35 +15,41 @@
...
-                 min_bid_age=ConstSettings.IAASettings.MIN_BID_AGE):
-        self.engines = [
-            TwoSidedEngine('High -> Low', higher_market, lower_market, min_offer_age,
-                           min_bid_age, self),
-            TwoSidedEngine('Low -> High', lower_market, higher_market, min_offer_age,
-                           min_bid_age, self),
-        ]
+                 min_bid_age=ConstSettings.IAASettings.MIN_BID_AGE,
+                 do_create_engine=True):
+        if do_create_engine:
```
Hassan754, line 32: Ummm I don't find this approach clean tbh
In case this is here in order to control what type of engines to create alongside the mro, I suppose we can create a dedicated method that selects the wanted engine, and we can overwrite it later on

src/gsy_e/models/strategy/area_agents/settlement_agent.py
```diff
@@ -1,3 +1,20 @@
+"""
+Copyright 2018 Grid Singularity
+This file is part of D3A.
```
Hassan754, line 3: Reminder to update the newly added modules
```suggestion
This file is part of Grid Singularity Exchange.
```

src/gsy_e/models/strategy/area_agents/one_sided_engine.py
```diff
@@ -186,7 +194,7 @@ def event_offer_traded(self, *, trade):
             except OfferNotFoundException:
                 pass
             except MarketException as ex:
-                self.owner.log.error("Error deleting InterAreaAgent offer: {}".format(ex))
+                self.owner.log.error("Error deleting InterAreaAgent offer: %s", ex)
```
Hassan754, line 197: Nit
```suggestion
                self.owner.log.exception("Error deleting InterAreaAgent offer:")
```

src/gsy_e/models/area/__init__.py
spyrostz: I also take advantage of these methods in many places in D3ASIM-3633, and I think that they are good to be there. The disadvantage of one single method would be that the caller code would have to also import the market type enum and compare against it in order to differentiate the market type. Therefore more imports and more verbose IMO. However it is a matter of preference IMO

tests/strategies/area_agents/test_future_agent.py
```diff
@@ -0,0 +1,54 @@
+"""
+Copyright 2018 Grid Singularity
+This file is part of gsy_e.
```
Hassan754, line 3: AFAIS it should be `part of Grid Singularity Exchange

Hassan754 (approved):
Left some comments, non blockers and can be addressed in a followup PR if accepted, LGTM

src/gsy_e/models/area/__init__.py
Hassan754: Makes sense, and I agree that it's entirely subjective.

src/gsy_e/models/area/event_dispatcher.py
hannesdiedrich: @Hassan754 please read carefully, I explicitly left an inline comment in response to a comment from Faizan. This needs to be like I implemented.

Hassan754: @hannesdiedrich I'm not sure I get you, how is 
```python
if condition_a and condition_b:
    do_something()
```
different than

```python
if condition_a
   if condition_b:
     do_something()
```

src/gsy_e/models/strategy/area_agents/future_agent.py
hannesdiedrich: Ok, I can move the code to `delete_engines`

src/gsy_e/models/strategy/area_agents/two_sided_agent.py
hannesdiedrich: ok, let me try this.

src/gsy_e/models/area/event_dispatcher.py
hannesdiedrich: The problem is the else block
```python
if condition_a and condition_b:
    do_something()
else:
    do_something_else()
```
is different from 
```python
if condition_a
   if condition_b:
    do_something()
else:
    do_something_else()
```
In the first case, if one of the conditions is not try, it will automatically go into the else block. But this is not what we want. we actually want
```python
if condition_a
   if condition_b:
     do_something()
   else:
      do_nothing()
else:
    do_something_else()
```
I you like i can add the specific else block or convert the outer else block into an and check explicitly to for the rest of the market types.

Hassan754: Ah I see, I completely overlooked this
In this case, you have the freedom to select any approach you like

hannesdiedrich: My proposal:
```python
if condition_a and condition_b:
       do_something()
if condition_c:
    do_something_else()
```

Hassan754: Seems great! only convert the second if to a `elif` and it should be fine

hannesdiedrich: this will make pylint complain. :-( 
I will go the following way actually:
```python
if condition_a and condition_b:
       do_something()
       return
do_something_else()
```

src/gsy_e/models/area/__init__.py
hannesdiedrich: Good idea! I was also a bit insecure here. Your solution seems more secure.

src/gsy_e/models/area/event_dispatcher.py
Hassan754: I tried this and pylint didn't complain
```python
        if market_type == AvailableMarketTypes.FUTURE and self.future_agent:
        elif market_type != AvailableMarketTypes.FUTURE:
```

src/gsy_e/models/area/__init__.py
hannesdiedrich: Actually, now I remember, why I decided to to it like this: If we call it multiple times per market_slot, it will add an unnecessary load overhead. So, I would vote for keeping it like this.

Hassan754: How often do we access these market ids lists? we can create properties for them which will ensure that they're always up to date (and no computation will be done unless we do access the property)

hannesdiedrich: They are access on every market event and area event (so on every tick, market_cycle, offer, bid, accept_offer, etc) So a LOT of times.

src/gsy_e/models/strategy/area_agents/inter_area_agent.py
```diff
@@ -42,6 +43,13 @@ def __init__(self, *, owner, higher_market, lower_market,
         self.lower_market = lower_market
         self.min_offer_age = min_offer_age
 
+        self._create_engines()
+        self.name = make_iaa_name(owner)
+        self.uuid = owner.uuid
+
+    def _create_engines(self):
```
Hassan754, line 50: 👍

src/gsy_e/models/area/__init__.py
Hassan754: We may still cache them, however let's leave it for now

Hassan754 (approved):
LGTM

PR #1268, opened 2021-10-28
 Refactor external strategies styling

Hassan754:


spyrostz (approved):
LGTM

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -39,227 +46,286 @@
...
+            self.redis, self.channel_prefix, self.connected,
+            self._get_transaction_id(payload))
+
+    def _update_connection_status(self) -> None:
+        """Update the connected flag to sync it with the _registered flag.
 
-    def _unregister(self, payload):
-        self._connected = unregister_area(self.redis, self.channel_prefix, self.connected,
-                                          self._get_transaction_id(payload))
+        If the client unregistered (_registered=False), pop the asset from the aggregator mapping
```
hannesdiedrich, line 218: Also the other direction applies right? That is why you renamed the function, right?
So I would propose:
```suggestion
        Change assets' connection status including the connection to the aggregator.
```

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -58,13 +58,15 @@ def _get_transaction_id(payload):
...
-                                        area_uuid=self.area.uuid)
+        self._connected = ExternalStrategyConnectionManager.register(
+            self._redis_communicator, self.channel_prefix,
+            self.connected, self._get_transaction_id(payload),
+            area_uuid=self.area.uuid)
 
     def _unregister(self, payload):
-        self._connected = unregister_area(self._redis_communicator, self.channel_prefix,
-                                          self.connected, self._get_transaction_id(payload))
+        self._connected = ExternalStrategyConnectionManager.unregister(
```
hannesdiedrich, line 67: Shouldn't this be
```suggestion
        self._is_registered = ExternalStrategyConnectionManager.unregister(
```
now?

Hassan754: @hannesdiedrich Actually no, I was not refactoring the `redi_external_market_connection`, I only updated this function name in the refactoring
So this flag is not touched here (I only renamed the flag in the external strategies)

Weirdly though, I took a look at the `redi_external_market_connection` _connected flag just now, and it seems that it's not used at all (we just set it to True/False but we do not reference it anywhere)
I suppose this is not harmful at the moment but should indeed be refactored in a separate PR, do you agree?

src/d3a/models/strategy/external_strategies/__init__.py
Hassan754: Right, Adapted!

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: I see, this is a bug. The register and unregister functionality does not work for the market connection. The `self.connected` flag of the area is never checked in any of the market commands.
`_register` always worked before because `self.connected` is always False.
And apparently nobody ever called `_unregister` which would have revealed the error. We probably only used the aggregator connection lately
Please open a bug story actually. It is not crucial, as for the REST connection via d3a-web we have another layer of security, but still, the direct connection should be working correctly even though nobody is using it. Thanks!

hannesdiedrich (approved):
After bug ticket is created, LGTM

hannesdiedrich (commented):
@Hassan754 Please run the d3a-api-client integration tests with these changes locally and check whether they are passing before pushing this ok? Thanks a lot!

src/d3a/models/area/redis_external_market_connection.py
Hassan754: Sounds good, I have created the ticket!
https://gridsingularity.atlassian.net/browse/D3ASIM-3678

Hassan754:
> @Hassan754 Please run the d3a-api-client integration tests with these changes locally and check whether they are passing before pushing this ok? Thanks a lot!

Thanks for the reminder, they are passing!

PR #1269, opened 2021-10-28
D3ASIM-3626: Add flag for Degrees of Freedom

fievelk:
Add a flag to enable/disable Degrees of Freedom.

Note: we define as DoF the `attributes` and `requirements` arguments of the orders.

## Steps

To test the changes you can do the following
1. Modify `test_redis_aggregator` setup in d3a-api-client by adding DoF to the sent orders. For example:
    ```python
    self.add_to_batch_commands.bid_energy(
        area_uuid=area_uuid, price=30, energy=energy, 
        attributes={"some-attribute": 12})
    ```

2. Run d3a and d3a-api-client:
    ```shell
    d3a run --setup strategy_tests.external_devices --no-export --seed 0 --enable-external-connection --disable-dof --paused
    d3a-api-client -l INFO run --setup test_redis_aggregator
    ```

You should now be able to see that the DoF have been correctly stripped off the orders. To do that, you can check the returned output of the batch aggregator command in d3a-api-client. It should contain a message like the following:
> {..., 'message': "The following arguments are not supported for this market and have been removed from your order: ['attributes']."}


spyrostz (approved):
LGTM

Hassan754 (dismissed):
LGTM

src/gsy_e/models/strategy/__init__.py
```diff
@@ -458,6 +459,15 @@ def __init__(self):
 
     parameters = None
 
+    @property
+    def simulation_config(self) -> SimulationConfig:
+        """Return the SimulationConfiguration used by the current market area.
+
+        NOTE: this configuration is currently unique and shared across all areas, but this might
+        change in the future (each area could have its own configuration).
```
hannesdiedrich, line 467: 👍  This is why I would not vote for adding this property now. Leave it for now but I predict that we will revert it soonish.

src/d3a/models/strategy/__init__.py
```diff
@@ -231,6 +232,14 @@ def __init__(self):
 
     parameters = None
 
+    @property
+    def simulation_config(self) -> SimulationConfig:
```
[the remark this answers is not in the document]
hannesdiedrich: > If anything changes in the interface of any of the intermediate objects, everything breaks.

But the wrapping property will not help with that, it will fail too, right?

spyrostz (approved):
LGTM

src/gsy_e/models/strategy/external_strategies/pv.py
```diff
@@ -315,6 +315,13 @@ def _list_offers_aggregator(self, arguments: Dict) -> Dict:
         return response
 
     def _offer_aggregator(self, arguments: Dict) -> Dict:
+        response_message = ""
+        arguments, filtered_fields = self.filter_degrees_of_freedom_arguments(arguments)
+        if filtered_fields:
+            response_message = (
+                "The following arguments are not supported for this market and have been "
+                f"removed from your order: {filtered_fields}.")
+
```
hannesdiedrich, line 324: This looks to me that it could also be put into a decorator and consequently used more easily in multiple places. What do you think?

src/gsy_e/models/strategy/external_strategies/dof_filter.py
```diff
@@ -0,0 +1,24 @@
...
+
+        The order dictionary is modified in-place.
+        Args:
+            order: a dictionary of arguments describing a bid or an offer.
+        """
+        filtered_fields = []
+        for field in cls.FIELDS:
+            if order.get(field):
+                order.pop(field)
+                filtered_fields.append(field)
```
hannesdiedrich, line 22: How about this?
```suggestion
            if order.pop(field, None):
                filtered_fields.append(field)
```

hannesdiedrich (commented):
Two comments. LGTM in general

src/gsy_e/models/strategy/__init__.py
fievelk: I agree that this will need to be changed. When this happens I suggest to just rename it to `area_config` or something similar. I wouldn't remove it completely: I think the property itself would still be useful and better than using `self.area.config`.

src/d3a/models/strategy/__init__.py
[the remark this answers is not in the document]
fievelk: @hannesdiedrich The property wouldn't "save" us from this, but it would make our life easier. It would allow us to fix the breaking bug by just modifying one line (the content of the property). If we didn't have it, we would need to change all the calls in the entire codebase.
Also, suppose we use properties to alias long chains of calls. Once we start doing it, we'll find out that even the content of our properties won't be a long chain of calls. That's because it will contain calls to aliases. E.g.:
```python
@property
def simulation_config(self):
    return self.area.config

@property
def grid_fee_type(self):
    # Here we abbreviated even more. It would have been `self.area.config.grid_fee_type`
    return self.simulation_config.grid_fee_type
```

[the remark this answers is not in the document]
hannesdiedrich: ok, convinced!

src/gsy_e/models/strategy/external_strategies/pv.py
fievelk: We've briefly talked about it [here](https://github.com/gridsingularity/gsy-e/pull/1269#discussion_r738560687). @Hassan754 suggested to leave it as it is for now, since new changes will come soon.

src/gsy_e/models/strategy/external_strategies/dof_filter.py
fievelk: It looks good to me. I remember fiddling with this method for a bit but right now I don't see why we shouldn't use your solution. Thanks!

fievelk: @hannesdiedrich I've followed your comment in 1394ca7

hannesdiedrich:
Once the tests pass, LGTM

fievelk:
> Once the tests pass, LGTM

@hannesdiedrich tests are failing because this PR has gsy-framework dependencies. I tested it locally and it works. I'll do it again now.

Hassan754 (approved):

spyrostz (approved):

PR #1272, opened 2021-11-01
External strategy abc

spyrostz:
This PR deals with the following: 
- moving the forecast and measurement part of the external strategies from ExternalMixin to their own mixin, in order for the ExternalMixin to include only code that is always necessary for every mixin (need to extend this in order to abstract BidExternalMixin and OfferExternalMixin too)
- declaring the dependencies of the mixins as type hints, in order to have more clarity over which members does the mixed-in  class need to have in order for the mixin to work correctly. 

The initial approach to declare abstract methods for all the needed members did not work out unfortunately, because the Mixins need to precede the class that inherits them in the MRO order (in order to override methods that the mixed-in strategy class defines, e.g. event_tick). Therefore by declaring these methods as abstract the concrete class would end up with abstract method implementations, therefore raising an error when calling them.

### DIFF
--- a/src/d3a/models/strategy/external_strategies/pv.py
+++ b/src/d3a/models/strategy/external_strategies/pv.py
@@ -38,7 +38,7 @@ if TYPE_CHECKING:
 class PVExternalMixin(ExternalMixin):
 
     state: "PVState"
-    offers: Offers
+    offers: "Offers"
     can_offer_be_posted: Callable
     post_offer: Callable
     set_produced_energy_forecast_kWh_future_markets: Callable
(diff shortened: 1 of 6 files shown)


Hassan754 (approved):
LGTM, Thanks!

src/d3a/models/strategy/external_strategies/pv.py
```diff
@@ -17,21 +17,33 @@
...
+
 
 class PVExternalMixin(ExternalMixin):
+
+    state: "PVState"
+    offers: "Offers"
+    can_offer_be_posted: Callable
+    post_offer: Callable
+    set_produced_energy_forecast_kWh_future_markets: Callable
+    _delete_past_state: Callable
```
hannesdiedrich, line 45: Just to rephrase what you wrote in the description of the PR:
With these type hints you are making sure that the `PVExternalMixin` is aware of these methods and even if the `ExternalMixin` is not implementing them, no error is raised by pycharm, right?

spyrostz: Precisely, but more importantly, the future developer will know that these are the methods that need to be implemented by the mixed-in class. It is probably an overkill for the user of the class since we all more or less know that this class needs to be mixed-in to a subclass of BaseStrategy/BidEnabledStrategy, however for the future maintainer it might facilitate him when refactoring this class.

PR #1275, opened 2021-11-01
D3ASIM-3667: Add interval for future market clearing times

Hassan754:


src/d3a/d3a_core/util.py
```diff
@@ -486,3 +487,28 @@ def is_time_slot_in_past_markets(time_slot: DateTime, current_time_slot: DateTim
...
         return time_slot < current_time_slot
+
+
+class FutureMarketCounter:
+    """Hold a time counter for the future market.
+
+    In the future market, we only want to clear in a predefined interval.
+    """
+    def __init__(self):
+        self.last_time_dispatched = pendulum.now()
```
spyrostz, line 498: I think that this will not work out as expected. It might work out in the case of Canary Networks, where the simulation time is following the real clock time, however it will not work for simulation and collaborations that are running in accelerated mode. In order to achieve that you need to pass the simulation clock time. Easiest way to get it is via the market.now member, which contains the simulation datetime.

src/d3a/d3a_core/util.py
```diff
@@ -486,3 +487,28 @@ def is_time_slot_in_past_markets(time_slot: DateTime, current_time_slot: DateTim
...
+
+    @property
+    def is_time_for_clearing(self) -> bool:
+        """Compare current time with the latest time clearing was dispatched.
+
+        Returns True if the FUTURE_MARKET_CLEARING_INTERVAL has
+        already passed since the last dispatch time.
+        """
+        current_time = pendulum.now()
```
spyrostz, line 507: Same as above, you should compare the simulation times and not the actual real times. Therefore it would be better if is_time_for_clearing is a method that receives the current_time as an argument

src/d3a/d3a_core/util.py
```diff
@@ -486,3 +487,28 @@ def is_time_slot_in_past_markets(time_slot: DateTime, current_time_slot: DateTim
...
+        duration_in_s = (current_time - self.last_time_dispatched).seconds
+        duration_in_min = duration_in_s // 60
+        x = ConstSettings.FutureMarketSettings.FUTURE_MARKET_CLEARING_INTERVAL
```
spyrostz, line 510: Probably a suffix _MINUTES or _M would be good in the setting name to be more explicit.

spyrostz (commented):
Some comments, will LGTM once resolved

spyrostz (approved):
All comments resolved, LGTM

src/d3a/d3a_core/util.py
```diff
@@ -486,3 +486,30 @@ def is_time_slot_in_past_markets(time_slot: DateTime, current_time_slot: DateTim
...
+    def __call__(self, last_time_dispatched: DateTime):
+        """To be called when activating the counter."""
+        self.last_time_dispatched = last_time_dispatched
+
+    def is_time_for_clearing(self, current_time: DateTime) -> bool:
+        """Compare current time with the latest time clearing was dispatched.
+
+        Returns True if the FUTURE_MARKET_CLEARING_INTERVAL_MINUTES has
+        already passed since the last dispatch time.
+        """
```
spyrostz, line 508:
```suggestion
        """
        if not self.last_time_dispatched: 
            return True
```
In order to immediately mark that it is time for clearing if the original value is None.

Also minor but it would be good to be added, last_time_dispatched can probably be private.

Hassan754: @spyrostz seems great, do you think that we can also get rid of the `__call__` method and set the `last_time_dispatched` in the first function call as following  
```python
if not self._last_time_dispatched: 
            self._last_time_dispatched = current_time
            return True

```

Hassan754: I suppose the only difference we will get is at the first call, do we want to match future markets from the very beginning or do we need to wait the interval before that ?

spyrostz: @Hassan754 even better, I like your suggestion even more!

src/d3a/d3a_core/util.py
```diff
@@ -486,3 +486,29 @@ def is_time_slot_in_past_markets(time_slot: DateTime, current_time_slot: DateTim
...
+        if not self._last_time_dispatched:
+            self._last_time_dispatched = current_time
+            return True
+        duration_in_s = (current_time - self._last_time_dispatched).seconds
```
fievelk, line 508: Wouldn't this be the same?
```suggestion
        duration_in_s = (current_time - self._last_time_dispatched).minutes
```

fievelk (dismissed):

src/d3a/d3a_core/util.py
Hassan754: Of course! Thanks for catching that, it was a remnant of several design and implementation changes and got overlooked by me

spyrostz (approved):
LGTM

fievelk (approved):

faizan2590 (approved):
lgtm

PR #1277, opened 2021-11-02
Feature/d3 asim 3633

spyrostz:


src/d3a/constants.py
```diff
@@ -67,3 +67,12 @@ class SettlementTemplateStrategiesConstants:
...
     UPDATE_INTERVAL_MIN = 5
+
+
+class FutureTemplateStrategiesConstants:
+    INITIAL_BUYING_RATE = 0
+    FINAL_BUYING_RATE = 50
+    INITIAL_SELLING_RATE = 50
+    FINAL_SELLING_RATE = 0
+
+    UPDATE_INTERVAL_MIN = 5
```
faizan2590, line 78:
```suggestion
class FutureTemplateStrategiesConstants(SettlementTemplateStrategiesConstants):
```

src/d3a/models/strategy/future/strategy.py
```diff
@@ -0,0 +1,253 @@
...
+            return
+        strategy.post_offer(
+            market=strategy.area.future_markets,
+            replace_existing=False,
+            energy=available_energy_kWh,
+            price=available_energy_kWh * self._offer_updater.initial_rate[time_slot],
+            time_slot=time_slot
+        )
+
+    def _post_prosumer_first_bid_offer(
```
faizan2590, line 183:
```suggestion
    def _post_prosumer_first_order(
```

faizan2590 (commented):
some minor comments, other than that LGTM

src/d3a/models/market/two_sided.py
```diff
@@ -103,6 +103,9 @@ def bid(self, price: float, energy: float, buyer: str, buyer_origin: str,
         if energy <= 0:
             raise InvalidBid()
 
+        if time_slot is None:
+            time_slot = self.time_slot
```
hannesdiedrich, line 107: Great! So now all of the order objects should have a not-None `time_slot` member. 👍

src/d3a/models/base.py
```diff
@@ -31,8 +32,17 @@
...
+class AssetType(Enum):
+    PRODUCER = 0
+    CONSUMER = 1
+    PROSUMER = 2
+    UNDEFINED = 3
+
+
 class AreaBehaviorBase:
     """Base class used by area behaviour defining classes `BaseStrategy`"""
+    asset_type: AssetType = AssetType.UNDEFINED
```
hannesdiedrich, line 44: This contradicts with the property you implemented below, right?
IMO this can/should be removed here.

src/d3a/setup/d3a-settings.json
```diff
@@ -98,7 +98,6 @@
         10000
       ],
       "MIN_ALLOWED_SOC": 10,
-      "SELL_ON_MOST_EXPENSIVE_MARKET": false,
```
hannesdiedrich, line 101: This needs a small PR on d3a-interface, right?

tests/strategies/area_agents/test_future_engine.py
```diff
@@ -32,21 +34,24 @@
 
 
 @pytest.fixture(name="future_engine")
-def future_engine_fixture():
+def future_engine_fixture() -> Generator[FutureEngine, None, None]:
```
hannesdiedrich, line 37: Why is this typing needed? Isn't always a `FutureEngine` object yielded?

hannesdiedrich (commented):
Left some comments. In general, LGTM

tests/strategies/area_agents/test_future_engine.py
spyrostz: Yes, but when we declare types we are obliged to define the return type, and not the yield type. Therefore for generators, the return type is the following https://docs.python.org/3/library/typing.html#typing.Generator , YieldType=FutureEngine, SendType=None (because we do not send anything to the generator), ReturnType=None (because we have no return in the generator function)

src/d3a/models/base.py
spyrostz: Indeed, removing this. Thanks!

src/d3a/constants.py
spyrostz: Fixing, thanks!

src/d3a/models/strategy/future/strategy.py
spyrostz: Because post_producer_first_offer and post_consumer_first_bid have different methods in their respective state classes to read the available energy for sell / buy (get_energy_requirement / get_available_energy instead of get_available_energy_to_buy/sell). Since you mentioned it though I'll try to refactor it to avoid the duplication of the post_first_bid/offer

src/d3a/models/strategy/__init__.py
```diff
@@ -848,11 +853,12 @@ def _get_traded_bids_from_market(self, market_id: str) -> List[Bid]:
...
         return self._traded_bids[market_id]
 
-    def are_bids_posted(self, market_id: str) -> bool:
+    def are_bids_posted(self, market_id: str, time_slot: DateTime = None) -> bool:
         """Checks if any bids have been posted in the market slot with the given ID."""
         if market_id not in self._bids:
             return False
-        return len(self._bids[market_id]) > 0
+        return len([bid for bid in self._bids[market_id]
+                    if time_slot is None or bid.time_slot == time_slot]) > 0
```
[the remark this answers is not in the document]
spyrostz: Fixing

src/d3a/setup/d3a-settings.json
spyrostz: Hmmm, seems like I forgot to create the respective PR in the d3a-interface, doing it now, thanks!

src/d3a/models/strategy/future/strategy.py
```diff
@@ -0,0 +1,253 @@
...
+        Args:
+            initial_buying_rate: Initial rate of the future bids
+            final_buying_rate: Final rate of the future bids
+            initial_selling_rate: Initial rate of the future offers
+            final_selling_rate: Final rate of the future offers
+        """
+        super().__init__()
+
+        self._update_interval = FutureTemplateStrategiesConstants.UPDATE_INTERVAL_MIN
+        self._bid_updater = FutureTemplateStrategyBidUpdater(
```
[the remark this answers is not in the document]
spyrostz: Hmmm, I think this is better dealt with a new Interface class + a factory method. Reason for that is because if I use if / else branching, I would have to repeat this if / else on the event_tick method where the bid/offer updater are updating their clock / energy rates. Let me give it a try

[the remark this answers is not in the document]
spyrostz: Refactored this to use a factory method, that will provide empty implementations depending on the strategy type.

tests/strategies/area_agents/test_future_engine.py
hannesdiedrich: Very interesting! Thanks a lot for explaining and the link!

src/d3a/models/strategy/future/strategy.py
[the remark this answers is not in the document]
spyrostz: I know, I was also in between while implementing this. The only reason I did that was in order to potentially reuse it in template strategies in the future, for instance for the case of the storage strategy on one-sided market (where the bid_updater should not be constructed at all). But yeah, as you said, factory classes have unavoidable boilerplate.

src/d3a/models/strategy/future/strategy.py
hannesdiedrich: I guess the refactoring was not looking better, right?

integration_tests/environment.py
```diff
@@ -70,6 +70,7 @@ def before_scenario(context, scenario):
     context.simdir = "./gsy_e-simulation/integration_tests/"
     os.makedirs(context.simdir, exist_ok=True)
     context.resource_manager = contextlib.ExitStack()
+    GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 0
```
Hassan754, line 73: Shouldn't this be reverted in the `after_scenario`?

src/gsy_e/constants.py
```diff
@@ -15,11 +15,13 @@
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-# Need to import required settings from gsy-framework in order to be available in gsy_e,
-# thus avoiding accessing the gsy-framework constants.
+# Need to import required settings from d3a-interface in order to be available in d3a,
```
Hassan754, line 18:
```suggestion
# Need to import required settings from gsy-framework in order to be available in d3a,
```

src/d3a/models/strategy/future/strategy.py
spyrostz: Darn, I think that github has messed up the commit diffs again and you cannot see the correct changes. The refactoring looked great, you can find it here https://github.com/gridsingularity/gsy-e/pull/1277/files#diff-bf32dee2ce4eaa87404d14ed3015f3a70bf17ee8d4d87a2132ae68cbae7e0793R170 . Essentially I completely removed the prosumer method and called post_producer_first_offer and _post_consumer_first_bid independently

integration_tests/environment.py
spyrostz: Indeed, however after your comment I thought that since we want this setting to be applied globally it should be only in before_all (in order for this to be valid for all the integration tests, essentially to be applied once). I'll move it, thanks!

src/gsy_e/models/strategy/__init__.py
```diff
@@ -524,12 +524,17 @@ def post_first_offer(self, market: "OneSidedMarket", energy_kWh: float,
             energy=energy_kWh,
         )
 
+    def get_posted_offers(self, market: "OneSidedMarket",
+                          time_slot: Optional[DateTime] = None) -> List[Offer]:
+        """Get list of posted bids from a market"""
```
Hassan754, line 529:
```suggestion
        """Get list of posted offers from a market"""
```

src/d3a/models/strategy/future/strategy.py
hannesdiedrich: 👀  weird indeed, looking into it

src/gsy_e/models/strategy/future/strategy.py
```diff
@@ -0,0 +1,258 @@
...
+                required_energy_kWh = strategy.state.get_energy_requirement_Wh(time_slot) / 1000.0
+                self._post_consumer_first_bid(strategy, time_slot, required_energy_kWh)
+            elif strategy.asset_type == AssetType.PRODUCER:
+                available_energy_kWh = strategy.state.get_available_energy_kWh(time_slot)
+                self._post_producer_first_offer(strategy, time_slot, available_energy_kWh)
+            elif strategy.asset_type == AssetType.PROSUMER:
+                available_energy_sell_kWh = strategy.get_available_energy_to_sell_kWh(time_slot)
+                available_energy_buy_kWh = strategy.get_available_energy_to_buy_kWh(time_slot)
+                self._post_producer_first_offer(strategy, time_slot, available_energy_sell_kWh)
+                self._post_consumer_first_bid(strategy, time_slot, available_energy_buy_kWh)
```
hannesdiedrich, line 174: Assigning a new variable is only for better reading of the code, right? Or is there another reason to not pass the `strategy.get_available_energy_to_sell_kWh(time_slot)` and `strategy.get_available_energy_to_buy_kWh(time_slot)` into the `_post_producer_first_offer` and `_post_consumer_first_bid` directly?

src/gsy_e/models/strategy/__init__.py
```diff
@@ -669,6 +674,7 @@ def update_offer_rates(self, market: "OneSidedMarket", updated_rate: float) -> N
                 continue
             try:
                 # Delete the old offer and create a new equivalent one with an updated price
+                time_slot = offer.time_slot or iterated_market.time_slot
```
Hassan754, line 677: shouldn't they both be the same?

src/gsy_e/models/strategy/future/strategy.py
spyrostz: Only for readability purposes. If you prefer it that way I can get rid of the intermediate variables though.

src/gsy_e/models/strategy/future/strategy.py
```diff
@@ -0,0 +1,258 @@
+"""
+Copyright 2018 Grid Singularity
+This file is part of D3A.
```
Hassan754, line 3:
```suggestion
This file is part of Grid Singularity Exchange.
```

tests/area/test_markets.py
```diff
@@ -81,8 +81,10 @@ def test_market_rotation_is_successful_keep_past_markets(self, area_fixture):
         ticks_per_slot = area_fixture.config.slot_length / area_fixture.config.tick_length
         area_fixture.current_tick = ticks_per_slot
         area_fixture.cycle_markets()
+        # The expected future market count is market slots for the whole simulation duration
+        # -2 to not take into account the first 2 market slots due to the cycle_markets call
```
hannesdiedrich, line 85: Sorry, I do not get that. What changed that 
```python
expected_number_of_future_markets = (
                GlobalConfig.future_market_duration / area_fixture.config.slot_length + 1)
```
is not True any more? The number of future markets is not dependant on the sim duration AFAIU.

tests/market/test_future.py
```diff
@@ -43,10 +46,12 @@ def active_future_market() -> FutureMarkets:
             grid_fees=GridFee(grid_fee_percentage=area.grid_fee_percentage,
                               grid_fee_const=area.grid_fee_constant),
             name=area.name)
-    future_market.create_future_markets(DEFAULT_CURRENT_MARKET_SLOT, DEFAULT_SLOT_LENGTH)
+    future_market.create_future_markets(
+        DEFAULT_CURRENT_MARKET_SLOT, DEFAULT_SLOT_LENGTH, area.config)
     yield future_market
 
     GlobalConfig.future_market_duration = orig_future_market_duration
+    GlobalConfig.start_date = orig_start_date
```
hannesdiedrich, line 54: I think you do not need that because you are not overwriting the `GlobalConfig.start_date` earlier  but only the `area.config.start_date`.

src/gsy_e/models/strategy/future/strategy.py
```diff
@@ -0,0 +1,258 @@
...
+                if strategy.are_offers_posted(market.id):
+                    strategy.update_offer_rates(market, self.get_updated_rate(time_slot))
+
+
+class FutureMarketStrategyInterface:
+    """Dummy/empty class that does not provide concrete implementation of the methods.
+    Is needed in order to disable the implementation of the future market strategy
+    when future markets are disabled by configuration."""
+    def __init__(self, *args, **kwargs):
+        pass
```
Hassan754, line 87: Is this really needed?

tests/strategies/test_strategy_pv.py
```diff
@@ -322,11 +334,13 @@ def area_test66():
 
 @pytest.fixture()
 def pv_test66(area_test66):
+    GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 0
```
hannesdiedrich, line 337: Is this for making the tests run faster or in order to not have to populate the `future_markets` member in the area mock/fixture?

tests/area/test_markets.py
spyrostz: Doch, I was also surprised at first however it is true. Think of a simulation with 24 hours duration, with future market horizon of 24 hours. In this case, if we do not limit the future markets in the simulation duration, we would end up with future markets being created up until 2 days from the start of the simulation (because future markets are created on every market slot, at the 23:45 we would create a future market for the next day at 23:45). However this does not make sense, because as long as a user sets the simulation duration to 1 day, he only cares about what happens during this day and not after that, even if future markets perform trades on the next day. 
The -2 is simple to deduce, it is because the first future market (00:00 does not exist for the future markets, so 00:15) has already been recycled by the cycle_market call.

hannesdiedrich (commented):
Some questions, comments. LGTM once addressed.

src/gsy_e/models/strategy/future/strategy.py
hannesdiedrich: ok, then leave it as is.

src/gsy_e/models/strategy/update_frequency.py
```diff
@@ -80,12 +114,17 @@ def delete_past_state_values(self, current_market_time_slot):
...
 
     @staticmethod
-    def get_all_markets(area):
-        return area.all_markets
+    def get_all_markets(area: "Area") -> List["OneSidedMarket"]:
+        """Get list of available markets. Defaults to only the spot market."""
+        return [area.spot_market]
+
+    @staticmethod
+    def get_all_time_slots(area: "Area") -> List[DateTime]:
```
Hassan754, line 122: This probably can be private

tests/area/test_markets.py
hannesdiedrich: Ahhh. yes this is what you meant in a personal chat today. You are referring to this change: right?
https://github.com/gridsingularity/gsy-e/pull/1277/files#diff-9f03ce1c1dbc11b431a78472cf3160e5beea1f89caa9ba8e6ce5b5858b45ad7eR143

Can you please push your gsy-framework branch so that i can see the function `is_time_slot_in_simulation_duration` there too? Thanks!

spyrostz: This is not framework actually, it is in gsy_e.gsy_e_core.util. It is only a range comparison between start/end date + the CN case.

src/gsy_e/models/strategy/__init__.py
spyrostz: For the future market not really, because iterated_market.time_slot is None, and offer.time_slot is the actual future time_slot that the offer has been placed to.

tests/area/test_markets.py
hannesdiedrich: Ah darn, indeed.

src/gsy_e/models/strategy/future/strategy.py
```diff
@@ -0,0 +1,258 @@
...
+        Args:
+            strategy: Strategy object of the asset
+
+        Returns: None
+
+        """
+        if not strategy.area.future_markets:
+            return
+        self._bid_updater.update_and_populate_price_settings(strategy.area)
+        self._offer_updater.update_and_populate_price_settings(strategy.area)
```
Hassan754, line 162: What if the `_bid_updater`, `_offer_updater` are instances of `TemplateStrategyUpdaterInterface`?  This `update_and_populate_price_settings` doesn't exist there AFAIS

src/gsy_e/constants.py
spyrostz: Fixing, thanks!

tests/area/test_markets.py
hannesdiedrich: I leave this comment unresolved for others to read along. But for me it is resolved. :-)

src/gsy_e/models/strategy/__init__.py
spyrostz: Fixing

src/gsy_e/models/strategy/future/strategy.py
spyrostz: Fixing

src/gsy_e/models/strategy/future/strategy.py
spyrostz: I think it is a remnant of the implementation, will give it a try without it, thanks!

tests/market/test_future.py
spyrostz: Unfortunately if I do not revert it followup tests fail. Looks like the reference is updated and not the value of the start_date.

hannesdiedrich: ok.

src/gsy_e/models/strategy/future/strategy.py
spyrostz: This is exactly the point of constructing the TemplateStrategyUpdaterInterface . Essentially the factory method is used in order to disable the implementation of this class for some asset types, therefore to still call the methods but do nothing at all. For asset types that need these, the factory method will deal with creating not the interface but the actual updater class.

tests/strategies/future/test_future_strategy.py
```diff
@@ -0,0 +1,165 @@
...
+        future_strategy_fixture.area.current_tick = 0
+        future_strategy_fixture.area.config = Mock()
+        future_strategy_fixture.area.config.tick_length = duration(seconds=15)
+
+    def test_event_market_cycle_posts_bids_load(self) -> None:
+        """Validate that market cycle event posts bids and offers for the load strategy"""
+        future_strategy_fixture = LoadHoursStrategy(100)
+        future_strategy = FutureMarketStrategy(future_strategy_fixture.asset_type, 10, 50, 50, 20)
+        self._setup_strategy_fixture(future_strategy_fixture)
+        if isinstance(future_strategy_fixture, LoadHoursStrategy):
```
Hassan754, line 72: would this be False at all?

src/gsy_e/models/strategy/future/strategy.py
Hassan754: But I don't see where we're defining the `update_and_populate_price_settings` for example in the `TemplateStrategyUpdaterInterface`, so I would not predict that it will do nothing, I would predict that it will break for calling up a method that doesn't exist
Thanks for bearing with me, I could have missed it

tests/strategies/future/test_future_strategy.py
```diff
@@ -0,0 +1,165 @@
...
+        future_strategy_fixture.update_offer_rates = Mock()
+        future_strategy_fixture.area = Mock()
+        future_strategy_fixture.area.future_markets = self.future_markets
+        future_strategy_fixture.area.current_tick = 0
+        future_strategy_fixture.area.config = Mock()
+        future_strategy_fixture.area.config.tick_length = duration(seconds=15)
+
+    def test_event_market_cycle_posts_bids_load(self) -> None:
+        """Validate that market cycle event posts bids and offers for the load strategy"""
+        future_strategy_fixture = LoadHoursStrategy(100)
```
Hassan754, line 69: The name is a bit misleading, It could mean that this instance is a `FutureMarketStrategy` instance when it's not; And reviewing becomes confusing when we have `future_strategy_fixture` and `future_strategy`

Hassan754 (approved):
LGTM

src/gsy_e/models/strategy/__init__.py
Hassan754: Ah you're right, I always miss that
Thanks!

src/gsy_e/models/strategy/update_frequency.py
spyrostz: Will do so thanks!

src/gsy_e/models/strategy/future/strategy.py
spyrostz: There is probably something wrong with the diff for you too like with Hannes before. It is the first method of the TemplateStrategyUpdaterInterface https://github.com/gridsingularity/gsy-e/pull/1277/files/05a66ff4e693d529f6dcb0b9d936ad1929cc6412#diff-b9af14d73fabe24d5f22776a3741af78f0a06e27159aa95519bd6e82304cc336R38

tests/strategies/future/test_future_strategy.py
spyrostz: Good catch, this is a remnant of a parameterized testcase. Removing this

tests/strategies/future/test_future_strategy.py
spyrostz: Indeed, therefore I would go with the rename:
future_strategy_fixture -> load_Strategy_fixture (and pv_strategy_fixture in other tests)

tests/strategies/test_strategy_pv.py
spyrostz: The reason why I did that specifically for this test was because this test was actually reading profiles outside the range of one day. If you check the test implementation, you will see that the goal of this is to check that the energy production of the PV during the night is lower than the production during the night. Unfortunately this is done with this check:
`elif time > afternoon_time:`
which takes into account timestamps from the next day (because taking into account the future markets the profile is 2 days long). Quick fix was to completely disregard the future markets in this test (because this is not the goal of the test at the very end)

faizan2590 (approved):
LGTM

src/gsy_e/models/strategy/future/strategy.py
Hassan754: Weird indeed, I can now see it but for some reason these methods were not shown yesterday, good then

tests/strategies/future/test_future_strategy.py
Hassan754: I'm not sure if the differences are not clearly shown but AFAIS this check is still here and in other parts of the tests

Hassan754: Ah forget it, I can now see that it was fixed, Thanks!

Hassan754 (approved):
LGTM

tests/strategies/future/test_future_strategy.py
spyrostz: Github has started to behave erratically after all these changes :D Thanks!

PR #1278, opened 2021-11-02
D3ASIM-3662: Fix the market slot logging

Hassan754:


spyrostz (approved):
LGTM

fievelk (approved):

fievelk:
@Hassan754 it would be great if you could add a unit test to check that we don't break this in the future.

Hassan754:
> @Hassan754 it would be great if you could add a unit test to check that we don't break this in the future.

I share with you same concern, however It looks that the `simulation.py` module needs to be refactored soon and the module itself only has 1 function tested, which seems risky for such high level module
I suggest creating a ticket dedicated to test and refactor this module, and in any case if you believe that I should invest some time writing a unittest for the fix introduced even now, let me know, otherwise let me create a dedicated ticket

fievelk:
> > @Hassan754 it would be great if you could add a unit test to check that we don't break this in the future.
> 
> I share with you same concern, however It looks that the `simulation.py` module needs to be refactored soon and the module itself only has 1 function tested, which seems risky for such high level module I suggest creating a ticket dedicated to test and refactor this module, and in any case if you believe that I should invest some time writing a unittest for the fix introduced even now, let me know, otherwise let me create a dedicated ticket

In general I would find it better to add tests in the context of the PR. However, you're right when you say that this module needs some refactoring; trying to add tests now would take a long time. So your suggested solution works for me in this case.

Hassan754:
> > > @Hassan754 it would be great if you could add a unit test to check that we don't break this in the future.
> > 
> > 
> > I share with you same concern, however It looks that the `simulation.py` module needs to be refactored soon and the module itself only has 1 function tested, which seems risky for such high level module I suggest creating a ticket dedicated to test and refactor this module, and in any case if you believe that I should invest some time writing a unittest for the fix introduced even now, let me know, otherwise let me create a dedicated ticket
> 
> In general I would find it better to add tests in the context of the PR. However, you're right when you say that this module needs some refactoring; trying to add tests now would take a long time. So your suggested solution works for me in this case.

Thanks a lot, you can find the ticket [here](https://gridsingularity.atlassian.net/browse/D3ASIM-3684), I already prioritized it to be included in the upcoming 2-3 weeks sprints

PR #1279, opened 2021-11-03
Feature/d3 asim 3669: Rebranding

faizan2590:


.coveragerc
```diff
@@ -1,18 +1,18 @@
 [run]
 branch = True
-source = d3a
+source = gsy_e
```
spyrostz, line 3: I think that the _ is not needed. You can use `gsye` as is in order to be more concise, while at the same time the repository name can be gsy-e . That way the "ugly" _ will be avoided throughout the source code.

faizan2590: I used the same convention as for other repos like `d3a-api-client` is the repo name and the main package is names `d3a_api_client`. Though I am okay with your suggestion too

spyrostz: after offline discussion, leave it as is

PR #1280, opened 2021-11-03
Feature/d3 asim 3675

spyrostz:
- Added a dictionary on the unsettled energy deviation as part of the asset_info dict that is reported via the client every market cycle. 
- Fixed bug where the energy deviation was not populated for the LoadExternalMixin and PVExternalMixin but only for the Forecast-related Mixins


src/d3a/models/strategy/settlement/strategy.py
```diff
@@ -200,6 +207,15 @@ def event_offer_traded(self, strategy: BidEnabledStrategy,
             strategy.state.decrement_unsettled_deviation(
                 trade.offer_bid.energy, market.time_slot)
 
+    def get_unsettled_deviation_dict(self, strategy: "BidEnabledStrategy") -> Dict:
+        return {
+            "unsettled_deviation_kWh": {
+                format_datetime(time_slot): strategy.state.get_unsettled_deviation_kWh(
```
hannesdiedrich, line 213: As this is d3a-api-client user facing, i suggest to use the `DATE_TIME_FORMAT` instead of the UI format. Thanks!

hannesdiedrich (commented):
One comment, once addressed, LGTM

src/d3a/models/strategy/settlement/strategy.py
spyrostz: will change it thanks!

hannesdiedrich (dismissed):
LGTM

.pylintrc
```diff
@@ -145,7 +145,9 @@ disable=invalid-name,
         deprecated-sys-function,
         exception-escape,
         comprehension-escape,
-        missing-module-docstring
+        missing-module-docstring,
+        too-many-ancestors,
```
fievelk, line 149: I think the `too-many-ancestors` warning is useful for new code and shouldn't be completely disabled.

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -300,8 +298,9 @@ def _get_market_from_time_slot(self, time_slot: DateTime) -> Market:
             return market
         market = self.area.get_settlement_market(time_slot)
         if not market:
-            raise Exception(f"Timeslot {time_slot} is not currently in the spot, future or "
-                            f"settlement markets")
+            raise MarketException(
+                f"Timeslot {time_slot} is not currently in the spot, future or "
+                f"settlement markets")
```
Hassan754, line 303: Nit
```suggestion
                "settlement markets")
```

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -317,7 +316,9 @@ def redis(self) -> "ExternalConnectionCommunicator":
     @property
     def _device_info_dict(self) -> Dict:
         """Return the asset info."""
-        return {}
+        return {
+            **self._settlement_market_strategy.get_unsettled_deviation_dict(self)
```
Hassan754, line 320: What is expected to happen when a mixed in class doesn't define this `_settlement_market_strategy` ?

Hassan754 (approved):
LGTM

src/d3a/models/strategy/external_strategies/__init__.py
spyrostz: Exception. However, as discussed offline, this is how mixins work, and that's why we use them only in very limited context (essentially only for classes that inherit from BaseStrategy / BidEnabledStrategy, therefore we know that this member will be part of the concrete class).

.pylintrc
spyrostz: I disagree on that. Class ancestry is not some nit or small part of the code that can be overlooked (e.g. too-many-arguments) but it is one of the first things that will be checked by the class maintainer. Therefore, if this error is raised, it  is almost guaranteed that it is deliberate and with full knowledge of the implementer, and also extremely explicit to the PR reviewer (I doubt that this will go unnoticed in a PR). Therefore no need for a linter to enforce a rule on that, and litter all mixin classes with the disable-pylint statement.

fievelk: I still disagree on this, to be honest. I don't think that all of our current `too-many-ancestors` warnings are caused by deliberate choices that couldn't have been avoided.
Overlooking parts of the code is something that happens constantly when we're in a rush; using a linter is also a way to tell the developer to avoid coding in a hurry (or to add comments to notify readers when this is not possible). Disabling linter warnings on a module basis is ok for me, if I have to choose between the two situations.
So I personally still vote to keep this warning enabled. I'll leave it to the other reviewers anyway, as they might have different opinions.

spyrostz: Still not persuaded by the comment. The only occurrences of too-many-ancestors that I have seen in the codebase are the Strategies / ExternalMixin (and I am also not sure about the Market classes, but I doubt that we have reached the limit yet), which is a design decision to implement it this way, therefore will not change. If you find another such occurrence then we can talk more about it, because personally I have not seen any.

spyrostz: Warnings that pop up only in cases that we know are OK (which IMO is one of them, as I said this warning is only in one place in the codebase which we know is ok, and if new code is added that violates this we will definitely notice in the PR and make sure that it makes sense or not) are not really helpful, and will only end up polluting our codebase with pylint: disable with no benefit at all. 

In addition to having no benefit of leaving it there, this is contrary to the main goal of Pylint, which is to tailor it to your codebase's needs (that's why it is feasible to disable its warnings in so many ways after all). Essentially Pylint's goal is to not treat all its warnings as errors, but to explicitly select which ones should be treated as errors and which should not according to the codebase. 

BTW this comment is not only specific to this warning. IMO it probably makes sense to disable other warning types too, either because we disagree with them, or because they do not fit to our codebase (for instance, one such candidate is too-broad-exception, but I will leave this discussion for another comment :D )

src/d3a/models/strategy/external_strategies/__init__.py
```diff
@@ -546,5 +547,6 @@ def market_info_dict(self) -> Dict:
                     if self.area.current_market else None,
                     "total_cost": self.energy_traded_costs(self.area.current_market.id)
                     if self.area.current_market else None},
-                "asset_bill": self.device.stats.aggregated_stats.get("bills")
+                "asset_bill": self.device.stats.aggregated_stats.get("bills"),
+
```
faizan2590, line 551: Is it also due to pylint?

faizan2590 (dismissed):
LGTM

src/d3a/models/strategy/external_strategies/__init__.py
spyrostz: Unfortunately ended up here during the implementation, Pylint does not complain about that. Should not hurt to have here though

Hassan754 (approved):

src/d3a/models/strategy/__init__.py
```diff
@@ -635,10 +644,15 @@ def can_settlement_offer_be_posted(
             offer_energy, offer_price, unsettled_energy_kWh, market,
             replace_existing=replace_existing)
 
+    @property
+    def spot_market(self) -> "Market":
+        """Return the spot_market member of the area."""
+        return self.area.spot_market
```
fievelk, line 650: Nit: I favour the idea of adding this property. It would be great if we could adapt the rest of the codebase (replacing the old `area.spot_market` occurrences with the new property).

src/d3a/models/strategy/__init__.py
```diff
@@ -451,9 +454,14 @@ def __init__(self):
         self._allowed_disable_events = [AreaEvent.ACTIVATE, MarketEvent.OFFER_TRADED]
         self.event_responses = []
         self._market_adapter = market_strategy_connection_adapter_factory()
+        self._settlement_market_strategy = self._create_settlement_market_strategy()
 
     parameters = None
 
+    @classmethod
+    def _create_settlement_market_strategy(cls):
+        return SettlementMarketStrategyInterface()
```
fievelk, line 463: What would happen if we call `settlement_market_strategy_factory` here as well? Wouldn't it be working the same way?

tests/strategies/settlement/test_settlement_strategy.py
```diff
@@ -154,3 +154,17 @@ def test_event_bid_trade_updates_energy_deviation(self, strategy_fixture):
...
         assert strategy_fixture.state.get_unsettled_deviation_kWh(self.time_slot) == 14
+
+    @pytest.mark.parametrize(
+        "strategy_fixture", [LoadHoursStrategy(100), PVStrategy()])
+    def test_get_unsettled_deviation_dict(self, strategy_fixture):
+        self._setup_strategy_fixture(strategy_fixture, True, False)
+        strategy_fixture.state.set_energy_measurement_kWh(15, self.time_slot)
+        unsettled_deviation_dict = self.settlement_strategy.get_unsettled_deviation_dict(
+            strategy_fixture)
+        from d3a_interface.utils import format_datetime
```
fievelk, line 165: Nit: this can be moved on top

fievelk (approved):
Some minor questions, LGTM.

src/d3a/models/strategy/__init__.py
spyrostz: I'll merge it now and open a separate PR for this and the other comment. Heads up though, I will disable pylint for this

PR #1289, opened 2021-11-09
Timezone integration tests

spyrostz:
Solved the failing "Strategies follow yearly user profile saved in the DB" integration test, which could be reproduced only in local environments and not on CI nor production. The root cause of the bug was that the profile timeseries that are stored in the database were timezone-aware, therefore there are conversion methods that translate any timezone in the DB to UTC. However, the default Python timezone methods (`astimezone`) do not take into account DST, therefore in any environment with a local timezone with DST, the integration tests were failing. CI and production remained unaffected because they were running on UTC time anyways. 
Solution was to move from Python timezone methods to pendulum timezone methods, which solved the problem and translated the timezones taking into account DST as expected.

### DIFF
--- a/integration_tests/steps/storage_test.py
+++ b/integration_tests/steps/storage_test.py
@@ -20,7 +20,6 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 # pylint: disable=no-name-in-module
 from behave import then
 from d3a_interface.read_user_profile import read_arbitrary_profile, InputProfileTypes
-from pendulum import duration
 
 from d3a.constants import DEFAULT_PRECISION
 from d3a.setup.strategy_tests.storage_strategy_break_even_hourly import (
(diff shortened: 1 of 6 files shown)


integration_tests/steps/storage_test.py
```diff
@@ -61,22 +60,20 @@ def step_impl(context):
...
-                expected_time_stamp = start_time_stamp + duration(hours=trade.creation_time.hour)
                 if trade.seller == name:
                     assert (round(trade.offer_bid.energy_rate, 2) >=
-                            round(final_selling_rate[expected_time_stamp], 2))
+                            round(final_selling_rate[market.time_slot], 2))
                     trades_sold.append(trade)
                 elif trade.buyer == name:
                     assert (round(trade.offer_bid.energy_rate, 2) <=
-                            round(final_buying_rate[expected_time_stamp], 2))
+                            round(final_buying_rate[market.time_slot], 2))
```
Hassan754, line 76: 🥇

Hassan754 (approved):

fievelk (approved):

PR #1291, opened 2021-11-10
Feature/D3ASIM-3678: Refactor the market external connection module

Hassan754:
This PR can be reviewed commit by commit, the 3 included commits deal with:

1. Fix the majority of pylint warnings + minor structural improvements
2. The actual fix of the registration operation
3. Deal with the TODO comment in the set_grid_fees callback
4. Test the module


src/gsy_e/models/area/__init__.py
```diff
@@ -352,10 +351,6 @@ def cycle_markets(self, _trigger_event=True, _market_cycle=False, deactivate=Fal
         if deactivate:
             return
 
-        if self.should_update_child_strategies is True:
```
Hassan754, line 355: I removed this check because it looks like the sole reason behind its existence is to update child strategies after updating the parent one in the set_grid_fees callback of the redis external market
As I have refactored the mentioned method to use the `area_reconfigure_event` now, there was no need for the `should_update_child_strategies` flag anymore

src/d3a/models/area/redis_external_market_connection.py
```diff
@@ -104,47 +103,36 @@ def sub_to_external_channels(self) -> None:
...
         grid_fees_response_channel = f"{self.channel_prefix}/response/grid_fees"
         payload_data = (
             payload["data"] if isinstance(payload["data"], dict)
             else json.loads(payload["data"]))
         try:
-            validate_area(grid_fee_percentage=payload_data.get("fee_percent", None),
-                          grid_fee_constant=payload_data.get("fee_const", None))
+            self.area.area_reconfigure_event(
+                grid_fee_constant=payload_data.get("fee_percent", None),
+                grid_fee_percentage=payload_data.get("fee_const", None))
```
hannesdiedrich: You are setting the grid-fees before you have validated them.
The comment in line 107 was reffering to replace 
```python
self.area.grid_fee_percentage = payload_data["fee_percent"]
```
and
```python
self.area.grid_fee_constant = payload_data["fee_const"]
```
with 
```python
self.area.area_reconfigure_event(
                grid_fee_constant=payload_data.get("fee_percent", None),
                grid_fee_percentage=payload_data.get("fee_const", None))
```
The validation here and below should stay the sam (could be refactored of course), though.
BTW, this method can/should be unittested for exactly the validation (especially against the GlobalConfigFeeType).

hannesdiedrich (commented):
One important comment, will review again once resolved.

src/d3a/models/area/redis_external_market_connection.py
Hassan754: @hannesdiedrich I have already created dedicated unit tests to test the behavior upon submitting wrong GlobalConfigFeeType , you can check the `test_set_grid_fees_callback_conflicting_grid_fee_type_config` and `test_set_grid_fees_callback_valid_grid_fee_type_config`.

Furthermore, per your first point regarding the validation of the fees, the validation already exists in the area's [area_reconfigure_event](https://github.com/gridsingularity/gsy-e/blob/832a3799bd323dcb1abe5a783b234744b61dcb0c/src/gsy_e/models/area/__init__.py#L211) method, this is why I removed the redundant `validate_area` call from here.

Hassan754:
PS. The integration tests are passing on gsy-e-sdk

src/d3a/models/area/redis_external_market_connection.py
hannesdiedrich: Darn, I was trapped inside one commit. Reviewed it now. Thanks a lot for the extensive tests!

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #1292, opened 2021-11-10
Feature/d3 asim 3669: Rebranding

faizan2590:


spyrostz (dismissed):
LGTM

.gitignore
```diff
@@ -1,12 +1,12 @@
...
 dump.rdb
 
-# Byte-compiled / optimized / DLL files
+# Byte-compiled / optimized / DLL files.github/workflows/gsy-e_ci.yaml
 __pycache__/
 *.py[cod]
 *$py.class
 
-# JetBeans PyCharm files
+# JetBeans PyCharm files.github/workflows/gsy-e_ci.yaml
```
mclrch, line 9:
```suggestion
# JetBeans PyCharm files
```

mclrch (approved):
Two minor fixes
LGTM

.gitignore
faizan2590: Thanks, got added due to quick global searches

.gitignore
```diff
@@ -92,19 +92,19 @@ ENV/
 .ropeproject
 
 *.DS_Store
-.d3a/
+.gsy-e/
```
fievelk, line 95: Are we using `gsy-e` or `gsy_e` for folders? For example, you're using `gsy_e` below.

faizan2590: `gsy-e`is the repo name as well as cli commands

`gsy_e` is the main package

spyrostz: As far as I see this can be deleted though. None of our software creates a .d3a/ folder, and we do not have a specific purpose for this. I would recommend to get rid of this line, either before or after the merge of this PR. 
FYI, gsy_e would have the effect of git to ignore all our codebase :D

faizan2590: Will adapt later, in order to not delay anymore

fievelk:
One general comment:
- there are several occurrences of `gsy_e-interface` and `gsy_e-web` that should be refactored to `gsy-framework` and `gsy-web`. They mostly (probably always) occur in strings or comments. I only commented a few of them but there are more.

Besides this, I left some specific comments.

PR #1301, opened 2021-11-11
 Fix future markets traversal in myco external client

Hassan754:


src/gsy_e/models/myco_matcher/myco_external_matcher.py
```diff
@@ -142,7 +145,7 @@ def match_recommendations(self, **kwargs) -> None:
         self._recommendations = []
         self.myco_ext_conn.publish_json(channel, response_data)
 
-    def publish_simulation_id(self, message):
+    def publish_simulation_id(self, _):
```
spyrostz, line 148: I assume this is because you want to conform to the Interface class method arguments, but Pylint complains, right ? Since you replaced it with _ already it is okay, but in general I would prefer to disable the `unused-argument` Pylint warning for the whole module. Nothing to do now, just a note for future occurrences in order to avoid the hassle (and also to be able to see the available argument names in the method definition instead of _).

spyrostz (approved):
LGTM

src/gsy_e/models/myco_matcher/myco_external_matcher.py
Hassan754: Right, this was to silent pylint
I thought about disabling this warning but I suppose that we should avoid disabling this specific warning for the whole module and should only disable it for each use case, I'll do this in the future, Thanks!

fievelk: @spyrostz @Hassan754 I would suggest that we disable warnings on a module level only when there are several of them (because it could mean that the module was designed to be in that way). If it's just one or two lines I'd rather disable them only on those lines.

spyrostz: I would not hesitate to do this even for the whole module actually. This is one of the warnings that are very easy to track because Pycharm / VSCode linters consistently show these errors, and one has to be very careless to ignore it. Matter of preference though, both options will work fine for me.

spyrostz: @fievelk this is usually the case with interfaces though. In the typical usecase that an interface inherits from an ABC and provides empty implementation in its methods (in order to be compatible to the factory design pattern), all the Interface methods will report this warning. Therefore instead of polluting the source code with multiple `# pylint: disable` statements it is much cleaner to disable it in the whole module.

fievelk: I personally find it useful; I don't work with PyCharm, so this might depend on the IDE in use.

Besides the IDE, I think that the warning makes sense in most situations. The parameters defined in an interface should be respected in all the classes that implement that interface. If this doesn't happen, the subclasses are slightly different than what the interface dictates, and they're implementing a different interface. I'm definitely ok with accepting this in some cases (especially existing code), but I think that in general the pylint warning makes sense.

fievelk (approved):

PR #1303, opened 2021-11-12
D3ASIM-3553: Rename Area to Market

fievelk:
This is the first PR that handles the renaming of "Area" into "Market". This name should be used in every part of the code that affects either researchers or UI users. Internally, the `Area` class can be preserved because it represents a different concept (since it represents both markets and assets).

### DIFF
--- a/src/gsy_e/models/area/__init__.py
+++ b/src/gsy_e/models/area/__init__.py
@@ -237,18 +237,19 @@ class Area:
 
         self._set_grid_fees(grid_fee_constant, grid_fee_percentage)
         self.throughput = throughput
-        self._update_descendants_strategy_prices()
+        self.update_descendants_strategy_prices()
 
         return None
 
-    def _update_descendants_strategy_prices(self):
+    def update_descendants_strategy_prices(self):
+        """Recursively update the strategy prices of all descendants of the area."""
         try:
             if self.strategy is not None:
                 self.strategy.event_activate_price()
             for child in self.children:
-                child._update_descendants_strategy_prices()
+                child.update_descendants_strategy_prices()
         except Exception:
-            log.exception("area._update_descendants_strategy_prices failed.")
+            log.exception("area.update_descendants_strategy_prices failed.")
             return
 
     def _set_grid_fees(self, grid_fee_const, grid_fee_percentage):
--- a/src/gsy_e/models/area/__init__.py
+++ b/src/gsy_e/models/area/__init__.py
@@ -102,6 +102,10 @@ class AreaChildrenList(list):
 # pylint: disable=too-many-instance-attributes
 # pylint: disable=too-many-public-methods
 class Area:
+    """Generic class to define both market areas and devices.
+
+    Important: this class should not be used in setup files. Please use Market or Asset instead.
+    """
 
     # pylint: disable=too-many-arguments
     def __init__(self, name: str = None, children: List["Area"] = None,
@@ -668,3 +672,17 @@ class Area:
     def is_market_future(self, market_id: str) -> bool:
         """Return True if market_id belongs to a FUTURE market."""
         return market_id == self.future_markets.id
+
+
+class Market(Area):
+    """Class to define geographical market areas that can contain children (areas or assets)."""
+
+
+class Asset(Area):
+    """Class to define assets (devices). These instances cannot contain children."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if self.children:
+            raise ValueError(f"{self.__class__.__name__} instances can't have children.")
--- a/src/gsy_e/models/area/__init__.py
+++ b/src/gsy_e/models/area/__init__.py
@@ -16,18 +16,18 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from logging import getLogger
-from typing import List, Dict, Optional, TYPE_CHECKING
+from typing import List, Dict, Optional, Union, TYPE_CHECKING
 from uuid import uuid4
 
 from gsy_framework.area_validator import validate_area
 from gsy_framework.constants_limits import ConstSettings, GlobalConfig
 from gsy_framework.enums import SpotMarketTypeEnum
+from gsy_framework.exceptions import GSyAreaException, GSyDeviceException
 from gsy_framework.utils import key_in_dict_and_not_none
 from pendulum import DateTime, duration, today
 from slugify import slugify
 
 import gsy_e.constants
-from cached_property import cached_property
 from gsy_e.gsy_e_core.blockchain_interface import blockchain_interface_factory
 from gsy_e.gsy_e_core.device_registry import DeviceRegistry
 from gsy_e.gsy_e_core.exceptions import AreaException
@@ -40,17 +40,18 @@ from gsy_e.models.area.redis_external_market_connection import RedisMarketExtern
 from gsy_e.models.area.stats import AreaStats
 from gsy_e.models.area.throughput_parameters import ThroughputParameters
 from gsy_e.models.config import SimulationConfig
+from gsy_e.models.market.future import FutureMarkets
 from gsy_e.models.market.market_structures import AvailableMarketTypes
 from gsy_e.models.strategy import BaseStrategy
 from gsy_e.models.strategy.external_strategies import ExternalMixin
-from gsy_e.models.market.future import FutureMarkets
+
 log = getLogger(__name__)
 
-
 if TYPE_CHECKING:
-    from d3a.models.market import Market
+    from d3a.models.market import MarketBase
 
 
+# pylint: disable=fixme
 # TODO: As this is only used in the unittests, please remove it here and replace the usages
 #       of this class with gsy-framework.constants_limits.GlobalConfig class:
 DEFAULT_CONFIG = SimulationConfig(
@@ -79,6 +80,8 @@ def check_area_name_exists_in_parent_area(parent_area, name):
 
 
 class AreaChildrenList(list):
+    """Class to define the children of an area."""
+
     def __init__(self, parent_area, *args, **kwargs):
         self.parent_area = parent_area
         super().__init__(*args, **kwargs)
@@ -96,15 +99,18 @@ class AreaChildrenList(list):
         super().insert(index, item)
 
 
+# pylint: disable=too-many-instance-attributes
+# pylint: disable=too-many-public-methods
 class Area:
 
+    # pylint: disable=too-many-arguments
     def __init__(self, name: str = None, children: List["Area"] = None,
                  uuid: str = None,
                  strategy: BaseStrategy = None,
                  config: SimulationConfig = None,
                  budget_keeper=None,
                  balancing_spot_trade_ratio=ConstSettings.BalancingSettings.SPOT_TRADE_RATIO,
-                 event_list=[],
+                 event_list=None,
                  grid_fee_percentage: float = None,
                  grid_fee_constant: float = None,
                  external_connection_available: bool = False,
@@ -130,6 +136,7 @@ class Area:
             raise AreaException("A leaf area can not have children.")
         self.strategy = strategy
         self._config = config
+        event_list = event_list if event_list is not None else []
         self.events = Events(event_list, self)
         self.budget_keeper = budget_keeper
         if budget_keeper:
@@ -148,6 +155,7 @@ class Area:
 
     @property
     def name(self):
+        """Return the name of the area."""
         return self.__name
 
     @name.setter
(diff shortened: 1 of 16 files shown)


src/gsy_e/models/area/__init__.py
```diff
@@ -646,3 +672,17 @@ def is_market_balancing(self, market_id: str) -> bool:
...
+
+
+class Asset(Area):
+    """Class to define assets (devices). These instances cannot contain children."""
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        if self.children:
+            raise ValueError(f"{self.__class__.__name__} instances can't have children.")
```
fievelk, line 688: This is the core part of the renaming. We create a new `Market` class that has the same functionality as the current `Area` class. Moreover, the `Asset` class is basically the same but it can't be instantiated with children.

Note: in theory, it's possible to create an instance of `Asset` without children and later assign them with `asset.childer = ...`. I tried to create setters for it but it would get messy. If you want I can work on it, but it'll need some refactoring.

src/gsy_e/models/area/__init__.py
```diff
@@ -221,22 +235,25 @@ def area_reconfigure_event(self, **kwargs):
...
 
         self._set_grid_fees(grid_fee_constant, grid_fee_percentage)
         self.throughput = throughput
-        self._update_descendants_strategy_prices()
+        self.update_descendants_strategy_prices()
+
+        return None
 
-    def _update_descendants_strategy_prices(self):
+    def update_descendants_strategy_prices(self):
```
spyrostz, line 248: Why changing this method to public ? From a quick search this is not called from anywhere else in the code, therefore it makes sense to stay private.

src/gsy_e/models/area/__init__.py
```diff
@@ -221,22 +235,25 @@ def area_reconfigure_event(self, **kwargs):
...
+        """Recursively update the strategy prices of all descendants of the area."""
         try:
             if self.strategy is not None:
                 self.strategy.event_activate_price()
             for child in self.children:
-                child._update_descendants_strategy_prices()
+                child.update_descendants_strategy_prices()
         except Exception:
-            log.exception("area._update_descendants_strategy_prices failed.")
+            log.exception("area.update_descendants_strategy_prices failed.")
```
spyrostz, line 256: If this is the reason for the change from private to public method, it would be much better to change the exception log instead to something like: `Failed to update strategy prices for the descendants of the area.`

src/gsy_e/models/area/__init__.py
spyrostz, line 688: Just for future reviewers' clarification, I assume that `self.__class__.__name__` is here in order to support any future renames / rebranding of the class, right ?

tests/area/test_area.py
```diff
@@ -46,88 +62,72 @@ def setup_method(self):
...
-                               self.config, None, grid_fee_percentage=1)
-        self.area_child.parent = self.area
-        self.area.children = [self.area_child]
-        self.area.grid_fee_percentage = 1
-        self.dispatcher = AreaDispatcher(self.area)
-        self.stats = AreaStats(self.area._markets, self.area)
-
-    def teardown_method(self):
+    @staticmethod
+    def teardown_method():
```
spyrostz, line 66: This will work with pytest, however beware when stumbling upon existing unittest.TestCase tests. In the latter, setup and teardown methods have to be normal methods and not static methods

spyrostz (approved):
One comment, LGTM though

src/gsy_e/models/area/__init__.py
hannesdiedrich: Leaving it like you propose works for me. If a user tries to add a children to an Asset, it is his/her own fold IMO.
One minor thing though, how about first checking for the children and only then call the `super().__init__()` ?

tests/area/test_area.py
```diff
@@ -46,88 +62,72 @@ def setup_method(self):
...
-        self.area._markets.create_new_spot_market(
-            current_time, AvailableMarketTypes.SPOT, self.area)
-        current_time = today(tz=constants.TIME_ZONE).add(minutes=2*self.config.slot_length.minutes)
-        self.area._markets.rotate_markets(current_time)
-        assert len(self.area.past_markets) == 1
-        assert (list(self.area.past_markets)[-1].time_slot ==
-                today(tz=constants.TIME_ZONE).add(minutes=self.config.slot_length.minutes))
-
-    def test_keep_past_markets(self):
+        area = Area(name="Street", children=[Area(name="House")],
```
hannesdiedrich, line 86: A general question: Internally in the code we will use `MarketBase` and `Area` exchangeably right?

hannesdiedrich: I am asking because of the serialising of configurations from the gsy-web side:
https://github.com/gridsingularity/gsy-e/blob/03f8ea4240b8b6678394be787b4a18c1b2a1b130/src/gsy_e/gsy_e_core/area_serializer.py#L136
Where we could also make the code a bit more explicit. 
IMO, no need to change all tests like here though.

hannesdiedrich (commented):
Two comments. LGTM in general

src/gsy_e/models/area/__init__.py
```diff
@@ -130,6 +140,7 @@ def __init__(self, name: str = None, children: List["Area"] = None,
             raise AreaException("A leaf area can not have children.")
         self.strategy = strategy
         self._config = config
+        event_list = event_list if event_list is not None else []
```
Hassan754, line 143: Nit
```suggestion
        event_list = event_list or []
```

src/gsy_e/models/area/__init__.py
```diff
@@ -221,22 +235,25 @@ def area_reconfigure_event(self, **kwargs):
                             export_capacity_kVA=export_capacity_kVA
                         )
 
-        except Exception as ex:
+        except (GSyAreaException, GSyDeviceException) as ex:
```
Hassan754, line 238: What about these exceptions, won't we have to rename them or did you leave them as a follow up PR?

Hassan754 (approved):
LGTM

tests/area/test_area.py
hannesdiedrich: Can you please either adapt the `area_serializer.py` or explain me, why it does make sense? Thanks!

src/gsy_e/models/area/__init__.py
hannesdiedrich: > One minor thing though, how about first checking for the children and only then call the `super().__init__()` ?
Can you please do this? IMO this saves the "effort" of initialisation of the super class if the user accidentally provided a list of children.

hannesdiedrich: It was discussed offline that it only makes sense now. Basically: `self.children` is not populated  before the `super().__init__()`  is called.

tests/area/test_area.py
hannesdiedrich: Ok, I see. But we also have only-backend-users, too. Do I get it right: They will use the `Market` class from now on (as the new wiki will show) and old user-setups will work because we still support `Area`, right? 
The only point were I predict confusion is when the users look into our default/test setup files which are not in sync with the new wiki. But I guess  our costumer service can explain then :-)

fievelk: @hannesdiedrich Yes, this was discussed on slack a few days ago. Actually, that conversation didn't conclude and I worked on the other changes in the meanwhile. 
I'd personally update the existing setups with the new syntax. If you want me to amend all existing setups I can work on it (maybe in a new PR since there are lots of setups).

src/gsy_e/models/area/__init__.py
Hassan754: This is right, and I feel that it would minimize the chances of us passing ANY non None object and it ending up being assigned to the event list, so imo it makes sense to adapt the proposal

src/gsy_e/models/area/__init__.py
Hassan754: Umm I'm okay with sticking to the requirements but I prefer to also get rid of these terms completely to not create confusions in the future (even for us current developers)
No need to do it in context of this PR, but I vote for doing it as a follow up one

src/gsy_e/models/market/__init__.py
```diff
@@ -59,7 +59,7 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
-class Market:
+class MarketBase:
```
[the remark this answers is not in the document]
Hassan754: Although both `Market` and `MarketBase` are correct terms, in my opinion we should distinguish them (I personally would still be confused between them)
I propose to name the market which we trade in `EnergyMarket` or something else

src/gsy_e/models/area/__init__.py
fievelk: If the user passes a non-expected None object, I think it's good to let the execution crash. Also, `if param is not None else []` is just the safe equivalent of `= []`, which is what I wanted to fix. If that's ok for you, I would keep this as it is.

src/gsy_e/models/area/__init__.py
fievelk: Ok, I can work on creating a new PR to fix this.

src/gsy_e/models/market/__init__.py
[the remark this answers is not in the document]
fievelk: I think `EnergyMarket` would have the same ambiguity of `MarketBase`, in the sense that it's still easy to mix it with the "area market". I agree that the difference between `Market` and `MarketBase` is not clear though.
However, please let's discuss this on slack and move on with this PR. We talked about the `MarketBase` name a few days ago and we decided to move on with it. Changing it now requires time and more testing, and the rebranding for the "area" -> "market" is quite urgent.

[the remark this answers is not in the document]
Hassan754: This is quite fine for me but I thought that rebranding again after merging this PR won't be possible, imo it's either now or never
Anyway, let's not drag this PR on, lgtm for now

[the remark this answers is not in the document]
fievelk: The `MarketBase` class is an internal class which is not exposed to end-users, so it can be renamed as we will. The "important" class for the rebranding is the new `Market` class, which will stay unchanged.

tests/area/test_area.py
hannesdiedrich: This probably needs to be decided in the discussion. I can leave my 5 cents to it if you link me to it (privately in slack). Thanks!

src/gsy_e/models/area/__init__.py
spyrostz: That's okay then, it makes sense to be public in this case! Thanks for explaining!

tests/area/test_area.py
spyrostz: To take this discussion from the top and clarify the most important thing first: old Market -> MarketBase , new Market -> class that inherits from Area and can be used by the setup files for rebranding purposes, which inherits from Area. These are 2 totally distinct classes and in no way can they be used interchangeably. The latter could be told for new Market and Area (since the former inherits from the latter).
For the rest of the discussion, I personally do not mind to rename all setup files from Area to Market / Asset classes, however I find it a bit futile since the researchers have their own scripts and do not use our integration tests ones IIRC (for instance, see setup/api_setup or setup/myco_setup)

spyrostz (approved):
LGTM

src/gsy_e/models/market/__init__.py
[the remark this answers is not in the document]
faizan2590: Renaming `Area` --> `Market`. Old `Market` class could be renamed to `MarketUnit`, the term 'unit' might aid in some clarity that each area(now called market) is split into smaller units based on time slot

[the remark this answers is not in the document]
fievelk: @faizan2590 thanks. Maybe even `MarketSlot` would be quite straightforward (at least in my opinion). However, I would suggest you to please create a poll on slack and keep this for a separate PR.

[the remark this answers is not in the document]
spyrostz: I would like to raise awareness on the fact that the Market class is not a class that can operate on its own, and therefore should not be constructed. Essentially it is the base class in the market hierarchy, and all other market classes (OneSidedMarket, TwoSidedMarket etc.) inherit from this class. Therefore, `Base` in `MarketBase` is not a random term, it really means that this is a base class and implies that should not be constructed. 
I raised this point because all the suggestions (MarketUnit, MarketSlot, EnergyMarket) ignore this aspect. IMO Base is necessary to be part of this class name. I also agree that it is better to take this discussion to slack since it has nothing to do with the ticket itself.

spyrostz (approved):
LGTM

faizan2590 (approved):
Screenshot 2021-11-22 at 18 47 24

Minor, but title should be "Rename `Market` to `MarketBase`

LGTM anyways

PR #1304, opened 2021-11-16
GSYE-6: Fix myco external matcher

Hassan754:


src/gsy_e/models/myco_matcher/myco_external_matcher.py
```diff
@@ -84,19 +84,25 @@ def publish_orders(self, message):
             for market in area_data["markets"]:
                 self.area_markets_mapping.update(
                     {f"{area_uuid}-{market.time_slot_str}": market})
-                market_offers_bids_list_mapping[area_uuid] = self._get_orders(market, filters)
+                if area_uuid not in market_orders_list_mapping:
+                    market_orders_list_mapping[area_uuid] = {}
+                market_orders_list_mapping[area_uuid].update(
+                    self._get_orders(market, filters))
```
Hassan754, line 90: The problem was that we were re-assigning the value of this key in each iteration instead of updating the content of the dictionary to add new time slot data
This was the direct cause of the failing ci tests in the myco sdk

src/gsy_e/models/myco_matcher/myco_external_matcher.py
```diff
@@ -84,19 +84,25 @@ def publish_orders(self, message):
             for market in area_data["markets"]:
                 self.area_markets_mapping.update(
                     {f"{area_uuid}-{market.time_slot_str}": market})
-                market_offers_bids_list_mapping[area_uuid] = self._get_orders(market, filters)
+                if area_uuid not in market_orders_list_mapping:
+                    market_orders_list_mapping[area_uuid] = {}
```
fievelk, line 88: Do we need to add the empty area to the mapping? Couldn't we just leave it out?

src/gsy_e/models/myco_matcher/myco_external_matcher.py
```diff
@@ -84,19 +84,25 @@ def publish_orders(self, message):
...
+                market_orders_list_mapping[area_uuid].update(
+                    self._get_orders(market, filters))
         self.area_uuid_markets_mapping = {}
         # TODO: change the `bids_offers` key and the channel to `orders`
         response_data.update({
-            "bids_offers": market_offers_bids_list_mapping,
+            "bids_offers": market_orders_list_mapping,
```
fievelk, line 105: Nit (unrelated to this PR): it would be good to change the `bids_offers` key to `orders`, as we discussed a while ago.

fievelk (approved):

src/gsy_e/models/myco_matcher/myco_external_matcher.py
Hassan754: In the next line we're guaranteed to add orders to this empty area mapping.
I added the key here in order to aggregate orders of area later on (An area can have orders in different time slots)

src/gsy_e/models/myco_matcher/myco_external_matcher.py
Hassan754: I agree, let me create the PRs for fix after merging this fix, however I'm a bit concerned about backward compatibility, but let's fix it now rather than later

hannesdiedrich (approved):
LGTM

PR #1306, opened 2021-11-18
Feature/d3 asim 3672 myco matcher infinite loop

spyrostz:
Please look at individual commit messages for clarifications on each one of the fixes.

### DIFF
--- a/src/gsy_e/gsy_e_core/simulation.py
+++ b/src/gsy_e/gsy_e_core/simulation.py
@@ -26,7 +26,6 @@ from time import sleep, time, mktime
 
 import psutil
 from gsy_framework.constants_limits import ConstSettings, GlobalConfig
-from gsy_framework.exceptions import GSyException
 from gsy_framework.kafka_communication.kafka_producer import kafka_connection_factory
 from gsy_framework.utils import format_datetime, str_to_pendulum_datetime
 from numpy import random
@@ -142,7 +141,7 @@ class Simulation:
             log.debug("Using setup module '%s'", self.setup_module_name)
         except (ModuleNotFoundError, ImportError) as ex:
             raise SimulationException(
-                "Invalid setup module '{}'".format(self.setup_module_name))
+                "Invalid setup module '{}'".format(self.setup_module_name)) from ex
 
     def _init(self, slot_length_realtime, seed, paused, pause_after, redis_job_id, enable_bc):
         self.paused = paused
--- a/src/gsy_e/gsy_e_core/simulation.py
+++ b/src/gsy_e/gsy_e_core/simulation.py
@@ -24,7 +24,6 @@ from importlib import import_module
 from logging import getLogger
 from time import sleep, time, mktime
 
-import click
 import psutil
 from gsy_framework.constants_limits import ConstSettings, GlobalConfig
 from gsy_framework.exceptions import GSyException
@@ -141,9 +140,9 @@ class Simulation:
                 sys.path.append(ConstSettings.GeneralSettings.SETUP_FILE_PATH)
                 self.setup_module = import_module("{}".format(self.setup_module_name))
             log.debug("Using setup module '%s'", self.setup_module_name)
-        except ImportError as ex:
+        except (ModuleNotFoundError, ImportError) as ex:
             raise SimulationException(
-                "Invalid setup module '{}'".format(self.setup_module_name)) from ex
+                "Invalid setup module '{}'".format(self.setup_module_name))
 
     def _init(self, slot_length_realtime, seed, paused, pause_after, redis_job_id, enable_bc):
         self.paused = paused
@@ -608,9 +607,9 @@ def run_simulation(setup_module_name="", simulation_config=None, simulation_even
                 redis_job_id=saved_sim_state["general"]["simulation_id"],
                 **kwargs
             )
-
-    except GSyException as ex:
-        raise click.BadOptionUsage(ex.args[0])
+    except SimulationException as ex:
+        log.error(ex)
+        return
 
     if saved_sim_state is not None and \
             saved_sim_state["areas"] != {} and \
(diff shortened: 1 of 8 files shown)


src/gsy_e/models/myco_matcher/myco_internal_matcher.py
```diff
@@ -67,6 +67,7 @@ def match_recommendations(self, **kwargs):
                     continue
                 while True:
```
hannesdiedrich, line 68: I feel like I asked this question already before: Why is this `while True` loop here in the first place? Wouldn't calling the `self._get_matches_recommendations()` and `market.match_recommendations()` without it also work?
I think it is worth to add an inline comment about it.

spyrostz: I am not sure, but to my eyes the rationale behind this is to iterate this process multiple times, until the available recommendations from residual orders are exhausted. For instance, if the first recommendation set produces trades, then some residual orders will be created which might be potential matches, therefore the process has to be repeated in order to match those etc. I'll reflect this on a comment. 
Without this fix, the problem was that if no trade was performed in the market.match_recommendations call, then the _get_matches_recommendations would return exactly the same pairs, which would then not be matched by market.match_recommendations, thus running to infinity.

src/gsy_e/models/market/future.py
```diff
@@ -106,6 +120,16 @@ def delete_orders_in_old_future_markets(self, current_market_time_slot: DateTime
...
 
+        self.offers = self._remove_old_orders_from_dict(self.offers, current_market_time_slot)
+        self.bids = self._remove_old_orders_from_dict(self.bids, current_market_time_slot)
+
+        self.offer_history = self._remove_old_orders_from_list(
+            self.offer_history, current_market_time_slot)
+        self.bid_history = self._remove_old_orders_from_list(
+            self.bid_history, current_market_time_slot)
+        self.trades = self._remove_old_orders_from_list(
+            self.trades, current_market_time_slot)
```
hannesdiedrich, line 131: Thanks a lot! I did not think about these. 🤦

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -608,9 +607,9 @@ def run_simulation(setup_module_name="", simulation_config=None, simulation_even
                 redis_job_id=saved_sim_state["general"]["simulation_id"],
                 **kwargs
             )
-
-    except GSyException as ex:
-        raise click.BadOptionUsage(ex.args[0])
+    except SimulationException as ex:
+        log.error(ex)
+        return
```
hannesdiedrich, line 612: I am not sure about this. AFAIS, this will break our error handling in the rq_job_handler:
https://github.com/gridsingularity/gsy-e/blob/0854c61564645deaaf12beb954a4e42c23a735f7/src/gsy_e/gsy_e_core/rq_job_handler.py#L126
I think we need to re-raise the exception here.

spyrostz: Actually it will not, because I explicitly caught SimulationException, which is only raised in this context by ModuleNotFoundError or ImportError. All other exceptions will reach the rq_job_handler , only without wrapping them up in click.BadOptionUsage errors. And this is actually one of the fixes, because click does not make sense to use in non-cli mode, it should only be used when someone runs the software through cli. Therefore its exceptions should not be used here, and we can use the exceptions that the code itself produces.

src/gsy_e/models/strategy/__init__.py
```diff
@@ -909,7 +909,8 @@ def get_posted_bids(self, market: "Market", time_slot: Optional[DateTime] = None
 
     def _assert_market_type_on_bid_event(self, market_id):
         assert (ConstSettings.IAASettings.MARKET_TYPE == SpotMarketTypeEnum.TWO_SIDED.value or
-                self.area.is_market_future(market_id)), (
+                self.area.is_market_future(market_id) or
+                self.area.is_market_settlement(market_id)), (
             "Invalid state, cannot receive a bid if single sided market is globally configured or "
             "if it is not a future market bid.")
```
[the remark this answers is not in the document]
spyrostz: I somehow knew that the commit message was not enough to explain this. By default, future and settlement markets are two-sided. They also have the ability to operate in parallel with one-sided spot market. In this case though, we will receive events from bid, delete bid etc. while running one-sided markets. This assertion used to validate that, when running an one-sided market, we do not receive any bid-related events, which is not true anymore as said above, since the future and settlement markets will send their own bids. Therefore we needed to change the assertion to accept bids from future and settlement markets even during one-sided spot market. 
Regarding the message per se, I think that the only thing that needs to change is the second line `if it is not a future or settlement market bid.` . Other than that looks clear to my eyes

src/gsy_e/models/myco_matcher/myco_internal_matcher.py
hannesdiedrich: I understand now, Thanks a lot! Can you maybe add a small comment here, e.g. `Perform matching until all recommendations and their residuals are handled.` (there is a probably a better way to formulate this) . Thanks!

src/gsy_e/gsy_e_core/simulation.py
hannesdiedrich: I understand now. Thanks a lot!

Hassan754 (dismissed):
Looking at the ticket description I suppose there are follow up PRs, in this case LGTM

src/gsy_e/models/myco_matcher/myco_internal_matcher.py
spyrostz: I think I will go with your recommendation, sounds clear enough to me. Thanks!

src/gsy_e/models/strategy/__init__.py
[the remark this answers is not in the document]
hannesdiedrich: Thanks a lot for being patient and explain! For some reason my logic failed me earlier. I understand now what `_assert_market_type_on_bid_event` is handling and agree with your slight change of the assert message.

Hassan754 (approved):

hannesdiedrich (approved):
LGTM

PR #1307, opened 2021-11-18
D3ASIM-3553: Rename Inter Area Agent to Market Agent

fievelk:
Rename Inter Area Agent to Market Agent.

Small note: right now we use `ConstSettings.IAASettings.VALUE` to reach all settings. If we provide a shortcut alias for it in the future (e.g. `ConstSettings.VALUE`) similar PRs could be a thousand lines shorter.


src/gsy_e/gsy_e_core/util.py
```diff
@@ -158,8 +159,8 @@ def get_char(self, timeout=0):
         return False
 
 
-def make_iaa_name(owner):
-    return f"IAA {owner.name}"
+def make_ma_name(owner):
+    return f"MA {owner.name}"
```
fievelk, line 163: This method is the equivalent of this on GSy-framework: https://github.com/gridsingularity/gsy-framework/pull/299/files/728943715ee3cddca74c7eaa0204dca63aa2bf85#diff-e0a72d7786b41a5338bfdd2b84dd8ecc663518e42e3c82e2038dc65af7c4ab91R363-R365
The difference is that here we deal with objects and not dictionaries. IMO this could be unified (so that we would have only one source of truth). 
If you want, I can work on it. I would keep it for a separate PR though.

src/gsy_e/gsy_e_core/util.py
```diff
@@ -272,6 +269,10 @@ def convert_nested_settings(class_object, class_name, settings_dict):
         for key, value in dict(class_object.__dict__).items():
             if key.startswith("__"):
                 continue
+
+            if class_name not in settings_dict:
+                settings_dict[class_name] = {}
```
fievelk, line 274: These lines fix a bug with the way we were populating the `settings_dict`. The previous version of this method would fail if the configuration class (e.g. `MASettings`) did not have at least one parameter with values. So in a case like this:
```
class IAASettings(MASettings):
    pass
```
the method was raising exceptions. These two lines fix the issue.

src/gsy_e/gsy_e_core/util.py
hannesdiedrich: IMO not important and can be dealt with in the future,

src/gsy_e/setup/gsy_e_settings.json
```diff
@@ -201,7 +201,7 @@
       "FINAL_SELLING_RATE": 0,
       "MAX_WIND_TURBINE_OUTPUT_W": 160
     },
-    "IAASettings": {
+    "MASettings": {
```
hannesdiedrich, line 204: This suggests that you did not run `fab write_default_settings_file` which updates the default settings file automatically. AFAIU, also the reason for the "bug" here: https://github.com/gridsingularity/gsy-e/pull/1307/files#r753275470

src/gsy_e/models/area/event_dispatcher.py
```diff
@@ -57,16 +57,16 @@ class AreaDispatcher:
     dicts with interarea agents for each market type.
     """
     def __init__(self, area: "Area"):
-        self._inter_area_agents: Dict[DateTime, OneSidedAgent] = {}
+        self._market_agents: Dict[DateTime, OneSidedAgent] = {}
```
hannesdiedrich, line 60: how about taking the opportunity and rename this to `self._spot_market_agents` to be consistent with the other namings?

hannesdiedrich (commented):
Two minor suggestions, LGTM in general

src/gsy_e/setup/gsy_e_settings.json
fievelk: I didn't run it, but I'm not sure that this would have solved the issue. I ran it now and it didn't change the lines that I manually edited.
BTW: as I ran it, it added new lines (e.g. for the PV azimuth settings etc). I suppose that this command is not used that often, and it's not described in our internal wiki. When should we run it?

hannesdiedrich: `it's not described in our internal wiki.` --> indeed. This we reminded ourselves in the PRs if somebody forgot to run it as soon as the ConstSettings  or GlobalConfig changed. (also the answer to `When should we run it`) 

I am not sure how to handle this, this can indeed be part of our workflow. (Fortunately, nobody is using this feature AFAIK. Users normally change the Settings in the setup file directly. So this is not breaking stuff)

src/gsy_e/models/area/event_dispatcher.py
fievelk: I've addressed this in d993287

src/gsy_e/setup/gsy_e_settings.json
fievelk: I've done it in d993287.

spyrostz: Late to the discussion here, however I think that it is a good time to add it as a part of fab sync in order to be generated automatically on every sync and to not forget it. No need to do it now, I think that a followup ticket would be more appropriate.

spyrostz (approved):
LGTM

Hassan754 (approved):

PR #1308, opened 2021-11-18
Feature/d3 asim 3672

spyrostz:
Added future markets simulation results to CSV files and to the endpoint buffer dicts.

### DIFF
--- a/src/gsy_e/gsy_e_core/export.py
+++ b/src/gsy_e/gsy_e_core/export.py
@@ -25,7 +25,7 @@ import shutil
 from collections import namedtuple
 from copy import deepcopy
 from functools import reduce  # forward compatibility for Python 3
-from typing import Dict, Tuple, List, Mapping
+from typing import Dict, Tuple, List, Mapping, TYPE_CHECKING
 
 import plotly.graph_objs as go
 from gsy_framework.constants_limits import ConstSettings, GlobalConfig, DATE_TIME_FORMAT
@@ -47,6 +47,9 @@ from gsy_e.models.market.market_structures import (AvailableMarketTypes,
                                                    PAST_MARKET_TYPE_FILE_SUFFIX_MAPPING)
 from gsy_e.models.state import ESSEnergyOrigin
 from gsy_e.models.strategy.storage import StorageStrategy
+
+if TYPE_CHECKING:
+    from gsy_e.models.market.future import FutureMarkets
 
 _log = logging.getLogger(__name__)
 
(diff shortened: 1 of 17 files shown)


src/gsy_e/gsy_e_core/export.py
```diff
@@ -444,20 +511,28 @@ def plot_all_unmatched_loads(self):
         output_file = os.path.join(plot_dir, f"unmatched_loads_{root_name}.html")
         PlotlyGraph.plot_bar_graph(plot_desc, output_file)
 
-    def plot_ess_soc_history(self, area, subdir):
+
+class PlotESSSocHistory:
```
faizan2590, line 515:
```suggestion
class PlotEssSocHistory:
```
OR
```suggestion
class PlotESSSOCHistory:
```

src/gsy_e/gsy_e_core/export.py
```diff
@@ -301,13 +339,34 @@ def _export_area_clearing_rate(self, area, directory, file_suffix, is_first) ->
...
             _log.exception("Could not export area market_clearing_rate")
 
-    def _export_offers_bids_trades_to_csv_files(self, past_markets: List, market_member: str,
-                                                directory: dir, filename: str, labels: Tuple,
+    @staticmethod
+    def _export_future_offers_bid_trades_to_csv_files(
+            future_markets: "FutureMarkets", market_member: str, file_path: dir,
+            labels: Tuple, is_first: bool = False) -> None:
+        """ Export files containing individual future offers, bids (*-bids*/*-offers*.csv files).
+        """
```
faizan2590, line 347:
```suggestion
        """ Export files containing individual future offers, bids (*-bids*/*-offers*.csv files)."""
```
OR
```suggestion
        """ 
        Export files containing individual future offers, bids (*-bids*/*-offers*.csv files).
        """
```
NITS

src/gsy_e/gsy_e_core/export.py
```diff
@@ -708,48 +739,137 @@ def _plot_avg_trade_price(self, area_list: list, subdir: str):
...
 
-    def plot_energy_trade_profile_hr(self, area: Area, subdir: str):
+
+class PlotOrderInfo:
+    """Create plot for the order high resolution information"""
+
+    def __init__(self, endpoint_buffer):
+        self._endpoint_buffer = endpoint_buffer
+
+    def plot_per_area_per_market_slot(self, area, plot_dir):
```
faizan2590, line 783:
```suggestion
    def plot_per_area_per_market_slot(self, area: "Area", plot_dir: str):
```
NITS

faizan2590 (approved):
Some nits, other than that LGTM

src/gsy_e/gsy_e_core/export.py
```diff
@@ -301,6 +335,27 @@ def _export_area_clearing_rate(self, area, directory, file_suffix, is_first) ->
...
+        """
+        file_path = self._file_path(directory, filename)
+        try:
+            with open(file_path, "a") as csv_file:
+                writer = csv.writer(csv_file)
+                if is_first:
+                    writer.writerow(labels)
+                if not future_markets.market_time_slots:
+                    return
+                time_slot = future_markets.market_time_slots[0]
```
hannesdiedrich, line 355: Why are you only selecting the first future market here? Is it because it already contains all the offers and bids and trades because it will be closed in the next market cycle?

I predict that researchers will like to look at the state of the other future markets too (e.g. when pausing the simulation). But lets wait until they ask for it. It will not be a big deal to add.

hannesdiedrich (commented):
One clarification and one request to add a integrationtest for testing the correct export of the future market files like `Test offers, bids trades and stats are exported also for settlement markets.` . IMO you could also enable the future markets for that test too and just add for the files in there. 
LGTM in general

src/gsy_e/gsy_e_core/export.py
spyrostz: Unfortunately it is a bit trickier, because I wanted to avoid duplicate entries. If I iterated through all market slots here, then all bids and offers from all future markets would be written to the csv files on every slot. This means that bids and offers that have been already recorded would be rewritten to the csv files. 
This can be avoided by filtering bids and offers through creation time, however I think it is cleaner to only write them to CSVs when the time slot is closing. If the researchers demand it then we can always enforce the workaround I mentioned, however I do not think they will, because I am not sure how frequently do they pause the simulation and look at the CSV files content (I would expect that they run the simulation first, analyse its results later).

src/gsy_e/gsy_e_core/export.py
spyrostz: Since you like it I will change it, however 7 capital letters in a row look weird to me.

src/gsy_e/gsy_e_core/export.py
spyrostz: Fixed

src/gsy_e/gsy_e_core/export.py
spyrostz: Fixing

src/gsy_e/gsy_e_core/export.py
hannesdiedrich: I see. OK. Definitely agree that filtering for creation time would be an overkill for now. Thanks!

src/gsy_e/gsy_e_core/export.py
faizan2590: `PlotEssSocHistory `  was the alternate proposal

spyrostz: Indeed, misread it, does not look weird enough to me though to go through another iteration. Leaving it as is

faizan2590 (approved):
LGTM

faizan2590 (approved):

hannesdiedrich (approved):
LGTM

PR #1310, opened 2021-11-19
D3ASIM-3690: Do not update bid if updated_rate is the same as the exi…

hannesdiedrich:
…sting one.


src/gsy_e/models/strategy/__init__.py
```diff
@@ -774,9 +774,10 @@ def update_bid_rates(self, market: "TwoSidedMarket", updated_rate: float) -> Non
         """Replace the rate of all bids in the market slot with the given updated rate."""
         existing_bids = list(self.get_posted_bids(market))
         for bid in existing_bids:
+            if abs(bid.energy_rate - updated_rate) <= FLOATING_POINT_TOLERANCE:
+                continue
             assert bid.buyer == self.owner.name
-            if bid.id in market.bids.keys():
-                bid = market.bids[bid.id]
```
hannesdiedrich, line 779: I did not see the need for this additional lookup so I removed it and the tests pass. However, please let me know if you think this is important and explain why. Thanks!

spyrostz: One could even go further and use `market.delete_bid(bid)` instead of `market.delete_bid(bid.id)`. Outside of the scope of the change though therefore leave it as is.

spyrostz (approved):
LGTM

integration_tests/steps/pv_tests.py
```diff
@@ -17,9 +17,8 @@
 """
 from behave import then
 from math import isclose
-from gsy_framework.constants_limits import ConstSettings
+from gsy_framework.constants_limits import ConstSettings, DEFAULT_PRECISION
```
fievelk, line 20: What would you think about removing `DEFAULT_PRECISION` and directly using `limit_float_precision`?

fievelk (dismissed):

integration_tests/steps/pv_tests.py
hannesdiedrich: yes, will do it.

fievelk (approved):

PR #1312, opened 2021-11-23
Fix wiki local redirect and missing image

fievelk:
The local version of the mkdocs wiki is currently automatically redirecting to the online website, which is not optimal for development. This PR fixes the issue by providing the relative URL to the redirection page, rather than its absolute online URL.

Other fixes:
- add missing redirect plugin requirement
- upload missing development-path.png image


wiki/mkdocs.yml
```diff
@@ -129,4 +129,5 @@ plugins:
 
     - redirects:
         redirect_maps:
-            '': 'https://gridsingularity.github.io/gsy-e/documentation/'
+            # When trying to reach the generic index page, redirect to the documentation page
+            'index.md': 'documentation.md'
```
hannesdiedrich, line 133: This will also work when deployed, right?

fievelk: There's only one way to know 😬 It should. In any case, it would be easy to revert it.

faizan2590 (approved):

andreabertolini1995 (approved):
LGTM

PR #1313, opened 2021-11-23
fix_post_bid_on_one_sided: Added _assert_market_type_on_bid_event to …

hannesdiedrich:
…remove_bid_from_pending and post_bid in the BaseStrategy class in order to return a readable message if external clients post bids on the one-sided-market.


src/gsy_e/models/strategy/__init__.py
```diff
@@ -752,6 +752,7 @@ def post_bid(
         Returns: The bid posted to the market
 
         """
+        self._assert_market_type_on_bid_event(market.id)
```
fievelk, line 755: The name of the method states that this should be executed only after a bid event, in fact before this PR it was only called in `event_bid_*` methods.
I'd recommend changing the name (e.g. something like `validate_bid_market_type` or `bid_can_happen_in_market`, or anything else).
Also, I think this method is a good candidate to become a decorator. What do you think?

hannesdiedrich: I agree on the renaming and would choose `validate_bid_can_be_posted_on_market`. The decorator solution does not work because the interfaces of the functions that this method is used in is different (sometime the BaseMarket object is pass, sometime only the market id).

Hassan754 (commented):
LGTM after addressing Pierpaolo's comment

src/gsy_e/models/strategy/__init__.py
spyrostz: I agree with the general renaming idea, however I would recommend to stick to assert in the name to denote that an AssertionError is expected as the only "output" of the method. `_assert_bid_can_be_posted_on_market` for instance would work for me

spyrostz (dismissed):
LGTM

src/gsy_e/models/strategy/__init__.py
spyrostz: That's exactly the reason why it should be an assertion, that this should never be triggered by end-user settings.

faizan2590 (approved):
LGTM

fievelk (approved):
LGTM. Conflicts are probably related to the IAA refactoring.

PR #1317, opened 2021-11-23
Redirect to the online version of the wiki

fievelk:
Unfortunately, the previous attempts did not work for both production and local versions. We revert to the initial version, with the redirect pointing to the online website.


Hassan754 (approved):

andreabertolini1995 (approved):
LGTM

PR #1320, opened 2021-11-24
Fixed _aggregator_command_callback_mapping.

hannesdiedrich:


Hassan754 (approved):

src/gsy_e/models/strategy/external_strategies/__init__.py
```diff
@@ -500,7 +500,7 @@ def _aggregator_command_callback_mapping(self) -> Dict:
             "offer": self._offer_aggregator,
             "delete_offer": self._delete_offer_aggregator,
             "list_offers": self._list_offers_aggregator,
-            "device_info": self._device_info,
+            "device_info": self._device_info_aggregator,
```
fievelk, line 503: Do we have a way to make tests fail when this is not correct?

fievelk (approved):

src/gsy_e/models/strategy/external_strategies/__init__.py
fievelk: On another side: maybe this is a good chance to rename this to `_asset_info_aggregator`. It's an internal implementation detail, but we could slowly adapt our internal names as well if we're working on them. Anyway, secondary.

hannesdiedrich: I will ask Ravineesh to check for error logs in the gsy-e-sdk scripts when connecting to the exchange.

faizan2590 (approved):

PR #1321, opened 2021-11-24
 Allow simulations to be iterated slot by slot to help with debugging

Hassan754:


fievelk:
How does this work? Does one need to manually change the boolean value in the code?

Hassan754:
> How does this work? Does one need to manually change the boolean value in the code? If so, I'd suggest to at least add a cli argument for it.

Yes, we need to explicitly change the boolean value
I did think about adding a cli flag for it but I wasn't sure if it's okay to expose it to all users, what do you think?

hannesdiedrich (commented):
I fail to see the need for this. Why do you pause on every slot? 
We also have `--pause-at` cli flag

Hassan754:
> -pause-a

I personally found it useful when debugging, for example when comparing the states at the end of each time slot, and I thought it would be a nice to have feature

fievelk:
> I did think about adding a cli flag for it but I wasn't sure if it's okay to expose it to all users, what do you think?

I personally like this feature and I found myself in situations in which I would have liked to have it. Regarding end users I'm not sure. I would personally prefer to have a cli option for it, because it would be documented and more stable (maybe even tested when we test the cli and simulation modules).

src/gsy_e/gsy_e_core/cli.py
```diff
@@ -106,7 +108,7 @@ def main(log_level):
                 "(orders can't contain attributes/requirements)."))
 def run(setup_module_name, settings_file, duration, slot_length, tick_length,
         cloud_coverage, compare_alt_pricing, enable_external_connection, start_date,
-        pause_at, slot_length_realtime, enable_dof: bool, **kwargs):
+        pause_at, iterative: bool, slot_length_realtime, enable_dof: bool, **kwargs):
```
fievelk, line 111: I think we could find a different word for it though. "Iterative" is not exactly the same as "step-by-step" I think, but I might be wrong.

Hassan754: I see, I'm also not sure myself but I think that we can find some candidates [here](https://www.merriam-webster.com/thesaurus/step-by-step)

fievelk:
@hannesdiedrich @Hassan754 what do we want to do with this PR? Should we work on it?

Hassan754:
> @hannesdiedrich @Hassan754 what do we want to do with this PR? Should we work on it?

@fievelk I'll find some other term instead of 'iterative' and then ask for another round of review

hannesdiedrich (approved):
I changed my mind for this feature. I think it can be helpful and leave a LGTM here. 
I am also not sure whether `iterative` is a good name for the option. How about `--pause-after-each-slot`?

Hassan754:
> I changed my mind for this feature. I think it can be helpful and leave a LGTM here. I am also not sure whether `iterative` is a good name for the option. How about `--pause-after-each-slot`?

I'm okay with your proposal, @fievelk what's your input on this?

Hassan754:
> I would be ok with Hannes' suggestion. I have two more options that I found looking at synonyms of "step-by-step":
> 
> * stepwise
> * incremental
> 
> Any of the three works for me.

I like these three, @hannesdiedrich it's your call then

hannesdiedrich:
> I like these three, @hannesdiedrich it's your call then

I vote for incremental then. :-)

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -99,7 +99,7 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig,
 
         self.setup_module_name = setup_module_name
         self.is_stopped = False
-
+        self._incremental_slots = incremental
```
fievelk, line 102: Nit: At first glance, because of the naming, I would expect this variable to contain slots (not a boolean). Why not name it `incremental`? Anyway, not a big deal.

fievelk (dismissed):

src/gsy_e/gsy_e_core/simulation.py
Hassan754: Well both comments make sense, I missed the second one, let find a common ground

Hassan754: What do you have on `self._increment_slots` ?

fievelk: I'd suggest keeping it as it is now so that we can merge it. Any name might sound "incomplete" somehow, so let's give up to language ambiguity :)

hannesdiedrich (approved):
LGTM

PR #1322, opened 2021-11-24
GSYE-61: Added support for live events by updating the bid / offer up…

spyrostz:
…daters on the event_activate on all strategies, and not only on the event_market_cycle. Added extra checks in the profile populate methods in order to avoid timeslots that span outside the range of the simulation duration.


hannesdiedrich (approved):
Interesting! LGTM

src/gsy_e/models/strategy/future/strategy.py
```diff
@@ -92,11 +96,16 @@ def event_market_cycle(self, strategy: "BidEnabledStrategy") -> None:
     def event_tick(self, strategy: "BidEnabledStrategy") -> None:
         """Base class method for handling the tick"""
 
+    def update_and_populate_price_settings(self, strategy: "BidEnabledStrategy") -> None:
```
Hassan754, line 99: I can see that all strategies type hints throughout the module are set to `BidEnabledStrategy` when they can also be `OfferEnabledStrategy`, so for these usages I suppose it's safe to use `BaseStrategy`
```suggestion
    def update_and_populate_price_settings(self, strategy: "BaseStrategy") -> None:
```

src/gsy_e/models/strategy/future/strategy.py
```diff
@@ -146,6 +158,14 @@ def __init__(self, asset_type: AssetType,
             initial_buying_rate, final_buying_rate, asset_type
         )
 
+    def update_and_populate_price_settings(self, strategy):
+        """
+        Updates and populates the price settings of the bid / offer updaters. Should be called
```
Hassan754, line 163: Nit
```suggestion
        Update and populate the price settings of the bid / offer updaters. Should be called
```

Hassan754 (approved):
LGTM

Hassan754:
I retriggered the ci tests and they seem to be failing, LGTM once fixed

Hassan754 (dismissed):

src/gsy_e/models/strategy/__init__.py
```diff
@@ -456,6 +457,7 @@ def __init__(self):
         self.event_responses = []
         self._market_adapter = market_strategy_connection_adapter_factory()
         self._settlement_market_strategy = self._create_settlement_market_strategy()
+        self._future_market_strategy = self._create_future_market_strategy()
```
hannesdiedrich, line 460: 👍

src/gsy_e/models/strategy/update_frequency.py
```diff
@@ -169,6 +172,11 @@ def _calculate_number_of_available_updates_per_slot(self) -> int:
 
     def update_and_populate_price_settings(self, area: "Area") -> None:
         """Populate the price profiles for every available time slot."""
+
+        # Handling the case where future markets are disabled during a simulation.
+        if self._time_slot_duration_in_seconds <= 0:
```
hannesdiedrich, line 177: I really do not understand this. Why would the `GlobalConfig.slot_length.seconds` be changed while runtime. Can you please explain?

tests/strategies/future/test_future_strategy.py
```diff
@@ -40,19 +40,21 @@ class TestFutureMarketStrategy:
 
     def setup_method(self) -> None:
         """Preparation for the tests execution"""
+        self._original_future_markets = GlobalConfig.FUTURE_MARKET_DURATION_HOURS
```
hannesdiedrich, line 43: NIT
```suggestion
        self._original_future_markets_duration = GlobalConfig.FUTURE_MARKET_DURATION_HOURS
```

hannesdiedrich (dismissed):
One question, one minor comment. LGTM

src/gsy_e/models/strategy/update_frequency.py
spyrostz: Not the slot length, but the GlobalConfig.FUTURE_MARKET_DURATION_HOURS . I encountered this when running parameterised unit tests. In this case, the future markets were enabled at the time of test collection, therefore the parameterised strategies in the tests were created with future market strategy enabled. During the test run though, some tests changed the GlobalConfig.FUTURE_MARKET_DURATION_HOURS to 0, and because the parameterised strategies were already constructed, their _future_market_strategy member contained a FutureMarketStrategy class and not a FutureMarketStrategyInterface (which essentially disables the future markets). This effect is similar to having constructed the strategy classes and while the simulation is running, to change the value of GlobalConfig.FUTURE_MARKET_DURATION_HOURS to 0. In this case, the market_cycle of the FutureMarketStrategy will still be called (because it was constructed on the test collection)  but the self._time_slot_duration_in_seconds would be 0 (because the GlobalConfig.FUTURE_MARKET_DURATION_HOURS == 0 at the time of the test run) , and this would raise the exception that we perceived in the CI. 

In order to alleviate this, I fixed the unit tests teardown methods to not revert the future market duration to 0 but to the previously configured value. This part here is an extra sanity check, in case the future markets are disabled while the simulation is running, then do not process the timeslot at all.

hannesdiedrich: Thanks a lot for the explanation! The missing link was that the `FutureTemplateStrategyBidUpdater` and  `FutureTemplateStrategyOfferUpdater` overwrite the `_time_slot_duration_in_seconds` . I only looked at the implementation 10 lines above which made me wonder.

hannesdiedrich (approved):
LGTM

Hassan754 (approved):

PR #1326, opened 2021-11-30
Bug/gsye 60 Fix future market controlling and disable future market template strategy for external strategies

hannesdiedrich:


fievelk:
TLDR: LGTM. I just want to share something interesting I found :)

Regarding the `GlobalConfig` issue with `FUTURE_MARKET_DURATION_HOURS` and `future_market_duration`, there would be a way to fix the current behaviour without changing the interface. Usually we could do this by using properties. However, since we're dealing with class variables, this is not that easy. But there's a way to use getters/setters also for class variables.

Properties (and their setters) are defined on a class, but they only affect the instances of that class. So the trick is to create a metaclass for the class. A metaclass is "the class of the class". In this way, you define the property on the metaclass, and this will affect the instance of that metaclass (which is your "real" class). An example:

```python
from pendulum import duration

class MetaGlobalConfig(type):
    @property
    def future_market_duration(cls):
        return duration(hours=cls.FUTURE_MARKET_DURATION_HOURS)

    @future_market_duration.setter
    def future_market_duration(cls, value):
        cls.FUTURE_MARKET_DURATION_HOURS = value

class GlobalConfig(metaclass=MetaGlobalConfig):
    FUTURE_MARKET_DURATION_HOURS = 24

print(GlobalConfig.future_market_duration)
# 1 day
GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 48
print(GlobalConfig.future_market_duration)
# 2 days
GlobalConfig.future_market_duration = 96
print(GlobalConfig.future_market_duration)
# 4 days
print(GlobalConfig.FUTURE_MARKET_DURATION_HOURS)
# 96
```
This is not the best thing in the world, but it's quite fun.

fievelk (approved):

Hassan754 (approved):

hannesdiedrich:
> TLDR: LGTM. I just want to share something interesting I found :)
> 
> Regarding the `GlobalConfig` issue with `FUTURE_MARKET_DURATION_HOURS` and `future_market_duration`, there would be a way to fix the current behaviour without changing the interface. Usually we could do this by using properties. However, since we're dealing with class variables, this is not that easy. But there's a way to use getters/setters also for class variables.
> 
> Properties (and their setters) are defined on a class, but they only affect the instances of that class. So the trick is to create a metaclass for the class. A metaclass is "the class of the class". In this way, you define the property on the metaclass, and this will affect the instance of that metaclass (which is your "real" class). An example:
> 
> ```python
> from pendulum import duration
> 
> class MetaGlobalConfig(type):
>     @property
>     def future_market_duration(cls):
>         return duration(hours=cls.FUTURE_MARKET_DURATION_HOURS)
> 
>     @future_market_duration.setter
>     def future_market_duration(cls, value):
>         cls.FUTURE_MARKET_DURATION_HOURS = value
> 
> class GlobalConfig(metaclass=MetaGlobalConfig):
>     FUTURE_MARKET_DURATION_HOURS = 24
> 
> print(GlobalConfig.future_market_duration)
> # 1 day
> GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 48
> print(GlobalConfig.future_market_duration)
> # 2 days
> GlobalConfig.future_market_duration = 96
> print(GlobalConfig.future_market_duration)
> # 4 days
> print(GlobalConfig.FUTURE_MARKET_DURATION_HOURS)
> # 96
> ```
> 
> This is not the best thing in the world, but it's quite fun.

I see that we have quite different concepts of "fun". :-D 
But thanks for this example!

PR #1327, opened 2021-12-02
 Fix reraising an exception in order to not swallow the message

Hassan754:


src/gsy_e/models/market/two_sided.py
```diff
@@ -318,7 +318,7 @@ def match_recommendations(
                 #  on a matching algorithm setting
                 if is_external_matching_enabled():
                     # re-raise exception to be handled by the external matcher
-                    raise InvalidBidOfferPairException from invalid_bop_exception
+                    raise invalid_bop_exception
```
Hassan754, line 321: Here we're conditionally raising a new exception, this is swallowing the exception message and not forwarding it to the higher level code, so I reraised the same exception

fievelk (approved):

spyrostz (approved):
LGTM

PR #1328, opened 2021-12-02
Fixed unit tests failure caused by the disabling of the future market…

spyrostz:
…s by default.

### DIFF
--- a/tests/area/test_markets.py
+++ b/tests/area/test_markets.py
@@ -31,8 +31,9 @@ from gsy_e.models.strategy.storage import StorageStrategy
 
 class TestMarketRotation:
 
+    @staticmethod
     @pytest.fixture
-    def area_fixture(self):
+    def area_fixture():
         original_registry = DeviceRegistry.REGISTRY
         DeviceRegistry.REGISTRY = {
             "General Load": (33, 35),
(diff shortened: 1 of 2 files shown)


tests/area/test_markets.py
```diff
@@ -49,15 +50,18 @@ def area_fixture(self):
         ConstSettings.BalancingSettings.ENABLE_BALANCING_MARKET = False
         ConstSettings.SettlementMarketSettings.ENABLE_SETTLEMENT_MARKETS = False
         gsy_e.constants.RETAIN_PAST_MARKET_STRATEGIES_STATE = False
+        original_future_markets = GlobalConfig.FUTURE_MARKET_DURATION_HOURS
```
spyrostz, line 53: Caching the previous value of the setting and restoring it right after the test has been completed.

tests/area/test_markets.py
```diff
@@ -74,8 +78,10 @@ def test_market_rotation_is_successful(self, area_fixture):
         assert len(area_fixture.past_markets) == 1
         assert len(area_fixture.all_markets) == 1
 
+    @staticmethod
     @patch("gsy_e.constants.RETAIN_PAST_MARKET_STRATEGIES_STATE", True)
-    def test_market_rotation_is_successful_keep_past_markets(self, area_fixture):
+    def test_market_rotation_is_successful_keep_past_markets(area_fixture):
+        GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 24
```
spyrostz, line 84: This is the actual fix, this test depended on the future market hours having a value of 24.

hannesdiedrich (dismissed):
LGTM

faizan2590 (dismissed):
LGTM

tests/area/test_markets.py
fievelk: Shouldn't we restore the original value here as well? I know that `24` is the default, but if we change the default value this might create issues, right?

fievelk (approved):

Hassan754 (approved):

PR #1330, opened 2021-12-06
 Wrap the orders mappings in future market

Hassan754:
Actions covered by the implementation of this design:
```python
self.bids = {..}  # This will create a FutureOrders object and create relevant mappings
self.bids.pop(bid_id)  # This will pop this item from the {id: bid} and {time_slot: [bids_list]}
del self.bids[bid_id]  # Same as above
self.bids[bid_id] = bid  # Add the bid to the two mappings
self.bids.update({....})  # Same as above
```

### DIFF
--- a/src/gsy_e/models/market/future.py
+++ b/src/gsy_e/models/market/future.py
@@ -15,7 +15,7 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-# pylint: disable=too-many-instance-attributes, too-many-arguments, too-many-locals
+# pylint: disable=too-many-instance-attributes, too-many-arguments, too-many-locals, no-member
 from collections import UserDict
 from copy import deepcopy
 from logging import getLogger
@@ -55,10 +55,9 @@ class FutureOrders(UserDict):
         self.slot_order_mapping[order.time_slot].append(order)
 
     def __delitem__(self, order_id):
-        order = self.data.get(order_id)
+        order = self.data.pop(order_id, None)
         if order:
             self.slot_order_mapping[order.time_slot].remove(order)
-        super().__delitem__(order_id)
 
     def expire(self, current_market_time_slot: DateTime) -> Dict:
         """Remove old orders (time_slot in the past)."""
--- a/src/gsy_e/models/market/future.py
+++ b/src/gsy_e/models/market/future.py
@@ -15,7 +15,7 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-# pylint: disable=too-many-instance-attributes, too-many-arguments, too-many-locals, no-member
+# pylint: disable=too-many-arguments, too-many-locals, no-member
 from collections import UserDict
 from copy import deepcopy
 from logging import getLogger
@@ -59,21 +59,6 @@ class FutureOrders(UserDict):
         if order:
             self.slot_order_mapping[order.time_slot].remove(order)
         del self.data[order_id]
-
-    def expire(self, current_market_time_slot: DateTime) -> Dict:
-        """Remove old orders (time_slot in the past)."""
-        data = {}
-        slot_order_mapping = deepcopy(self.slot_order_mapping)
-        for order_id, order in self.data.items():
-            if order.time_slot > current_market_time_slot:
-                data[order_id] = order
-
-        for time_slot in self.slot_order_mapping.keys():
-            if time_slot <= current_market_time_slot:
-                del slot_order_mapping[time_slot]
-
-        self.data = data
-        self.slot_order_mapping = slot_order_mapping
 
 
 class FutureMarkets(TwoSidedMarket):
@@ -165,11 +150,22 @@ class FutureMarkets(TwoSidedMarket):
             order for order in order_list if order.time_slot > current_market_time_slot
         ]
 
+    def _expire_orders(self, orders: "FutureOrders", current_market_time_slot: DateTime) -> None:
+        """Remove old orders (time_slot in the past)."""
+        for order_id, order in deepcopy(list(orders.items())):
+            if order.time_slot <= current_market_time_slot:
+                if isinstance(order, Offer):
+                    self.delete_offer(order_id)
+                else:
+                    self.delete_bid(order_id)
+        for time_slot in deepcopy(list(orders.slot_order_mapping.keys())):
+            if time_slot <= current_market_time_slot:
+                del orders.slot_order_mapping[time_slot]
+
     def delete_orders_in_old_future_markets(self, current_market_time_slot: DateTime) -> None:
         """Delete order and trade buffers."""
-
-        self.offers.expire(current_market_time_slot)
-        self.bids.expire(current_market_time_slot)
+        self._expire_orders(self.offers, current_market_time_slot)
+        self._expire_orders(self.bids, current_market_time_slot)
 
         self.offer_history = self._remove_old_orders_from_list(
             self.offer_history, current_market_time_slot)
--- a/src/gsy_e/models/market/future.py
+++ b/src/gsy_e/models/market/future.py
@@ -193,16 +193,6 @@ class FutureMarkets(TwoSidedMarket):
             future_time_slot = future_time_slot.add(minutes=slot_length.total_minutes())
 
     @lock_market_action
-    def get_bids_per_slot(self, time_slot: DateTime) -> List[Bid]:
-        """Return list of bids for a specific market slot."""
-        return deepcopy(self.slot_bid_mapping[time_slot])
-
-    @lock_market_action
-    def get_offers_per_slot(self, time_slot: DateTime) -> List[Offer]:
-        """Return list of offers for a specific market slot."""
-        return deepcopy(self.slot_offer_mapping[time_slot])
-
-    @lock_market_action
     def bid(self, price: float, energy: float, buyer: str, buyer_origin: str,
             bid_id: Optional[str] = None,
             original_price: Optional[float] = None,
--- a/src/gsy_e/models/market/future.py
+++ b/src/gsy_e/models/market/future.py
@@ -16,12 +16,13 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 # pylint: disable=too-many-instance-attributes, too-many-arguments, too-many-locals
+from collections import UserDict
 from copy import deepcopy
 from logging import getLogger
-from typing import Dict, List, Union, Optional, Tuple, TYPE_CHECKING
+from typing import Dict, List, Optional, TYPE_CHECKING
 
 from gsy_framework.constants_limits import ConstSettings, GlobalConfig, DATE_TIME_FORMAT
-from gsy_framework.data_classes import Bid, Offer, Trade, BaseBidOffer, TradeBidOfferInfo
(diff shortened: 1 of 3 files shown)


src/gsy_e/models/market/future.py
```diff
@@ -41,6 +42,37 @@ class FutureMarketException(Exception):
...
+        super().__init__(*args, **kwargs)
+
+    def __setitem__(self, order_id, order):
+        self.data[order_id] = order
+        if order.time_slot not in self.slot_order_mapping:
+            self.slot_order_mapping[order.time_slot] = []
+        self.slot_order_mapping[order.time_slot].append(order)
+
+    def __delitem__(self, order_id):
+        order = self.data.get(order_id)
```
hannesdiedrich, line 58: You probably meant
```suggestion
        order = self.data.pop(order_id)
```
right?

src/gsy_e/models/market/future.py
```diff
@@ -100,25 +162,11 @@ def _remove_old_orders_from_list(order_list: List, current_market_time_slot: Dat
...
-                                            self.slot_bid_mapping, Bid)
-        self._delete_order_dict_market_slot(current_market_time_slot,
-                                            self.slot_offer_mapping, Offer)
-        self._delete_order_dict_market_slot(current_market_time_slot,
-                                            self.slot_trade_mapping, Trade)
 
-        self.offers = self._remove_old_orders_from_dict(self.offers, current_market_time_slot)
-        self.bids = self._remove_old_orders_from_dict(self.bids, current_market_time_slot)
+        self.offers.expire(current_market_time_slot)
+        self.bids.expire(current_market_time_slot)
```
hannesdiedrich, line 169: what happened with the trades?

Hassan754: The `trades` in its nature (inherited from the MarketBase) is a list, not a dictionary like the case of bids and offers
I kept dealing with the trades as usual [here](https://github.com/gridsingularity/gsy-e/blob/c860f7d45957f736c97b7a1fecdd6a73e2b63844/src/gsy_e/models/market/future.py#L177)

Hassan754: However, the slot_trades_mapping is now a dynamic property, so it's always in sync with the trades list (I checked the usages of this mapping and it seems to be only used when exporting the data)

Hassan754:
Note, if this design is fine for you, we can scale it up and do the same to all markets, while still allowing each market to have its own specific caching needs

src/gsy_e/models/market/future.py
```diff
@@ -41,6 +42,39 @@ class FutureMarketException(Exception):
...
+    """Special mapping object to keep track of a future market's orders."""
+    def __init__(self, *args, **kwargs):
+        self.slot_order_mapping = {}
+        super().__init__(*args, **kwargs)
+
+    def __setitem__(self, order_id, order):
+        self.data[order_id] = order
+        if order.time_slot not in self.slot_order_mapping:
+            self.slot_order_mapping[order.time_slot] = []
+        self.slot_order_mapping[order.time_slot].append(order)
```
Hassan754, line 55: Each time we add an item to this FutureOrders dictionary, this dunder method will keep both mappings in sync

src/gsy_e/models/market/future.py
```diff
@@ -41,6 +42,39 @@ class FutureMarketException(Exception):
...
+
+    def __delitem__(self, order_id):
+        order = self.data.pop(order_id, None)
+        if order:
+            self.slot_order_mapping[order.time_slot].remove(order)
```
Hassan754, line 60: Each time we delete an item from this FutureOrders dictionary (`del` or `pop`), this dunder method will keep both mappings in sync

src/gsy_e/models/market/future.py
```diff
@@ -54,9 +88,46 @@ def __init__(self, bc: Optional[NonBlockchainInterface] = None,
...
+
+    @property
+    def offers(self) -> FutureOrders:
+        """Return the {offer_id: offer} mapping."""
+        return self._offers
+
+    @offers.setter
+    def offers(self, orders) -> None:
+        """Wrap the setter of _orders in order to build a FutureOrders object."""
+        self._offers = FutureOrders(orders)
```
Hassan754, line 102: Each time we try to reassign the offers dict `market.offers = {"id1": offer1}`, this setter will get rid of the typical dictionary and will use our FutureOrders dictionary

src/gsy_e/models/market/future.py
```diff
@@ -71,13 +142,6 @@ def market_time_slots(self) -> List[DateTime]:
         """Return list of all time slots of future markets."""
         return list(self.slot_bid_mapping.keys())
 
-    def open_bids_and_offers(self, **kwargs) -> Tuple[List, List]:
```
Hassan754, line 74: This method seems to be outdated, there are no usages for it

src/gsy_e/models/market/future.py
```diff
@@ -54,9 +88,46 @@ def __init__(self, bc: Optional[NonBlockchainInterface] = None,
...
+        """Return the {time_slot: [offers_list]} mapping."""
+        return self.offers.slot_order_mapping
+
+    @property
+    def slot_trade_mapping(self) -> Dict:
+        """Return the {time_slot: [trades_list]} mapping."""
+        mapping = {time_slot: [] for time_slot in self.slot_bid_mapping.keys()}
+        for trade in self.trades:
+            mapping[trade.time_slot].append(trade)
+        return mapping
```
hannesdiedrich, line 131: The is not very efficient. So only the lookup in this property will take a lot of computational effort. Why not keeping the `slot_trade_mapping` as it was before because there are not two buffers that need to be synced?

Hassan754: @hannesdiedrich IMO there is a decision to be made here
First of all, the trades is a list and not a dictionary, so the solution introduced in this PR isn't relevant in the trades case.
To fix this, I can implement another class which acts like a list (and keeps track of the slot_trade_mapping).
I wouldn't suggest keeping it as it was because we'll have to pay attention on adding the trade to the list and to the slot_trade_mapping every time (and removing them in other cases).

I decided to go with this property solution because the slot_trade_mapping is only accessed once, when exporting the data (significantly less times than adding new trade to the list), so imo it's not much computation

the slightly more computation needed is balanced with the less memory used to store the mapping, both are minors

hannesdiedrich: Ok, You are probably right, if we only need the `self.slot_trade_mapping` only for the export now, leave it as is. We can add more logic later, if we need to access this property inside the market too.

src/gsy_e/models/market/future.py
hannesdiedrich: Wow, then the only usage if `self. slot_bid_mapping` and `self. slot_offer_mapping` is gone AFAIS. I wonder why we added them in the first place.
@spyrostz I see that we only export the trades (or statistics of them) for the future markets. I remember that we decided to postpone the export of the bids and offers per future market slot until somebody asks for it right?

If all the above is true, I wonder why the current change is needed. I was under the impression that we needed this to fix a bug because of not synced buffers. But as the order mappings are not used somewhere else, why is this needed @Hassan754 ?

hannesdiedrich (commented):
I think we need to discuss again the purpose of this. finished reviewing so far.

src/gsy_e/models/market/future.py
Hassan754: @hannesdiedrich we're using these mappings in `orders_per_slot`, `get_offers_per_slot` and `get_bids_per_slot` .

I decided to introduce this feature when debugging a future market issue, in the end it turned out that this sync isn't the problem, but it's a nice and secure to add improvement to our codebase, and as you can see this saved us lots of overriding methods just for the sake of also adding the order to the slot_offer_mapping and slot_bid_mapping

hannesdiedrich: Ah I see now, `orders_per_slot` is really the only place where we need them and it is important for the matching algorithm, right? Then ignore me.
BTW; `get_bids_per_slot ` and `get_offers_per_slot` do not seem to be used anywhere.

Hassan754: Interesting! 2 more methods to be removed !

tests/market/test_future.py
```diff
@@ -238,3 +249,73 @@ def test_orders_per_slot(future_market):
...
+
+    @staticmethod
+    def test_offers_setter(future_market, offer):
+        """Test reassigning the offers member of the future market."""
+        assert isinstance(future_market.offers, FutureOrders)
+        future_market.offers = {
+            str(offer.id): offer}
+        assert isinstance(future_market.offers, FutureOrders)
+        assert future_market.offers[str(offer.id)] == offer
+
```
hannesdiedrich, line 261: Lets add the check here, too. Or what do you think?
```suggestion
assert offer in offers.slot_order_mapping[offer.time_slot]

```

hannesdiedrich: And in the `test_bids_setter`, too.

hannesdiedrich (dismissed):
One minor comment left. LGTM after addressed.

hannesdiedrich (approved):
LGTM

src/gsy_e/models/market/future.py
```diff
@@ -54,9 +74,46 @@ def __init__(self, bc: Optional[NonBlockchainInterface] = None,
                          readonly=readonly, grid_fee_type=grid_fee_type,
                          grid_fees=grid_fees, name=name, in_sim_duration=True)
 
-        self.slot_bid_mapping: Dict[DateTime, List[Bid]] = {}
-        self.slot_offer_mapping: Dict[DateTime, List[Offer]] = {}
-        self.slot_trade_mapping: Dict[DateTime, List[Trade]] = {}
+        self._offers = FutureOrders()
+        self._bids = FutureOrders()
```
spyrostz, line 78: Can you leave a small clarification comment here, that the super().__init__() should always precede these, in order to not mistakenly move them before the constructor in the future? If the constructor was moved under these, they would be replaced by normal Python dicts (since they would be defined by the base class constructor)

spyrostz (approved):
One minor comment, LGTM

faizan2590 (approved):

src/gsy_e/models/market/future.py
Hassan754: @spyrostz actually these two lines can be completely removed, because this class defines the property setter and once we try to assign the value in the parent class `self.offers = {}` it will resort to the setter which in its turn will transform the dict object into a FutureOrders object, so we're safe
I left these two here in the constructor in order to make it easier to understand the changes and to make sure that these private members are defined in the __init__, but if you believe I should remove/add explicit comment, let me know

spyrostz: The problem is though that other market types can operate directly on `_bids` and `_offers` members (IIRC bid() or offer() methods do this `self._bids[bid.id] = bid`). Therefore even if the `bids` / `offers` properties are overridden, they would serve the original `_bids` and `_offers` in this case, which would be normal dicts. Please correct me if I have not seen something that I should have here.

Hassan754: No other market type have the `_bids` nor the `_offers` members, AFAIS these members are indeed defined in strategies and market agents, so they should not interfere (I suppose)

spyrostz: You are right, of course! I obviously confused it with the strategy members, and indeed since the members are `bids` and `offers` they can work even without it. Thanks for bearing with me !

Hassan754: Thanks a lot! I will be moving this PR forward then

PR #1331, opened 2021-12-06
Bug/gsye 74 Added ability to send settlement offers and bids and fixed issue in can_bid_be_posted and can_offer_be_posted

hannesdiedrich:


src/gsy_e/models/strategy/external_strategies/__init__.py
```diff
@@ -316,6 +320,117 @@ def _get_market_from_time_slot(self, time_slot: DateTime) -> MarketBase:
                 f"settlement markets")
         return market
 
+    def _offer_aggregator_impl(
```
spyrostz, line 323: I am leaving the comment here in order to ideally not discuss it in this PR since we have discussed it already: I think that moving this method to its own offer mixin would be good in the long run, however it should not be done in the context of this ticket since we have another open ticket that deals with this.

spyrostz: I have a followup question though: why was it necessary to be moved in the ExternalMixin in the context of this PR ? I assume to avoid code duplication, right ?

src/gsy_e/models/strategy/external_strategies/__init__.py
```diff
@@ -316,6 +320,117 @@ def _get_market_from_time_slot(self, time_slot: DateTime) -> MarketBase:
...
+
+            response = {
+                "command": "offer",
+                "status": "ready",
+                "offer": offer.to_json_string(replace_existing=replace_existing),
+                "transaction_id": arguments.get("transaction_id"),
+                "area_uuid": self.device.uuid,
+                "message": response_message}
+        except (AssertionError, GSyException):
+            logging.exception("Failed to post PV offer.")
```
spyrostz, line 370: Please remove the PV reference since it is on the ExternalMixin which has no knowledge of the strategy type. Alternatively, you can do the following `f"Failed to post offer on strategy {self.__class__.__name__}"` to get the class name and have the indication which strategy was affected.

src/gsy_e/models/strategy/external_strategies/__init__.py
hannesdiedrich: Yes, this was my try to be DRY and to re-use this code in the external PV and in the Load.

src/gsy_e/models/strategy/external_strategies/__init__.py
```diff
@@ -316,6 +320,117 @@ def _get_market_from_time_slot(self, time_slot: DateTime) -> MarketBase:
...
+                requirements=arguments.get("requirements")
+            )
+            response = {
+                "command": "bid", "status": "ready",
+                "bid": bid.to_json_string(replace_existing=replace_existing),
+                "area_uuid": self.device.uuid,
+                "transaction_id": arguments.get("transaction_id"),
+                "message": response_message}
+        except (AssertionError, GSyException):
+            logging.exception("Error when handling bid on area %s", self.device.name)
```
spyrostz, line 425: Even better than my class name proposal on the previous comment, you could copy the same error message on the logging.exception in the corresponding offer exception log.

src/gsy_e/models/strategy/external_strategies/__init__.py
hannesdiedrich: Ah yes, thanks!

src/gsy_e/models/strategy/__init__.py
```diff
@@ -374,13 +374,13 @@ def can_offer_be_posted(
 
         if replace_existing:
             # Do not consider previous open offers, since they would be replaced by the current one
-            posted_offer_energy = self.sold_offer_energy(market.id, time_slot)
```
spyrostz, line 377: I really wonder why did this not show up till now. I assume because this was only called from the external strategies and the template strategies that we usually test our code against do not use it right ?

spyrostz (dismissed):
LGTM

src/gsy_e/models/strategy/__init__.py
hannesdiedrich: I asked myself the same and concluded equally. This may even fix some issues the research team still have with the external strategies.

fievelk: I'm not sure I understood the fix. In my head, if the new offer has `replace_existing`, it should replace only the open offers that have not been traded yet, not all offers. So the energy should be equal to the energy of the offers that have been traded so far, not `0`. Could you please explain this part?

hannesdiedrich: Yes, you are right, and this was/is already take care of in line 383.
When setting `replace_existing=True` we only ignore the already posted offers, as they will be replaces in the following. I hope this explains it. Happy to be more verbose if you like.

fievelk: On line 383 we added the `time_slot` parameter to `sold_offer_energy()`. But this shouldn't change the behaviour as far as I know (it only affects future markets). I think I need the verbose explanation :)

src/gsy_e/models/strategy/external_strategies/load.py
```diff
@@ -291,58 +292,60 @@ def event_offer(self, *, market_id, offer) -> None:
...
-            bid = self.post_bid(
-                market,
-                arguments["price"],
-                arguments["energy"],
-                replace_existing=replace_existing,
-                attributes=arguments.get("attributes"),
-                requirements=arguments.get("requirements")
-            )
+            if self.area.is_market_settlement(market.id):
+                assert self.state.can_post_settlement_offer(market.time_slot), (
```
fievelk, line 300: I think that this assertion can fail even if the implementation is correct (because it depends on the user/client input). Since this is new code, I'd suggest using exceptions instead of assertions.

src/gsy_e/models/strategy/__init__.py
hannesdiedrich: Ha, I just had another (appr. 1000th) look at this and indeed, in line 383 we subtract the `sold_offer_energy` from the posted_energy which made the `replace_existing=True` case work, because we first add and then subtract the same number. But for the `replace_existing=False` case. we were not taking into account the `sold_offer_energy` .
So, this will be my proposal to fix both cases:
```python
    if replace_existing:
            posted_offer_energy = 0.0
        else:
            posted_offer_energy = self.posted_offer_energy(market.id, time_slot)

        posted_energy = (offer_energy
                         + posted_offer_energy
                         + self.sold_offer_energy(market.id, time_slot))
```
What do you think?

src/gsy_e/models/strategy/external_strategies/load.py
hannesdiedrich: Yeah, I was thinking the same, just copying from the old implementation. Will do like you propose.

src/gsy_e/models/strategy/__init__.py
fievelk: So if `replace_existing is True`, then our posted energy is sum of the energy of the 
- current offer
- the previous offers we've already traded.

If `replace_existing is False`, then our posted energy is the sum of the energy of the 
- current offer
- the previous offers we haven't traded yet
- the previous offers we have already traded.

Is this correct? If this is right, the snippet you posted seems good to me 👍

hannesdiedrich: Yes, that is correct. There are tests that I will extend to catch all corner cases. Thanks!

src/gsy_e/models/strategy/external_strategies/__init__.py
```diff
@@ -344,13 +349,13 @@ def _offer_aggregator_impl(
...
-                market,
-                time_slot=time_slot,
-                replace_existing=replace_existing)
+            if not self.can_offer_be_posted(arguments["energy"],
+                                            arguments["price"],
+                                            available_energy,
+                                            market,
+                                            time_slot=time_slot,
+                                            replace_existing=replace_existing):
+                raise OrderCanNotBePosted()
```
fievelk, line 358: Nit: I'd recommend using the other indentation style (called hanging indentation, as [mentioned in PEP 8](https://www.python.org/dev/peps/pep-0008/#indentation)). I originally thought it was just a personal preference, but I found out that it has some objective advantages:
1. it reduces the amount of empty white space, so it's less likely to break the max length of a line;
2. it avoids flake8 errors. If we rename `can_offer_be_posted` to something else with more/less characters, flake8 will raise errors regarding the indentation of all the arguments. This wouldn't happen using hanging indents.

fievelk: Also, I think that you don't need to instantiate the exception. This would work:
```python
raise OrderCanNotBePosted
```

tests/strategies/external/utils.py
```diff
@@ -15,17 +15,19 @@
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+# pylint: disable=missing-function-docstring
```
fievelk, line 18: Nit: this module is not a test module, but rather a utils module (for tests). So I wouldn't disable this warning.

fievelk (approved):

fievelk (dismissed):

src/gsy_e/models/strategy/__init__.py
```diff
@@ -380,7 +380,7 @@ def can_offer_be_posted(
 
         posted_energy = (offer_energy
```
spyrostz, line 381: Very nit, but in order to be 100% clear and to not fall into similar bugs in the future, I would rename this variable to posted_and_traded_offer_energy

spyrostz (approved):
One nit comment that can be skipped, LGTM

src/gsy_e/models/strategy/__init__.py
hannesdiedrich: I agree and will change it.

spyrostz (approved):

fievelk (approved):

PR #1332, opened 2021-12-07
GSYE-85: Fix the storage state bug when the future markets are enabled

Hassan754:
The first commit is the actual fix whereas the second one is a pylint warnings fixes

I did enable future markets in the integration tests and all features passed

### DIFF
--- a/src/gsy_e/models/state.py
+++ b/src/gsy_e/models/state.py
@@ -21,7 +21,7 @@ from enum import Enum
 from math import isclose, copysign
 from typing import Dict, Optional, List
 
-from gsy_framework.constants_limits import ConstSettings
+from gsy_framework.constants_limits import ConstSettings, GlobalConfig
 from gsy_framework.utils import (
     convert_pendulum_to_str_in_dict, convert_str_to_pendulum_in_dict, convert_kW_to_kWh)
 from gsy_framework.utils import limit_float_precision
@@ -700,7 +700,7 @@ class StorageState(StateInterface):
         Simulate actual Energy flow by removing pledged storage and adding bought energy to the
         used_storage
         """
-        if all_future_time_slots:
+        if GlobalConfig.FUTURE_MARKET_DURATION_HOURS:
             # In case the future market is enabled, the current time_slot will already be populated
             # So we need to reset this time_slot values to defaults
             self._delete_time_slot(current_time_slot)
--- a/src/gsy_e/models/state.py
+++ b/src/gsy_e/models/state.py
@@ -19,14 +19,14 @@ from abc import ABC, abstractmethod
 from collections import namedtuple
 from enum import Enum
 from math import isclose, copysign
-from typing import Dict, Optional
+from typing import Dict, Optional, List
 
 from gsy_framework.constants_limits import ConstSettings
 from gsy_framework.utils import (
     convert_pendulum_to_str_in_dict, convert_str_to_pendulum_in_dict, convert_kW_to_kWh)
+from gsy_framework.utils import limit_float_precision
 from pendulum import DateTime
 
-from gsy_framework.utils import limit_float_precision
 from gsy_e.constants import FLOATING_POINT_TOLERANCE
 from gsy_e.gsy_e_core.util import is_time_slot_in_past_markets, write_default_to_dict
 
@@ -76,6 +76,7 @@ class ProsumptionInterface(StateInterface, ABC):
         self._unsettled_deviation_kWh: Dict[DateTime, float] = {}
         self._forecast_measurement_deviation_kWh: Dict[DateTime, float] = {}
 
+    # pylint: disable=unused-argument, no-self-use
     def _calculate_unsettled_energy_kWh(
             self, measured_energy_kWh: float, time_slot: DateTime) -> float:
         """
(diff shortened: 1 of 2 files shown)


src/gsy_e/models/state.py
```diff
@@ -673,7 +700,10 @@ def market_cycle(self, past_time_slot, current_time_slot: DateTime, all_future_t
         Simulate actual Energy flow by removing pledged storage and adding bought energy to the
         used_storage
         """
-        self.add_default_values_to_state_profiles(all_future_time_slots)
+        # In case the future market is enabled, the current time_slot will already be populated
+        # So we need to reset this time_slot values to defaults
+        self._delete_time_slot(current_time_slot)
+        self.add_default_values_to_state_profiles([current_time_slot, *all_future_time_slots])
```
hannesdiedrich, line 706: Maybe we can save the effort of both steps here with for non-enabled future markets. What do you think?

Hassan754: Makes sense, I disabled this process for simulations that don't enable future markets

Hassan754:
> Please add unittests for this change. Once added, LGTM

I was trying to add such unit tests for this altered method but AFAIS we're not currently testing the state classes, which is risky
I propose creating a refactoring ticket to write tests for all the states modules.

As for the fix introduced in this PR, the integration tests [ticket](https://gridsingularity.atlassian.net/browse/GSYE-27) I'm working on will make sure everything is running as expected (this is how I first noticed the bug)

If the above proposal works for you, let me create the ticket

src/gsy_e/models/state.py
```diff
@@ -668,11 +695,18 @@ def add_default_values_to_state_profiles(self, future_time_slots):
...
         """
         Simulate actual Energy flow by removing pledged storage and adding bought energy to the
         used_storage
         """
+        if GlobalConfig.FUTURE_MARKET_DURATION_HOURS:
+            # In case the future market is enabled, the future orders have to be deleted once
+            # the market becomes a spot market
+            del self.offered_buy_kWh[current_time_slot]
+            del self.offered_sell_kWh[current_time_slot]
+            all_future_time_slots.insert(0, current_time_slot)
```
spyrostz, line 709: I think that this is not really required. If you backtrack `all_future_time_slots` you will see that it ends up in this method of the future markets:

```
    @property
    def market_time_slots(self) -> List[DateTime]:
        """Return list of all time slots of future markets."""
        return list(self.slot_bid_mapping.keys())
```
This returns a list with the dict keys, which is indeed the future markets, but the original object (the dict) is not modified, only the copy of the keys that are created. Also, it is not really correct to add the time slot of the spot market as the first timeslot of the future markets, therefore I think that you should not force this behavior here. 

Even without this line though, I think that the fix will work well.

src/gsy_e/models/state.py
```diff
@@ -668,11 +695,18 @@ def add_default_values_to_state_profiles(self, future_time_slots):
...
+    def market_cycle(self, past_time_slot, current_time_slot: DateTime,
+                     all_future_time_slots: List[DateTime]):
         """
         Simulate actual Energy flow by removing pledged storage and adding bought energy to the
         used_storage
         """
+        if GlobalConfig.FUTURE_MARKET_DURATION_HOURS:
+            # In case the future market is enabled, the future orders have to be deleted once
+            # the market becomes a spot market
+            del self.offered_buy_kWh[current_time_slot]
```
spyrostz, line 707: Thank you, that's perfect and exactly what I had in mind when we discussed this! I would prefer instead of deleting the key to have normal assignments to zero `self.offered_buy_kWh[current_time_slot] = 0.` , however it is a matter of personal preference therefore feel free to leave it as is.

spyrostz (dismissed):
2 comments, neither about the correct functionality of the fix which is good to go IMO. LGTM

src/gsy_e/models/state.py
Hassan754: Hand to hand with the below proposal (assigning it to 0) seems to be working indeed, adapted!

hannesdiedrich:
> I propose creating a refactoring ticket to write tests for all the states modules.

@Hassan754  It needs to be done indeed, but will probably never get scheduled, so I propose sticking to our rule to not reduce the coverage and add a unittest for the reset `self.offered_buy_kWh[current_time_slot]` and `self.offered_sell_kWh[current_time_slot]` in the `state.market_cycle()` and then open the mentioned ticket for all other unites in this module. What do you think?

Hassan754:
> > I propose creating a refactoring ticket to write tests for all the states modules.
> 
> @Hassan754 It needs to be done indeed, but will probably never get scheduled, so I propose sticking to our rule to not reduce the coverage and add a unittest for the reset `self.offered_buy_kWh[current_time_slot]` and `self.offered_sell_kWh[current_time_slot]` in the `state.market_cycle()` and then open the mentioned ticket for all other unites in this module. What do you think?

We have a better chance nowadays in reaching these tickets after we decided to always 1 ticket of this kind in our sprints
Anw, If we can push this fix PR I'll write a reminder to myself to write the test you mentioned on monday, it would be great, I'll make sure I address it asap, otherwise I'll create the tests first thing in the morning tomorrow in the context of this PR

Hassan754:
> > I propose creating a refactoring ticket to write tests for all the states modules.
> 
> @Hassan754 It needs to be done indeed, but will probably never get scheduled, so I propose sticking to our rule to not reduce the coverage and add a unittest for the reset `self.offered_buy_kWh[current_time_slot]` and `self.offered_sell_kWh[current_time_slot]` in the `state.market_cycle()` and then open the mentioned ticket for all other unites in this module. What do you think?

I created a [parent ticket](https://gridsingularity.atlassian.net/browse/GSYE-91) to test all the states, and several child tickets one for each state or a group of states

tests/test_state/test_storage_state.py
```diff
@@ -0,0 +1,35 @@
+from unittest.mock import patch
+
+from pendulum import now
+
+from gsy_e.models.state import StorageState
+
+
+class TestStorageState:
```
hannesdiedrich, line 8: ❤️

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #1333, opened 2021-12-08
loadhours.py deal with pylint styling warnings

Hassan754:
The pylint warnings are fixed commit by commit

### DIFF
--- a/src/gsy_e/models/strategy/load_hours.py
+++ b/src/gsy_e/models/strategy/load_hours.py
@@ -367,7 +367,7 @@ class LoadHoursStrategy(BidEnabledStrategy):
         return self._is_market_active(market) and self.state.can_buy_more_energy(market.time_slot)
 
     def _offer_comes_from_different_seller(self, offer):
-        return offer.seller != self.owner.name and offer.seller != self.area.name
+        return offer.seller not in [self.owner.name, self.area.name]
 
     def _set_alternative_pricing_scheme(self):
         if ConstSettings.MASettings.AlternativePricing.PRICING_SCHEME != 0:
--- a/src/gsy_e/models/strategy/load_hours.py
+++ b/src/gsy_e/models/strategy/load_hours.py
@@ -15,7 +15,6 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-import traceback
 from collections import namedtuple
 from logging import getLogger
 from typing import Union, Dict  # NOQA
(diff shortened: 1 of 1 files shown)


src/gsy_e/models/strategy/load_hours.py
```diff
@@ -488,19 +496,19 @@ def _is_market_active(self, market):
...
         if hrs_per_day is None:
             hrs_per_day = len(hrs_of_day)
-        if hrs_of_day is None:
-            hrs_of_day = list(range(24))
+
         self.hrs_of_day = hrs_of_day
         self._initial_hrs_per_day = hrs_per_day
 
-        if not all([0 <= h <= 23 for h in hrs_of_day]):
+        if not all((0 <= h <= 23 for h in hrs_of_day)):
```
hannesdiedrich, line 511: AFAIK, you do not need these additional brackets.

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -114,10 +119,9 @@ def state(self) -> LoadState:
         return self._state
 
     def _init_price_update(self, fit_to_limit, energy_rate_increase_per_update, update_interval,
-                           use_market_maker_rate, initial_buying_rate, final_buying_rate):
+                           initial_buying_rate, final_buying_rate):
 
         # Instantiate instance variables that should not be shared with child classes
-        self.use_market_maker_rate = use_market_maker_rate
```
hannesdiedrich, line 120: Have a look at the comment here. The initialisation of `self.use_market_maker_rate` seems to be here for a good reason.

hannesdiedrich (commented):
Two comments, LGTM once addressed.

src/gsy_e/models/strategy/load_hours.py
Hassan754: This could be right only in 1 case, if the extending class will override the `_init_price_update` method to implement some different behavior, however AFAIS there are no overriding methods of it

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -114,10 +119,9 @@ def state(self) -> LoadState:
...
         self.fit_to_limit = fit_to_limit
```
spyrostz, line 125: After looking at the code for the comment in the line right above, I found out that this member can be completely deleted. Feel free to do so, thanks!

spyrostz (commented):
2 comments, will LGTM once resolved.

src/gsy_e/models/strategy/load_hours.py
hannesdiedrich: I have to admit that I was too lazy to look deeper and find that out myself. Will do in the future though. Thanks a lot for explaining in such detail!

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #1335, opened 2021-12-09
GSYE-52: Remove dependency to sold offers in can_offer_be_posted beca…

hannesdiedrich:
…use it was already accounted for in the available_energy

### DIFF
--- a/src/gsy_e/models/strategy/__init__.py
+++ b/src/gsy_e/models/strategy/__init__.py
@@ -378,11 +378,10 @@ class Offers:
         else:
             posted_offer_energy = self.posted_offer_energy(market.id, time_slot)
 
-        posted_and_traded_offer_energy = (offer_energy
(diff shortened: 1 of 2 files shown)


spyrostz (approved):
LGTM. I think that on a cooldown period we definitely need to better harmonize State classes' management and hopefully get rid of the `sold_offer_energy` method (because IMO since the State class already reflects the traded energy, we already do the same calculation in 2 places)

Hassan754 (approved):

hannesdiedrich:
> LGTM. I think that on a cooldown period (when we have more time for refactoring issues) we definitely need to better harmonize State classes' management and hopefully get rid of the `sold_offer_energy` method (because IMO since the State class already reflects the traded energy, we already do the same calculation in 2 places)

Indeed, I was thinking the same. Thanks!

PR #1338, opened 2021-12-10
GSYE-73: Use energy buy rate of InfiniteBus as feed-in tariff.

faizan2590:
Earlier a constant value was being used to calculate feed-in tariff as reference for SavingsKPI, this PR targets it to be linked with energy buy rate of `InfiniteBus`

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -26,7 +26,8 @@ from gsy_framework.utils import get_json_dict_memory_allocation_size
 from pendulum import DateTime
 
 from gsy_e.gsy_e_core.sim_results.offer_bids_trades_hr_stats import OfferBidTradeGraphStats
-from gsy_e.gsy_e_core.util import get_market_maker_rate_from_config
+from gsy_e.gsy_e_core.util import (
+    get_market_maker_rate_from_config, get_feed_in_tariff_rate_from_config)
 from gsy_e.models.strategy.commercial_producer import CommercialStrategy
 from gsy_e.models.strategy.finite_power_plant import FinitePowerPlant
 from gsy_e.models.strategy.infinite_bus import InfiniteBusStrategy
@@ -48,6 +49,8 @@ _NO_VALUE = {
 }
 
 
+# pylint: disable=too-many-instance-attributes, unidiomatic-typecheck, too-many-branches
+# pylint: disable=logging-too-many-args, fixme
 class SimulationEndpointBuffer:
     """Handles collecting and buffering of all results for all areas."""
 
@@ -84,16 +87,16 @@ class SimulationEndpointBuffer:
 
         message_size = get_json_dict_memory_allocation_size(result_report)
         if message_size > 64000:
-            logging.error(f"Do not publish message bigger than 64 MB, "
-                          f"current message size {message_size / 1000.0} MB.")
+            logging.error("Do not publish message bigger than 64 MB, "
+                          "current message size MB.", (message_size / 1000.0))
             return {}
-        logging.debug(f"Publishing {message_size} KB of data via Redis.")
+        logging.debug("Publishing %s KB of data via Redis.", message_size)
         return result_report
 
     @staticmethod
     def _structure_results_from_area_object(target_area: "Area") -> Dict:
         """Add basic information about the area in the area_tree_dict."""
-        area_dict = dict()
+        area_dict = {}
         area_dict["name"] = target_area.name
         area_dict["uuid"] = target_area.uuid
         area_dict["parent_uuid"] = (target_area.parent.uuid
@@ -182,7 +185,8 @@ class SimulationEndpointBuffer:
                 "market_fee": area.future_markets.market_fee,
                 "const_fee_rate": (area.future_markets.const_fee_rate
                                    if area.future_markets.const_fee_rate is not None else 0.),
-                "feed_in_tariff": GlobalConfig.FEED_IN_TARIFF,
+                "feed_in_tariff": get_feed_in_tariff_rate_from_config(
+                    area.future_markets, time_slot=time_slot),
                 "market_maker_rate": get_market_maker_rate_from_config(
                     area.future_markets, time_slot=time_slot)
             }
@@ -203,10 +207,11 @@ class SimulationEndpointBuffer:
         stats_dict["market_fee"] = market.market_fee
         stats_dict["const_fee_rate"] = (market.const_fee_rate
                                         if market.const_fee_rate is not None else 0.)
-        stats_dict["feed_in_tariff"] = GlobalConfig.FEED_IN_TARIFF
+        stats_dict["feed_in_tariff"] = get_feed_in_tariff_rate_from_config(market)
         stats_dict["market_maker_rate"] = get_market_maker_rate_from_config(market)
         return stats_dict
 
+    # pylint: disable=too-many-branches
     def _populate_core_stats_and_sim_state(self, area: "Area") -> None:
         """Populate all area statistics and state into self.flattened_area_core_stats_dict and
         self.simulation_state."""
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -86,7 +86,7 @@ class SimulationEndpointBuffer:
         message_size = get_json_dict_memory_allocation_size(result_report)
         if message_size > 64000:
             logging.error("Do not publish message bigger than 64 MB, "
-                          "current message size MB.", (message_size / 1000.0))
+                          "current message size %s MB.", (message_size / 1000.0))
             return {}
         logging.debug("Publishing %s KB of data via Redis.", message_size)
         return result_report
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -48,7 +48,7 @@ _NO_VALUE = {
 
 
 # pylint: disable=too-many-instance-attributes
-# pylint: disable=logging-too-many-args, fixme
+# pylint: disable=logging-too-many-args
 class SimulationEndpointBuffer:
     """Handles collecting and buffering of all results for all areas."""
 
@@ -122,7 +122,6 @@ class SimulationEndpointBuffer:
 
     def generate_result_report(self) -> Dict:
         """Create dict that contains all statistics that are sent to the gsy-web."""
-        # TODO: In D3ASIM-2288, add unix_time=True to convert_pendulum_to_str_in_dict
         return {
             "job_id": self.job_id,
             "current_market": self.current_market_time_slot_str,
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -30,9 +30,7 @@ from gsy_e.gsy_e_core.util import (
     get_market_maker_rate_from_config, get_feed_in_tariff_rate_from_config)
 from gsy_e.models.strategy.commercial_producer import CommercialStrategy
 from gsy_e.models.strategy.finite_power_plant import FinitePowerPlant
-from gsy_e.models.strategy.infinite_bus import InfiniteBusStrategy
 from gsy_e.models.strategy.load_hours import LoadHoursStrategy
-from gsy_e.models.strategy.market_maker_strategy import MarketMakerStrategy
 from gsy_e.models.strategy.pv import PVStrategy
 from gsy_e.models.strategy.smart_meter import SmartMeterStrategy
 from gsy_e.models.strategy.storage import StorageStrategy
@@ -49,7 +47,7 @@ _NO_VALUE = {
 }
 
 
-# pylint: disable=too-many-instance-attributes, unidiomatic-typecheck, too-many-branches
+# pylint: disable=too-many-instance-attributes
 # pylint: disable=logging-too-many-args, fixme
 class SimulationEndpointBuffer:
     """Handles collecting and buffering of all results for all areas."""
@@ -283,18 +281,18 @@ class SimulationEndpointBuffer:
                 for trade in area.strategy.trades[area.parent.current_market]:
                     core_stats_dict["trades"].append(trade.serializable_dict())
 
-        elif type(area.strategy) == FinitePowerPlant:
-            core_stats_dict["production_kWh"] = area.strategy.energy_per_slot_kWh
-            if area.parent.current_market is not None:
-                for trade in area.strategy.trades[area.parent.current_market]:
-                    core_stats_dict["trades"].append(trade.serializable_dict())
-
-        elif type(area.strategy) in [InfiniteBusStrategy, MarketMakerStrategy, CommercialStrategy]:
-            if area.parent.current_market is not None:
-                core_stats_dict["energy_rate"] = (
-                    area.strategy.energy_rate.get(area.parent.current_market.time_slot, None))
-                for trade in area.strategy.trades[area.parent.current_market]:
-                    core_stats_dict["trades"].append(trade.serializable_dict())
+        elif isinstance(area.strategy, CommercialStrategy):
+            if isinstance(area.strategy, FinitePowerPlant):
+                core_stats_dict["production_kWh"] = area.strategy.energy_per_slot_kWh
+                if area.parent.current_market is not None:
+                    for trade in area.strategy.trades[area.parent.current_market]:
+                        core_stats_dict["trades"].append(trade.serializable_dict())
+            else:
+                if area.parent.current_market is not None:
+                    core_stats_dict["energy_rate"] = (
+                        area.strategy.energy_rate.get(area.parent.current_market.time_slot, None))
+                    for trade in area.strategy.trades[area.parent.current_market]:
+                        core_stats_dict["trades"].append(trade.serializable_dict())
 
         self.flattened_area_core_stats_dict[area.uuid] = core_stats_dict
 
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -182,8 +182,8 @@ class SimulationEndpointBuffer:
                 "market_fee": area.future_markets.market_fee,
                 "const_fee_rate": (area.future_markets.const_fee_rate
                                    if area.future_markets.const_fee_rate is not None else 0.),
-                "feed_in_tariff": get_feed_in_tariff_rate_from_config(
-                    area.future_markets, time_slot=time_slot),
+                "feed_in_tariff": get_feed_in_tariff_rate_from_config(area.future_markets,
+                                                                      time_slot=time_slot),
                 "market_maker_rate": get_market_maker_rate_from_config(
                     area.future_markets, time_slot=time_slot)
             }
--- a/src/gsy_e/gsy_e_core/util.py
+++ b/src/gsy_e/gsy_e_core/util.py
@@ -76,11 +76,8 @@ setLoggerClass(TraceLogger)
 log = getLogger(__name__)
 
 
-# pylint: disable=useless-super-delegation
 class TaggedLogWrapper(LoggerAdapter):
     """TaggedLogWrapper"""
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
 
     def process(self, msg, kwargs):
         msg = f"[{self.extra}] {msg}"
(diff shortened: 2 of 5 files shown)


src/gsy_e/gsy_e_core/util.py
```diff
@@ -443,21 +475,39 @@ def get_market_maker_rate_from_config(next_market, default_value=None, time_slot
...
-    else:
-        return GlobalConfig.market_maker_rate
+    return GlobalConfig.market_maker_rate
+
+
+def get_feed_in_tariff_rate_from_config(next_market, default_value=None, time_slot=None):
+    """Get feed in tariff rate from config."""
+    if next_market is None:
+        return default_value
+    if isinstance(GlobalConfig.FEED_IN_TARIFF, dict):
```
hannesdiedrich, line 485: When will the `GlobalConfig.FEED_IN_TARIFF` be a dict? I do not see any mechanism that converts from float to dict. Can you maybe help me?

src/gsy_e/models/strategy/infinite_bus.py
```diff
@@ -71,6 +73,13 @@ def _set_global_market_maker_rate(self):
         elif self.energy_rate is not None:
             GlobalConfig.market_maker_rate = self.energy_rate
 
+    def _set_global_feed_in_tariff_rate(self):
+        if self.buying_rate_profile is not None:
+            GlobalConfig.FEED_IN_TARIFF = read_arbitrary_profile(
+                InputProfileTypes.IDENTITY, self.buying_rate_profile)
+        elif self.energy_buy_rate is not None:
+            GlobalConfig.FEED_IN_TARIFF = self.energy_buy_rate
```
faizan2590, line 81: This is the actual change to set feed-in tariff as per buying rate of InfiniteBus.

src/gsy_e/gsy_e_core/util.py
```diff
@@ -443,21 +475,39 @@ def get_market_maker_rate_from_config(next_market, default_value=None, time_slot
                 raise e
         return find_object_of_same_weekday_and_time(GlobalConfig.market_maker_rate,
                                                     time_slot)
-    else:
-        return GlobalConfig.market_maker_rate
+    return GlobalConfig.market_maker_rate
+
+
+def get_feed_in_tariff_rate_from_config(next_market, default_value=None, time_slot=None):
```
hannesdiedrich, line 481: `next_market` is of `BaseMarkt` type right?
```suggestion
def get_feed_in_tariff_rate_from_config(next_market: "BaseMarket", default_value=None, time_slot=None):
```

src/gsy_e/gsy_e_core/util.py
```diff
@@ -443,21 +475,39 @@ def get_market_maker_rate_from_config(next_market, default_value=None, time_slot
...
+    """Get feed in tariff rate from config."""
+    if next_market is None:
+        return default_value
+    if isinstance(GlobalConfig.FEED_IN_TARIFF, dict):
+        if time_slot is None:
+            try:
+                time_slot = next_market.time_slot
```
hannesdiedrich, line 488: `FutureMarkets` objects also have a `time_slot` attribute, so this check will not be triggered.

src/gsy_e/gsy_e_core/util.py
hannesdiedrich: I see your comment now: https://github.com/gridsingularity/gsy-e/pull/1338/files#r767532113 Thanks!

faizan2590: https://github.com/gridsingularity/gsy-e/pull/1338/files#r767532113

src/gsy_e/models/strategy/infinite_bus.py
```diff
@@ -71,6 +73,13 @@ def _set_global_market_maker_rate(self):
         elif self.energy_rate is not None:
             GlobalConfig.market_maker_rate = self.energy_rate
 
+    def _set_global_feed_in_tariff_rate(self):
+        if self.buying_rate_profile is not None:
+            GlobalConfig.FEED_IN_TARIFF = read_arbitrary_profile(
+                InputProfileTypes.IDENTITY, self.buying_rate_profile)
```
hannesdiedrich, line 79: This is handled in `_read_or_rotate_profiles` already, so AFAIS, you can remove this if block.

hannesdiedrich (commented):
Left some comments, will review again after addressed and unit tests are added. Thanks

src/gsy_e/gsy_e_core/util.py
faizan2590: I haven't considered future market for now for SavingsKPI and would leave it for time being to not make it too complex, and once SavingsKPI has been fully tested for spot market, future market could be integrated late as per need

hannesdiedrich: I understood that, but the above check is not checking it the `next_market` is a future market. 
The area class has a `is_market_future` that could be handy, but only if you have access to the area.

Hassan754: You can also check whether the `next_market.time_slot is None`, up until now this is only possible in the future market

src/gsy_e/gsy_e_core/util.py
```diff
@@ -91,42 +94,43 @@ def trace(self, msg, *args, **kwargs):
...
             self.allowed_formats = "'XXh', 'XXd', 'XXdYYh'"
-        elif type == 'M:S':
+        elif type == "M:S":
             self.re = INTERVAL_MS_RE
             self.allowed_formats = "'XXm', 'XXs', 'XXmYYs', 'XX:YY'"
         else:
-            raise ValueError("Invalid type. Choices: 'H:M', 'M:S', 'D:H'")
+            raise ValueError("Invalid date_type. Choices: 'H:M', 'M:S', 'D:H'")
 
+    # pylint: disable=inconsistent-return-statements
```
Hassan754, line 133: Same as above

src/gsy_e/models/strategy/infinite_bus.py
```diff
@@ -30,7 +30,9 @@
 from gsy_e.models.strategy.commercial_producer import CommercialStrategy
 
 
+# pylint: disable=missing-class-docstring, too-many-instance-attributes, too-many-arguments
 class InfiniteBusStrategy(CommercialStrategy, BidEnabledStrategy):
+    """Implemenntationn for infinite bus to participate in GSy Exchange."""
```
Hassan754, line 35: Butterfly keyboards, I know the struggle :(
```suggestion
    """Implementation for infinite bus to participate in GSy Exchange."""
```

Hassan754 (commented):
General comment, AFAIS we disabled several relatively easy to fix warnings in this PR, this is not efficient, if the PR is urgent we can completely avoid fixing or dealing with warnings, but disabling warnings will deeply bury them until someone decides to deal with them
Moreover, writing primitive docstrings (copying the exact same method/class name) is also useless, we should either write a meaningful one or not at all
will review again once the comments are addressed

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -48,6 +49,8 @@
 }
 
 
+# pylint: disable=too-many-instance-attributes, unidiomatic-typecheck, too-many-branches
+# pylint: disable=logging-too-many-args, fixme
```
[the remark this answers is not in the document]
faizan2590: Using `type` rather than `isinstance` was intentional. Replacing  `type(area.strategy) == FinitePowerPlant` check with `isinstance`  as `FinitePowerPlant` inherits `CommercialStrategy`, using `isinstance` in this case would lead to classifying `CommercialStrategy` as   ´FinitePowerPlant` too, same goes for other cases too in this module. Using type helps to catch the main class excluding the derived classes

```
class CommercialStrategy:
    pass
class FinitePowerPlant(CommercialStrategy):
    pass

ib = FinitePowerPlant()

isinstance(ib, FinitePowerPlant)
>>> True
isinstance(ib, CommercialStrategy)
>>> True

```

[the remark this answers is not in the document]
spyrostz: You can resolve it though with nested `if isinstance()` : 
```
if isinstance(strategy, CommercialStrategy):
    if isinstance(strategy, FinitePowerPlant):
       ....
    if isinstance(strategy, MarketMakerStrategy):
       ....
``` 
I also think it is more readable, because it reflects the class hierarchy in the if blocks, contrary to `type` which flattens these and the first-time viewer of the code is unaware that these types are related. 
Also, in your example you did not try this:
```
cs = CommercialStrategy()
isinstance(cs, FinitePowerPlant)
```
If the above returns false (which to my knowledge it will return false), then it is possible to deal the situation with isinstance and without type.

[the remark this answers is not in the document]
spyrostz: We also discussed that whenever we come across TODO comments, we evaluate them and either remove them or fix them or add refactoring tickets. For this case there already is a refactoring ticket https://gridsingularity.atlassian.net/browse/D3ASIM-2288 which has been deprecated already (because we do not save unix times in the results but datetime strings anyway). Therefore the ticket should be closed, the TODO comment should be removed, and the pylint: disable=fixme should also be removed.

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -84,16 +87,16 @@ def prepare_results_for_publish(self) -> Dict:
 
         message_size = get_json_dict_memory_allocation_size(result_report)
         if message_size > 64000:
-            logging.error(f"Do not publish message bigger than 64 MB, "
-                          f"current message size {message_size / 1000.0} MB.")
+            logging.error("Do not publish message bigger than 64 MB, "
+                          "current message size MB.", (message_size / 1000.0))
```
spyrostz, line 91: You have missed the %s

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -203,10 +207,11 @@ def _read_market_stats_to_dict(market: "MarketBase") -> Dict:
         stats_dict["market_fee"] = market.market_fee
         stats_dict["const_fee_rate"] = (market.const_fee_rate
                                         if market.const_fee_rate is not None else 0.)
-        stats_dict["feed_in_tariff"] = GlobalConfig.FEED_IN_TARIFF
+        stats_dict["feed_in_tariff"] = get_feed_in_tariff_rate_from_config(market)
         stats_dict["market_maker_rate"] = get_market_maker_rate_from_config(market)
         return stats_dict
 
+    # pylint: disable=too-many-branches
```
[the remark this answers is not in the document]
spyrostz: Or remove the class-level warning in favor of this one. 

On another note, from what I see here Pylint has a point, this should in reality be refactored to a dispatching class / module since most of the if / elif blocks do not access state from the class and can be pure functions.

src/gsy_e/gsy_e_core/util.py
```diff
@@ -136,47 +140,52 @@ def convert(self, value, param, ctx):
...
         return False
 
 
 def make_ma_name(owner):
+    """Make market agent name."""
     return f"MA {owner.name}"
 
 
 def make_ba_name(owner):
+    """Make balanncing agent name."""
```
spyrostz, line 173: Typo: `balancing`

src/gsy_e/gsy_e_core/util.py
```diff
@@ -136,47 +140,52 @@ def convert(self, value, param, ctx):
...
     return f"BA {owner.name}"
 
 
 def make_sa_name(owner):
+    """Make settement agent name."""
```
spyrostz, line 178: Typo: `settlement`

src/gsy_e/gsy_e_core/util.py
```diff
@@ -284,12 +295,13 @@ def convert_nested_settings(class_object, class_name, settings_dict):
...
         return const_settings
-    except Exception:
+    except Exception as ex:
         raise SyntaxError("Error when serializing the const settings file. Incorrect "
-                          "setting structure.")
+                          "setting structure.") from ex
 
 
 def retry_function(max_retries=3):
+    """Retry function."""
```
spyrostz, line 304: `Decorator that retries the execution of the function until it returns without raising an exception`

src/gsy_e/gsy_e_core/util.py
```diff
@@ -299,16 +311,18 @@ def wrapped(*args, **kwargs):
 
 
 def recursive_retry(functor, retry_count, max_retries, *args, **kwargs):
+    """Recursive retry."""
```
spyrostz, line 314: `Recursive function that retries the execution of "functor" function if an exception is raised from it.`

src/gsy_e/models/strategy/infinite_bus.py
```diff
@@ -104,7 +113,9 @@ def _read_or_rotate_profiles(self, reconfigure=False):
                         profile_uuid=self.energy_rate_profile_uuid)))
 
         self._set_global_market_maker_rate()
+        self._set_global_feed_in_tariff_rate()
 
+    # pylint: disable=fixme
```
spyrostz, line 118: I think that the TODO comments in this file have to either be deleted, or better clarified in order to be better understood.

spyrostz (commented):
Have left some comments, will review once addressed. Similar to what @Hassan754 already commented on, the function docstrings need to either be improved or deleted completely. If the docstring is just a copy of the function name then it is better to skip the warning completely. Also, in case of urgencies, `SKIP=pylint` should be more favorable over `# pylint: disable=` , for the reasons that Hassan mentioned, in order for these warnings to be dealt with later.

src/gsy_e/gsy_e_core/util.py
faizan2590: I assume, you meant following,

```suggestion
def get_feed_in_tariff_rate_from_config(next_market: "MarketBase", default_value=None, time_slot=None):
```

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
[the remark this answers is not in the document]
faizan2590: I would prefer get it dealt with later with an small refactoring ticket as this PR targeted to solve bug also that its one of blocker to deploy

src/gsy_e/models/strategy/infinite_bus.py
faizan2590: I would suggest to make a refactoring ticket to monitor its relation to memory usage as simply a TODO proved to be overlooked here

Hassan754 (dismissed):
LGTM once others' comments are resolved

src/gsy_e/gsy_e_core/util.py
```diff
@@ -443,21 +481,40 @@ def get_market_maker_rate_from_config(next_market, default_value=None, time_slot
...
+        return default_value
+    if isinstance(GlobalConfig.FEED_IN_TARIFF, dict):
+        if time_slot is None:
+            try:
+                time_slot = next_market.time_slot
+            except AttributeError as e:
+                logging.exception("time_slot parameter is required for future markets.")
+                raise e
+        return find_object_of_same_weekday_and_time(GlobalConfig.FEED_IN_TARIFF,
+                                                    time_slot) or 0.
```
hannesdiedrich, line 500:
```suggestion
        time_slot = time_slot or next_market.time_slot
        if not time_slot:
            return 0
        return find_object_of_same_weekday_and_time(GlobalConfig.FEED_IN_TARIFF,
                                                    time_slot) or 0.
```

hannesdiedrich: Reason for my suggestion is that AFAIS, the exception that you are re-rasing  is not caught anywhere and would make the sim crash.

hannesdiedrich:
> Moreover, writing primitive docstrings (copying the exact same method/class name) is also useless, we should either write a meaningful one or not at all

I totally agree and thought we discussed this already multiple times. I really hope we can overcome this bad practice in future PRs and can omit these kind of comments.

hannesdiedrich (commented):
One last comment and the reminder to address the pylint comments in an follow up PR (right after this is pushed).

src/gsy_e/gsy_e_core/util.py
faizan2590: Rather than completely skipping the exception to go un notice, I would at least leave it with with `logging.error`

src/gsy_e/gsy_e_core/util.py
```diff
@@ -443,21 +481,40 @@ def get_market_maker_rate_from_config(next_market, default_value=None, time_slot
                 raise e
         return find_object_of_same_weekday_and_time(GlobalConfig.market_maker_rate,
                                                     time_slot)
-    else:
-        return GlobalConfig.market_maker_rate
+    return GlobalConfig.market_maker_rate
+
+
+def get_feed_in_tariff_rate_from_config(next_market: "MarketBase",
+                                        default_value=None, time_slot=None):
```
hannesdiedrich, line 488: I just saw this: Here you set the default_value to None but before you return 0 if the time_slot is not found. Can you please explain?

src/gsy_e/gsy_e_core/util.py
hannesdiedrich: Ok, agreed. But this case should never happen. If the next_market is a future market (i know, not for now), the time_slot parameter should be given. So actually, it would be worth an assertion here (as it would be an implementation error).
Please add the future market handling in the follow up PR. Otherwise this will end up in another hard to find bug. Thanks!

src/gsy_e/models/strategy/infinite_bus.py
spyrostz: It got overlooked/ignored because there was no clear action that had to be done here. If there is suspicion that there is a memory leak here, then this needs to be identified. If not, then the comment needs to be deleted and if / when we have a memory leak we will look at it in detail. 
Moreover, from a brief look at the code, this dict is only populated on construction. Therefore even if this `del` fails to deallocate the object due to reference counting, there is no chance that this dict will grow in time, therefore taking into account that in the vast majority of the simulations there is at most one infinite bus, this will never leak.

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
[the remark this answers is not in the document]
spyrostz: I do not think that this needs a ticket, an afterburner PR after this one would be sufficient

src/gsy_e/gsy_e_core/util.py
faizan2590: Took it as copy to `get_market_maker_rate_from_config`, it should either be some default or 0. In this case I will set it as zero

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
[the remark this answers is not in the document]
faizan2590: Okay will handle this and future market handling for feedin tariff as after burner PR

src/gsy_e/models/strategy/infinite_bus.py
faizan2590: Okay, to conclude, still its seems better to delete an unused profile here, even in this case where almost un-noticeable memory optimisation is being achieved, considering the fact that there is only one `InfiniteBus` in a simulation

spyrostz: The del should stay as it is. What I was talking about above was that there cannot be a memory leak, because this variable is only assigned once and never reassigned. Therefore even if the delete does not work, the size of this variable will never grow because there are no more assignments to it. Therefore a memory leak is not possible in this case. 

Therefore, the TODO comments are not needed here. Please delete them and remove the `# pylint: disable=fixme `

faizan2590: https://github.com/gridsingularity/gsy-e/pull/1338/commits/46fb16eca6742c005f891c418f990d94fcf665bf

spyrostz (approved):

src/gsy_e/gsy_e_core/util.py
spyrostz: @hannesdiedrich interesting indeed, I am fairly certain that I did not add it just for preventive reasons, but that I stumbled upon it. I suspect though that when I did the first implementation of this feature the future markets were a work in progress, and it might be that time_slot was missing as an argument back then. 
Nevertheless, I agree to go with your suggestion and to make it more iron-clad with an assertion. Thanks for catching it!

faizan2590: One problem with this suggestion is that at this point `time_slot = next_market.time_slot` it would have already raised an `AttributeError` exception and the next line where `time_slot ` is asserted won't be executed. SO, I propose it like following if we want it to be iron clad with an assertion

```
def get_feed_in_tariff_rate_from_config(next_market: "MarketBase", time_slot=None):
    """Get feed in tariff rate from config."""
    if next_market is None:
        return 0.
    if isinstance(GlobalConfig.FEED_IN_TARIFF, dict):
        if time_slot is None:
            try:
                time_slot = next_market.time_slot
            except AttributeError:
                assert False, "time_slot parameter is missing to get feed-in tariff"
        return find_object_of_same_weekday_and_time(GlobalConfig.FEED_IN_TARIFF,
                                                    time_slot) or 0.
    return GlobalConfig.FEED_IN_TARIFF
```

hannesdiedrich: @faizan2590 please have a look at my suggestion. It does not contain the try/except block. Also, I explained why the Assertion in you shared code will never be triggered.

faizan2590: Sorry, yes overlooked this part ` If this is not provided AND the market does not have a time_slot member, then this Exception is raised.
However, as mentioned above, also the future market has a time_slot parameter (it is just set to None). So, the AttributeError will never be triggered.`

I will adapt as per your suggestion

faizan2590: Adapted: https://github.com/gridsingularity/gsy-e/pull/1338/commits/7e6a6030b42388f904aeabb6f3e004382348fd45

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #1339, opened 2021-12-14
Fix create_future_markets to avoid creating any future market if the global future duration is 0

Hassan754:
2 Bugs were fixed in context of this PR

1. When the `FUTURE_MARKETS_DURATION_HOURS` == 0, we shouldn't create any future market but in reality, we were.
2. We were prolonging the future time slots starting from the first future market, which is not correct, for instance: FUTURE_MARKETS_DURATION_HOURS = 1, current_time_slot = 00:00:00, future_time_slot = 00:15:00, we were assigning the most_future_slot to be equal to 1 hour from the 00:15:00 slot -> 01:15:00, whereas it should be 01:00:00

### DIFF
--- a/src/gsy_e/models/market/future.py
+++ b/src/gsy_e/models/market/future.py
@@ -178,6 +178,8 @@ class FutureMarkets(TwoSidedMarket):
                               slot_length: duration,
                               config: "SimulationConfig") -> None:
         """Add sub dicts in order dictionaries for future market slots."""
+        if not GlobalConfig.FUTURE_MARKET_DURATION_HOURS:
+            return
         future_time_slot = current_market_time_slot.add(minutes=slot_length.total_minutes())
         most_future_slot = (future_time_slot +
                             duration(hours=GlobalConfig.FUTURE_MARKET_DURATION_HOURS))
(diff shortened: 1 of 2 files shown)


src/gsy_e/models/market/future.py
```diff
@@ -178,8 +178,10 @@ def create_future_markets(self, current_market_time_slot: DateTime,
                               slot_length: duration,
                               config: "SimulationConfig") -> None:
         """Add sub dicts in order dictionaries for future market slots."""
+        if not GlobalConfig.FUTURE_MARKET_DURATION_HOURS:
+            return
```
Hassan754, line 182: First bug fix

src/gsy_e/models/market/future.py
```diff
@@ -178,8 +178,10 @@ def create_future_markets(self, current_market_time_slot: DateTime,
...
         future_time_slot = current_market_time_slot.add(minutes=slot_length.total_minutes())
-        most_future_slot = (future_time_slot +
+        most_future_slot = (current_market_time_slot +
```
Hassan754, line 184: Second bug fix

spyrostz (approved):
LGTM

hannesdiedrich (approved):
LGTM

PR #1341, opened 2021-12-15
 Handle the deletion of orders in future markets

Hassan754:
The fix included in this PR is basically for when we expire old orders, we're calling `delete_offer` or `delete_bid` which will delete the order from the market's cache and will dispatch the deletion to  market agents (so that forwarded orders get deleted too); and due to the fact the the other markets are already deleting their old orders too, we will end up with `OfferNotFoundException` or `BidNotFoundException`, this PR catches the exception gracefully

### DIFF
--- a/src/gsy_e/models/market/future.py
+++ b/src/gsy_e/models/market/future.py
@@ -19,7 +19,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 from collections import UserDict
 from copy import deepcopy
 from logging import getLogger
-from typing import Dict, List, Optional, TYPE_CHECKING
+from typing import Dict, List, Optional, TYPE_CHECKING, Union
 
 from gsy_framework.constants_limits import ConstSettings, GlobalConfig, DATE_TIME_FORMAT
 from gsy_framework.data_classes import Bid, Offer
@@ -27,6 +27,7 @@ from gsy_framework.utils import is_time_slot_in_simulation_duration
 from pendulum import DateTime, duration
 
 from gsy_e.gsy_e_core.blockchain_interface import NonBlockchainInterface
+from gsy_e.gsy_e_core.exceptions import OfferNotFoundException, BidNotFoundException
 from gsy_e.models.market import GridFee
 from gsy_e.models.market import lock_market_action
 from gsy_e.models.market.two_sided import TwoSidedMarket
@@ -230,3 +231,25 @@ class FutureMarkets(TwoSidedMarket):
                               dispatch_event, adapt_price_with_fees, add_to_history,
                               seller_origin_id, seller_id, attributes, requirements, time_slot)
         return offer
+
+    @lock_market_action
+    def delete_bid(self, bid_or_id: Union[str, Bid]) -> None:
+        """Delete a bid from the market and dispatch the listeners (MA)"""
+        try:
+            super().delete_bid(bid_or_id)
+        except BidNotFoundException:
+            # In the future agents/markets, the expiration of old orders is already handled
+            # So the delete_bid may raise an BidNotFoundException when trying to dispatch
+            # the deletion event of bid that was already deleted elsewhere too
+            pass
+
+    @lock_market_action
+    def delete_offer(self, offer_or_id: Union[str, Offer]) -> None:
+        """Delete an offer from the market and dispatch the listeners (MA)"""
+        try:
+            super().delete_offer(offer_or_id)
+        except OfferNotFoundException:
+            # In the future agents/markets, the expiration of old orders is already handled
+            # So the delete_offer may raise an OfferNotFoundException when trying to dispatch
+            # the deletion event of offer that was already deleted elsewhere too
+            pass


hannesdiedrich (commented):
I am not sure about this. The `delete_offer` and `delete_bid` methods are not only used for expired orders. They are only used for deleting orders after they are traded (which actually happens often).
But let us understand why this is needed here. Are the `BidNotFoundException` and `OfferNotFoundException` raised when the `FutureMarkets._expire_orders` is called or when a trade happened?

Hassan754:
> I am not sure about this. The `delete_offer` and `delete_bid` methods are not only used for expired orders. They are only used for deleting orders after they are traded (which actually happens often). But let us understand why this is needed here. Are the `BidNotFoundException` and `OfferNotFoundException` raised when the `FutureMarkets._expire_orders` is called or when a trade happened?

Let's say that a market called `House1` forwarded a bid to the higher market called `Grid`, when we start expiring orders we start by the highest market (`Grid`), when this forwarded bid is deleted from the Grid, no listeners are interested in this event (the listeners are the parties that have the forwarded orders, not the original ones).
So later on when the `House1` wants to expire its orders, it will delete this bid, and this will trigger a deletion chain (trying to delete this forwarded bid from the other markets), the Grid is a listener and it will receive this deletion trigger, tries to delete it from its cache `self.bids.pop(bid_id)`, well at this stage the bid is already deleted so this statement will raise the `BidNotFoundException` -> this was happening in literally all market cycles when I was testing.

The other use of the `delete_bid`/`delete_offer` when we make a trade is safe because these orders are guaranteed to be there when we try to pop them

spyrostz:
> Let's say that a market called `House1` forwarded a bid to the higher market called `Grid`, when we start expiring orders we start by the highest market (`Grid`), when this forwarded bid is deleted from the Grid, no listeners are interested in this event (the listeners are the parties that have the forwarded orders, not the original ones). So later on when the `House1` wants to expire its orders, it will delete this bid, and this will trigger a deletion chain (trying to delete this forwarded bid from the other markets), the Grid is a listener and it will receive this deletion trigger, tries to delete it from its cache `self.bids.pop(bid_id)`, well at this stage the bid is already deleted so this statement will raise the `BidNotFoundException` -> this was happening in literally all market cycles when I was testing.

I think that this is not exactly right, or at least it does not exactly reflect the situation. If memory serves, even if an offer / bid is deleted on the Grid market, the event will be dispatched to its children too. Therefore the offer / bid should be deleted from all markets that it has been forwarded to (in theory at least). 
I think that if this is indeed the case, the problem is with the IAA/dispatcher and not with the market, and IMO even the future markets should not swallow these exceptions because these are potentially very useful. The reason why I am mentioning `if this is indeed the case` is because in theory future agent is a copy of the IAA, and we do not experience this behavior with the spot markets.

Hassan754:
> > Let's say that a market called `House1` forwarded a bid to the higher market called `Grid`, when we start expiring orders we start by the highest market (`Grid`), when this forwarded bid is deleted from the Grid, no listeners are interested in this event (the listeners are the parties that have the forwarded orders, not the original ones). So later on when the `House1` wants to expire its orders, it will delete this bid, and this will trigger a deletion chain (trying to delete this forwarded bid from the other markets), the Grid is a listener and it will receive this deletion trigger, tries to delete it from its cache `self.bids.pop(bid_id)`, well at this stage the bid is already deleted so this statement will raise the `BidNotFoundException` -> this was happening in literally all market cycles when I was testing.
> 
> I think that this is not exactly right, or at least it does not exactly reflect the situation. If memory serves, even if an offer / bid is deleted on the Grid market, the event will be dispatched to its children too. Therefore the offer / bid should be deleted from all markets that it has been forwarded to (in theory at least). I think that if this is indeed the case, the problem is with the IAA/dispatcher and not with the market, and IMO even the future markets should not swallow these exceptions because these are potentially very useful. The reason why I am mentioning `if this is indeed the case` is because in theory future agent is a copy of the IAA, and we do not experience this behavior with the spot markets.

There is probably a misunderstanding here, the orders' events will probably be dispatched to children areas in case they are interested (if this child contains a forwarded order), but what if it's vice versa? if the original order is in the child and the forwarded order resides in the parent area, taking into consideration that the expiration of orders happens in top down flow, when the forwarded order is deleted before the original one this will not be dispatched to the original market AFAIK

spyrostz:
> 

If the order is deleted from the parent area and was forwarded by the child, then the notification should have been broadcasted already by the parent to the child, and the child should then delete the forwarded offer, following to its source. In general, no matter where the order is deleted (to the parent or to the child), once it happens the deletion should propagate and delete all forwarded orders. TLDR, the system should never be in a case where the forwarded order is deleted on some markets and not on some others. If you really encountered this behavior, I would recommend creating a setup file that specifically reproduces it and open a ticket that will explicitly deal with this, because it is a serious bug and should not be dealt by just ignoring these exceptions.

Hassan754:
> > 
> 
> If the order is deleted from the parent area and was forwarded by the child, then the notification should have been broadcasted already by the parent to the child, and the child should then delete the forwarded offer, following to its source. In general, no matter where the order is deleted (to the parent or to the child), once it happens the deletion should propagate and delete all forwarded orders. TLDR, the system should never be in a case where the forwarded order is deleted on some markets and not on some others. If you really encountered this behavior though, I would recommend creating a setup file that specifically reproduces it and open a ticket that will explicitly deal with this, because it is a serious bug and should not be dealt by just ignoring these exceptions.

Does this mean that if a market that holds a copy of an offer, can delete the offer to its source by deleting it from its own cache?

spyrostz:
> Does this mean that if a market that holds a copy of an offer, can delete the offer to its source by deleting it from its own cache?

First of all this is definitely not a cache, the market objects really hold the offers of the market. Also in this scenario the market does not hold a copy of the offer, but a separate offer with possibly different price due to grid fees, that maintains an association with the originating offer. This is maintained in the MarketAgents / Engines by `forwarded_offers` dict if memory serves. Every time that an offer is deleted, you might have seen it from the one sided market that this `self._notify_listeners` is called. This is responsible for notifying the associated Market Area and the IAA of this and the neighboring areas that a delete offer event has occurred. The IAA engine is then capturing this event and calling the event_offer_deleted method, which checks if the offer is forwarded, and if it is, it deletes its source offer too.

Hassan754:
@spyrostz  A future market can hold two types of orders:

1. originals (are created within this market)
2. forwarded (are posted by the future agents)

When we expire the orders , and I wanted to re-mention that the expiration will happen area after area from top to down in the hierarchy , we will delete the two types hence we delete an original order alongside the forwarded copies of this order in other markets-> This is OK and it's running smoothly
However, if we delete a copy of an order, will this dispatch the deletion to the other markets? (I highly doubt that)

And to answer you, we are dealing with orders in the future markets distinctly, the `expiration` process is the responsible part here

Hassan754:
Let's resort to using examples, in case we have the following setup `Grid (parent) -> House (child)`.
Step 1: If the House market posts an offer and forwards it to the Grid market
Step 2: The `time_slot` of this order becomes outdated so we will
Step 3: Delete the offer in the Grid market (because we will start the events dispatching from the root area)
Step 4: This will trigger the notify listeners and will look for places where the agent of Grid forwarded the offer (in our case None, so nothing should happen here)
Step 5: It's the House turn to expire orders, it will delete this offer and as you mentioned it knows that this offer was forwarded to the Grid market, therefore the Grid agent will get called by the `event_offer_deleted` which will try to delete this forwarded offer and BOOM -> this will raise an exception because the offer is already deleted in step 3

spyrostz:
> @spyrostz A future market can hold two types of orders:
> 
> 1. originals (are created within this market)
> 2. forwarded (are posted by the future agents)
> 
> When we expire the orders , and I wanted to re-mention that the expiration will happen area after area from top to down in the hierarchy , we will delete the two types hence we delete an original order alongside the forwarded copies of this order in other markets-> This is OK and it's running smoothly However, if we delete a copy of an order, will this dispatch the deletion to the other markets? (I highly doubt that)

It will (assuming that you use delete_bid and delete_offer for deleting the bids/offers), I described you the exact code flow on the comment above. If for some reason this did not happen under some conditions, then it should be fixed. Furthermore, I want to emphasize again, these are not copies, the orders differ a lot from one market to another. 

> 
> And to answer you, we are dealing with orders in the future markets distinctly, the `expiration` process is the responsible part here

Yes, but if delete_bid and delete_offer are overridden as this PR intends to do, the exceptions will not be reported also for the normal flow and not just for the expiration process.

spyrostz:
> Let's resort to using examples, in case we have the following setup `Grid (parent) -> House (child)`. Step 1: If the House market posts an offer and forwards it to the Grid market -> Step 2: The `time_slot` of this order becomes outdated so we will -> Step 3: Delete the offer in the Grid market (because we will start the events dispatching from the root area) Step 4: This will trigger the notify listeners and will look for places where the agent of Grid forwarded the offer (in our case None, so nothing should happen here) Step 5: It's the House turn to expire orders, it will delete this offer and as you mentioned it knows that this offer was forwarded to the Grid market, therefore the Grid agent will get called by the `event_offer_deleted` which will try to delete this forwarded offer and BOOM -> this will raise an exception because the offer is already deleted in step 3

You are incorrect in step 4. Give a look at the Agent / Engine difference, there are 2 engines that track forwarded offers from / to. Essentially the agent is bidirectional and contains mappings for both directions, both House -> Grid and Grid -> House. Therefore the forwarded offer from the Grid to the House WILL be detected and will be deleted in step 4 already.

Hassan754:
In simple words, I don't think that it makes sense for ANY market to be able to delete an order from its origin (unless a trade occurred), but vice versa, when an original order gets deleted all forwarded orders should be deleted too

Hassan754:
> > Let's resort to using examples, in case we have the following setup `Grid (parent) -> House (child)`. Step 1: If the House market posts an offer and forwards it to the Grid market -> Step 2: The `time_slot` of this order becomes outdated so we will -> Step 3: Delete the offer in the Grid market (because we will start the events dispatching from the root area) Step 4: This will trigger the notify listeners and will look for places where the agent of Grid forwarded the offer (in our case None, so nothing should happen here) Step 5: It's the House turn to expire orders, it will delete this offer and as you mentioned it knows that this offer was forwarded to the Grid market, therefore the Grid agent will get called by the `event_offer_deleted` which will try to delete this forwarded offer and BOOM -> this will raise an exception because the offer is already deleted in step 3
> 
> You are incorrect in step 4. Give a look at the Agent / Engine difference, there are 2 engines on each agent that track forwarded offers from / to. Essentially the agent is bidirectional and contains mappings for both directions, both House -> Grid and Grid -> House. Therefore the forwarded offer from the Grid to the House WILL be detected and will be deleted in step 4 already.

So if an offer was forwarded to 10 markets, ANY market among these have the ability to delete this offer from the 9 other markets?

spyrostz:
> So if an offer was forwarded to 10 markets, ANY market among these have the ability to delete this offer from the 9 other markets?

At least at the moment yes, this is the case, and the reason for that is what started this discussion: to treat all orders equally no matter if it is a forwarded or a normal order, and to not have the system in a partial state where only some forwarded offers are deleted and not all of them (which would be the case if we followed the approach to not forward the deletions - some markets would have forwarded offers and others would not). To put it in another light though, the fact that a market has the ability to do that (which is reasonable because after all, we are not selling individual markets, but one system, therefore a delete operation should not only partially delete), does not mean that it does it. The only case where the market will delete forwarded orders is when the market closes (essentially the case that you describe at the moment).

Because you scared me though, you did not answer above: did you actually see this or you suspect it ?

Hassan754:
> Because you scared me though, you did not answer above: did you actually see this or you suspect it ?

TBH, if you try to run a simulation after activating the future markets, you will immediately face it, give a try to the default setup for instance(`GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 6`)
`d3a run` -> This will continuously raise this error at the beginning of each market slot

Hassan754:
I don't know where we're heading with our system, but I still believe that this whole deletion design has to be reconsidered and precautions have to be made, imagine having the ability to delete your neighbors offers :/

spyrostz:
> I don't know where we're heading to with our system, but I still believe that this whole deletion design has to be reconsidered and precautions have to be made, imagine having the ability to delete your neighbors offers :/

In some form this will always be the case as long as we provide a system and not individual markets. That's why we have authentication and we prohibit people from accessing and even viewing orders other than their own.

spyrostz:
> > Because you scared me though, you did not answer above: did you actually see this or you suspect it ?
> 
> TBH, if you try to run a simulation after activating the future markets, you will immediately face it, give a try to the default setup for instance(`GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 6`) `d3a run` -> This will continuously raise this error at the beginning of each market slot

Then the root cause needs to be actually debugged and not covered by the exception, as I mentioned above. Can be in the context of another ticket, but it has to be debugged because this is a deviation from the spot market case as far as I understand.

Hassan754:
> > > Because you scared me though, you did not answer above: did you actually see this or you suspect it ?
> > 
> > 
> > TBH, if you try to run a simulation after activating the future markets, you will immediately face it, give a try to the default setup for instance(`GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 6`) `d3a run` -> This will continuously raise this error at the beginning of each market slot
> 
> Then the root cause needs to be actually debugged and not covered by the exception, as I mentioned above. Can be in the context of another ticket, but it has to be debugged because this is a deviation from the spot market case as far as I understand.

In my defense I did lots of debugging and got many headaches in return as it was indeed weird and keeping up with the system is tiring, the discovery I ended up with is the one I described in the description of the ticket (so the changes in this PR were meant to fix the bug not cover it), but now I'm worried after your statements and the fact that this my assumption was not realistic, I will give it another look and get back to you with what I find

spyrostz:
> @spyrostz by the way, I did look again and I insist that the deletion is not bidirectional and the bug I described indeed makes sense, please have a look at the `event_offer_deleted` `src/gsy_e/models/strategy/market_agents/one_sided_engine.py:207`, the `if offer_info.source_offer.id == offer.id:` condition may illustrate this

You may have a point there, but I need to check it out to verify whether this is the case. I am tempted to say that the delete event will broadcast to the child IAAs and they will handle the deletion of the order, however I cannot say that with certainty unless I test it, that's why I was pushing in the other comments to isolate the issue and find the exact root cause.

spyrostz:
> > > I don't know where we're heading to with our system, but I still believe that this whole deletion design has to be reconsidered and precautions have to be made, imagine having the ability to delete your neighbors offers :/
> > 
> > 
> > In some form this will always be the case as long as we provide a system and not individual markets, I do not think that this is something irrational. That's why we have authentication and we prohibit people from accessing and deleting orders other than their own. To put it even further, in the deployed system a delete request to a market (like the one you describe) would have not even reached the simulation. Furthermore, I was under the impression that external agents /aggregators are not allowed direct access to the market by only via strategies, therefore we are only talking about internal market operation here.
> 
> AFAIK the users can list all orders available in their markets (even if they're not theirs), including ones that were forwarded from other markets, but I'm not sure about the deletion permissions though, and both cases don't change the fact that the system do allow this action

I do not agree. Where did we really add direct access to markets in the system ? To my knowledge, access to the market is routed through strategies, and IIRC the strategy is checked first, in order to validate that the order that is about to be deleted was posted by the strategy. If I am wrong, then we need to revisit the whole external strategy / aggregator implementation, the markets have really nothing to do with it.

spyrostz:
> > > > I don't know where we're heading to with our system, but I still believe that this whole deletion design has to be reconsidered and precautions have to be made, imagine having the ability to delete your neighbors offers :/
> > > 
> > > 
> > > In some form this will always be the case as long as we provide a system and not individual markets, I do not think that this is something irrational. That's why we have authentication and we prohibit people from accessing and deleting orders other than their own. To put it even further, in the deployed system a delete request to a market (like the one you describe) would have not even reached the simulation. Furthermore, I was under the impression that external agents /aggregators are not allowed direct access to the market by only via strategies, therefore we are only talking about internal market operation here.
> > 
> > 
> > AFAIK the users can list all orders available in their markets (even if they're not theirs), including ones that were forwarded from other markets, but I'm not sure about the deletion permissions though, and both cases don't change the fact that the system do allow this action
> 
> I do not agree. Where did we really add direct access to markets in the system ? To my knowledge, access to the market is routed through strategies, and IIRC the strategy is checked first, in order to validate that the order that is about to be deleted was posted by the strategy. If I am wrong, then we need to revisit the whole external strategy / aggregator implementation, the markets have really nothing to do with it though.

Hassan754:
> > @spyrostz by the way, I did look again and I insist that the deletion is not bidirectional and the bug I described indeed makes sense, please have a look at the `event_offer_deleted` `src/gsy_e/models/strategy/market_agents/one_sided_engine.py:207`, the `if offer_info.source_offer.id == offer.id:` condition may illustrate this
> 
> You may have a point there, but I need to check it out to verify whether this is the case. I am tempted to say that the delete event will broadcast to the child IAAs and they will handle the deletion of the order, however I cannot say that with certainty unless I test it, that's why I was pushing in the other comments to isolate the issue and find the exact root cause.

Unfortunately the terminologies are not unified between us, we should distinguish between `target`/`source` and `parent`/`child`, when I say that the deletion is not bidirectional I mean that it will only delete a target order if the source order is deleted, and not vice versa. As for the parent/child they both can interchangeably play the role of a target/source market

spyrostz:
> Unfortunately the terminologies are not unified between us, we should distinguish between `target`/`source` and `parent`/`child`, when I say that the deletion is not bidirectional I mean that it will only delete a target order if the source order is deleted, and not vice versa. As for the parent/child they both can interchangeably play the role of a target/source market

That's what I was trying to mention before. source - target are specific terms for the IAA Engines, because they need to be bidirectional (one engine from the parent to the child, and a second one from the child to the parent) . Therefore it does not make sense to refer to this with child - parent terms, since this implies a direction, while the engine does not.

hannesdiedrich:
I finally found the time to try it out myself (setting `GlobalConfig.FUTURE_MARKET_DURATION_HOURS = 6` with the command: `gsy-e -l DEBUG run --setup two_sided_market.default_2a -t 60s -s 60m --seed 0`) and like to summarise what I found out (not sure whether this has been mentioned here before, sorry If I repeat):

This is the chain of the event for the above mentioned example:
1. At the beginning of the market slot, `delete_orders_in_old_future_markets` is triggered which deletes the `H1 Storage1` offer in the Grid market which triggers also the deletion of the original offer in the House 1 market.
2. The first step however also dispatches an OFFER_DELETED event which is consumed by the MarketAgent between the Grid and House market (one of its engines). This will trigger the deletion of the offer that was already deleted in step 1 and obviously fail in the with the mentioned OfferNotFoundException.

AFAIS, the solution is quite simple for this part: 
For the expiration of the orders we need to omit the dispatching of an OFFER_DELETED or BID_DELETED event as the FutureMarkets class will deal with the correct deletion itself. In the  FutureEngine, the `clean_up_order_buffers` methods handles the deletion of the orders in the buffers already, so we are covered there, too. 
So, I propose to add a "dispatch_event" flag to the `delete_offer` and `delete_bid` methods in the market that toggles the event dispatching or you overwrite these methods in the Future markets and omit the event dispatching there.
(I personally prefer the first proposal as it reduces code duplication).

I implemented quickly and tried it out. It revealed another issue:
The future strategy tries to update the offer rate in the `update_offer_rates` and tries to delete the old offer first before adding the updated one. The old offer however was already deleted by the future market because it is from an expired future market slot. This is a result of the fact that the strategy also keeps track of the posted orders. By deleting the order from the market side, the two buffers go out of sync. However, this is only due to the nature of the future markets and is not a wrong design IMO.
Here we indeed somehow need to ignore cases where the order can not be found in the market (like Hassan proposed in this PR). 

Another solution for the latter would be to only expire orders in the future market one time_slot later. Meaning to only delete the 12:00 future market slot when the current spot market slot is 12:15, which would result in a concept of a "Past-Future Market". With this the strategies still would have access to the old orders in the inactive future market but nobody can trade there. But while writing this I think it is not a good approach actually.

What do you all think @spyrostz @Hassan754

Hassan754:
@hannesdiedrich In the point number 1, AFAIS when we delete the orders in the grid they are not deleted in the House1 (the OfferDeleted is dispatched but it will delete all orders forwarded **FROM** the Grid, not **TO** it).
So when we try to expire orders in the House1 market, it will dispatch the OfferDeleted event and this will delete orders forwarded FROM House 1 (the ones in Grid market in our case), but yes, here it's already late because the forwarded orders were already deleted in the Grid.
I also validate that the `update_offer_rates` issue exists and I touched it in some simulation runs

hannesdiedrich:
@spyrostz @Hassan754 we should continue this discussion.
Let me copy paste what I wrote Spyros in a private chat some time ago:
IMO the market should expire the orders, because we do not delete all orders and their forwarded orders before the market is closed in the spot market either. We just need to tell the future strategies that there are outdated orders.

spyrostz:
> @spyrostz @Hassan754 we should continue this discussion. Let me copy paste what I wrote Spyros in a private chat some time ago: IMO the market should expire the orders, because we do not delete all orders and their forwarded orders before the market is closed in the spot market either. We just need to tell the future strategies that there are outdated orders.

@hannesdiedrich I have not forgotten this, please give me some more time to debug a behavior I stumbled upon the event dispatcher that is probably the root cause for this issue. Will ping you as soon as I find out exactly what happens.

PR #1342, opened 2021-12-16
GSYE-95: Update energy requirements for loadhours in future markets

Hassan754:
[LoadHoursStrategy]: The energy requirements of future time slots were not getting updated as they were supposed to be .this is why they were not placing bids in the future markets, this PR adds support for this functionality

### DIFF
--- a/src/gsy_e/models/area/event_dispatcher.py
+++ b/src/gsy_e/models/area/event_dispatcher.py
@@ -223,7 +223,8 @@ class AreaDispatcher:
     @staticmethod
     def _create_agent_object(owner: "Area", higher_market: MarketBase,
                              lower_market: MarketBase, market_type: AvailableMarketTypes
-                             ) -> Union[OneSidedAgent, SettlementAgent, BalancingAgent]:
+                             ) -> Union[OneSidedAgent, SettlementAgent,
+                                        BalancingAgent, FutureAgent]:
         agent_constructor_arguments = {
             "owner": owner,
             "higher_market": higher_market,
@@ -325,6 +326,8 @@ class AreaDispatcher:
         self._delete_past_agents(self._spot_agents)
         self._delete_past_agents(self._balancing_agents)
         self._delete_past_agents(self._settlement_agents)
+        if self._future_agent:
+            self._future_agent.delete_engines()
 
     def _delete_past_agents(
             self, market_agent_member: Dict[DateTime, Union[
--- a/src/gsy_e/models/market/future.py
+++ b/src/gsy_e/models/market/future.py
@@ -207,7 +207,8 @@ class FutureMarkets(TwoSidedMarket):
                                         "method in future markets.")
         bid = super().bid(price=price, energy=energy, buyer=buyer, buyer_origin=buyer_origin,
                           bid_id=bid_id, original_price=original_price,
-                          add_to_history=True, adapt_price_with_fees=adapt_price_with_fees,
+                          add_to_history=add_to_history,
+                          adapt_price_with_fees=adapt_price_with_fees,
                           buyer_origin_id=buyer_origin_id, buyer_id=buyer_id,
                           attributes=attributes, requirements=requirements, time_slot=time_slot)
         return bid
--- a/src/gsy_e/models/market/future.py
+++ b/src/gsy_e/models/market/future.py
@@ -205,7 +205,7 @@ class FutureMarkets(TwoSidedMarket):
                                         "method in future markets.")
         bid = super().bid(price=price, energy=energy, buyer=buyer, buyer_origin=buyer_origin,
                           bid_id=bid_id, original_price=original_price,
-                          add_to_history=False, adapt_price_with_fees=adapt_price_with_fees,
+                          add_to_history=True, adapt_price_with_fees=adapt_price_with_fees,
                           buyer_origin_id=buyer_origin_id, buyer_id=buyer_id,
                           attributes=attributes, requirements=requirements, time_slot=time_slot)
         return bid
--- a/src/gsy_e/models/market/one_sided.py
+++ b/src/gsy_e/models/market/one_sided.py
@@ -329,7 +329,7 @@ class OneSidedMarket(MarketBase):
         if already_tracked is False:
             self._update_stats_after_trade(trade, offer)
             log.info(f"{self._debug_log_market_type_identifier}[TRADE][OFFER] "
-                     f"[{self.name}] [{self.time_slot_str}] {trade}")
+                     f"[{self.name}] [{trade.time_slot}] {trade}")
 
         # TODO: Use non-blockchain non-event-driven version for now for both blockchain and
         # normal runs.
--- a/src/gsy_e/models/strategy/load_hours.py
+++ b/src/gsy_e/models/strategy/load_hours.py
@@ -19,7 +19,7 @@ from collections import namedtuple
 from logging import getLogger
 from typing import Union, Dict, List  # NOQA
 
-from gsy_framework.constants_limits import ConstSettings
+from gsy_framework.constants_limits import ConstSettings, GlobalConfig
 from gsy_framework.data_classes import Offer
 from gsy_framework.enums import SpotMarketTypeEnum
 from gsy_framework.exceptions import GSyDeviceException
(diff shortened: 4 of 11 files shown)


src/gsy_e/models/area/event_dispatcher.py
```diff
@@ -325,6 +326,8 @@ def event_market_cycle(self) -> None:
         self._delete_past_agents(self._spot_agents)
         self._delete_past_agents(self._balancing_agents)
         self._delete_past_agents(self._settlement_agents)
+        if self._future_agent:
+            self._future_agent.delete_engines()
```
hannesdiedrich, line 330: Thanks so much! I am really surprised what we forgot in the initial round.

src/gsy_e/models/market/future.py
```diff
@@ -207,7 +207,7 @@ def bid(self, price: float, energy: float, buyer: str, buyer_origin: str,
                                         "method in future markets.")
         bid = super().bid(price=price, energy=energy, buyer=buyer, buyer_origin=buyer_origin,
                           bid_id=bid_id, original_price=original_price,
-                          add_to_history=False, adapt_price_with_fees=adapt_price_with_fees,
+                          add_to_history=True, adapt_price_with_fees=adapt_price_with_fees,
```
hannesdiedrich, line 210: Not sure, but shouldn't this be:
```suggestion
                          add_to_history= add_to_history, adapt_price_with_fees=adapt_price_with_fees,
```

src/gsy_e/models/market/one_sided.py
```diff
@@ -329,7 +329,7 @@ def accept_offer(self, offer_or_id: Union[str, Offer], buyer: str, *,
         if already_tracked is False:
             self._update_stats_after_trade(trade, offer)
             log.info(f"{self._debug_log_market_type_identifier}[TRADE][OFFER] "
-                     f"[{self.name}] [{self.time_slot_str}] {trade}")
+                     f"[{self.name}] [{trade.time_slot}] {trade}")
```
hannesdiedrich, line 332: 👍

src/gsy_e/models/strategy/predefined_load.py
```diff
@@ -110,7 +110,7 @@ def event_market_cycle(self):
         self._read_or_rotate_profiles()
         super().event_market_cycle()
 
-    def _update_energy_requirement_future_markets(self):
+    def _update_energy_requirement_spot_market(self):
```
hannesdiedrich, line 113: NIT: Now that we have two different methods, I would propose to add a wrapper methods that contains both in order not to call `_update_energy_requirement_future_markets` inside of `_update_energy_requirement_spot_market`
```python
def _update_energy_requirement_in_state(self):
    self._update_energy_requirement_spot_market()
    self._update_energy_requirement_future_markets()
```

tests/strategies/market_agents/test_future_agent.py
```diff
@@ -48,12 +48,3 @@ def test__init__creates_engines_of_correct_type(future_agent: FutureAgent) -> No
...
             assert isinstance(engine, FutureEngine)
-
-    @staticmethod
-    def test_event_market_cycle_calls_engine_clean_up_buffers(future_agent: FutureAgent) -> None:
-        """Test whether event_market_cycle calls clean_up_buffers of all engines."""
-        for engine in future_agent.engines:
-            engine.clean_up_order_buffers = Mock()
-        future_agent.event_market_cycle()
-        for engine in future_agent.engines:
-            engine.clean_up_order_buffers.assert_called_once()
```
hannesdiedrich, line 59: CAn you please explain? This method is still used, right?

hannesdiedrich (commented):
Left some comments, will review again. Thanks!

src/gsy_e/models/market/future.py
Hassan754: Not really, in the future market case we have to always set it to True because the `offer_history` and `bid_history` is what we use while exporting data

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -512,12 +512,22 @@ def assign_hours_of_per_day(self, hrs_of_day: List[int], hrs_per_day: List[int])
             raise ValueError("Length of list 'hrs_of_day' must be greater equal 'hrs_per_day'")
 
     def _update_energy_requirement_future_markets(self):
+        if not GlobalConfig.FUTURE_MARKET_DURATION_HOURS:
+            return
+        for time_slot in self.area.future_market_time_slots:
+            desired_energy_Wh = (
+                self.energy_per_slot_Wh
+                if self._allowed_operating_hours(time_slot) else 0.0)
```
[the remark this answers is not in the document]
Hassan754: Can you help me understand the `hrs_per_day` variable and what it's keeping track of? AFAIS in the constructor we're defining its type as `Dict[int, int]` whereas in the `assign_hours_of_per_day` we're dealing with it as a list of integers
Something seems off here

tests/strategies/market_agents/test_future_agent.py
Hassan754: We already have several tests covering the cleanup behavior in `tests/strategies/market_agents/test_future_engine.py`,  `test_clean_up_order_buffers_removes_all_traces_of_unneeded_bids` , `test_clean_up_order_buffers_removes_all_traces_of_unneeded_offers`

src/gsy_e/models/strategy/load_hours.py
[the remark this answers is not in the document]
hannesdiedrich: `self.hrs_per_day` contains the hours per day in a day-hours mapping. So, if there are 5 simulation days and you set the constructor parameter `hrs_per_day` to 3, the Dict would look like:
```python
self.hrs_per_day = {1:3 , 2:3, 3:3, 4:3, 5:3}
```
AFAIS, in the assign_hours_of_per_day we are not handling it as a list. Please show we exactly where. Thanks

src/gsy_e/models/market/future.py
hannesdiedrich: We do not want to put all the bids into the `bid_history`. Have a look at the `split_bid` method, the bid that is boing matched, should not be part of the history, as the original bid already is. The residual is set to be `add_to_history=True`. This is why you need to inherit the setting here, as I suggested.

src/gsy_e/models/strategy/load_hours.py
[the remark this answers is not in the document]
Hassan754: This type hint is in the method header: `hrs_per_day: List[int]`, and have a look at the assignment: 
```python
        if hrs_per_day is None:
            hrs_per_day = len(hrs_of_day) # Here we're assigning its value to be an integer
```

src/gsy_e/models/market/future.py
Hassan754: Ah I see, you're right, adapted

src/gsy_e/models/strategy/load_hours.py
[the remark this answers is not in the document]
hannesdiedrich: I see now, this method is used for assigning `self.hrs_of_day` and `self._initial_hrs_per_day` . The latter is later assigned to `self.hrs_per_day` in `add_entry_in_hrs_per_day`.
The type hint is indeed wrong, it should be `int`.

hannesdiedrich (approved):
LGTM

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -512,12 +515,31 @@ def assign_hours_of_per_day(self, hrs_of_day: List[int], hrs_per_day: List[int])
...
+    def _update_energy_requirement_in_state(self):
+        self._update_energy_requirement_spot_market()
+        self._update_energy_requirement_future_markets()
+
+    def _update_energy_requirement_spot_market(self):
         self.energy_per_slot_Wh = convert_W_to_Wh(
             self.avg_power_W, self.simulation_config.slot_length)
         desired_energy_Wh = (self.energy_per_slot_Wh
                              if self._allowed_operating_hours(self.area.spot_market.time_slot)
                              else 0.0)
```
Hassan754, line 540: I have the feeling that we should here substract the energy traded in this slot (in the future)

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #1347, opened 2021-12-21
GSYE-59: Adapted savings percentage KPI in the integration test, the …

spyrostz:
…introduction of M-N matches results in increased trade from the storage compared to the recursive 1-1 matching method that was used till now, thus increasing the savings percentage.


hannesdiedrich (approved):
LGTM

Hassan754 (approved):

PR #1348, opened 2021-12-22
GSYE-105: Add market_type to the bid/offer commands

Hassan754:


src/gsy_e/models/strategy/external_strategies/__init__.py
```diff
@@ -367,11 +367,13 @@ def _offer_aggregator_impl(
                 "offer": offer.to_json_string(replace_existing=replace_existing),
                 "transaction_id": arguments.get("transaction_id"),
                 "area_uuid": self.device.uuid,
+                "market_type": market.__class__.__name__,
```
Hassan754, line 370: This will be either `OneSided` `TwoSided`, `FutureMarket` or `SettlementMarket`

fievelk, line 370: I would recommend creating a property for this to hide the implementation detail and the nested calls. I would suggest for example `type_name` (since `type` is a reserved keyword in Python) or `type_`

fievelk (commented):
A personal opinion about this: as a user, I would find it a bit confusing to have these possible outputs: `FutureMarkets`, `SettlementMarkets` and then also `OneSidedMarket` and `TwoSidedMarket`. Instead of the last two, I would expect something like `SpotMarket`. 
That's also because a `TwoSidedMarket` can still be future/settlement, so it doesn't exactly solve the confusion I think.

hannesdiedrich:
> 1. as a user, I would find it a bit confusing to have these possible outputs: `FutureMarkets`, `SettlementMarkets` and then also `OneSidedMarket` and `TwoSidedMarket`. Instead of the last two, I would expect something like `SpotMarket`.
>    That's also because a `TwoSidedMarket` can still be spot/future/settlement (in theory, not in our classes) so it doesn't solve the confusion of the end user, I think.

I agree with that. I suggest to create a method/property that translates the class name into a user-friendly string (only distinguishes between future, settlement and spot). 
Additionally, all this could be handled solely on the sdk side, because we know the settlement market slots there. The future market slots could be destinguished simply by comparing the time_slot parameter with the current time. But maybe I am missing something here.

Hassan754:
@fievelk @hannesdiedrich Thanks for your suggestions, I'll create a property in the market classes to return a client side name, I prefer this over doing it in the client side to not bind these names to any logic which can change overtime, and also this is not scalable imo

src/gsy_e/models/market/two_sided.py
```diff
@@ -429,3 +429,8 @@ def replace_recommendations_with_residuals(recommendation: Dict):
             recommendations = [replace_recommendations_with_residuals(recommendation)
                                for recommendation in recommendations]
         return recommendations
+
+    @staticmethod
+    def representation_name():
+        """Return the market type representation."""
+        return "Spot Market"
```
fievelk, line 436: I think this is not needed, since `TwoSidedMarket` inherits from `OneSidedMarket`.

fievelk:
I'm not completely sold on the `representation_name` naming of the property (I think that `type_name` might have been a bit clearer). However, this is not a big deal for me.
I just left one comment about the inheritance of the property in two sided markets.

src/gsy_e/models/market/two_sided.py
Hassan754: It indeed does, I have removed from here

fievelk (approved):

hannesdiedrich (approved):
LGTM

PR #1352, opened 2022-01-05
Bug/gsye 123

spyrostz:
The root cause of the bug was that the _recommendations buffer on the External Myco Matcher was not cleared under certain occasions (more specifically, when the full recommendation set was rejected for some reason - for instance, when the Myco was not fast enough and was sending recommendations for a market slot after the slot was deleted / replaced with a new one). This had the effect that the buffer was always populated, thus always rejecting any new recommendations that might arrive by the Myco. The fix was to clear the buffer even when the recommendation set was fully rejected.

In the context of the PR, I also added pylint docstring exceptions for test functions and classes. This will be dealt by this ticket for all repos https://gridsingularity.atlassian.net/browse/GSYE-63, however since these Pylint warnings got in the way I added the fix for gsy-e.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -477,7 +477,7 @@ name-group=
 
 # Regular expression which should only match function or class names that do
 # not require a docstring.
-no-docstring-rgx=^_
+no-docstring-rgx=^_|^test_|^Test|^step_impl|^setup_method|^teardown_method
 
 # List of decorators that produce properties, such as abc.abstractproperty. Add
 # to this list to register other decorators that produce valid properties.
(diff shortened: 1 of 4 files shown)


.pylintrc
```diff
@@ -477,7 +477,7 @@ name-group=
 
 # Regular expression which should only match function or class names that do
 # not require a docstring.
-no-docstring-rgx=^_
+no-docstring-rgx=^_|^test_|^Test|^step_impl|^setup_method|^teardown_method
```
hannesdiedrich, line 480: ❤️

hannesdiedrich (approved):
Thanks for the explanation. LGTM

Hassan754 (dismissed):
Thanks a lot for the fix, It seems to be hard to spot, LGTM

Hassan754 (approved):

hannesdiedrich (approved):
LGTM

PR #1354, opened 2022-01-10
 Remove make_ma/ba/sa/fa_name functions and their usages

Hassan754:


src/gsy_e/models/strategy/market_agents/two_sided_engine.py
```diff
@@ -50,8 +50,6 @@ def __repr__(self):
     def _forward_bid(self, bid):
         if bid.buyer == self.markets.target.name:
             return None
-        if self.owner.name == self.markets.target.name:
-            return None
```
Hassan754, line 54: These lines were removed as a fix for the market agents' integration tests, I couldn't see the rational behind their existence in the first place.
FYI @spyrostz @hannesdiedrich

spyrostz: Are you certain that the integration tests pass with this change, and that the same bid does not get recreated to the same area? I need to run this part of the code to tell you with certainty, but I vaguely recall that this condition ensured that the bid will not be recreated on the same market. For instance, if this agent/engine is controlled by market House 1 and the forward_bid is called from the engine higher_market -> House 1 with a bid that has already been created in House 1, then the same bid would be recreated on the House 1 market. Again, this is from my memory, so take this with a grain of salt. 
Could be that more code has been added meanwhile that makes this condition useless, however I think that it is good to have a second layer of protection here. Why did this condition need to be removed in the first place in order for integration tests to pass ?

Hassan754: @spyrostz AFAIK the source and target markets are actual markets, not agents; therefore their names shouldn't have MA... prefixes before their names.
Looking again at the lines I removed:
```python
if self.owner.name == self.markets.target.name:
```
the owner name used to have a prefix (MA House 1 for instance) whereas the `self.markets.target.name` shouldn't have such prefix (even if the target market is House 1, this condition will still be True). 
This is why I believe that this condition wasn't actually needed

hannesdiedrich:
@Hassan754 I forgot to ask you last time: Did you try to run the whole workflow (from UI->gsy-web->gsy-e), so to start a simulation from the UI locally with these changes? If not, it would be nice in order to be 100% sure that there are no code fragments in the gsy-web or gsy-framework that depend on the changes made here. Thanks!

faizan2590:
> @Hassan754 I forgot to ask you last time: Did you try to run the whole workflow (from UI->gsy-web->gsy-e), so to start a simulation from the UI locally with these changes? If not, it would be nice in order to be 100% sure that there are no code fragments in the gsy-web or gsy-framework that depend on the changes made here. Thanks!

Since he I guess not able to run UI locally, I did it on his request, so the test simulation finished successfully.

https://user-images.githubusercontent.com/33482427/150497354-4471c405-89d4-4f2f-9f1c-589f350246c6.mov

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #1355, opened 2022-01-11
GSYE-109: Send settlement market orders to external myco

fievelk:
Enable the external matching for orders in settlement markets. The `"settlement_markets"` content was already available to the external myco matcher, but it wasn't being used to send orders to the myco SDK.

### DIFF
--- a/src/gsy_e/models/myco_matcher/myco_external_matcher.py
+++ b/src/gsy_e/models/myco_matcher/myco_external_matcher.py
@@ -99,7 +99,7 @@ class MycoExternalMatcher(MycoMatcherInterface):
                     continue
 
                 # Cache the market (needed while matching)
-                for market in area_data["markets"]:
+                for market in area_data["markets"] + area_data.get("settlement_markets", []):
                     self.area_markets_mapping.update(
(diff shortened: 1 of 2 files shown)


src/gsy_e/models/myco_matcher/myco_external_matcher.py
```diff
@@ -97,6 +97,7 @@ def _publish_orders(self):
                 if filtered_areas_uuids and area_uuid not in filtered_areas_uuids:
                     # Client is uninterested in this Area -> skip
                     continue
+
                 # Cache the market (needed while matching)
                 for market in area_data["markets"]:
```
Hassan754, line 102:
```suggestion
                for market in area_data["markets"] + area_data["settlement_markets"]:
```

Hassan754 (commented):
Left one proposal comment, the settlement markets are to be dealt with the same as normal spot markets because for these we're creating one object for each slot (different than the future markets which is one object holding all slots).
Once the comment is resolved, LGTM

Hassan754 (dismissed):

Hassan754 (approved):

faizan2590 (approved):

PR #1357, opened 2022-01-12
GSYE-108: Remove loss per hours mechanism in storage assets

fievelk:
The "loss per hour" mechanism was used to gradually reduce the energy stored by storage assets. Since this mechanism is generating issues and users have no control over it, we are deprecating it. 
We might think of re-introducing it in the future if a good use-case scenario for it will arise.

### DIFF
--- a/src/gsy_e/models/strategy/external_strategies/storage.py
+++ b/src/gsy_e/models/strategy/external_strategies/storage.py
@@ -441,7 +441,7 @@ class StorageExternalMixin(ExternalMixin):
         if not self.connected and not self.is_aggregator_controlled:
             super().event_tick()
         else:
-            self.state.tick(self.area, self.spot_market.time_slot)
+            self.state.check_state(self.spot_market.time_slot)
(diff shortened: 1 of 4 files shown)


hannesdiedrich (approved):
Did I mention I like to deprecate stuff? Thanks for this beauty! :-) 
LGTM

src/gsy_e/models/strategy/external_strategies/storage.py
```diff
@@ -441,7 +441,7 @@ def event_tick(self) -> None:
         if not self.connected and not self.is_aggregator_controlled:
             super().event_tick()
         else:
-            self.state.tick(self.area, self.spot_market.time_slot)
+            self.state.check_state(self.spot_market.time_slot)
```
spyrostz, line 444: Thanks for maintaining this! It is really more or less a failsafe/sanity check, but has been very useful for detecting issues over the years

spyrostz (approved):
LGTM, I also share Hannes' sentiments !

PR #1361, opened 2022-01-17
GSYE-111: Validate the required energy of recommendations

Hassan754:
When the recommendation is created based on orders requirements, we have to use the values inside these requirements for validation Instead of validating the original members of the orders.

### DIFF
--- a/src/gsy_e/models/market/two_sided.py
+++ b/src/gsy_e/models/market/two_sided.py
@@ -330,20 +330,29 @@ class TwoSidedMarket(OneSidedMarket):
 
     @staticmethod
     def _validate_requirements_satisfied(
-            bid: Bid, offer: Offer, clearing_rate: float = None,
-            selected_energy: float = None) -> None:
+            recommendation: BidOfferMatch) -> None:
         """Validate if both trade parties satisfy each other's requirements.
 
         :raises:
             InvalidBidOfferPairException: Bid offer pair failed the validation
         """
-        if ((offer.requirements or bid.requirements) and
-                not RequirementsSatisfiedChecker.is_satisfied(
-                    offer=offer, bid=bid, clearing_rate=clearing_rate,
-                    selected_energy=selected_energy)):
-            # If no requirement dict is satisfied
+        offer_requirement = {}
+        bid_requirement = {}
+        if (recommendation.matching_requirements or {}).get("offer_requirement"):
+            offer_requirement = recommendation.matching_requirements["offer_requirement"]
+        if (recommendation.matching_requirements or {}).get("bid_requirement"):
+            bid_requirement = recommendation.matching_requirements["bid_requirement"]
+        if not (
+                RequirementsSatisfiedChecker.is_bid_requirement_satisfied(
+                    recommendation.offer, recommendation.bid, bid_requirement,
+                    recommendation.trade_rate, recommendation.selected_energy) and
+                RequirementsSatisfiedChecker.is_offer_requirement_satisfied(
+                    recommendation.offer, recommendation.bid, offer_requirement,
+                    recommendation.trade_rate, recommendation.selected_energy)):
+            # If requirements are not satisfied
             raise InvalidBidOfferPairException(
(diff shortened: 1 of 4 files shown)


src/gsy_e/models/market/two_sided.py
```diff
@@ -337,48 +330,59 @@ def match_recommendations(
...
+        selected_energy = recommendation.selected_energy
+        clearing_rate = recommendation.trade_rate
+        market_offer = self.offers.get(recommendation.offer["id"])
+        market_bid = self.bids.get(recommendation.bid["id"])
+
+        if not (market_offer and market_bid):
+            # If not all offers bids exist in the market, skip the current recommendation
+            raise InvalidBidOfferPairException("Not all bids and offers exist in the market.")
+        bid_energy = recommendation.bid_energy
+        offer_energy = market_offer.energy
```
hannesdiedrich, line 371: As both variables are only used once in this method, you can omit declaring the additional variables here, right?

src/gsy_e/models/market/two_sided.py
```diff
@@ -337,48 +330,59 @@ def match_recommendations(
...
         if selected_energy > bid_energy:
             raise InvalidBidOfferPairException(
                 f"Energy traded {selected_energy} is higher than bids energy {bid_energy}.")
```
hannesdiedrich, line 374: Why are we returning the `bid_energy` here and the `offer_energy` a few lines below in the bid_energy_rate but don't do it in the error messages for the rates?

tests/test_two_sided_market.py
```diff
@@ -391,56 +431,16 @@ def test_matching_list_gets_updated_with_residual_offers(self):
...
-    for p in patches:
-        p.start()
-    yield
-    for p in patches:
-        p.stop()
-
-
 class TestTwoSidedMarketMatchRecommendations:
     """Class Responsible for testing two sided market's matching functionality."""
-    def test_match_recommendations_fake_offer_bid(self, market, two_sided_market_matching):
```
hannesdiedrich, line 411: Why was this removed?

hannesdiedrich (commented):
Two minor questions, LGTM once answered.

tests/test_two_sided_market.py
Hassan754: I have moved the validation of existence of orders from the `match_recommendations` to the `validate_bid_offer_match` method, thus this test was removed and another test was added, please have a look at the `test_validate_bid_offer_match_orders_do_not_exist`

hannesdiedrich: I see, thanks and sorry for overseeing!

hannesdiedrich (approved):

tests/test_two_sided_market.py
```diff
@@ -280,13 +315,14 @@ def test_market_trade_bid_partial(
...
+    @staticmethod
     def test_market_accept_bid_emits_bid_split_on_partial_bid(
-            self, called, market=TwoSidedMarket(bc=MagicMock(), time_slot=pendulum.now())):
+            called, market=TwoSidedMarket(bc=MagicMock(), time_slot=pendulum.now())):
         market.add_listener(called)
         bid = market.bid(20, 20, "A", "A")
         trade_offer_info = TradeBidOfferInfo(1, 1, 1, 1, 1)
         trade = market.accept_bid(bid, energy=1, trade_offer_info=trade_offer_info)
-        assert all([ev != repr(MarketEvent.BID_DELETED) for c in called.calls for ev in c[0]])
+        assert all((ev != repr(MarketEvent.BID_DELETED) for c in called.calls for ev in c[0]))
```
spyrostz, line 325: Minor, but I think that not even the second set of brackets is needed here.

spyrostz (approved):
One minor comment, LGTM

tests/test_two_sided_market.py
Hassan754: Thanks! I'll fix this in my other in progress PR

PR #1364, opened 2022-01-25
Refactor styling warnings in the markets modules

Hassan754:


hannesdiedrich (dismissed):
LGTM

tests/market/test_market.py
```diff
@@ -335,26 +335,6 @@ def test_market_acct_multiple(market=OneSidedMarket(bc=NonBlockchainInterface(st
...
-
-    assert market.avg_offer_price == 2
-
-
-@pytest.mark.parametrize("market",
-                         [OneSidedMarket(bc=MagicMock(), time_slot=now()),
-                          BalancingMarket(bc=MagicMock(), time_slot=now()),
-                          SettlementMarket(bc=MagicMock(), time_slot=now())])
-def test_market_avg_offer_price_empty(market):
-    assert market.avg_offer_price == 0
```
spyrostz, line 355: Since these tests were deleted, and by a brief look at the code, I saw that there are some unused members in the market class that we could get rid of. These are `_avg_offer_price`, `min_offer_price`, `max_offer_price`. Also `_update_min_max_avg_offer_prices` method should be deleted as far as I see. Please double check in order to validate that I did not miss any use of these members outside of the class, I did not see any such occurrences from a first look though.

spyrostz (commented):
One comment, will LGTM once dealt with, the rest of the PR looks good

tests/market/test_market.py
Hassan754: Thanks for looking them up! you're right!
I pushed their deprecation

spyrostz (approved):
LGTM

PR #1365, opened 2022-01-25
GSYE-155: Read or rotate the rate profiles at any time the set_parame…

spyrostz:
…ters is called, no matter if the values of the set_parameters are None. Added extra warning message and another read/rotate profiles in the _populate_profiles method in order to cover the case that the rate profiles were not read correctly and to try again.

### DIFF
--- a/src/gsy_e/models/strategy/update_frequency.py
+++ b/src/gsy_e/models/strategy/update_frequency.py
@@ -15,6 +15,7 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+import logging
 from typing import TYPE_CHECKING, Callable, List
 
 from gsy_framework.constants_limits import ConstSettings, GlobalConfig
@@ -23,6 +24,7 @@ from gsy_framework.utils import (find_object_of_same_weekday_and_time,
                                  is_time_slot_in_simulation_duration)
 from pendulum import duration, DateTime, Duration
 
+import gsy_e.constants
 from gsy_e.gsy_e_core.global_objects_singleton import global_objects
 from gsy_e.gsy_e_core.util import write_default_to_dict, is_time_slot_in_past_markets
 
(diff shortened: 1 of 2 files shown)


src/gsy_e/models/strategy/update_frequency.py
```diff
@@ -225,24 +243,17 @@ def set_parameters(self, *, initial_rate: float = None, final_rate: float = None
...
-            should_update = True
         if fit_to_limit is not None:
             self.fit_to_limit = fit_to_limit
-            should_update = True
         if update_interval is not None:
             self.update_interval = update_interval
-            should_update = True
-        if should_update:
-            self._read_or_rotate_rate_profiles()
+        self._read_or_rotate_rate_profiles()
```
hannesdiedrich, line 256: I was first afraid that this would increase the computational effort and the number of queries to the profile DB. But there is an if check and the population is skipped if the current time_slot is not part of the profile. If this is only a corner case and this change does not mean that the profiles are still only populated every simulation-week (and not every market slot), all good.

hannesdiedrich (approved):
One clarification, LGTM

hannesdiedrich:
@spyrostz should we make this part of the next deploy?

src/gsy_e/models/strategy/update_frequency.py
spyrostz: This change is not really that different from the former implementation. The only difference between the 2 is that the old implementation would not call the `_read_or_rotate_profiles` method if `set_parameters` was called without arguments (because this is the only condition where should_update remains False). As for the impact to the queries to the DB, it is not different from the previous implementation in the sense that it performs queries only when the current time_slot is not part of the profile, as you mentioned. So I think that we are good here

spyrostz:
> @spyrostz should we make this part of the next deploy?

Does not hurt to have in the deploy, let me notify the PR reviewers again.

hannesdiedrich (approved):
LGTM

Hassan754 (approved):

PR #1366, opened 2022-01-26
Bug/gsye 150 Send external events and responses via redis queue

hannesdiedrich:
In order to make sure that only one consumer on the gsy-web side is handling each of the events and responses sent by the gsy-e, the communication was moved from redis pubsub to redis queue. This was done for both myco matcher and external devices and aggregators. For local simulation runs, the original pubsub way is kept. The constants.SEND_EVENTS_RESPONSES_TO_SDK_VIA_RQ is handling the switch between the two ways.
Also part of this PR: Centralised the queue names in the gsy-framework and reused them here.

### DIFF
--- a/src/gsy_e/gsy_e_core/cli.py
+++ b/src/gsy_e/gsy_e_core/cli.py
@@ -133,8 +133,6 @@ def run(setup_module_name, settings_file, duration, slot_length, tick_length,
                 external_connection_enabled=enable_external_connection,
                 enable_degrees_of_freedom=enable_dof)
 
-        gsy_e.constants.SEND_EVENTS_RESPONSES_TO_SDK_VIA_RQ = False
-
         if compare_alt_pricing is True:
             ConstSettings.MASettings.AlternativePricing.COMPARE_PRICING_SCHEMES = True
             # we need the seconds in the export dir name
--- a/src/gsy_e/gsy_e_core/cli.py
+++ b/src/gsy_e/gsy_e_core/cli.py
@@ -24,13 +24,12 @@ import click
 from click.types import Choice
 from click_default_group import DefaultGroup
 from colorlog.colorlog import ColoredFormatter
-from pendulum import DateTime, today
-
 from gsy_framework.constants_limits import ConstSettings
 from gsy_framework.exceptions import GSyException
 from gsy_framework.settings_validators import validate_global_settings
+from pendulum import DateTime, today
 
-from gsy_e.constants import DATE_FORMAT, DATE_TIME_FORMAT, TIME_FORMAT, TIME_ZONE
+import gsy_e.constants
 from gsy_e.gsy_e_core.simulation import run_simulation
 from gsy_e.gsy_e_core.util import (
     DateType, IntervalType, available_simulation_scenarios, convert_str_to_pause_after_interval,
@@ -85,7 +84,8 @@ _setup_modules = available_simulation_scenarios
               help="Start simulation in paused state")
 @click.option("--pause-at", type=str, default=None,
               help="Automatically pause at a certain time. "
-              f"Accepted Input formats: ({DATE_FORMAT}, {TIME_FORMAT}) [default: disabled]")
+              f"Accepted Input formats: ({gsy_e.constants.DATE_FORMAT}, "
+                   f"{gsy_e.constants.TIME_FORMAT}) [default: disabled]")
 @click.option("--repl/--no-repl", default=False, show_default=True,
               help="Start REPL after simulation run.")
 @click.option("--no-export", is_flag=True, default=False, help="Skip export of simulation data")
@@ -96,9 +96,10 @@ _setup_modules = available_simulation_scenarios
               help="Compare alternative pricing schemes")
 @click.option("--enable-external-connection", is_flag=True, default=False,
               help="External Agents interaction to simulation during runtime")
-@click.option("--start-date", type=DateType(DATE_FORMAT),
-              default=today(tz=TIME_ZONE).format(DATE_FORMAT), show_default=True,
-              help=f"Start date of the Simulation ({DATE_FORMAT})")
+@click.option("--start-date", type=DateType(gsy_e.constants.DATE_FORMAT),
+              default=today(tz=gsy_e.constants.TIME_ZONE).format(gsy_e.constants.DATE_FORMAT),
+              show_default=True,
+              help=f"Start date of the Simulation ({gsy_e.constants.DATE_FORMAT})")
 @click.option("--enable-dof/--disable-dof",
               is_flag=True, default=True,
               help=(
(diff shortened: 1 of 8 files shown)


spyrostz (dismissed):
LGTM

Hassan754 (dismissed):
LGTM

Hassan754 (dismissed):

src/gsy_e/gsy_e_core/cli.py
```diff
@@ -132,10 +133,13 @@ def run(setup_module_name, settings_file, duration, slot_length, tick_length,
                 external_connection_enabled=enable_external_connection,
                 enable_degrees_of_freedom=enable_dof)
 
+        gsy_e.constants.SEND_EVENTS_RESPONSES_TO_SDK_VIA_RQ = False
```
fievelk, line 136: Since the default is already `False`, do you think we can omit this line?

src/gsy_e/gsy_e_core/redis_connections/redis_area_market_communicator.py
```diff
@@ -103,59 +153,64 @@ def __init__(self, is_enabled):
...
         self.aggregator.publish_all_commands_responses(self)
         self.aggregator.publish_all_events(self)
 
 
-class BlockingCommunicator(RedisCommunicator):
-    def __init__(self):
-        super().__init__()
-        self.lock = Lock()
+class RQExternalConnectionCommunicator(ExternalConnectionCommunicator, RQResettableCommunicator):
+    """Communicator for sending messages using redis queue including utils for aggregator."""
```
fievelk, line 203: Why don't we just override `publish_json` here and avoid mixins? Something like:
```python
class RQExternalConnectionCommunicator(ExternalConnectionCommunicator):
    def publish_json(self, channel: str, data: Dict) -> None:
        ...
```
Then we could remove the `RQResettableCommunicator` class. What do you think?

src/gsy_e/gsy_e_core/rq_job_handler.py
```diff
@@ -17,12 +19,16 @@
...
 
-def decompress_and_decode_queued_strings(queued_string):
+def decompress_and_decode_queued_strings(queued_string: bytes) -> Dict:
+    """Decompress and decode data sent via redis queue."""
     return pickle.loads(decompress(queued_string))
 
 
-def launch_simulation_from_rq_job(scenario, settings, events, aggregator_device_mapping,
-                                  saved_state, job_id):
+# pylint: disable=too-many-arguments, too-many-locals
```
fievelk, line 27: Note that this will disable these warnings on the entire module, from this point onwards.

src/gsy_e/gsy_e_core/cli.py
hannesdiedrich: Probably, yes I wanted to be explicit, but no need, will remove it.

src/gsy_e/gsy_e_core/redis_connections/redis_area_market_communicator.py
hannesdiedrich: The `ExternalConnectionCommunicator` and `RQExternalConnectionCommunicator` are inheriting from the `RQResettableCommunicator`, in order to remove the `RQResettableCommunicator` we would need to move all code from there into the `ExternalConnectionCommunicator`. But they have different implementations of e.g. `sub_to_response`. 
I did not want to duplicate code which is why I created the RQExternalConnectionCommunicator that inherits all from ExternalConnectionCommunicator and overwrites the `publish_json` with the  `RQResettableCommunicator.publish_json`
Let me know if there is a more cleaner way. Thanks

fievelk: I'm not sure if I'm missing something. `ExternalConnectionCommunicator` and `RQExternalConnectionCommunicator` inherit from the same class (`ResettableCommunicator`). `RQResettableCommunicator` redefines only one method (`publish_json`). My proposal is to delete `RQResettableCommunicator` and directly redefine `publish_json` inside of `RQExternalConnectionCommunicator` (as in the snippet I put in the comment above).
`RQExternalConnectionCommunicator` would still inherit from `ExternalConnectionCommunicator`, so we wouldn't have to duplicate code.
Sorry if I'm not seeing something!

fievelk: I'm not sure if I'm missing something. `ExternalConnectionCommunicator` and `RQResettableCommunicator` inherit from the same class (`ResettableCommunicator`). `RQResettableCommunicator` redefines only one method (`publish_json`). My proposal is to delete `RQResettableCommunicator` and directly redefine `publish_json` inside of `RQExternalConnectionCommunicator` (as in the snippet I put in the comment above).
`RQExternalConnectionCommunicator` would still inherit from `ExternalConnectionCommunicator`, so we wouldn't have to duplicate code.
Sorry if I'm not seeing something!

hannesdiedrich: I understood, what you were proposing, but we need the `RQResettableCommunicator` for the myco communication. `ExternalConnectionCommunicator` and `RQExternalConnectionCommunicator` is used for the aggregator communication and overwrite methods, just like `sub_to_channel` from their baseclass `ResettableCommunicator`. 
Sorry If I misunderstood something here. Please let me know @fievelk

src/gsy_e/gsy_e_core/redis_connections/redis_area_market_communicator.py
```diff
@@ -15,86 +15,136 @@
...
+        """Wait until response of send message was received."""
+        start_time = time()
+        while not response_received_callback() and \
+                (time() - start_time < REDIS_PUBLISH_RESPONSE_TIMEOUT):
+            with self.lock:
+                self.pubsub.get_message(timeout=REDIS_POLL_TIMEOUT)
+
+
 class ResettableCommunicator(RedisCommunicator):
+    """Communicator for sending messages using redis pubsub including."""
```
fievelk, line 97: Should "including" be followed by something? Also, could you please specify in the docstring what "resettable" means?

src/gsy_e/gsy_e_core/redis_connections/redis_area_market_communicator.py
fievelk: Ahh ok, here's what I hadn't seen! I see your point now. I usually try to avoid mixins, that's probably why I got suspicious. Thanks!

src/gsy_e/gsy_e_core/redis_connections/redis_area_market_communicator.py
hannesdiedrich: https://github.com/gridsingularity/gsy-e/pull/1366/commits/66b747ac0df704e94b3f2d895a279ed233b2bad2

fievelk (approved):

spyrostz (approved):
LGTM

PR #1367, opened 2022-01-27
GSYE-122: As tester, I would like to move all integration tests to a separate repo.

ERafaelMartinez:
This PR aims to address GSYE-122 ticket. The changes suggested by it apply only to gsy-e repository.


tox.ini
```diff
@@ -95,3 +95,29 @@ setenv =
...
+                      echo
+					  bash
+commands =
+		python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
+		pip-sync requirements/tests.txt requirements/pandapower.txt
+		pip install -e .
+		pip uninstall -y gsy-framework
+		pip install git+https://github.com/gridsingularity/gsy-framework@{env:BRANCH:master}
+		rm -rf gsy-backend-integration-tests
+		rm -rf integration_tests
```
hannesdiedrich, line 117: You do not need these as the worker will not have any artefacts from old runs in it.

tox.ini
```diff
@@ -95,3 +95,29 @@ setenv =
...
+		mkdir integration_tests
+		git clone --branch {env:BRANCH:master} git@github.com:gridsingularity/gsy-backend-integration-tests.git
+		cp -r ./gsy-backend-integration-tests/gsy_e_tests/integration_tests/ ./integration_tests/
+		bash -c 'for f in ./integration_tests/*.feature; do behave --tags=-slow --tags=-disabled "$f" || exit 1; done'
+		rm -rf gsy-backend-integration-tests
+		rm -rf integration_tests
```
hannesdiedrich, line 123: You probably want this for local runs to clean up right? Lets see how we can get rid of it (see the next comments).

tox.ini
```diff
@@ -95,3 +95,29 @@ setenv =
...
+		pip-sync requirements/tests.txt requirements/pandapower.txt
+		pip install -e .
+		pip uninstall -y gsy-framework
+		pip install git+https://github.com/gridsingularity/gsy-framework@{env:BRANCH:master}
+		rm -rf gsy-backend-integration-tests
+		rm -rf integration_tests
+		mkdir integration_tests
+		git clone --branch {env:BRANCH:master} git@github.com:gridsingularity/gsy-backend-integration-tests.git
+		cp -r ./gsy-backend-integration-tests/gsy_e_tests/integration_tests/ ./integration_tests/
+		bash -c 'for f in ./integration_tests/*.feature; do behave --tags=-slow --tags=-disabled "$f" || exit 1; done'
```
hannesdiedrich, line 121: How about:
```suggestion
		git clone --branch {env:BRANCH:master} git@github.com:gridsingularity/gsy-backend-integration-tests.git 
		bash -c 'for f in ./gsy-backend-integration-tests/gsy_e_tests/*.feature; do behave --tags=-slow --tags=-disabled "$f" || exit 1; done'
```

tox.ini
```diff
@@ -95,3 +95,29 @@ setenv =
...
+[testenv:ci_gsye_122]
+basepython = python3.8
+passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH BRANCH
+deps = pip-tools
+whitelist_externals = git
+                      /bin/rm
+                      /bin/cp
+                      /bin/mkdir
+                      echo
+					  bash
```
hannesdiedrich, line 109: Do you really need to whitelist all of these?

tox.ini
```diff
@@ -95,3 +95,29 @@ setenv =
 commands =
     pip install -e .
 	behave --tags=-disabled integration_tests
+
+
+[testenv:ci_gsye_122]
```
hannesdiedrich, line 100: As discussed offline, please put all your code in the `[testenv:ci]` so it gets automatically run in the github actions. Thanks!

tox.ini
hannesdiedrich: Can you please try out providing a temporary directory with `envtmpdir` 
https://tox.wiki/en/latest/config.html#conf-envtmpdirm ? AFAIU you do not need to clean up then. Thanks!

hannesdiedrich (commented):
Left some comments. Let me know when I can review again. Thanks

tox.ini
ERafaelMartinez: Oh, I see. However I recall running the ci command locally. Will this not be a possibility anymore??

tox.ini
ERafaelMartinez: I'll give it a try! Thanks for the resource

tox.ini
ERafaelMartinez: Probably not all of them, after addressing the use of a `envtmpdir`, and also previous comment on which we got rid of the cp ... line command.
I also forgot to get rid of `echo` which is not used at all. I'll fix this on the following commits .

tox.ini
ERafaelMartinez: On the first try,  I did something similar to what you suggest. However the references to other modules on the step files and the path to other files were affected and for it to work they needed to be edited. The difference though, was that the gsy-e project was not held on the root of the project, but on a subfolder, causing the problem mentioned above.
This time, gsy-e it is indeed on the root and therefore what you suggest works, so I'll change it. (I discovered the reason why after trying it out again)

ERafaelMartinez:
I think the rename of the job from `test-build `to `gsy-e-ci` affected one of the requirements after the merge. The tests passed now though.

tox.ini
```diff
@@ -104,4 +114,7 @@ whitelist_externals =
 commands =
     pip install -e .
 	git clone --branch {env:BRANCH:master} git@github.com:gridsingularity/gsy-backend-integration-tests.git {envtmpdir}/gsy-backend-integration-tests
-	bash -c 'for f in {envtmpdir}/gsy-backend-integration-tests/gsy_e_tests/integration_tests/*.feature; do behave --tags=-disabled "$f" || exit 1; done'
+	bash -c 'cp -r {envtmpdir}/gsy-backend-integration-tests/gsy_e_tests/integration_tests/ ./integration_tests/'
+	bash -c 'for f in ./integration_tests/*.feature; do behave --tags=-disabled "$f" || exit 1; done'
+commands_post =
+	bash -c '[ -z "{env:INTEGRATION_TESTS_REPO_URL:}" ] && rm -rf ./integration_tests/ || :'
```
hannesdiedrich, line 120: This will make the use of `envtmpdir` pointless AFAIS. Also, where do you set the `INTEGRATION_TESTS_REPO_URL` env variable for local tox runs?

ERafaelMartinez: I understand why you say it.
However, if we were to let here the previous approach and try to fix the import of the module (which I tried), more import references will have also to be updated, because the `envtmpdir` lives in a rout in which the import does not work as it used to. For example, after fixing the import of the function on integration_tests.py module, then it ran with the following Exception related to the export path:

```
File "/Users/GSy/Workspace/gsy-e/.tox/integrationtests/lib/python3.8/site-packages/behave/model.py", line 1329, in run
match.run(runner.context)
File "/Users/GSy/Workspace/gsy-e/.tox/integrationtests/lib/python3.8/site-packages/behave/matchers.py", line 98, in run
self.func(context, *args, **kwargs)
File ".tox/integrationtests/tmp/gsy-backend-integration-tests/gsy_e_tests/integration_tests/steps/integration_tests.py", line 681, in run_sim
raise Exception(er)


 Exception: [Errno 2] No such file or directory: '/Users/GSy/Workspace/gsy-e/gsy_e-simulation/integration_tests/two_sided_market.infinite_bus/2022-01-31T15:49:29/raw_data/2022-01-31T03:00.json'
```

And even though the files were created, they would not be read. 

If  `behave ./integration_tests/` is used on the testenv scenario instead of `bash -c 'for f in ./integration_tests/*.feature; do behave --tags=-disabled "$f" || exit 1; done'` , all of the tests are executed and you can see more than one fail do to similar reasons. For this to be "fixed" differently, it would be required to update multiple files and path references. On the first approach I came up with (the open discussed offline on a meeting) I had done this in a way but can get tricky.

IMO this might appear to be un-neat, but it is indeed more straight forward. Also, with this command, the cleanup is only done when run locally, and as it is "inside" a bash command, it does not need to whitelist anything else apart from `bash`.

---------

The  `INTEGRATION_TESTS_REPO_URL` env variable was set by Marcelo on the github workflow. If this variable is passed, then the execution would be online. Otherwise, it is local. `[ -z "{env:INTEGRATION_TESTS_REPO_URL:}" ]` returns True if the variable is not set (defaults to empty), and if True, `rm -rf ./integration_tests/` will then be executed. Otherwise, nothing happens.

tox.ini
```diff
@@ -42,11 +42,22 @@ commands =
 
 [testenv:integrationtests]
 basepython = python3.8
+passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH SOLC_BINARY BRANCH INTEGRATION_TESTS_REPO_URL
 deps =
 	-rrequirements/tests.txt
+setenv =
+	INT_TESTS_REPO_SSH_URL = git@github.com:gridsingularity/gsy-backend-integration-tests.git
+whitelist_externals =
```
fievelk, line 50: `whitelist_externals` has been deprecated and it's now `allowlist_externals`.

tox.ini
```diff
@@ -42,11 +42,22 @@ commands =
...
+	git
+	bash
+commands_pre =
```
fievelk, line 53: Are you sure that this is working? If I'm not mistaken, this has been introduced in `tox==3.4` and we're using `tox==3.24`. Could you please double-check?

tox.ini
ERafaelMartinez: Sure. I also thought about it but decided to stick with that previously implemented.

tox.ini
```diff
@@ -42,11 +42,22 @@ commands =
...
+	pip install -e .
+	git clone {env:INTEGRATION_TESTS_REPO_URL:{env:INT_TESTS_REPO_SSH_URL:}} {envtmpdir}/gsy-backend-integration-tests
 commands =
-    pip install -e .
-	behave --tags=-disabled integration_tests {posargs}
+	bash -c 'cp -r {envtmpdir}/gsy-backend-integration-tests/gsy_e_tests/integration_tests/ ./integration_tests/'
+	bash -c 'for f in ./integration_tests/*.feature; do behave --tags=-slow --tags=-disabled "$f" || exit 1; done'
```
fievelk, line 58: What's the difference between this and the following?
```
bash -c 'behave --tags=-slow --tags=-disabled integration_tests'
```

tox.ini
```diff
@@ -42,11 +42,22 @@ commands =
...
+whitelist_externals =
+	git
+	bash
+commands_pre =
+	pip install -e .
+	git clone {env:INTEGRATION_TESTS_REPO_URL:{env:INT_TESTS_REPO_SSH_URL:}} {envtmpdir}/gsy-backend-integration-tests
 commands =
-    pip install -e .
-	behave --tags=-disabled integration_tests {posargs}
+	bash -c 'cp -r {envtmpdir}/gsy-backend-integration-tests/gsy_e_tests/integration_tests/ ./integration_tests/'
```
fievelk, line 57: Could you please explain why we now need to use `bash -c`? Doesn't `cp -r ...` work directly?

tox.ini
```diff
@@ -42,11 +42,22 @@ commands =
...
+	bash -c 'for f in ./integration_tests/*.feature; do behave --tags=-slow --tags=-disabled "$f" || exit 1; done'
+commands_post =
+	bash -c '[ -z "{env:INTEGRATION_TESTS_REPO_URL:}" ] && rm -rf ./integration_tests/ || :'
```
fievelk, line 60: What happens if we just leave the `integration_tests` folder inside of `{envtmpdir}`? Wouldn't it be automatically deleted at the end of the tests?

tox.ini
fievelk: > The INTEGRATION_TESTS_REPO_URL env variable was set by Marcelo on the github workflow. If this variable is passed, then the execution would be online. Otherwise, it is local

Sorry, I still didn't understand this part :) What does it mean that the execution would be "online" vs "local"? In both cases we would still be cloning the same repository, right?

tox.ini
fievelk: Update: I now saw your other comment [here](https://github.com/gridsingularity/gsy-e/pull/1367/files#r795839825). I suppose it answers my question.

fievelk (commented):
@hannesdiedrich @spyrostz @ERafaelMartinez I have one main question concerning the general requirements. I had understood that the integration tests would have been moved to a separate repo (and that's done). But I also thought that we would have completely removed them from the `tox` files of the individual repositories. In this PR I see that Edo has been striving to clone and run those tests here on gsy-e too. Is this needed? Couldn't we just run them from their own `gsy-backend-integration-tests` repo?

tox.ini
ERafaelMartinez: This implementation executes each `*.feature` file one by one, so if any integration test fail on the` *.feature` file being executed, the bash command exits and therefore testenv (ci/integrationtests) fails. If the tests on `*.feature` file pass, then it executes the following `*.feature` file on the directory.

On the command you posted, the tests are executed one after the other, regardless of wether they failed or not.

I guess the first approach saves resources, interrupting the process after the first test fails. The second one executes all of them. Both testenv approaches would fail at the end. 
I guess one of them is just somehow more efficient.

ERafaelMartinez: A downside for the first approach is that you could not get the information of all of the tests that failed due to changes. You would just know something broke, maybe fix it and then run it again until you find out if anything different fails, and so on.

I took the first approach because it was already implemented when I started working on this ticket, but I'm ok with both of them. Actually I ocasionaly used the second just to save me some time on identifying all of the failing tests...

tox.ini
ERafaelMartinez: As I explained in a previous comment, I encountered myself with the `bash -c 'for f in ./integration_tests/*.feature; do behave --tags=-slow --tags=-disabled "$f" || exit 1; done'` command, and therefore this `bash` command was whitelisted. 
Originally I thought of using `cp -r ...` directly, but then I noticed it could be also achieved by using `bash`, and would thus only be needed to whitelist `bash` only instead of `cp`, `bash`, and probably other commands.

tox.ini
fievelk: In my personal preference and workflow, I would actually go for the second one. As you said, I would find it better to see all failures at once to be able to fix them. Moreover, it might be that one single test doesn't give me enough information to understand what's going wrong, while seeing all the failing tests might provide more context and information to infer the real cause of the failures. @hannesdiedrich what do you think of this?

tox.ini
ERafaelMartinez: In this case, by "online" we understand the tests being executed by a Github workflow.
By "local", we understand running them directly on the cli by means of `tox -e integrationtests` or `tox -e ci`.

And yes, in both cases we would clone the same repository, but if cloned "online", as `gsy-backend-integration-tests repo` is private, it is cloned by using an https url, with some credentials attached to it. 
On the other hand, if executed locally, the clone can be done by means of a ssh url. Trying to clone "online" with the ssh url resuted in a _no right permissions_ error.

Regarding removing `./integration_tests` if the tests are run locally, it is to leave the local repository clean of that folder which is only temporarily needed to execute the tests correctly. If ran "online", AFAIU, it would be in a "virtual machine", which would be reset after the workflow execution...

tox.ini
fievelk: I see...but I suppose that whitelisting commands is done to actually keep the allowed commands under control, to avoid allowing them all. It might just be a matter of style: I would personally find it cleaner to just add `cp` to the whitelist and avoid the `bash -c '...'` part. But feel free to keep it as you like it more.

tox.ini
fievelk: Could we clone in both cases using HTTPS, so that we don't have to define two variables and two flows? In gsy-web we have [this line](https://github.com/gridsingularity/gsy-e-sdk/blob/dfa138a90703e9376b92c801fa878b6f3de26ad5/integration_tests/build_test_containers.sh#L11) that uses it and I think it works well.

Regarding the deletion of `./integration_tests`: I would personally just delete it also during the github action. I think that it's good to keep the same execution flow in all our environments, as much as possible. And also because this would simplify the code without slowing it down (I suppose that the deletion is very fast).

tox.ini
hannesdiedrich: The early exit was implemented by spyros [here](https://github.com/gridsingularity/gsy-e/commit/91339e147f3074c3a33a32b6159ca230b3ebd6a4) and the explanation is [here](https://github.com/gridsingularity/gsy-e/pull/1363#issue-1110512892).
Basically the runtime is much lower if you early return.
I agree with that. If you want to debug the failing intehgration tests, you can run `behave --tags=-slow --tags=-disabled integration_tests` just locally IMO.

tox.ini
hannesdiedrich: > Could we clone in both cases using HTTPS, so that we don't have to define two variables and two flows? In gsy-web we have [this line](https://github.com/gridsingularity/gsy-e-sdk/blob/dfa138a90703e9376b92c801fa878b6f3de26ad5/integration_tests/build_test_containers.sh#L11) that uses it and I think it works well.

No, because with HTTPS you need to first login to github and with the SSH approach you can provide a key that we can pass through the github action env.
As Eduardo correctly mentions, for public repos, this is all not a problem and we could indeed use only HTTPS, but for private ones, we need to authenticate.

tox.ini
fievelk: I'm not totally convinced, but it's just because I have a different personal preference. So it's fine to keep it as it is. Just one thing: behave already provides this feature out of the box ([`--stop`](https://behave.readthedocs.io/en/stable/behave.html#cmdoption-stop)), so I think the code could be simplified to just be:
```bash
behave --tags=-slow --tags=-disabled integration_tests --stop
```
I haven't tried it but it's supposed to work.

spyrostz: I agree with all of you, that from the tester's perspective he would like to see all integration tests that actually fail after a CI test run. However, from the CI worker's perspective, exiting on the first integration test that fails will save a lot of the computation time. Keep in mind we are talking only about the gsy-e integration tests, which take around 15 mins on a CI worker, and if the worker exits as soon as one of the first integration tests fail, it will leave the worker free for other test runs that are blocked during this time. 

The reason for creating this bash command, was essentially memory leaks, so my focus was not on the behave env but more on the Python interpreter that was running it. After inspecting the memory usage of behave locally, I saw that it was increasing gradually while the integration tests were running. This had the effect that the first integration tests were relatively fast, but the following tests were gradually becoming slower. After some time inspecting the behave internals (during which I did not find any variables that were not deallocated from us) I decided to follow the easy solution and recreate the behave / Python interpreter at the end of every feature file, in order to deallocate the memory and start from scratch. 

Keep in mind that these 2 issues that are discussed are not coupled, meaning that even the bash script can be adapted to run all the tests despite some feature files failing.

fievelk: Thanks for the explanation Spyros! Regarding the first point, I would then suggest using the "early stop" command only for the `ci` environment and removing it from the other environments that are not used by the CI. 
(Minor point, related to this: maybe our tox file could be refactored a bit. I have the feeling that we could reuse many parts of it to avoid code repetition. Not too important anyway).

hannesdiedrich (commented):
I had a quick look into the tox documentation, but could not find a way to reduce the code duplication in the tox file. Do you know how we could call one test env inside the other in order to only have one env that clones and calls the integration tests? 
If this is not feasible, LGTM

fievelk:
> I had a quick look into the tox documentation, but could not find a way to reduce the code duplication in the tox file. Do you know how we could call one test env inside the other in order to only have one env that clones and calls the integration tests?
If this is not feasible, LGTM

@hannesdiedrich there are some ways, I think. The global tox env is inherited by all other envs (which could be useful to avoid repetitions sometimes). Also, there is this section that might be useful: [Substitution for values from other sections](https://tox.wiki/en/latest/config.html#substitution-for-values-from-other-sections). I think that our tox files could be refactored a bit, but I would need to have a closer look to be sure.

hannesdiedrich (dismissed):
LGTM

fievelk (dismissed):

ERafaelMartinez:
Added new commits to comply with agreements reached on related PR's. Thanks for your review!

hannesdiedrich (approved):
LGTM

spyrostz (commented):
One comment, will LGTM once resolved

tox.ini
```diff
@@ -56,42 +66,59 @@ exclude = .tox,wiki,.cache,.d3a,.hypothesis,.pytest_cache,vagrant,requirements,v
...
+    pip install git+https://github.com/gridsingularity/gsy-framework@{env:GSY_FRAMEWORK_BRANCH:master}
+commands =
     flake8
-    coverage run -m py.test --random-order {posargs:tests}
+    coverage run -m pytest --random-order {posargs:tests}
     coverage combine
     coverage xml
     coverage report
+    git clone {env:INTEGRATION_TESTS_REPO:git@github.com:gridsingularity/gsy-backend-integration-tests.git} {envtmpdir}/gsy-backend-integration-tests
+    cp -r {envtmpdir}/gsy-backend-integration-tests/gsy_e_tests/integration_tests/ ./integration_tests/
```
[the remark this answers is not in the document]
ERafaelMartinez: I tried it out locally and works smoothly.
However, I'd like to mention that the reason behind removing the copied folder (or in this case, the softlink) on `commands_post` was to leave the working folder (cloned repo) clean (just as it was before the execution) to avoid any confusion or accidental commit of these files where working locally.
So, IMO, `commands_post` should continue to be there.
I'm open to know anyone's opinion on this matter. What do you think?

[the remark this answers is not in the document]
spyrostz: I agree, does not hurt to leave the post step here in order to end up with a clean repository. Thanks for dealing with it!

spyrostz (approved):
LGTM

fievelk (approved):

PR #1371, opened 2022-02-03
TESTS-125: Changed the algorithm for splitting the available energy f…

spyrostz:
…or current and future market slots to allocate as much as possible on the first market slot. Added unit tests for the clamp_energy_to_sell/buy methods.

### DIFF
--- a/tests/test_state/test_storage_state.py
+++ b/tests/test_state/test_storage_state.py
@@ -1,5 +1,6 @@
 # pylint: disable=protected-access
 from unittest.mock import patch
+import pytest
 
 from pendulum import now, duration
 
@@ -142,3 +143,18 @@ class TestStorageState:
             market_slot_list[3]: 5.0,
         }
         assert expected_energy_to_buy == storage_state.energy_to_buy_dict
+
+    @pytest.mark.parametrize("is_selling", [True, False])
+    def test_clamp_energy_asserts_battery_traded_more_than_energy_per_slot(self, is_selling):
+        storage_state, market_slot_list = self._setup_storage_state_for_clamp_energy()
+        storage_state._battery_energy_per_slot = 0.0
+        for time_slot in market_slot_list:
+            storage_state.offered_buy_kWh[time_slot] = 5.0
+            storage_state.pledged_buy_kWh[time_slot] = 15.0
+            storage_state.offered_sell_kWh[time_slot] = 5.0
+            storage_state.pledged_sell_kWh[time_slot] = 15.0
+        with pytest.raises(AssertionError):
+            if is_selling:
+                storage_state._clamp_energy_to_sell_kWh(market_slot_list)
+            else:
+                storage_state._clamp_energy_to_buy_kWh(market_slot_list)
--- a/tests/test_state/test_storage_state.py
+++ b/tests/test_state/test_storage_state.py
@@ -1,6 +1,7 @@
+# pylint: disable=protected-access
 from unittest.mock import patch
 
-from pendulum import now
+from pendulum import now, duration
 
 from gsy_e.models.state import StorageState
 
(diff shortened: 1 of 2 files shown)


hannesdiedrich (dismissed):
LGTM

tests/test_state/test_storage_state.py
```diff
@@ -33,3 +34,111 @@ def test_market_cycle():
...
+        assert expected_energy_to_sell == storage_state.energy_to_sell_dict
+
+    def test_clamp_energy_to_sell_kWh_respects_energy_offers_trades(self):
+        storage_state, market_slot_list = self._setup_storage_state_for_clamp_energy()
+
+        storage_state._battery_energy_per_slot = 25.0
+        for time_slot in market_slot_list:
+            storage_state.offered_sell_kWh[time_slot] = 5.0
+            storage_state.pledged_sell_kWh[time_slot] = 15.0
+        storage_state._clamp_energy_to_sell_kWh(market_slot_list)
```
ERafaelMartinez, line 93: I understand that it would make no sense to offer/promise more energy than it is able to deliver, and therefore the quantities you use on your tests respect `_battery_energy_per_slot >= offered_sell_kWh + pledged_sell_kWh` condition. 
Therefore, I think of a test in which this is not like that. For example this same test but with pledges and offers of 15.0 kWh each. In this case `storage_dict[time_slot] = self._max_offer_energy_kWh(time_slot) = -5`, and afterwards, the available_energy_for_all_slots would end up growing after `available_energy_for_all_slots -= -5`, and make it less likely to be assigned to `energy_to_sell[future_time_slot]` on future time_slots. Also, energy_to_sell_dict would have several negative values, which do have an interpretation on _there is some energy to bid for instead of sell/offer_ , however being this is the purpose of energy_to_buy_dict, I'm not sure this is the right behavior. 
Of course this scenario is hypothetical and would only be enabled by setting conditions which break this `_battery_energy_per_slot >= offered_sell_kWh + pledged_sell_kWh`.

However, I wonder how is this scenario avoided? In other words, how can we be sure that `_battery_energy_per_slot >= offered_sell_kWh + pledged_sell_kWh` always happens?

ERafaelMartinez (commented):
I only have one concern. Left a comment. Other than that LGTM.

tests/test_state/test_storage_state.py
spyrostz: @ERafaelMartinez to answer your last question, the scenario that you describe is avoided only by the calling code unfortunately. The way this class works from the outside (to be more precise, from the trading strategies) , is that the caller code first has to get the available energy to sell / buy (using the get_available_energy_to_buy/sell_kWh methods) which takes into account the battery energy per slot in order for the condition to not be violated, post the offer / bid with the available energy returned by the method, and then updates the pledged energy by calling register_energy_from_posted_offer/bid. That way it is guaranteed that the amount of energy that is placed as an offer or bid never violates this condition. 
I predict that on the next refactoring cycle this will be one of the most important things to refactor in this class, in order to ensure that this condition is never violated from inside this class. However for now I would go with Hannes' recommendation on the ticket comments, and add assertions, in order to at least get notified that this condition will never be violated. 
I'll also try to add unit tests that validate the assertion, I know that they will probably be deprecated soon, however I will add them in case the aforementioned refactoring is not done soon.

ERafaelMartinez: Thank you for explaining, and sorry for the nitpickiness

ERafaelMartinez (approved):
LGTM

tests/test_state/test_storage_state.py
spyrostz: > Thank you for explaining, and sorry for the nitpickiness

Please don't be thanks again for catching that cornercase !

hannesdiedrich (approved):
LGTM

PR #1372, opened 2022-02-08
TESTS-109: Create unit tests for the storage state.

ERafaelMartinez:
This PR contains unit-tests for StorageState class methods.

### DIFF
--- a/tests/test_state/test_storage_state.py
+++ b/tests/test_state/test_storage_state.py
@@ -600,9 +600,9 @@ class TestStorageState:
         current_time_slot, _, _ = self._initialize_time_slots()
         storage_state.add_default_values_to_state_profiles([current_time_slot])
         storage_state.energy_to_sell_dict[current_time_slot] = "test_energy_to_sell"
-        storage_state.offered_sell_kWh[current_time_slot] = "test_energy_active_in_bids"
+        storage_state.offered_sell_kWh[current_time_slot] = "test_energy_active_in_offers"
         storage_state.energy_to_buy_dict[current_time_slot] = "test_energy_to_buy"
-        storage_state.offered_buy_kWh[current_time_slot] = "test_energy_active_in_offers"
+        storage_state.offered_buy_kWh[current_time_slot] = "test_energy_active_in_bids"
         free_storage_mock = MagicMock(return_value="test_free_storage")
         used_storage_mock = PropertyMock()
         storage_state.free_storage = free_storage_mock
--- a/tests/test_state/test_storage_state.py
+++ b/tests/test_state/test_storage_state.py
@@ -110,8 +110,8 @@ class TestStorageState:
         storage_state.pledged_sell_kWh[past_time_slot] = 10
         storage_state.market_cycle(past_time_slot, current_time_slot, future_time_slots)
         assert storage_state.charge_history[current_time_slot] != storage_state.initial_soc
-        assert storage_state.charge_history_kWh[current_time_slot] != \
-               storage_state.initial_capacity_kWh
+        assert (storage_state.charge_history_kWh[current_time_slot] !=
+               storage_state.initial_capacity_kWh)
         assert storage_state.offered_history[current_time_slot] != "-"
 
     @staticmethod
--- a/tests/test_state/test_storage_state.py
+++ b/tests/test_state/test_storage_state.py
@@ -5,6 +5,21 @@ from pendulum import now, duration
 
 from gsy_e.models.state import StorageState, ESSEnergyOrigin
 
+SAMPLE_STATE = {
+    "pledged_sell_kWh": {},
+    "offered_sell_kWh": {},
+    "pledged_buy_kWh": {},
+    "offered_buy_kWh": {},
+    "charge_history": {},
+    "charge_history_kWh": {},
+    "offered_history": {},
+    "used_history": {},
+    "energy_to_buy_dict": {},
+    "energy_to_sell_dict": {},
+    "used_storage": 0.0,
+    "battery_energy_per_slot": 0.0,
+}
+
 
 class TestStorageState:
     """Test the StorageState class."""
@@ -49,11 +64,12 @@ class TestStorageState:
     def test_market_cycle_ess_share_time_series_dict(self):
         storage_state = StorageState(initial_soc=100,
                                      capacity=100,
-                                     initial_energy_origin=ESSEnergyOrigin.LOCAL)
+                                     initial_energy_origin=ESSEnergyOrigin.LOCAL,
+                                     max_abs_battery_power_kW=15)
         past_time_slot, current_time_slot, future_time_slots = self._initialize_time_slots()
         active_market_slot_time_list = [past_time_slot, current_time_slot, *future_time_slots]
         storage_state.add_default_values_to_state_profiles(active_market_slot_time_list)
-        energy = 10
+        energy = 0.3
         storage_state.activate(
             slot_length=duration(minutes=15), current_time_slot=current_time_slot)
         storage_state.register_energy_from_one_sided_market_accept_offer(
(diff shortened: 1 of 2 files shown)


tests/test_state/test_storage_state.py
```diff
@@ -36,6 +57,128 @@ def test_market_cycle():
...
+        storage_state.activate(
+            slot_length=duration(minutes=15), current_time_slot=current_time_slot)
+        storage_state.register_energy_from_one_sided_market_accept_offer(
+            energy, past_time_slot, ESSEnergyOrigin.LOCAL)
+        storage_state.register_energy_from_one_sided_market_accept_offer(
+            energy, past_time_slot, ESSEnergyOrigin.UNKNOWN)
+        storage_state.register_energy_from_one_sided_market_accept_offer(
+            energy, past_time_slot, ESSEnergyOrigin.UNKNOWN)
+        storage_state.market_cycle(past_time_slot, current_time_slot, future_time_slots)
+        expected_time_series = {ESSEnergyOrigin.LOCAL: storage_state.initial_capacity_kWh + energy,
```
hannesdiedrich, line 93: This is not a time series, only the value for the expected past_time_slot, right?
```suggestion
        expected_ess_share_last_market = {ESSEnergyOrigin.LOCAL: storage_state.initial_capacity_kWh + energy,
```

tests/test_state/test_storage_state.py
```diff
@@ -36,6 +57,128 @@ def test_market_cycle():
...
+        past_time_slot, current_time_slot, future_time_slots = self._initialize_time_slots()
+        active_market_slot_time_list = [past_time_slot, current_time_slot, *future_time_slots]
+        storage_state.add_default_values_to_state_profiles(active_market_slot_time_list)
+        storage_state.activate(
+            slot_length=duration(minutes=15), current_time_slot=current_time_slot)
+        storage_state.pledged_sell_kWh[past_time_slot] = 10
+        storage_state.market_cycle(past_time_slot, current_time_slot, future_time_slots)
+        assert storage_state.charge_history[current_time_slot] != storage_state.initial_soc
+        assert storage_state.charge_history_kWh[current_time_slot] != \
+               storage_state.initial_capacity_kWh
```
hannesdiedrich, line 111: Please avoid backslashes. Thanks!
```suggestion
        assert (storage_state.charge_history_kWh[current_time_slot] != 
               storage_state.initial_capacity_kWh)
```

tests/test_state/test_storage_state.py
```diff
@@ -158,3 +301,294 @@ def test_clamp_energy_asserts_battery_traded_more_than_energy_per_slot(self, is_
...
+        initial_registry_number = len(storage_state._used_storage_share)
+        first_origin = storage_state._used_storage_share[0].origin
+        storage_state._track_energy_sell_type(energy=energy_for_sale)
+        if len(storage_state._used_storage_share) != initial_registry_number - 1:
+            assert len(storage_state._used_storage_share) == initial_registry_number
+            assert storage_state._used_storage_share[0].origin == first_origin
+            assert isclose(
+                storage_state._used_storage_share[0].value, 0, abs_tol=FLOATING_POINT_TOLERANCE)
+        else:
+            assert len(storage_state._used_storage_share) == initial_registry_number - 1
```
hannesdiedrich, line 577: Can you please explain? Why do you need the else block here? I would expect the following:
```suggestion
        assert len(storage_state._used_storage_share) == initial_registry_number - 1
        assert len(storage_state._used_storage_share) == initial_registry_number
        assert storage_state._used_storage_share[0].origin == first_origin
        assert isclose(
                storage_state._used_storage_share[0].value, 0, abs_tol=FLOATING_POINT_TOLERANCE)
```

tests/test_state/test_storage_state.py
```diff
@@ -158,3 +301,294 @@ def test_clamp_energy_asserts_battery_traded_more_than_energy_per_slot(self, is_
...
+                              EnergyOrigin(ESSEnergyOrigin.UNKNOWN, 0.1)]
+        storage_state._used_storage_share = used_storage_share
+        return storage_state
+
+    def test_track_energy_bought_type_append_new_energy_origin(self):
+        storage_state = self._setup_storage_state_for_energy_origin_tracking()
+        initial_registry_number = len(storage_state._used_storage_share)
+        storage_state._track_energy_bought_type(energy=1.0, energy_origin=ESSEnergyOrigin.LOCAL)
+        assert len(storage_state._used_storage_share) == initial_registry_number + 1
+        assert isinstance(storage_state._used_storage_share[-1], EnergyOrigin)
```
hannesdiedrich, line 542: It would be better if you also test for the exact EnergyOrigin enum:
```suggestion
        assert storage_state._used_storage_share[-1].origin == EnergyOrigin. LOCAL
        assert storage_state._used_storage_share[-1].value == energy
```

tests/test_state/test_storage_state.py
```diff
@@ -158,3 +301,294 @@ def test_clamp_energy_asserts_battery_traded_more_than_energy_per_slot(self, is_
...
+
+    def test_delete_past_state_values_market_slot_in_past(self):
+        storage_state = StorageState()
+        past_time_slot, current_time_slot, future_time_slots = self._initialize_time_slots()
+        active_market_slot_time_list = [past_time_slot, current_time_slot, *future_time_slots]
+        storage_state.add_default_values_to_state_profiles(active_market_slot_time_list)
+        with patch("gsy_e.gsy_e_core.util.ConstSettings.SettlementMarketSettings."
+                   "ENABLE_SETTLEMENT_MARKETS", False):
+            storage_state.delete_past_state_values(current_time_slot)
+            assert storage_state.pledged_sell_kWh.get(past_time_slot) is None
```
hannesdiedrich, line 400: Can you please explain? Shouldn't there entries for all provided `add_default_values_to_state_profiles` , independent of the ENABLE_SETTLEMENT_MARKETS setting?

hannesdiedrich (commented):
Left some comment. LGTM in general. Thanks!

tests/test_state/test_storage_state.py
```diff
@@ -158,3 +301,294 @@ def test_clamp_energy_asserts_battery_traded_more_than_energy_per_slot(self, is_
...
+            slot_length=duration(minutes=15), current_time_slot=current_time_slot)
+        assert storage_state.get_soc_level(current_time_slot) == 1
+        storage_state.charge_history[current_time_slot] = 50
+        assert storage_state.get_soc_level(current_time_slot) == 0.5
+
+    def test_to_dict_keys_in_return_dict(self):
+        storage_state = StorageState()
+        current_time_slot, _, _ = self._initialize_time_slots()
+        storage_state.add_default_values_to_state_profiles([current_time_slot])
+        assert set(storage_state.to_dict(current_time_slot).keys()).issubset(SAMPLE_STATS.keys())
```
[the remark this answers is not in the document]
ERafaelMartinez: The reason why I implemented this in such a way is because I imagined a scenario where an additional key is added to the stats dict on `to_dict()` method. 

I thought that if that happens, it should not break the test, because all of the previous keys (which might already have some use somewhere else) would be unaffected. If one of the currently existing keys is removed, then the test would indeed fail.

However, I did notice that if the test were to work as I intended it to, the subset condition should be the other way around.

```suggestion
        assert set(SAMPLE_STATS.keys()).issubset(storage_state.to_dict(current_time_slot).keys())
```

What do you think? Does this make sense to you? 
Or should we let the test break if any additional key is added to the stats dict, so that it becomes evident that the tests should be also updated/refactored?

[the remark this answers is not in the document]
ERafaelMartinez: Sure, I'm on it. Could you please have a look at this comment [here](https://gridsingularity.atlassian.net/browse/TESTS-109?focusedCommentId=25260)?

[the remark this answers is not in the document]
ERafaelMartinez: On the mean time, a check for the returned values is added [here](https://github.com/gridsingularity/gsy-e/pull/1372/commits/88d1efe28017571f8659bc26ab27a9c934923726).

tests/test_state/test_storage_state.py
ERafaelMartinez: The methods here are supposed to test wether `delete_past_state_values()` method works properly. What it does is that it deletes key-value pairs from state profiles dicts (such as `self.pledge_sell_kWh`) if this are "in the past".

That is the reason for the names of these test methods `..._slot_not_in_past` and `..._slot_in_past`. 
The thing here is that "being in the past" (being a past state) can be relative depending on wether Settlement Markets are enabled or not.
When they are, then those market slots which happened less than an hour ago from `current_time_slot` are not past_states yet, and therefore should not be deleted when calling `delete_past_state_values()`. Therefore this lines:

```python
with patch("gsy_e.gsy_e_core.util.ConstSettings.SettlementMarketSettings."
             "ENABLE_SETTLEMENT_MARKETS", True):
      storage_state.delete_past_state_values(current_time_slot)
      assert storage_state.pledged_sell_kWh.get(past_time_slot) is not None
```

because `past_time_slot` is by construction only 15 minutes before current_time_slot, and therefore when the SettlementMarkets (SM) are enabled, is not technically a past state yet, and thus should not be deleted when calling the method.

On the other hand, if SM are disabled, then `past_time_slot` becomes indeed a past state and should therefore be deleted.

```python
with patch("gsy_e.gsy_e_core.util.ConstSettings.SettlementMarketSettings."
             "ENABLE_SETTLEMENT_MARKETS", False):
      storage_state.delete_past_state_values(current_time_slot)
      assert storage_state.pledged_sell_kWh.get(past_time_slot) is None
```

tests/test_state/test_storage_state.py
```diff
@@ -158,3 +301,294 @@ def test_clamp_energy_asserts_battery_traded_more_than_energy_per_slot(self, is_
...
+    def _setup_registration_test(self):
+        storage_state = StorageState()
+        current_time_slot, _, _ = self._initialize_time_slots()
+        storage_state.add_default_values_to_state_profiles([current_time_slot])
+        storage_state.activate(
+            slot_length=duration(minutes=15), current_time_slot=current_time_slot)
+        return storage_state, current_time_slot
+
+    @staticmethod
+    def _assert_negative_energy_raise_error(method, energy, time_slot):
```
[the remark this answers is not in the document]
ERafaelMartinez: I agree. Thanks!

tests/test_state/test_storage_state.py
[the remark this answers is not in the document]
hannesdiedrich: This way around it makes sense, indeed!

tests/test_state/test_storage_state.py
hannesdiedrich: Ok, thank you, so this is basically testing the utility function `is_time_slot_in_past_markets`, right?
Thanks a lot for the explanation, keep it as is then.

tests/test_state/test_storage_state.py
```diff
@@ -595,4 +595,25 @@ def test_to_dict_keys_in_return_dict(self):
         storage_state = StorageState()
         current_time_slot, _, _ = self._initialize_time_slots()
         storage_state.add_default_values_to_state_profiles([current_time_slot])
-        assert set(storage_state.to_dict(current_time_slot).keys()).issubset(SAMPLE_STATS.keys())
+        storage_state.energy_to_sell_dict[current_time_slot] = "test_energy_to_sell"
+        storage_state.offered_sell_kWh[current_time_slot] = "test_energy_active_in_bids"
+        storage_state.energy_to_buy_dict[current_time_slot] = "test_energy_to_buy"
+        storage_state.offered_buy_kWh[current_time_slot] = "test_energy_active_in_offers"
```
hannesdiedrich, line 605: Any reason why you did not use floats here? Is the a common practice for tests?

hannesdiedrich (approved):
One minor comments left. LGTM already

Hassan754 (approved):

tests/test_state/test_storage_state.py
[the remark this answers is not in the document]
Hassan754: Thanks!

tests/test_state/test_storage_state.py
ERafaelMartinez: There is no specific reason behind these other than I found somehow arbitrary to set a specific value to make the assertion with. I've seen this strategy in some testing on online forums, though. I'd say this is just a mater of personal taste. Any float-str value would have make the job.

hannesdiedrich (approved):
LGTM

spyrostz (approved):

PR #1373, opened 2022-02-10
 Soft refactoring of the grid fees modules

Hassan754:


src/gsy_e/models/market/grid_fees/constant_grid_fees.py
```diff
@@ -15,47 +15,37 @@
...
-        chain settles.
-        """
+            original_bid_rate, propagated_bid_rate, clearing_rate):
         return clearing_rate + (original_bid_rate - propagated_bid_rate)
 
-    def update_forwarded_bid_with_fee(self, source_bid, original_bid):
-        if source_bid is None:
-            return original_bid - self.grid_fee_rate
-        return source_bid - self.grid_fee_rate
+    def update_forwarded_bid_with_fee(self, source_price, original_price):
```
spyrostz, line 42: I would use `source_rate` and `original_rate` instead, to denote that these should be in cents / kWh. Same for the `update_forwarded_offer_with_fee`

src/gsy_e/models/market/grid_fees/constant_grid_fees.py
```diff
@@ -15,47 +15,37 @@
...
-    def update_incoming_bid_with_fee(self, source_bid, original_bid):
-        if source_bid is None:
-            return original_bid
-        return source_bid
+    def update_incoming_bid_with_fee(self, source_price, original_price):
+        return source_price or original_price
 
-    def update_incoming_offer_with_fee(self, source_offer_price, original_price):
-        if source_offer_price is None:
+    def update_incoming_offer_with_fee(self, source_price, original_price):
```
spyrostz, line 32: Same goes here, this method receives energy_rate as input arguments

spyrostz (commented):
Thanks for the refactoring! My only general comment is that the method arguments should be changed to reflect that they require energy rate and not price, once this is dealt with I will LGTM

src/gsy_e/models/market/grid_fees/constant_grid_fees.py
Hassan754: Thanks a lot for bringing this up, I overlooked it.
It seems that bid's methods were accepting rate whereas the offer's were expecting a price, I have just homogenized them

hannesdiedrich (approved):
LGTM

spyrostz (approved):
LGTM

PR #1374, opened 2022-02-10
GSYE-178: Enable grid fees accounting for DoF requirements

Hassan754:
We used to add grid fees to the orders' prices when they get forwarded to another market.
Now that the orders have other `price` source of truth (the ones in the requirements), when we add fees to the order.price we also have to add fees to all `price` key in all `order.requirements` list (otherwise offer seller may end up getting less than it requested)

### DIFF
--- a/src/gsy_e/models/market/two_sided.py
+++ b/src/gsy_e/models/market/two_sided.py
@@ -79,6 +79,22 @@ class TwoSidedMarket(OneSidedMarket):
         """
         return deepcopy(self.bids)
 
+    def _update_requirements_prices(self, bid):
+        requirements = []
+        for requirement in bid.requirements or []:
+            updated_requirement = {**requirement}
+            if "price" in updated_requirement:
+                if "original_price" not in updated_requirement:
+                    updated_requirement["original_price"] = updated_requirement["price"]
+
+                energy = updated_requirement.get("energy") or bid.energy
+                updated_price = (self.fee_class.update_incoming_bid_with_fee(
+                    updated_requirement["price"] / energy,
+                    updated_requirement["original_price"] / energy)) * energy
+                updated_requirement["price"] = updated_price
+            requirements.append(updated_requirement)
+        return requirements
+
     @lock_market_action
     def bid(self, price: float, energy: float, buyer: str, buyer_origin: str,
             bid_id: Optional[str] = None,
@@ -110,7 +126,8 @@ class TwoSidedMarket(OneSidedMarket):
                   self.now, price, energy, buyer, original_price, buyer_origin,
                   buyer_origin_id=buyer_origin_id, buyer_id=buyer_id,
                   attributes=attributes, requirements=requirements, time_slot=time_slot)
-
+        if adapt_price_with_fees:
+            bid.requirements = self._update_requirements_prices(bid)
         self.bids[bid.id] = bid
         if add_to_history is True:
             self.bid_history.append(bid)
--- a/src/gsy_e/models/strategy/market_agents/two_sided_engine.py
+++ b/src/gsy_e/models/strategy/market_agents/two_sided_engine.py
@@ -56,9 +56,9 @@ class TwoSidedEngine(MAEngine):
                     updated_requirement["original_price"] = updated_requirement["price"]
 
                 energy = updated_requirement.get("energy") or bid.energy
-                updated_price = (self.markets.source.fee_class.update_forwarded_bid_with_fee(
+                updated_price = self.markets.source.fee_class.update_forwarded_bid_with_fee(
                     updated_requirement["price"] / energy,
-                    updated_requirement["original_price"] / energy)) * energy
+                    updated_requirement["original_price"] / energy) * energy
                 updated_requirement["price"] = updated_price
             requirements.append(updated_requirement)
         return requirements
--- a/src/gsy_e/models/strategy/market_agents/two_sided_engine.py
+++ b/src/gsy_e/models/strategy/market_agents/two_sided_engine.py
@@ -47,6 +47,22 @@ class TwoSidedEngine(MAEngine):
         return "<TwoSidedPayAsBidEngine [{s.owner.name}] {s.name} " \
                "{s.markets.source.time_slot:%H:%M}>".format(s=self)
 
+    def _update_requirements_prices(self, bid):
+        requirements = []
+        for requirement in bid.requirements or []:
+            updated_requirement = {**requirement}
+            if "price" in updated_requirement:
+                if "original_price" not in updated_requirement:
+                    updated_requirement["original_price"] = updated_requirement["price"]
+
+                energy = updated_requirement.get("energy") or bid.energy
+                updated_price = (self.markets.source.fee_class.update_forwarded_bid_with_fee(
+                    updated_requirement["price"] / energy,
+                    updated_requirement["original_price"] / energy)) * energy
+                updated_requirement["price"] = updated_price
+            requirements.append(updated_requirement)
+        return requirements
+
     def _forward_bid(self, bid):
         if bid.buyer == self.markets.target.name:
             return None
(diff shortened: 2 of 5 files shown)


src/gsy_e/models/strategy/market_agents/two_sided_engine.py
```diff
@@ -47,6 +47,22 @@ def __repr__(self):
         return "<TwoSidedPayAsBidEngine [{s.owner.name}] {s.name} " \
                "{s.markets.source.time_slot:%H:%M}>".format(s=self)
 
+    def _update_requirements_prices(self, bid):
+        requirements = []
+        for requirement in bid.requirements or []:
+            updated_requirement = {**requirement}
+            if "price" in updated_requirement:
+                if "original_price" not in updated_requirement:
```
hannesdiedrich, line 55: Can you please explain why the `original_price` should be part of an requirement?

hannesdiedrich: I am sorry, but I think I also do not get the concept of a `price` requirement. Can you please explain when this could be useful?
When I post a bid like `Bid(energy=10, price=2)`, I already set the price, don't I? Why would I set an price requirement?

Hassan754: Sure, let me elaborate.
The goal of the degrees of freedom is to allow buyers to have flexibility in the buying process, for instance I can post a bid `Bid(energy=10, price=2, requirements=[{"preferred_partners": ["Hannes"], "price": 1}])`

Hassan754: So the bid in this case will be sold to "Hannes" for a lower price.
And as for the `original_price` requirement, it serves the same goal as the original_price attribute of the orders (denoting the price before adding/substracting grid fees)

hannesdiedrich: I see, thank you! So if the order was not forwarded yet, the "original_price" would not be part of the requirement, and from the moment of first forwarding, this member is in the requirement, right?
Do we have a validator for the requirements (I had a quick look but could not find it)? I mean to prohibit to let the user set the member "original_price" in order to make the grid fee calculation work. What do you think?

src/gsy_e/models/strategy/market_agents/two_sided_engine.py
```diff
@@ -47,6 +47,22 @@ def __repr__(self):
...
+                    updated_requirement["original_price"] = updated_requirement["price"]
+
+                energy = updated_requirement.get("energy") or bid.energy
+                updated_price = (self.markets.source.fee_class.update_forwarded_bid_with_fee(
+                    updated_requirement["price"] / energy,
+                    updated_requirement["original_price"] / energy)) * energy
```
hannesdiedrich, line 61: NIT: I found the bracket placing weird. As it works as you proposed, you can leave it if you like. 
```suggestion
                    updated_requirement["original_price"] / energy) * energy)
```

src/gsy_e/models/strategy/market_agents/two_sided_engine.py
Hassan754: > if the order was not forwarded yet, the "original_price" would not be part of the requirement, and from the moment of first forwarding, this member is in the requirement

That's correct!
Regarding the validator, I can now see its importance, thanks for foreseeing it!

Hassan754:
> I see that you only adapted the grid_fee calculation to the bids. Is there any reason not to do it for the offers?

This is because the price requirement is only supported for bids, see [here](https://github.com/gridsingularity/gsy-framework/blob/4d6009981bae87ad63aae81f4a9bfba1d72eaf9f/gsy_framework/matching_algorithms/requirements_validators.py#L107)

src/gsy_e/models/strategy/market_agents/two_sided_engine.py
Hassan754: @hannesdiedrich the validator needs some thorough thinking and designing to cover all corner cases, if you agree I'll add this in follow up PR after merging both this and [the other](https://github.com/gridsingularity/gsy-e/pull/1375) PRs

hannesdiedrich: Ok, I am also fine with adding a ticket for this and postpone it a bit.

hannesdiedrich:
> > I see that you only adapted the grid_fee calculation to the bids. Is there any reason not to do it for the offers?
> 
> This is because the price requirement is only supported for bids, see [here](https://github.com/gridsingularity/gsy-framework/blob/4d6009981bae87ad63aae81f4a9bfba1d72eaf9f/gsy_framework/matching_algorithms/requirements_validators.py#L107)

Your link is not pointing correctly AFAIU. But as you say, if this is only supported for bids, all good.

hannesdiedrich (approved):
LGTM

src/gsy_e/models/strategy/market_agents/two_sided_engine.py
hannesdiedrich: @Hassan754 Please add a note into the ticket to tell us to consider some time for the `thorough thinking and designing to cover all corner cases` Thanks!

src/gsy_e/models/market/two_sided.py
```diff
@@ -79,6 +79,22 @@ def get_bids(self) -> Dict:
...
+        for requirement in bid.requirements or []:
+            updated_requirement = {**requirement}
+            if "price" in updated_requirement:
+                if "original_price" not in updated_requirement:
+                    updated_requirement["original_price"] = updated_requirement["price"]
+
+                energy = updated_requirement.get("energy") or bid.energy
+                updated_price = (self.fee_class.update_incoming_bid_with_fee(
+                    updated_requirement["price"] / energy,
+                    updated_requirement["original_price"] / energy)) * energy
```
fievelk, line 93: Nit: given what the `update_incoming_bid_with_fee` method is doing (just selecting one of the two rates) these lines seem a bit redundant. They're basically doing:
```python
updated_price = updated_requirement["price"] or updated_requirement["original_price"]
```
This could be ok if the implementation of `update_incoming_bid_with_fee()` might change in the future. Regardless, these lines seem a bit redundant somehow and, since you wrote the `BaseClassGridFees` interface, you might be the right person to see if this can be changed or if it's better to keep it as it is.

fievelk: One more thing: `GridFees.update_incoming_bid_with_fee()` is not doing what the name implies. It is not updating anything, it's just selecting which rate to apply.

fievelk: First thing popping in my head: given the name, I would probably expect `update_incoming_bid_with_fee()` to take a `Bid` as an argument and modify it. This would internally do what you're doing in `_update_requirements_prices()`. Just an idea, you might find other better ones.

Hassan754: @fievelk I agree with your proposals and indeed these modules need to be further refactored, actually I did not write them I only fixed some stylings warnings in this [PR](https://github.com/gridsingularity/gsy-e/pull/1373).
After an offline discussion with Spyros, we agreed not to alter the logic to avoid introducing bugs, especially because we'll revisit and reimplement the entire fees modules after getting rid of the hierarchical setups we currently support (One more point to discuss in the upcoming workshop I assume)

src/gsy_e/models/market/two_sided.py
```diff
@@ -79,6 +79,22 @@ def get_bids(self) -> Dict:
         """
         return deepcopy(self.bids)
 
+    def _update_requirements_prices(self, bid):
+        requirements = []
+        for requirement in bid.requirements or []:
```
fievelk, line 84: Nit: I think we could change the [init method](https://github.com/gridsingularity/gsy-framework/blob/4d6009981bae87ad63aae81f4a9bfba1d72eaf9f/gsy_framework/data_classes.py#L58) to always provide an empty list for `Bid.requirements`, so that we can avoid this check.

src/gsy_e/models/strategy/market_agents/two_sided_engine.py
fievelk: > I see, thank you! So if the order was not forwarded yet, the "original_price" would not be part of the requirement, and from the moment of first forwarding, this member is in the requirement, right?

I think it might also be worth adding the `original_price` at the very beginning, before the order is forwarded. This might simplify our life and we could avoid needing to check every time if `"original_price"` is present or not.

fievelk (approved):
I left some comments. It might be good to address the complexity of `update_incoming_bid_with_fee()`, but I'll leave an LGTM and leave it up to you.

PR #1378, opened 2022-02-14
TESTS-110: Create unit tests for the ProsumptionInterface, Consumption and Production States

ERafaelMartinez:
This PR contains unit tests for ProsumptionInterface, ConsumptionState and ProductionState classes.

### DIFF
--- a/tests/test_state/test_prosumption_interface.py
+++ b/tests/test_state/test_prosumption_interface.py
@@ -1,0 +1,110 @@
+# pylint: disable=protected-access
+from unittest.mock import patch, MagicMock
+from typing import Dict
+import pytest
+from pendulum import now, DateTime
+
+from gsy_e.models.state import ProsumptionInterface
+
+
+class ProsumptionInterfaceHelper(ProsumptionInterface):
+    """Helper class to override unimplemented StateInterface's
+       abstract methods on ProsumptionInterface class."""
+
+    def get_state(self) -> Dict:
+        raise NotImplementedError
+
+    def restore_state(self, state_dict: Dict):
+        raise NotImplementedError
+
+    def delete_past_state_values(self, current_time_slot: DateTime):
+        raise NotImplementedError
+
+
+class TestProsumptionInterface:
+    """ Test ProsumptionInterface class"""
+
+    @staticmethod
+    def _setup_base_configuration():
+        return ProsumptionInterfaceHelper(), now()
+
+    def test_set_energy_measurement_kWh_dicts_not_empty(self):
+        prosumption_interface, current_time_slot = self._setup_base_configuration()
+        energy = 1
+        prosumption_interface.set_energy_measurement_kWh(
+            energy_kWh=energy, time_slot=current_time_slot
+        )
+        assert prosumption_interface.get_energy_measurement_kWh(
+            time_slot=current_time_slot) == energy
+        assert isinstance(prosumption_interface.get_forecast_measurement_deviation_kWh(
+            time_slot=current_time_slot), float)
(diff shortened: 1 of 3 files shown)


tests/test_state/test_prosumption_interface.py
```diff
@@ -0,0 +1,110 @@
...
+        raise NotImplementedError
+
+    def restore_state(self, state_dict: Dict):
+        raise NotImplementedError
+
+    def delete_past_state_values(self, current_time_slot: DateTime):
+        raise NotImplementedError
+
+
+class TestProsumptionInterface:
```
Hassan754, line 24: By the way, we have a `TestProsumptionInterface` residing in `tests/strategies/test_strategy_state.py`

Hassan754 (approved):

tests/test_state/test_prosumption_interface.py
ERafaelMartinez: Thanks for the tip. I added a couple of them just to not bypass it. Also removed the other file.

Hassan754 (approved):

tests/test_state/test_prosumption_interface.py
```diff
@@ -0,0 +1,125 @@
...
+        assert isinstance(prosumption_interface.get_forecast_measurement_deviation_kWh(
+            time_slot=current_time_slot), float)
+        assert isinstance(prosumption_interface.get_unsettled_deviation_kWh(
+            time_slot=current_time_slot), float)
+
+    @staticmethod
+    def _setup_configuration_for_settlement_posting(
+            energy_deviation=None, unsettled_deviation=None):
+        prosumption_interface = ProsumptionInterfaceHelper()
+        current_time_slot = now()
```
hannesdiedrich, line 48: You can do this here too, right?
```suggestion
   prosumption_interface, current_time_slot = self._setup_base_configuration()
```

tests/test_state/test_prosumption_interface.py
```diff
@@ -0,0 +1,125 @@
...
+        copysign_mock = MagicMock()
+        with patch("gsy_e.models.state.copysign", return_value=copysign_mock):
+            prosumption_interface, time_slot = self._setup_configuration_for_settlement_posting(
+                energy_deviation=energy_deviation, unsettled_deviation=unsettled_deviation)
+            if expected_response:
+                assert prosumption_interface.get_signed_unsettled_deviation_kWh(
+                    time_slot) == copysign_mock
+            else:
+                assert prosumption_interface.get_signed_unsettled_deviation_kWh(
+                    time_slot) is expected_response
```
hannesdiedrich, line 91: I think something is wrong here. The outcome of  `get_signed_unsettled_deviation_kWh` should either be a number or None. Here you are checking for None and `copysign_mock`. I would expect this implementation:
```suggestion
        assert prosumption_interface.get_signed_unsettled_deviation_kWh(
            time_slot) == expected_response
```

hannesdiedrich (commented):
Left two comments, LGTM once resolved.

tests/test_state/test_prosumption_interface.py
ERafaelMartinez: I understand what you mean. 
I agree with the fact that the method returns either None or a number. For the case on when it returns a number, this number is at the same time the return value for `copysign` function (being mocked). 
The condition block is to account for the case in which None value is expected, and when the number (mock) is.
For the sake of clearness I'll split this in two different test cases.

spyrostz (approved):
LGTM

Hassan754 (approved):
LGTM

PR #1386, opened 2022-02-16
GSYE-175: Moved updating / recycling of the profiles after the Area.c…

spyrostz:
…ycle_markets() method, in order for the current market time_slot to be taken into account while recycling the profile.


src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -334,11 +334,11 @@ def _execute_simulation(self, slot_resume, tick_resume, console=None):
                         f"{self.progress_info.elapsed_time} elapsed, "
                         f"ETA: {self.progress_info.eta}")
 
+            self.area.cycle_markets()
```
hannesdiedrich, line 337: Wow! Tricky one. Thanks a lot!

hannesdiedrich (approved):
LGTM

src/gsy_e/gsy_e_core/simulation.py
spyrostz: And extremely painful to debug unfortunately, since the DST was hiding it during the implementation (back in the summer DST was active, and we did not have the fix that adapted for daylight saving time, therefore this bug was hidden by it and appeared after the DST change)

Hassan754 (approved):
LGTM

PR #1389, opened 2022-02-21
 Update PV energy availability in future markets

Hassan754:
The PV's strategy was not updating the energy availability for the future market, this PR adds the setter method ` set_produced_energy_forecast_kWh_future_market` for the future markets.
There is also another added method ` set_produced_energy_forecast_kWh_in_state` that basically wraps both the setter calls for the spot market and the future markets

### DIFF
--- a/src/gsy_e/models/strategy/predefined_pv.py
+++ b/src/gsy_e/models/strategy/predefined_pv.py
@@ -92,9 +92,9 @@ class PVPredefinedStrategy(PVStrategy):
     def read_config_event(self):
         # this is to trigger to read from self.simulation_config.cloud_coverage:
         self.cloud_coverage = None
-        self.set_produced_energy_forecast_kWh_future_markets(reconfigure=True)
+        self.set_produced_energy_forecast_kWh_in_state(reconfigure=True)
 
-    def set_produced_energy_forecast_kWh_future_markets(self, reconfigure=True):
+    def set_produced_energy_forecast_kWh_spot_market(self, reconfigure=True):
         self._power_profile_index = self.cloud_coverage \
             if self.cloud_coverage is not None else self.simulation_config.cloud_coverage
         if reconfigure:
@@ -213,7 +213,7 @@ class PVUserProfileStrategy(PVPredefinedStrategy):
         if key_in_dict_and_not_none(kwargs, 'power_profile'):
             self._power_profile_input = kwargs['power_profile']
         self._read_or_rotate_profiles(reconfigure=True)
-        self.set_produced_energy_forecast_kWh_future_markets(reconfigure=True)
+        self.set_produced_energy_forecast_kWh_in_state(reconfigure=True)
 
     def event_market_cycle(self):
         self._read_predefined_profile_for_pv()
--- a/src/gsy_e/models/strategy/predefined_pv.py
+++ b/src/gsy_e/models/strategy/predefined_pv.py
@@ -17,7 +17,7 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import pathlib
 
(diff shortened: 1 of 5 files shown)


src/gsy_e/models/strategy/pv.py
```diff
@@ -221,17 +222,32 @@ def event_tick(self):
         self._settlement_market_strategy.event_tick(self)
         self._future_market_strategy.event_tick(self)
 
-    def set_produced_energy_forecast_kWh_future_markets(self, reconfigure=True):
+    def set_produced_energy_forecast_kWh_in_state(self, reconfigure=True):
+        self.set_produced_energy_forecast_kWh_spot_market(reconfigure)
+        self.set_produced_energy_forecast_kWh_future_market(reconfigure)
```
hannesdiedrich, line 227: I propose to remove the unit from all three method names to be consistent with the loads' equivalents. Also, the unit does not matter when calling the method (there will not be e.g. a `set_produced_energy_forecast_Wh_spot_market ` in the future).
Thanks!

src/gsy_e/models/strategy/predefined_pv.py
```diff
@@ -92,9 +92,9 @@ def __init__(
     def read_config_event(self):
         # this is to trigger to read from self.simulation_config.cloud_coverage:
         self.cloud_coverage = None
-        self.set_produced_energy_forecast_kWh_future_markets(reconfigure=True)
+        self.set_produced_energy_forecast_kWh_in_state(reconfigure=True)
 
-    def set_produced_energy_forecast_kWh_future_markets(self, reconfigure=True):
+    def set_produced_energy_forecast_kWh_spot_market(self, reconfigure=True):
```
hannesdiedrich, line 97: You need to reimplement `set_produced_energy_forecast_kWh_future_market` for the predefined PV, too, otherwise the gaussian profile will be saved for the future markets.
Another way would be to loop here again over the list of `[spot_market_slot, future_market_slot_1, future_market_slot_2,]` and call the `self.state.set_available_energy(available_energy_kWh, slot_time, reconfigure)
` for each

src/gsy_e/models/strategy/pv.py
```diff
@@ -221,17 +222,32 @@ def event_tick(self):
...
+    def set_produced_energy_forecast_kWh_future_market(self, reconfigure=True):
+        if not GlobalConfig.FUTURE_MARKET_DURATION_HOURS:
+            return
+        for slot_time in self.area.future_market_time_slots:
+            difference_to_midnight_in_minutes = slot_time.diff(
+                pendulum.datetime(
+                    slot_time.year, slot_time.month, slot_time.day)).in_minutes() % (60 * 24)
+            available_energy_kWh = self.gaussian_energy_forecast_kWh(
+                difference_to_midnight_in_minutes) * self.panel_count
+            self.state.set_available_energy(available_energy_kWh, slot_time, reconfigure)
```
hannesdiedrich, line 250: looks like this could be refactored into a method in order to reduce code duplication, OR you could follow my second proposal in https://github.com/gridsingularity/gsy-e/pull/1389/files#r810905663

hannesdiedrich (commented):
Left some comments, will review again.

src/gsy_e/models/strategy/predefined_pv.py
Hassan754: Thanks a lot for the proposal! I have adapted [here](https://github.com/gridsingularity/gsy-e/pull/1389/commits/e347f7e4505b98071c264021d85d12479eb03a6f)

hannesdiedrich (approved):
Thanks! LGTM

spyrostz (approved):

PR #1390, opened 2022-02-21
GSYE-154: Read the bid's requirement price when matching with DoF

Hassan754:
When the myco client matches a bid based on its price requirement, this requirement should outweigh the generic `price` member of the bid.
This PR deals with using `BidOfferMatch.bid_energy_rate` instead of the `Bid.energy_rate`, therefore handles figuring out where to read the price value from

### DIFF
--- a/src/gsy_e/models/market/grid_fees/__init__.py
+++ b/src/gsy_e/models/market/grid_fees/__init__.py
@@ -16,7 +16,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from abc import ABC, abstractmethod
-from typing import Tuple
+from typing import Tuple, Optional
 
 from gsy_framework.data_classes import TradeBidOfferInfo, Bid, Offer
 
@@ -48,37 +48,10 @@ class BaseClassGridFees(ABC):
         """Add fees for offer's rate when it's forwarded by another market."""
 
     @abstractmethod
-    def update_forwarded_bid_trade_original_info(
-            self, trade_original_info: TradeBidOfferInfo, market_bid: Bid) -> TradeBidOfferInfo:
-        """
-        When a forwarded bid gets matched in a target market, it will also get matched in
-        the source market with adjustments to the TradeBidOfferInfo of this trade.
-        This method deals with duplicating and updating the TradeBidOfferInfo values.
-        Args:
-            trade_original_info: TradeBidOfferInfo instance created in the target market
-            market_bid: the source bid instance that was cleared in a target market.
-
-        Returns: TradeBidOfferInfo
-        """
-
-    @abstractmethod
-    def update_forwarded_offer_trade_original_info(
-            self, trade_original_info: TradeBidOfferInfo, market_offer: Offer
-    ) -> TradeBidOfferInfo:
-        """
-        When a forwarded offer gets matched in a target market, it will also get matched in
-        the source market with adjustments to the TradeBidOfferInfo of this trade.
-        This method deals with duplicating and updating the TradeBidOfferInfo values.
-        Args:
-            trade_original_info: TradeBidOfferInfo instance created in the target market
-            market_offer: the source offer instance that was cleared in a target market.
-
-        Returns: TradeBidOfferInfo
-        """
-
-    @abstractmethod
-    def propagate_original_bid_info_on_offer_trade(self, trade_original_info):
-        """Add fees to the cleared bid trade.
+    def adapt_bid_fees_on_bid_trade(
+            self, trade_original_info: TradeBidOfferInfo, market_bid: Bid
+    ) -> Tuple[Optional[float], Optional[float], Optional[float]]:
+        """Add fees to the bid of the cleared bid trade.
 
         The reason we recalculate fees on trade even when we initially have these fees
         on the orders when they got forwarded is that the clearing rate is not always
@@ -86,8 +59,29 @@ class BaseClassGridFees(ABC):
         """
 
     @abstractmethod
-    def propagate_original_offer_info_on_bid_trade(self, trade_original_info, ignore_fees=False):
-        """Add fees to the cleared offer trade.
+    def adapt_offer_fees_on_offer_trade(
+            self, trade_original_info: TradeBidOfferInfo, market_offer: Offer
+    ) -> Tuple[Optional[float], Optional[float], Optional[float]]:
+        """Add fees to the offer of the cleared offer trade.
+
+        The reason we recalculate fees on trade even when we initially have these fees
+        on the orders when they got forwarded is that the clearing rate is not always
+        the offer or bid rate, It might change depending on the algorithm.
+        """
+
+    @abstractmethod
+    def adapt_bid_fees_on_offer_trade(
+            self, trade_original_info) -> Tuple[Optional[float], Optional[float], Optional[float]]:
+        """Add fees to the bid of the cleared offer trade.
+
+        The reason we recalculate fees on trade even when we initially have these fees
+        on the orders when they got forwarded is that the clearing rate is not always
+        the offer or bid rate, It might change depending on the algorithm.
+        """
+
+    @abstractmethod
+    def adapt_offer_fees_on_bid_trade(self, trade_original_info, ignore_fees=False):
+        """Add fees to the offer of the cleared bid trade.
 
         The reason we recalculate fees on trade even when we initially have these fees
         on the orders when they got forwarded is that the clearing rate is not always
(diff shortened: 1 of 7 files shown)


src/gsy_e/models/market/two_sided.py
```diff
@@ -344,7 +344,10 @@ def match_recommendations(
                     # re-raise exception to be handled by the external matcher
                     raise invalid_bop_exception
                 continue
-            original_bid_rate = market_bid.original_price / market_bid.energy
+            original_bid_rate = recommended_pair.bid_energy_rate - market_bid.accumulated_grid_fees
+            market_bid.update_price(
+                recommended_pair.bid_energy_rate * recommended_pair.selected_energy)
+            market_bid.original_price = original_bid_rate * recommended_pair.selected_energy
```
Hassan754, line 350: If the matching algorithm has used the requirements' price value, we should update the bid's generic price because latter in the code we frequently read this value

src/gsy_e/models/strategy/market_agents/two_sided_engine.py
```diff
@@ -149,8 +149,8 @@ def event_bid_traded(self, *, bid_trade):
 
             source_rate = bid_info.source_bid.energy_rate
             target_rate = bid_info.target_bid.energy_rate
-            assert abs(source_rate) + FLOATING_POINT_TOLERANCE >= abs(target_rate), \
-                f"bid: source_rate ({source_rate}) is not lower than target_rate ({target_rate})"
+            # assert abs(source_rate) + FLOATING_POINT_TOLERANCE >= abs(target_rate), \
+            #     f"bid: source_rate ({source_rate}) is not lower than target_rate ({target_rate})"
```
Hassan754, line 153: In the case of DoF, someone could decide to buy energy from a certain seller at a higher price, however this assertion here is being raised due to that, so I was wondering whether it's still relevant or we should delete it

src/gsy_e/models/market/grid_fees/__init__.py
```diff
@@ -48,46 +48,40 @@ def update_forwarded_offer_with_fee(self, source_rate: float, original_rate: flo
...
-        """
-        When a forwarded bid gets matched in a target market, it will also get matched in
-        the source market with adjustments to the TradeBidOfferInfo of this trade.
-        This method deals with duplicating and updating the TradeBidOfferInfo values.
-        Args:
-            trade_original_info: TradeBidOfferInfo instance created in the target market
-            market_bid: the source bid instance that was cleared in a target market.
-
-        Returns: TradeBidOfferInfo
+    def adapt_bid_fees_on_bid_trade(
```
Hassan754, line 51: I have noticed that in all occurrences in our codebase, we call either `update_forwarded_bid_trade_original_info` with `propagate_original_bid_info_on_offer_trade` or the `update_forwarded_offer_trade_original_info` with `propagate_original_offer_info_on_bid_trade`, in each of the mentioned methods we recreate a TradeBidOfferInfo object.
SO I simplified the interfaces and left the creation of this instance to the caller code.

src/gsy_e/models/market/two_sided.py
```diff
@@ -344,15 +344,16 @@ def match_recommendations(
                     # re-raise exception to be handled by the external matcher
                     raise invalid_bop_exception
                 continue
-            original_bid_rate = market_bid.original_price / market_bid.energy
+            original_bid_rate = recommended_pair.bid_energy_rate - market_bid.accumulated_grid_fees
```
hannesdiedrich, line 347: AFAIU, you are subtracting a price from a rate, which is not possible due to different units. Can you please explain?

src/gsy_e/models/market/two_sided.py
```diff
@@ -344,15 +344,16 @@ def match_recommendations(
...
                 trade_rate = original_bid_rate
             else:
                 trade_rate = self.fee_class.calculate_original_trade_rate_from_clearing_rate(
-                    original_bid_rate, market_bid.energy_rate, recommended_pair.trade_rate)
+                    original_bid_rate, recommended_pair.bid_energy_rate,
+                    recommended_pair.trade_rate)
             trade_bid_info = TradeBidOfferInfo(
                 original_bid_rate=original_bid_rate,
-                propagated_bid_rate=market_bid.energy_rate,
+                propagated_bid_rate=recommended_pair.bid_energy_rate,
```
hannesdiedrich, line 356: the `recommended_pair.bid_energy_rate` has already been adapted to grid_fees here right?

src/gsy_e/models/strategy/market_agents/two_sided_engine.py
hannesdiedrich: There is a bug in the assertion message `lower` should be `higher` right?
Apart from that, I also do not remember why this assertion is here. IMO it is already a bit late to assert here. It should be in the propagation already.
If we decide to remove it, please also do the same for the one sided engine: https://github.com/gridsingularity/gsy-e/blob/989ee1edee8b104efca334d92518575fc8268448/src/gsy_e/models/strategy/market_agents/one_sided_engine.py#L155

src/gsy_e/models/strategy/market_agents/one_sided_engine.py
```diff
@@ -157,22 +157,24 @@ def event_offer_traded(self, *, trade):
...
-                                       trade.fee_price / trade.traded_energy
+                    trade_offer_rate = trade.trade_rate - trade.fee_price / trade.traded_energy
                 else:
                     # trade_offer_rate not used in two sided markets, trade_bid_info used instead
                     trade_offer_rate = None
-                updated_trade_bid_info = \
-                    self.markets.source.fee_class.update_forwarded_offer_trade_original_info(
-                        trade.offer_bid_trade_info, offer_info.source_offer)
+                    trade.offer_bid_trade_info.original_offer_rate = (
+                        self.markets.source.fee_class.adapt_offer_fees_on_offer_trade(
```
hannesdiedrich, line 165: Isn't `adapt_offer_fees_on_offer_trade` returning a tuple? It does not return a trade_bid_info any more

src/gsy_e/models/market/grid_fees/constant_grid_fees.py
```diff
@@ -47,50 +47,35 @@ def update_forwarded_bid_with_fee(self, source_rate, original_rate):
...
         grid_fee_rate = self.grid_fee_rate if not ignore_fees else 0.0
         offer_rate = trade_original_info.propagated_offer_rate + grid_fee_rate
-        trade_offer_info = TradeBidOfferInfo(
-            original_bid_rate=None,
-            propagated_bid_rate=None,
-            original_offer_rate=trade_original_info.original_offer_rate,
-            propagated_offer_rate=offer_rate,
-            trade_rate=trade_original_info.trade_rate)
-        return trade_offer_info
+        return trade_original_info.original_offer_rate, offer_rate, trade_original_info.trade_rate
```
hannesdiedrich, line 78: The last element of the return tuple seems to be never used in the calls of `adapt_offer_fees_on_bid_trade`. IMO you can skip returning it.

hannesdiedrich (commented):

src/gsy_e/models/market/grid_fees/constant_grid_fees.py
Hassan754: You are right, we're not currently using it but I left it as this just in order to homogenize the `adapt_offer_fees_on_bid_trade` and `adapt_bid_fees_on_bid_trade` return value

src/gsy_e/models/strategy/market_agents/one_sided_engine.py
Hassan754: Indeed it's returning a tuple, I'm unpacking the returned values at the end of the line:
```python
                            trade.offer_bid_trade_info, offer_info.source_offer)[0])
```

src/gsy_e/models/market/two_sided.py
Hassan754: Yes, This is done when the bids are forwarded [here](https://github.com/gridsingularity/gsy-e/blob/989ee1edee8b104efca334d92518575fc8268448/src/gsy_e/models/strategy/market_agents/two_sided_engine.py#L57)

src/gsy_e/models/market/two_sided.py
Hassan754: Great catch, I've just fixed this.
Thanks!

src/gsy_e/models/strategy/market_agents/one_sided_engine.py
hannesdiedrich: Hmm, IMO this does not look nice. Unfortunately, I think the refactoring did not improve the readability. Just to be sure: It did not fix anything, it is only because of https://github.com/gridsingularity/gsy-e/pull/1390/files#r813677517 right?

Hassan754: @hannesdiedrich yes, the comment you mentioned is the reason for this refactoring. However, I have just had another look and I think that we can further simplify the interface to only return two elements, disregarding the trade_rate

LuisGSY (approved):
After testing in a community with several configurations (DoF), I conclude it seems to work as expected

PR #1392, opened 2022-02-28
GSYE-205: Enable residuals while DOF is active

Hassan754:


src/gsy_e/models/market/two_sided.py
```diff
@@ -442,16 +440,30 @@ def _replace_orders_with_residual_in_recommendations_list(
...
+            return recommendation
+
+        def _replace_recommendations_with_residuals(recommendation: Dict):
             if (recommendation["offer"]["id"] == offer_trade.offer_bid.id and
                     offer_trade.residual is not None):
                 recommendation["offer"] = offer_trade.residual.serializable_dict()
             if (recommendation["bid"]["id"] == bid_trade.offer_bid.id and
                     bid_trade.residual is not None):
                 recommendation["bid"] = bid_trade.residual.serializable_dict()
+                recommendation = _replace_matching_requirements(recommendation)
```
hannesdiedrich, line 463: why is this only added to the bid case?

Hassan754: This is because at the moment the `energy` requirement is only supported in the bid cases, so offers can't really have residual requirements

PR #1395, opened 2022-03-04
Bug/gsye 217 Omit calculating simulation results if not connected to gsy-web or no local export selected

hannesdiedrich:
Mein fix in 2d3095d04d0c69ac18bb4a5d7add901c4da19698

Unfortunately I could not think about a prettier solution.


src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -325,19 +327,19 @@ def update_stats(self, area: "Area", simulation_status: str,
...
-                ConstSettings.GeneralSettings.EXPORT_ENERGY_TRADE_PROFILE_HR):
-            self.offer_bid_trade_hr.update(area)
+            if (ConstSettings.GeneralSettings.EXPORT_OFFER_BID_TRADE_HR or
+                    ConstSettings.GeneralSettings.EXPORT_ENERGY_TRADE_PROFILE_HR):
+                self.offer_bid_trade_hr.update(area)
 
         self.result_area_uuids = set()
         self.update_results_area_uuids(area)
+
+        self.bids_offers_trades.clear()
```
spyrostz, line 342: I think this can be inside the `update_offer_bid_trade` method. And since this method is called only by this class, then I think that it is better to make it private, even though it is outside the scope of the ticket I think it would be good to fix

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -244,28 +244,34 @@ def update_area_stats(self, area, endpoint_buffer):
...
-                # for integration tests:
-                self.export.raw_data_to_json(
-                    self.area.current_market.time_slot_str,
-                    self.endpoint_buffer.flattened_area_core_stats_dict
-                )
-
-            self.file_stats_endpoint(self.area)
-
-        elif self.should_send_results_to_broker:
+        if self.should_send_results_to_broker:
```
spyrostz, line 252: Minor, but I think the change in the order will change the meaning of this if / elif block. To be more precise, the former behavior stated that "if there is no configuration to expose the simulation results at the end of the simulation, then check if the results should be sent to the broker", and the updated order is exactly the opposite. This will affect the scenario where the broker is online (therefore the should_send_results_to_broker variable will be true) and we want to export the results. With the previous implementation this worked without a problem, because the export check took precedence over the broker check, which is not the case now.

Regardless of the above, to my eyes it is optional since a CLI user will probably not have Kafka installed and running anyways.

spyrostz (commented):
2 comments, will LGTM once dealt with or discussed, thanks!

src/gsy_e/gsy_e_core/simulation.py
hannesdiedrich: Ah yes, let me try to also take this into account.

hannesdiedrich: @spyrostz 
I gave it another look and I think there is no need for a change, as the case you are mentioning :
> where the broker is online (therefore the should_send_results_to_broker variable will be true) and we want to export the results

We only want to export the results when we start the gsy-e from the cli, right? The `should_send_results_to_broker` already contains a check if the gsy-e is started from cli: https://github.com/gridsingularity/gsy-e/blob/2984dc21b1ddb6330e031617f6e676b649925265/src/gsy_e/gsy_e_core/simulation.py#L437

So, AFAIU, the first if block will not be triggered when we run the gsy-e locally but the Kafka connection is enabled.
What do you think?

spyrostz: You are right, I did not recall that we had already adapted it in order to explicitly define that the simulation was started from the CLI! Thanks for the clarification, it is good as it is!

spyrostz (approved):
LGTM

Hassan754:
@hannesdiedrich is this PR ready for review?

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -325,24 +327,24 @@ def update_stats(self, area: "Area", simulation_status: str,
...
         self.update_results_area_uuids(area)
-        self.update_offer_bid_trade()
 
-    def update_offer_bid_trade(self) -> None:
+        self._update_offer_bid_trade()
+
+    def _update_offer_bid_trade(self) -> None:
         """Populate self.bids_offers_trades with results from flattened_area_core_stats_dict
         (for local export of statistics)."""
+        self.bids_offers_trades.clear()
```
Hassan754, line 347: So basically we weren't clearing these before, did you find that by stumbling upon a bug or is this just a precaution (I'm afraid we need to be careful here)?

Hassan754 (commented):
Left one comment, LGTM once discussed/resolved

spyrostz (approved):
LGTM

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
hannesdiedrich: we need to clear the `bids_offers_trades` dict buffer from the last markets' bids, offers, trades before populating these with the new market results again. Have a look at master now, this step is done before calling _update_offer_bid_trade. I just moved it inside the method.

Hassan754: I see, in this case all good.

Hassan754 (approved):

PR #1396, opened 2022-03-04
GSYE-211: Forward offers' attributes and requirements

fievelk:
When being forwarded to a higher market, offers did not pass attributes and requirements to the creation of the new offer. This PR fixes the issue.

### DIFF
--- a/src/gsy_e/models/strategy/market_agents/one_sided_engine.py
+++ b/src/gsy_e/models/strategy/market_agents/one_sided_engine.py
@@ -51,6 +51,20 @@ class MAEngine:
             s=self
         )
 
+    def _update_offer_requirements_prices(self, offer):
+        requirements = []
+        for requirement in offer.requirements or []:
+            updated_requirement = {**requirement}
+            if "price" in updated_requirement:
+                energy = updated_requirement.get("energy") or offer.energy
+                original_offer_price = updated_requirement["price"] + offer.accumulated_grid_fees
+                updated_price = self.markets.target.fee_class.update_forwarded_offer_with_fee(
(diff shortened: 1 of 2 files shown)


src/gsy_e/models/strategy/market_agents/one_sided_engine.py
```diff
@@ -51,6 +51,20 @@ def __repr__(self):
             s=self
         )
 
+    def _update_offer_requirements_prices(self, offer):
```
Hassan754, line 54: Atm this is not needed for the offers because the don't support 'energy' nor 'price' requirement. However this is also not harmful and will be needed in case the business logic changes.

Hassan754 (approved):
lGTm

src/gsy_e/models/strategy/market_agents/one_sided_engine.py
fievelk: Thanks! Do we have some documentation/requirements about it where it would be possible to find this kind of information?

Hassan754: It's not explicitly mentioned but in the following [board](https://miro.com/app/board/o9J_lAuQITw=/) you can see that the requirements are mostly for the bids.
The supported requirements are stored in a constant variable [here](https://github.com/gridsingularity/gsy-framework/blob/fd57017b8e7847dabb23168537ab7ffefbf8c7f0/gsy_framework/matching_algorithms/requirements_validators.py#L107)

spyrostz (approved):
LGTM

PR #1397, opened 2022-03-14
GSYE-154: Read the bid's requirement price when matching with DoF

Hassan754:


src/gsy_e/models/market/two_sided.py
```diff
@@ -344,15 +344,17 @@ def match_recommendations(
                     # re-raise exception to be handled by the external matcher
                     raise invalid_bop_exception
                 continue
-            original_bid_rate = market_bid.original_price / market_bid.energy
-            if ConstSettings.MASettings.BID_OFFER_MATCH_TYPE == BidOfferMatchAlgoEnum.PAY_AS_BID:
+            original_bid_rate = recommended_pair.bid_energy_rate + (
```
hannesdiedrich, line 347: Is this really a `+` ? Shouldn't be the grid_fee rate subtracted from the `bid_energy_rate` in order to get the `original_bid_rate`?

src/gsy_e/models/market/two_sided.py
```diff
@@ -344,15 +344,17 @@ def match_recommendations(
...
+                    market_bid.accumulated_grid_fees / recommended_pair.bid_energy)
+            if ConstSettings.MASettings.BID_OFFER_MATCH_TYPE == (
+                    BidOfferMatchAlgoEnum.PAY_AS_BID.value):
                 trade_rate = original_bid_rate
```
hannesdiedrich, line 351: I probably still not get it right, but I would expect:
```suggestion
                trade_rate = recommended_pair.bid_energy_rate + ( market_bid.accumulated_grid_fees / recommended_pair.bid_energy)
```

src/gsy_e/models/market/two_sided.py
Hassan754: In the case of offers, this is true. However, the bid's original rate is higher than the propagated rate, so the original rate is indeed `propagated_rate`+ `fees`

src/gsy_e/models/market/two_sided.py
Hassan754: You're right, this was messed up after several rounds of adjustments.
It is now fixed

src/gsy_e/models/market/two_sided.py
hannesdiedrich: Sorry, I do not agree.
Lets say the original bid in the House market has the rate of 30ct/kWh.
If this bid is forwarded to the Community market with 2ct/kWh fee, the propagated bid rate is 32cent/kWh.
Consequently: original_bid_rate (30) = propagated_rate (32) - fee_rate (2)
Or do I get it totally wrong?

Hassan754: I think that it's the opposite, in your example the bid in the community market will be 28 cents (The original buyer isn#t willing to pay more than 30).
So the bid will be cleared at 28 cents/kwh, the community area will sell the energy to the bid buyer for 30 cents/kwh (SO its share of 2 cents/kwh is guaranteed)

hannesdiedrich: Thanks a lot for bering with me! I had to recapitulate how the grid fees were implemented. You are right of course, the seller should pay the fees and the market is called `pay as bid`, so the bid price should not be exceeded. Thanks again!

src/gsy_e/models/strategy/market_agents/one_sided_engine.py
```diff
@@ -165,11 +164,6 @@ def event_offer_traded(self, *, trade):
 
         if trade.offer_bid.id == offer_info.target_offer.id:
             # Offer was accepted in target market - buy in source
-            source_rate = offer_info.source_offer.energy_rate
-            target_rate = offer_info.target_offer.energy_rate
-            assert abs(source_rate) <= abs(target_rate) + FLOATING_POINT_TOLERANCE, \
```
spyrostz, line 170: Why remove this assertion? I find it very valuable in order to prevent bugs in the chain trade operation.

spyrostz (commented):
One comment, will LGTM once discussed or dealt with

src/gsy_e/models/strategy/market_agents/one_sided_engine.py
spyrostz: @Hassan754 I am aware that this condition can fail for the case of degrees of freedom, but I still think that these asserts can remain for the default case. You can have an if check before the assert in order to circumvent it if any dof has been set. That way the check will still apply for non-dof trades.

PR #1404, opened 2022-03-18
TESTS-138: Template PR message not showing when PR is created

ERafaelMartinez:
Recently, a PR template message file was created on gsy-e, gsy-e-sdk, gsy-myco-sdk and gsy-web repositories including a line which specifies wether a particular branch from GBIT repository should be called. This PR includes a fix for the template to load on a new PR to base branch.

This PR also includes, just for safety, a default cloning branch for GBIT repo implementation, considering the possibility that (by mistake) no GBIT cloning branch is specified.


.github/workflows/gsy-e_ci.yaml
```diff
@@ -21,9 +21,16 @@ jobs:
       env:
         prBody: ${{ steps.PR.outputs.pr_body }}
       run: |
-        echo "::set-output name=INTEGRATION_TESTS_BRANCH::$(echo -e $prBody | sed -n 's/.*\*\*INTEGRATION_TESTS_BRANCH\*\*=\([^ ]*\).*/\1/p')"
+        echo "::set-output name=PARSED_BRANCH::$(echo -e $prBody | sed -n 's/.*\*\*INTEGRATION_TESTS_BRANCH\*\*=\([^ ]*\).*/\1/p')"
```
hannesdiedrich, line 24: I propose to use a more explicit name, in order to not get confused later (when we maybe have to add more branches to the template),
```suggestion
        echo "::set-output name=PARSED_INTEGRATION_TESTS_BRANCH::$(echo -e $prBody | sed -n 's/.*\*\*INTEGRATION_TESTS_BRANCH\*\*=\([^ ]*\).*/\1/p')"
```

hannesdiedrich (commented):

.github/pull_request_template.md
```diff
@@ -0,0 +1,4 @@
+TODO: PR description and edit branch pointing to
+      gsy-backend-integration-tests repository.
+
+#INTEGRATION_TESTS_BRANCH=master
```
fievelk, line 4: This PR template is outdated

.github/workflows/gsy-e_ci.yaml
```diff
@@ -21,9 +21,16 @@ jobs:
...
       id: parse_branch
 
+    - name: validate parsed branch
+      env:
+        PARSED_BRANCH: ${{ steps.parse_branch.outputs.PARSED_BRANCH }}
+      run: |
+        echo "::set-output name=INTEGRATION_TESTS_BRANCH::$( [ -z $PARSED_BRANCH ] && echo -e 'master' || echo -e $PARSED_BRANCH)"
```
fievelk, line 31: Could you please explain why `echo` is needed in this part?
```sh
$( [ -z $PARSED_BRANCH ] && echo -e 'master' || echo -e $PARSED_BRANCH)
```

fievelk: In any case, I think this syntax might be more suitable:
```sh
${PARSED_BRANCH:-'master'}
```
(Based on bash's [Shell Parameter Expansion](https://www.gnu.org/software/bash/manual/bash.html#Shell-Parameter-Expansion))

.github/workflows/gsy-e_ci.yaml
spyrostz: The problem here is that this might need to be also set locally. As far as I see, this precedes the tox commands, therefore if we make this change, we will have to change tox to use PARSED_INTEGRATION_TESTS_BRANCH, which is not ideal for the local user IMO.

spyrostz (approved):
LGTM

.github/workflows/gsy-e_ci.yaml
ERafaelMartinez: @spyrostz 
This variable (PARSED_BRANCH / PARSED_INTEGRATION_TESTS_BRANCH) is only "auxiliar", meaning that it is only used inside the git workflow. It serves to evaluate wether anything was parsed from the PR message, and then in a next step, INTEGRATION_TESTS_BRANCH is set according to wether PARSED_BRANCH is null or not.

It is not entirely clear for me why you say this might be needed on the tox file. The variable that is used on the tox file is INTEGRATION_TESTS_BRANCH, which remains unchanged.

fievelk (approved):

fievelk (approved):

.github/workflows/gsy-e_ci.yaml
spyrostz: @ERafaelMartinez that's my bad, I thought while reviewing that it would also pass through tox and would be used by it. Since it is only used by the CI configuration then all good from my side. Thanks for the clarification here!

ERafaelMartinez: No worries! Thanks for reviewing :)

hannesdiedrich (approved):
LGTM

PR #1409, opened 2022-03-23
Gsye 233 simulation refactoring

spyrostz:
Various refactoring in order to hopefully simplify and split the concerns of the Simulation class. The following classes are created in the context of this PR: 
- SimulationStatusManager: handles the changes in the simulation status (e.g. pausing / stopping a simulation)
- SimulationTimeManager: handles simulation time tracking and related operations 
- SimulationSetup: handles the initial setup of the simulation (e.g. loading a setup module) and includes related parameters
- SimulationResultsManager: handles simulation results, and encapsulates both the Kafka and CSV files flow in one class. Also includes endpoint_buffer, kafka connection and other objects that are relevant to the simulation results. 
- SimulationExternalEvents: encapsulates live events + other signals that arrive from gsy-web during the runtime of the simulation, e.g. stop/pause simulation

There are many more things that can still be abstracted away from the main class that I did not deal with in order to decrease the size and the impact of the changes (first 3 that come to mind is the simulation state / restore from the database, global objects and simulation_config that should probably belong to the SimulationSetup class)

Unfortunately the changes did not allow me to achieve my original goal to drop the number of member variables of the Simulation class under 7 in order to remove the pylint warning (they are 9 at the moment), but still a decrease from 21 variables is a significant improvement. 

The PR is branched off https://github.com/gridsingularity/gsy-e/pull/1395 so it needs to be merged after it.

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-223


src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -52,14 +52,14 @@
...
-    def __init__(self, job_id, initial_params, area, should_export_plots):
+    def __init__(self, job_id, random_seed, area, should_export_plots):
         self.job_id = job_id
         self.result_area_uuids = set()
         self.current_market_time_slot_str = ""
         self.current_market_ui_time_slot_str = ""
         self.current_market_time_slot_unix = None
         self.current_market_time_slot = None
-        self.random_seed = initial_params["seed"] if initial_params["seed"] is not None else ""
+        self.random_seed = random_seed if random_seed is not None else ""
```
Hassan754, line 62: Nit
```suggestion
        self.random_seed = random_seed or ""
```

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -68,278 +69,462 @@ def __init__(self):
...
-        else:
-            self.export_subdir = export_subdir
+@dataclass
+class SimulationStatusManager:
+    """State of the Simulation class."""
+    paused: bool = False
+    pause_after: duration = None
+    timed_out: bool = False
+    stopped: bool = False
+    sim_status = "initializing"
```
Hassan754, line 113: DO you think creating an enum for the simulation statuses makes sense here?

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -68,278 +69,462 @@ def __init__(self):
...
+class SimulationSetup:
+    """Static simulation configuration."""
+    seed: int = 0
+    enable_bc: bool = False
+    use_repl: bool = False
+    setup_module_name: str = ""
+    started_from_cli: str = True
+
+    def __post_init__(self):
+        self.seed = self._set_random_seed(self.seed)
```
Hassan754, line 253: Nit, by convention setters should return nothing, so I propose moving the value assignment inside the `self._set_random_seed`

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -70,66 +70,195 @@ def __init__(self):
...
     timed_out: bool = False
     stopped: bool = False
-    paused_time: int = 0  # Time spent in paused state, in seconds
     sim_status = "initializing"
-    start_time: datetime = None
-    tick_time_counter: datetime = None
+    incremental: bool = False
 
+    @property
+    def status(self):
```
Hassan754, line 117: I see you created this property, can it include the `initializing` , `finished` and the other statuses as well?

Hassan754 (dismissed):
Left some comments, looks awesome, Thanks!

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -68,278 +69,462 @@ def __init__(self):
...
+        """
+        if self.incremental:
+            self.paused = True
+
+
+@dataclass
+class SimulationTimeManager:
+    """Handles simulation time management."""
+    start_time: datetime = now(tz=TIME_ZONE)
+    tick_time_counter: datetime = time()
```
hannesdiedrich, line 173:
```suggestion
    tick_time_counter: DateTime = time()
```
In order to make sure that pendulum datetime is used here and the `.add()` in `handle_slowdown_and_realtime` works. Thanks!

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -68,278 +69,462 @@ def __init__(self):
...
-            sleep(seconds_until_next_tick)
-
-        if self.slot_length_realtime:
-            self.tick_length_realtime_s = self.slot_length_realtime.seconds / \
-                                          self.simulation_config.ticks_per_slot
-        return slot_count, slot_resume, tick_resume
-
     def _execute_simulation(self, slot_resume, tick_resume, console=None):
-        self.current_expected_tick_time = self.run_start
         config = self.simulation_config
```
hannesdiedrich, line 516: I stumble over this a lot. How about taking the opportunity and either use `config` in the whole method or (my preference) use `self. simulation_config` only. What do you think?

src/gsy_e/gsy_e_core/simulation.py
hannesdiedrich, line 253: Can you please explain, why this can not be done when initialising the data class?

src/gsy_e/gsy_e_core/live_events.py
```diff
@@ -43,39 +41,46 @@ def __repr__(self):
...
+            if area_type is not None and area_type != "Area":
+                if area.strategy is None:
+                    return False
+                reactivate = False
+                if area_type == "MarketMaker":
+                    area.strategy = MarketMakerStrategy(**self.area_params)
+                    reactivate = True
+                elif area_type == "InfiniteBus":
+                    # pylint: disable=fixme
+                    # TODO: After hack to move this parameter casting at the web side
```
hannesdiedrich, line 65: Can you please either remove it and maybe create a ticket for this? Thanks!

src/gsy_e/gsy_e_core/simulation.py
spyrostz: I do not return something here actually. I am just calculating a new random seed value and reassigning to self.seed. Therefore I think it is fine as it is, and it it more clear because the member variable is mutated directly in post_init and is not hidden inside another method, therefore it is more clear.

src/gsy_e/gsy_e_core/redis_connections/redis_communication.py
```diff
@@ -24,33 +24,41 @@
...
+        # pylint: disable=too-many-arguments
         self._live_events = live_events
         self._simulation_id = simulation_id if simulation_id is not None else ""
-        self._configuration_id = gsy_e.constants.CONFIGURATION_ID
-        self._simulation = simulation
+        self._simulation_status = simulation_status
+        self._area = area
+        self._progress_info = progress_info
+        # Lambda is needed here in order to avoid storing the area as a member variable.
+        # Instead a closure is used that will capture the area variable in the function scope.
```
hannesdiedrich, line 59: Sorry, I do not get this comment. Is it maybe outdated and remnant of intermediate state?

hannesdiedrich (commented):
Thanks a lot for this refactoring! 
I left some rather minor comments. 
Additionally, it would be great if you could add types to the input and output parameters of the methods. Thanks!

src/gsy_e/gsy_e_core/simulation.py
spyrostz: Because the data class constructor only accepts values and cannot execute a function call and then pass the result to the member variable. In order to do the latter, you have to pass the variable you intend to use, and then use the post_init method to call the function that will mutate its value. At least that's what I got from the documentation of the dataclasses.
The obvious fix for this is to ditch the dataclass and use a normal class for that, and admittedly I was also wondering which one of the 2 is the correct solution. Therefore if you would prefer a normal class instead lead me know and I will change it.

src/gsy_e/gsy_e_core/redis_connections/redis_communication.py
spyrostz: It is indeed an outdated comment. I tried initially to get rid of the `self._area` member by capturing the constructor value in a closure/lambda, but after seeing that pylint allows us for more member variables that this class had I decided to remove this in order to be easier for the reader. I will remove the comment, thanks!

src/gsy_e/gsy_e_core/redis_connections/redis_communication.py
```diff
@@ -24,33 +24,41 @@
...
 
-REDIS_URL = os.environ.get('REDIS_URL', 'redis://localhost')
+REDIS_URL = os.environ.get("REDIS_URL", "redis://localhost")
 
 
 if TYPE_CHECKING:
     from gsy_e.models.area import Area
 
 
 class RedisSimulationCommunication:
```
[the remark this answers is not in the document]
spyrostz: that's a good idea, will do so.

src/gsy_e/gsy_e_core/simulation.py
hannesdiedrich: Thanks for the explanation! I am not sure. I think it is not worth to refactor again, leave like it is. Thanks!

Hassan754: I was talking about the `self._set_random_seed(self.seed)` statement, this is a call to the setter and therefore it should not return any value (the assignment should be inside the _set_random_seed)

src/gsy_e/gsy_e_core/simulation.py
spyrostz: Let me actually do it. I will self-contain it in one commit and hopefully it will not be that much of a fuss

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
spyrostz: Fixing, thanks!

src/gsy_e/gsy_e_core/simulation.py
spyrostz: Fixing, thanks!

src/gsy_e/gsy_e_core/simulation.py
spyrostz: Ok, now I reread your first comment `by convention setters should return nothing` and understand what you meant. Will do it, thanks for bearing with me!

Hassan754 (approved):

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -438,40 +450,39 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig,
...
 
         validate_const_settings_for_simulation()
 
-    def _init(self, redis_job_id):
+    def _init(self, redis_job_id: str) -> None:
         # has to be called before get_setup():
         global_objects.profiles_handler.activate()
 
-        self.area = self._setup.load_setup_module(self.simulation_config)
+        self.area = self._setup.load_setup_module(self._setup.config)
```
hannesdiedrich, line 468: Now that the `self._setup` already has access to the simulation_config itself, how about omitting passing it here again?

spyrostz: You are right I missed that, will fix it thanks!

hannesdiedrich (commented):
One last comment. LGTM once addressed

hannesdiedrich:
https://github.com/gridsingularity/gsy-e/pull/1409/commits/a94d1814c7d853f3e6ea99fc1b8e9fe04e187086 seems to be an empty commit. Is this supposed to be like this or is github failing me again?

hannesdiedrich (approved):
LGTM

spyrostz:
> [a94d181](https://github.com/gridsingularity/gsy-e/commit/a94d1814c7d853f3e6ea99fc1b8e9fe04e187086) seems to be an empty commit. Is this supposed to be like this or is github failing me again?

For me it shows the diff correctly as we speak, so I would attribute it to github yes.

Hassan754 (approved):

PR #1410, opened 2022-04-01
GSYE-229: Set current_tick to offer_age when the residual bid is trad…

hannesdiedrich:
…ed right after creating and before the bid_age could be populated

The following scenario led to this KeyError of [this line](https://github.com/gridsingularity/gsy-e/blob/2984dc21b1ddb6330e031617f6e676b649925265/src/gsy_e/models/strategy/market_agents/two_sided_engine.py#L239):

1. A bid is partially traded and is consequently split into `traded_bid` and `residual_bid`. 
2. The `residual_bid` is traded in the same tick and again, the `TwoSidedEngine.event_bid_split` method is called. At the end of the handling of the forwarded bids, the `bid_age` entry of the `residual_bid` is overwritten by the bid_age of the `traded_bid`. However, this entry does not exist yet, because the `event_tick` method has not been called in the meantime.

## Proposed changes

- If the `bid_age` entry for the `traded_bid` does not exist yet, write the `current_tick` to the bid_age entry for the residual_bid


spyrostz (approved):
LGTM

Hassan754 (approved):

PR #1411, opened 2022-04-01
GSYE-242: Added the event_activate_price method to the BaseStrategy c…

spyrostz:
…lass with an empty implementation, in order to provide a default implementation for all strategy types.

Exception was raised because the InfiniteBusStrategy was missing an implementation for the event_activate_price method 

## Proposed changes

Added the method in the BaseStrategy class in order to be inherited from all strategies and provide a default (empty) implementation.


Hassan754 (approved):

fievelk (approved):

hannesdiedrich (approved):
LGTM

PR #1412, opened 2022-04-01
Fix redis naming

spyrostz:

Unfortunately naming a package as `redis` is not allowed, since apparently redis-py does not play well with it https://github.com/redis/redis-py/issues/533#issuecomment-56084464 

## Proposed changes

Renamed the redis package to redis_connections, and removed the "connection" part of the underlying modules names


mclrch (approved):
LGTM

Hassan754 (approved):

PR #1413, opened 2022-04-04
TESTS-143: I want to reduce tox files redundancy on tox file testenvs.

ERafaelMartinez:

Modularization of the testenvs in tox file was improved for gsy-web repository. The changes on this repo will match up with that done on gsy-web.

## Proposed changes

Improve the modularization of the tox configuration environments.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -10,25 +10,19 @@ exclude = .tox,wiki,.cache,.d3a,.hypothesis,.pytest_cache,vagrant,requirements,v
 # environments unless overridden
 [testenv]
 basepython = python3.8
-passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH SOLC_BINARY INTEGRATION_TESTS_REPO INTEGRATION_TESTS_BRANCH
+passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH SOLC_BINARY GSY_FRAMEWORK_BRANCH
 allowlist_externals =
     git
     /bin/rm
     /bin/ln
     bash
-deps =
-    pip-tools
-    coverage
 commands_pre =
     pip install --upgrade pip
     pip install -rrequirements/tests.txt
 
-[testenv:covclean]
-skip_install = true
-commands = coverage erase
-
 [testenv:coverage]
 skip_install = true
+deps = coverage
 commands_pre =
     {[testenv]commands_pre}
     pip install -e .
@@ -38,12 +32,18 @@ commands =
     coverage report
     coverage xml
 
+[testenv:covclean]
+skip_install = true
+deps = {[testenv:coverage]deps}
+commands = coverage erase
+
 [testenv:check_readme]
 skip_install = true
 deps = readme_renderer
 commands = python setup.py check --restructuredtext --strict
 
 [testenv:integrationtests]
+passenv = {[testenv]passenv} INTEGRATION_TESTS_REPO INTEGRATION_TESTS_BRANCH
 commands_pre =
     {[testenv]commands_pre}
     pip install -e .
@@ -58,6 +58,8 @@ commands_post =
     rm ./integration_tests
 
 [testenv:ci]
+passenv = {[testenv:integrationtests]passenv}
+deps = {[testenv:coverage]deps}
 commands_pre =
     python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
     pip install -rrequirements/pandapower.txt
@@ -72,6 +74,7 @@ commands_post =
     {[testenv:integrationtests]commands_post}
 
 [testenv:test_dispatch_events_top_to_bottom]
+passenv = {[testenv:integrationtests]passenv}
 setenv =
     DISPATCH_EVENTS_BOTTOM_TO_TOP = False
 commands_pre =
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -2,58 +2,64 @@
...
-basepython = python3.8
 skip_install = true
-deps = coverage
 commands =
     coverage combine
     coverage report
 
+[testenv:codecoverage]
+commands_pre =
+    {[testenv]commands_pre}
```
fievelk, line 37: This should be already automatically inherited by the `testenv` environment, so I don't think we need to override it.
```suggestion
```

tox.ini
```diff
@@ -2,58 +2,64 @@
...
 
 [testenv:coverage]
-basepython = python3.8
 skip_install = true
-deps = coverage
 commands =
     coverage combine
     coverage report
 
+[testenv:codecoverage]
```
fievelk, line 35: I'd remove this `codecoverage` environment and just add everything inside `coverage`. I don't think we ever need to generate coverage reports without running pytest.

tox.ini
```diff
@@ -62,63 +68,29 @@ commands_post =
...
 commands_pre =
     python -c "import fcntl; fcntl.fcntl(1, fcntl.F_SETFL, 0)"
-    pip-sync requirements/tests.txt requirements/pandapower.txt
-    pip install -e .
+    pip-sync requirements/pandapower.txt
+    {[testenv:integrationtests]commands_pre}
     pip uninstall -y gsy-framework
     pip install git+https://github.com/gridsingularity/gsy-framework@{env:GSY_FRAMEWORK_BRANCH:master}
 commands =
     flake8
```
fievelk, line 80: I know it's the same, but this would show that we're re-using another env.
```suggestion
    {[testenv:flake8]commands}
```
(To be honest, I'm not completely sure that we need the `testenv:flake8` environment at all in our tox file).

tox.ini
```diff
@@ -62,63 +68,29 @@ commands_post =
...
-    -rrequirements/tests.txt
 setenv =
     DISPATCH_EVENTS_BOTTOM_TO_TOP = False
-allowlist_externals =
-    git
-    /bin/ln
-    /bin/rm
-    bash
+commands_pre =
+    {[testenv:integrationtests]commands_post}
```
fievelk, line 92:
```suggestion
commands_pre =
    {[testenv:integrationtests]commands_pre}
```

tox.ini
fievelk: Sorry, I just noticed that the `coverage` env didn't run pytest. Why do you think we need this `codecoverage` env?

fievelk (commented):
I've left some comments. Please make sure to try running tox with each test env individually, to double-check everything works.

tox.ini
ERafaelMartinez: I'm not sure TBH. The testenv was already there I just moved it up to be together with coverage testenv as they are "similar".

tox.ini
ERafaelMartinez: Ok. I'll adapt it. But if it is not necessary at all I could better remove it. 
Could it be maybe needed for the pre-commit workflow?

tox.ini
fievelk: I agree with you 👍

tox.ini
fievelk: Nope, that should work independently as far as I know. I think we can remove the `[testenv:flake8]` env.

The only part we should keep is the `[flake8]` one: https://github.com/gridsingularity/gsy-e/blob/228599c33ba283021d1357309bec3f71428da739/tox.ini#L61-L63

because that's actually the flake8 configuration we're using. I'm not particularly fond of having it inside the `tox.ini` file, but that's ok for the moment.

ERafaelMartinez: Ok, let's do it that way then. I agree with you.

ERafaelMartinez:
Adapted to comments. Thank you for your re-review.

tox.ini
```diff
@@ -2,123 +2,81 @@
...
+basepython = python3.8
+passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH SOLC_BINARY INTEGRATION_TESTS_REPO INTEGRATION_TESTS_BRANCH
+allowlist_externals =
+    git
+    /bin/rm
+    /bin/ln
+    bash
 deps =
     pip-tools
     coverage
```
fievelk, line 21: I know this was already like this, but are we sure that coverage should be here? I think there are environments that don't need it.

tox.ini
```diff
@@ -2,123 +2,81 @@
...
-passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH SOLC_BINARY INTEGRATION_TESTS_REPO
+basepython = python3.8
+passenv = LANG TERM LANGUAGE LC_ALL LD_LIBRARY_PATH SOLC_BINARY INTEGRATION_TESTS_REPO INTEGRATION_TESTS_BRANCH
+allowlist_externals =
+    git
+    /bin/rm
+    /bin/ln
+    bash
 deps =
     pip-tools
```
fievelk, line 20: You've removed the usage of `pip sync`, so I think that `pip-tool` is not required as a dependency anymore here. Could you please check?

fievelk (commented):
Left some comments.

tox.ini
ERafaelMartinez: You are right. I'll readapt here and in other repos PRs.

Hassan754 (approved):

PR #1414, opened 2022-04-04
Fix log levels when receiving incorrect orders

fievelk:

This error is currently spamming our logs, but it is often caused by an incorrect order sent by the user. Therefore, the message is just part of the regular flow and should probably not be logged as an exception, but rather as an INFO message.

## Proposed changes

- Change log level of the `OrderCanNotBePosted` exception handling to `INFO`.

### DIFF
--- a/src/gsy_e/models/strategy/external_strategies/__init__.py
+++ b/src/gsy_e/models/strategy/external_strategies/__init__.py
@@ -369,8 +369,14 @@ class ExternalMixin:
                 "area_uuid": self.device.uuid,
                 "market_type": market.type_name,
                 "message": response_message}
-        except (OrderCanNotBePosted, GSyException):
-            logging.exception("Error when handling offer on area %s", self.device.name)
+        except (OrderCanNotBePosted, GSyException) as ex:
+            if isinstance(ex, OrderCanNotBePosted):
+                # This can happen in the normal flow, when the user sends an incorrect offer
+                logging.info("Error when handling offer on area %s. %s", self.device.name, ex)
+            else:
+                # This is more unexpected and we might want to be notified
+                logging.exception("Error when handling offer on area %s. %s", self.device.name, ex)
+
             response = {
                 "command": "offer", "status": "error",
                 "market_type": market.type_name,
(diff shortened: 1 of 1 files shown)


src/gsy_e/models/strategy/external_strategies/__init__.py
```diff
@@ -369,8 +369,14 @@ def _offer_aggregator_impl(
...
                 "message": response_message}
-        except (OrderCanNotBePosted, GSyException):
-            logging.exception("Error when handling offer on area %s", self.device.name)
+        except (OrderCanNotBePosted, GSyException) as ex:
+            if isinstance(ex, OrderCanNotBePosted):
+                # This can happen in the normal flow, when the user sends an incorrect offer
+                logging.info("Error when handling offer on area %s. %s", self.device.name, ex)
+            else:
+                # This is more unexpected and we might want to be notified
+                logging.exception("Error when handling offer on area %s. %s", self.device.name, ex)
```
Hassan754, line 378: nit, maybe we can chain exception handlers
```suggestion
        except OrderCanNotBePosted as ex:
                 # This can happen in the normal flow, when the user sends an incorrect offer
                logging.info("Error when handling offer on area %s. %s", self.device.name, ex)
        except OrderCanNotBePosted as ex:
                 # This is more unexpected and we might want to be notified
                logging.exception("Error when handling offer on area %s. %s", self.device.name, ex)
```

Hassan754 (approved):
Left one comment, LGTM

src/gsy_e/models/strategy/external_strategies/__init__.py
fievelk: The issue with this is that we still need to execute these lines inside the exception: https://github.com/gridsingularity/gsy-e/blob/228599c33ba283021d1357309bec3f71428da739/src/gsy_e/models/strategy/external_strategies/__init__.py#L374-L380
I think that, if we chain exception handlers as you suggest, there's no way to do that without duplicating code. Or do you see an alternative?

hannesdiedrich (approved):
LGTM

src/gsy_e/models/strategy/external_strategies/__init__.py
Hassan754: I see, in this case all good. Thanks!

spyrostz: There is also the `finally` statement that could potentially be used here, but frankly this is one of the least frequently used features by me because I find the exception control flow much more confusing than with if / else's. It is feasible, but I would probably discourage you from trying it.

spyrostz (approved):

src/gsy_e/models/strategy/external_strategies/__init__.py
fievelk: I thought about it initially, but the `finally` is executed even when the exceptions are not raised. In our case, however, the code should only be executed when an exception is raised (it should not be executed when everything works smoothly).

PR #1416, opened 2022-04-13
Feature/gsye 254

spyrostz:

In order to be able to implement the coefficients based trading many parts of gsy-e need to be switched off, because they only apply to market trading.

## Proposed changes

Split different parts of the codebase in order to have separate flows according to whether the Coefficients based trading is selected or not.
-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-254


src/gsy_e/gsy_e_core/area_serializer.py
```diff
@@ -134,7 +132,7 @@ def optional(attr):
         grid_fee_percentage = description.get('grid_fee_percentage', None)
         grid_fee_constant = description.get('grid_fee_constant', None)
         area = Area(name, children, uuid, optional('strategy'), config,
-                    optional('budget_keeper'), grid_fee_percentage=grid_fee_percentage,
+                    grid_fee_percentage=grid_fee_percentage,
```
fievelk, line 135: Should we remove `budget_keeper` also from the `Area` class? E.g. https://github.com/gridsingularity/gsy-e/blob/1fb0a9900d63329ee6f387d81b98ac6fb08aa51a/src/gsy_e/models/area/__init__.py#L115
(and other usages)

src/gsy_e/models/config.py
```diff
@@ -129,3 +129,25 @@ def read_market_maker_rate(self, market_maker_rate):
         Reads market_maker_rate from arbitrary input types
         """
         self.market_maker_rate = read_and_convert_identity_profile_to_float(market_maker_rate)
+
+
+def create_simulation_config_from_global_config():
+    """
+    Create a SimulationConfig object from the GlobalConfig class members.
+    These 2 object are not currently in sync because qthe following parameters are missing from the
```
hannesdiedrich, line 137:
```suggestion
    These 2 object are not currently in sync because the following parameters are missing from the
```

src/gsy_e/models/area/__init__.py
```diff
@@ -48,20 +48,7 @@
...
-# TODO: As this is only used in the unittests, please remove it here and replace the usages
-#       of this class with gsy-framework.constants_limits.GlobalConfig class:
-DEFAULT_CONFIG = SimulationConfig(
-    sim_duration=duration(hours=24),
-    slot_length=duration(minutes=15),
-    tick_length=duration(seconds=1),
-    cloud_coverage=ConstSettings.PVSettings.DEFAULT_POWER_PROFILE,
-    start_date=today(tz=gsy_e.constants.TIME_ZONE),
-    capacity_kW=ConstSettings.PVSettings.DEFAULT_CAPACITY_KW
-)
```
hannesdiedrich, line 64: ❤️

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -403,7 +403,7 @@ def _populate_core_stats_and_sim_state(self, area: "Area"):
             else:
                 if area.parent.current_market is not None:
                     core_stats_dict["energy_rate"] = (
-                        area.strategy.energy_rate.get(area.current_time_slot, None))
+                        area.strategy.energy_rate.get(area.now, None))
```
hannesdiedrich, line 406: Hmm, not sure whether this will work, because `area.now` also contains the tick duration. How about using `self.current_market_time_slot` ?

src/gsy_e/models/area/__init__.py
```diff
@@ -152,12 +149,124 @@ def name(self, new_name):
...
+            for child in self.children:
+                child.update_descendants_strategy_prices()
+        except GSyException:
+            log.exception("area.update_descendants_strategy_prices failed.")
+            return
+
+
+# pylint: disable=too-many-instance-attributes
+# pylint: disable=too-many-public-methods
+class Area(AreaBase):
```
hannesdiedrich, line 248: In order to limit the disabling only to the class scope, please move these inside of the class.
Or did I get this wrong before?

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
hannesdiedrich: Sorry, only see the other changes now. Will remove this comment.

src/gsy_e/models/area/__init__.py
```diff
@@ -224,6 +238,10 @@ def update_descendants_strategy_prices(self):
             log.exception("area.update_descendants_strategy_prices failed.")
             return
 
+    def cycle_markets(self, current_time_slot: str):
+        """Perform operations that should be executed on market cycle."""
+        self._current_time_slot = current_time_slot
```
hannesdiedrich, line 243: You are assigning a `str` to `self._current_time_slot`, but the `now` property is supposed to return a `DateTime` object, which is needed for the `area.strategy.energy_rate.get(area.now, None))` lookup in the endpoint_buffer. AFAIU, you should pass a `DateTime` object to this method here in simulation.py

hannesdiedrich (commented):
One major and a few minor comments. LGTM in general. 
Thanks a lot for the refactoring!

src/gsy_e/models/area/__init__.py
spyrostz: You are correct, this indeed needs to be a datetime object. Fixing it now thanks!

src/gsy_e/models/area/__init__.py
spyrostz: Will fix this, it was like this before IIRC, that's why it went under my radar. I have the same recollection too, that in order for the pylint disables to be applied in the class scope they have to be declared inside the class.

src/gsy_e/models/config.py
spyrostz: Fixing, thanks!

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -350,3 +355,66 @@ def _update_offer_bid_trade(self) -> None:
...
+        if self.current_market_time_slot_str == "":
+            return
+
+        core_stats_dict = {"trades": []}
+        for trade in area.trades:
+            core_stats_dict["trades"].append(trade.serializable_dict())
+
+        if isinstance(area.strategy, SmartMeterStrategy):
+            core_stats_dict["smart_meter_profile_kWh"] = (
+                area.strategy.state.get_energy_at_market_slot(self.current_market_time_slot))
```
fievelk, line 378: I see that this method (`_populate_core_stats_and_sim_state()`) is mostly calling `area.strategy.state`. Do you think we could do something like `core_stats_dict = area.strategy.state.get_state()`? This might save us all the `if isinstance()` statements. What do you think?

spyrostz: I agree, and I will go even further to name the state method as `get_current_state` in order to get the very latest state in order to remove the dependency of the `self.current_market_time_slot`, which IIRC is already included in some state classes (I distinctly recall it being used in the StorageState class). Give me some time, I'll prepare something, and thanks for the comment! At the first stage I will change only the coefficient-related code, but I will try to also refactor the market code at the same time if it does not take ages to do so.

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -523,9 +543,55 @@ def _get_current_market_time_slot(self, slot_number: int) -> DateTime:
...
+            self.area.cycle_coefficients_trading(self.progress_info.current_slot_time)
+
+            global_objects.profiles_handler.update_time_and_buffer_profiles(
+                self._get_current_market_time_slot(slot_no))
+
+            self._results.update_and_send_results(
+                self.current_state, self.progress_info, self.area, self._status.status)
+            self._external_events.update(self.area)
+
+            gc.collect()
```
fievelk, line 567: Could you explain what objects are in memory and unreachable at this point? I've never used `gc.collect()` explicitly before.

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -523,9 +543,55 @@ def _get_current_market_time_slot(self, slot_number: int) -> DateTime:
...
+            self._status.handle_incremental_mode()
+
+        self._simulation_finish_actions(slot_count)
+
     def _execute_simulation(self, slot_resume: int, tick_resume: int,
                             console: NonBlockingConsole = None) -> None:
+        if ConstSettings.MASettings.MARKET_TYPE == SpotMarketTypeEnum.COEFFICIENTS:
+            self._execute_coefficient_simulation(slot_resume, tick_resume, console)
+        else:
+            self._execute_market_simulation(slot_resume, tick_resume, console)
```
fievelk, line 591: Nit: I would generally go with two separate classes (e.g. `CoefficientSimulation` and ` MarketSimulation`) rather than implementing two mutually-excluding functions in the same class. Is there a specific reason to go this way instead, or was it a matter of time?

src/gsy_e/gsy_e_core/simulation.py
spyrostz: This is admittedly not as useful in CPython, but more useful in Pypy due to its aggressive caching behavior. When we encountered issues with severe memory leaks in the gsy-e in the past, we observed that most of the leaks appeared in Pypy and not in CPython. Memory profiling revealed that Pypy in some cases did not delete even objects that were reference counted to zero, in order to cache them and be able to fetch them from its cache if needed later. At some point in time it indeed deallocated them, but not until it had exhausted the memory of the system. Therefore the simulation process was allocating the full amount of memory for prolonged periods of time, which affected the runtime performance of the process. In order to force Pypy to deallocate the memory, we were obliged to force the garbage collector to run on regular intervals, and the ideal place for this operation was on every market cycle. 

We observed also some minor improvements on the CPython version but I suspected that this was due to circular reference detection and not due to the actual garbage collection process, therefore I cannot tell with certainty. Also keep in mind that we run these tests with CPython and Pypy 3.5 version, therefore things might have changed meanwhile. Despite that, I would feel much safer if we left the GC call in place, because worst case scenario we will just trigger an extra collection cycle, which should have minimal effect in the runtime of the application. 

To the feature to be implemented though, I suspect that the memory management will be far far easier, because I expect to have much less cyclical references (due to lack of MarketAgent objects which held references to the Area objects). It remains to be seen.

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -350,3 +355,66 @@ def _update_offer_bid_trade(self) -> None:
...
                 k: area_result[k] for k in ("offers", "bids", "trades")}
+
+
+class CoefficientEndpointBuffer(SimulationEndpointBuffer):
+    """Calculate the endpoint results for the Coefficient based market."""
+
+    def _calculate_and_update_last_market_time_slot(self, area):
+        pass
+
+    def _populate_core_stats_and_sim_state(self, area: "Area"):
```
fievelk, line 366: Here the type is `Area`, but AFAIU this class will only be used for market simulations (not for coefficient simulations). Is this correct?

src/gsy_e/models/area/__init__.py
```diff
@@ -99,63 +87,70 @@ def insert(self, index, item):
         super().insert(index, item)
 
 
-# pylint: disable=too-many-instance-attributes
-# pylint: disable=too-many-public-methods
-class Area:
-    """Generic class to define both market areas and devices.
-
-    Important: this class should not be used in setup files. Please use Market or Asset instead.
+class AreaBase:
```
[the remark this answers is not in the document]
fievelk: Perfect!

fievelk (commented):
I've left some comments. Thanks for working on this!

hannesdiedrich (dismissed):
LGTM

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
spyrostz: Agreed, and fixed some other type mismatches in endpoint buffer along the way.

src/gsy_e/models/state.py
```diff
@@ -381,6 +385,12 @@ def _calculate_unsettled_energy_kWh(
                              self.get_available_energy_kWh(time_slot))
         return traded_energy_kWh - measured_energy_kWh
 
+    def get_results_dict(self, current_time_slot: DateTime) -> dict:
+        return {
+            "pv_production_kWh": self.get_energy_production_forecast_kWh(current_time_slot),
+            "available_energy_kWh": self.get_available_energy_kWh(current_time_slot)
+        }
```
fievelk, line 392: This looks quite similar to the `get_state()` method already implemented in the parent class (e.g. `ProductionState`). Do you think there could be a way to re-use that? The reason why I'm focusing on this is that I'd like to use `get_state()` as the only source of truth for all the "global recap" information about the `State` class. So that if we change `get_state()` this will also be reflected automatically in the `get_results_dict()` method without any additional effort.

fievelk (commented):
I've left one comment. It looks like tests are currently failing. I'll review again once this is addressed.

src/gsy_e/models/state.py
spyrostz: I don't think this is a good idea conceptually. The state of an object is in general way different from the parts of it that are exposed to the user (which is actually the simulation results). A prime example is the StorageState, where the internal state includes many parameters, but the results demand only the SOC. 
The alternative would be to just pass the state and do the filtering on the simulation results parsing, but this is a reactive approach that I do not like. IMO it is much better for the class itself to have different functions that report parts of the internal state for the different needed usecases.

spyrostz:
> I've left one comment. It looks like tests are currently failing. I'll review again once this is addressed.

The CI is failing because it cannot recognize the framework branch env variable correctly. Let me debug this further, however meanwhile you can ignore this since the tests pass locally.

src/gsy_e/models/state.py
fievelk: Ok, if the two states should be different then I agree 👍

fievelk (approved):

Hassan754 (approved):

PR #1420, opened 2022-04-19
TESTS-147: I want to work with a docker image with a mounted volume to allow code editing on the fly

ERafaelMartinez:

The purpose of this PR is to implement a workflow that allows us to use a dynamic `gsy-e` image at the moment of the execution of the integration tests for `gsy-e-sdk` and `gsy-myco-sdk` repositories. 
Additionally, pre-commit and pylintrc files as well as a pr-template-message are included.

## Proposed changes

- New docker-compose file to manage services needed for integration tests execution.
- Slight refactor to the Dockerfile and a bash script which depended on it.
- More changes proposed on GBIT-repo's PR.

**INTEGRATION_TESTS_BRANCH**=TESTS-147


docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+      - REDIS_URL=${REDIS_URL:-redis://redis.container:6379/}
+    volumes:
+      - ${GSY_E_TARGET_REPO:-integration_tests}:/gsy-e
+
+  redis:
+    image: redis:6.2.5
+    restart: always
+    container_name: gsy-e-tests-redis
+    ports:
+      - "${PORT:-3679}:${PORT:-6379}"
```
spyrostz, line 19: As a frequent user of redis-cli I would very much prefer to maintain the external port to 6379 so that it works for debugging even with this setup. Most of us are aware anyways that we need to stop any system-installed Redis / Postgres (does not apply for this PR of course) before running a docker-compose command that includes Redis.

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    hostname: ${HOST:-redis.container}
+    volumes:
+      - redis_volume:/data
+
+volumes:
+  integration_tests:
+    driver: local
+  redis_volume:
+    driver: local
```
spyrostz, line 28: How come we need these volumes here ? I can understand for integration_tests (because I assume that this compose file will be used when running the integration tests, therefore it needs to read them from somewhere) however I do not think that a Redis volume will be something of use here.

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    build:
+      dockerfile: Dockerfile
+      context: ${GSY_E_TARGET_REPO:-./}
+    container_name: ${CONTAINER_NAME:-gsy-e-tests}
+    entrypoint: ["gsy-e"]
+    command: ${COMMAND:-}
+    environment:
+      - REDIS_URL=${REDIS_URL:-redis://redis.container:6379/}
+    volumes:
+      - ${GSY_E_TARGET_REPO:-integration_tests}:/gsy-e
```
fievelk, line 12: Why is the default volume `integration_tests`? Shouldn't it be the complete gsy-e folder? That's what `GSY_E_TARGET_REPO` would target.

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+
+  redis:
+    image: redis:6.2.5
+    restart: always
+    container_name: gsy-e-tests-redis
+    ports:
+      - "${PORT:-3679}:${PORT:-6379}"
+    hostname: ${HOST:-redis.container}
```
fievelk, line 20: I'm afraid that `PORT` and `HOST` names might be too broad and they could clash with some other env variable.

docker-compose.yml
fievelk, line 28: Why is `volumes` declared at top-level? And could you explain why we need to specify the driver (and why `local` specifically)?

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
+services:
+  integration_tests:
```
fievelk, line 2: This is not just building the integration tests part, right? Could it be used to build the entire gsy-e?

docker-compose.yml
```diff
@@ -0,0 +1,32 @@
...
+    build:
+      dockerfile: Dockerfile
+      context: ${GSY_E_TARGET_REPO:-./}
```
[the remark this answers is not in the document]
ERafaelMartinez: `GSY_E_TARGET_REPO` is set either on gsy-e-sdk or gsy-myco-sdk at the moment of the integration tests execution. Nevertheless I understand your point and agree to it.

I'm working on something result of the discussion on GBIT with which I think I can get rid of this you point out.
I'll push the changes when it is ready.

docker-compose.yml
ERafaelMartinez: This is actually a typo. It should be set to "6379". Thanks for pointing it out!

docker-compose.yml
ERafaelMartinez: I thought of naming it because I saw that some hashed volumes were being left behind and they corresponded with the temporal volume the redis service creates. However these are only left behind if the normal execution of the integration tests is interrupted. At least I tend to run and stop it without necessarily waiting till everything is finished and therefore cleaned up.
By naming it, the redis service would always use the same volume, and on interruption, no hashed unused volumes would be left behind, as redis_volume would persist and be reused.

docker-compose.yml
ERafaelMartinez: The default volume is `integration_tests` for it to be possible to execute the tests without necessarily specifying GSY_E_TARGET_REPO variable. The data stored on integration_tests volume would correspond to the same data which is copied during the image build.

Therefore, if GSY_E_TARGET_REPO is not specified, the code under which the tests run is the image's default code, (most likely the master remote branch depending on how the image was built).

As you mention, if GSY_E_TARGET_REPO is specified, then the code under gsy-e folder of the target repo is the one that will be used for the tests execution.

docker-compose.yml
ERafaelMartinez: > I'm afraid that PORT and HOST names might be too broad and they could clash with some other env variable. Or am I wrong?

I agree with your concern. This would be fixed with more variable verbosity. However, as you then mentioned, these will hardly change, and we might want to set them explicitly in sake of simplicity.
I decided to add the flexibility just because the steps which build the image seemed to have that flexibility as well.

I would also go for setting those explicitly.

docker-compose.yml
ERafaelMartinez: > Why is volumes declared at top-level?

This is in order to create the docker volume which is used for a given service (apart from the mount type).


> And could you explain why we need to specify the driver (and why local specifically)?

AFAIU, think this is not strictly necessary. The local driver is the default value, so this is how would the default configuration should look like, if made explicit.

I originally did it like this in order to be "mentally" aware of how the values were set. 

This can be safely rewritten as follows, which I will update in a future commit.
```suggestion
volumes:
  integration_tests:
  redis_volume:
```

docker-compose.yml
ERafaelMartinez: This is actually building a working gsy-e image, which in particular will be used to test the sdks. That's the reason it is named as integration_tests.
However, the same exact image could be used to run the exchange for a different purpose.

ERafaelMartinez: So I would say it does serve to build the whole gsy-e.

However, as was mentioned on other comments, and after working on the comments left on GBIT I realized that adding this compose file on gsy-e is not really necessary.

ERafaelMartinez:
@fievelk @spyrostz 
Thank you both for your reviews.
As a result of addressing the comments on the GBIT PR, and also the ones left in this same PR, I noticed that these changes are not strictly necessary, and therefore decided to stop this PR.

I will not close it yet, as I want to keep track of the comments left here regarding the compose file, which I will relocate somewhere else.

docker-compose.yml
fievelk: I would rename this service `gsy-e`, since it can be used for multiple purposes (not just integration tests).

docker-compose.yml
fievelk: Ah sorry, I thought that `integratino_tests` here was the folder name, but it's actually the name of the volume. In that case, as I've mentioned in another comment, I'd just rename that volume to be `gsy-e` for clarity.

docker-compose.yml
fievelk: What's the progress on this? I think that the options are two AFAIU: either we rename these using more verbose names, or we just hardcode their values here without using variables.

docker-compose.yml
fievelk: > This can be safely rewritten as follows, which I will update in a future commit.

What's the current progress on this?

docker-compose.yml
ERafaelMartinez: I adapted to this comments on the compose file on GBIT's PR:
https://github.com/gridsingularity/gsy-backend-integration-tests/pull/33 

I opted for hardcoding the values.

docker-compose.yml
ERafaelMartinez: Adapted on GBIT's compose files.

docker-compose.yml
ERafaelMartinez: I saw your comment as well on GBIT, which has been replied to.

ERafaelMartinez:
Comments here, have been either replied to, or addressed on [GBIT's PR](https://github.com/gridsingularity/gsy-backend-integration-tests/pull/33).
Will close this PR now. Thank you!

PR #1426, opened 2022-04-28
GSYE-269: Set an empty list as a default children for new created area

cruky:

Sanitize the `Area`'s `children` parameter (e.g. like in `SimulationEndpointBuffer`) and set the default value always to an empty list.

## Proposed changes

Set an empty list as a default children for new created area in `area_from_dict`.

### DIFF
--- a/src/gsy_e/gsy_e_core/area_serializer.py
+++ b/src/gsy_e/gsy_e_core/area_serializer.py
@@ -22,15 +22,6 @@ from gsy_framework.utils import convert_pendulum_to_str_in_dict, key_in_dict_and
 from gsy_e.models.area import Area # NOQA
 from gsy_e.models.strategy import BaseStrategy
 from gsy_e.models.area.throughput_parameters import ThroughputParameters
-# pylint: disable=unused-import
-from gsy_e.models.strategy.market_maker_strategy import MarketMakerStrategy  # NOQA
-from gsy_e.models.strategy.commercial_producer import CommercialStrategy  # NOQA
-from gsy_e.models.strategy.pv import PVStrategy  # NOQA
-from gsy_e.models.strategy.storage import StorageStrategy  # NOQA
-from gsy_e.models.strategy.load_hours import LoadHoursStrategy # NOQA
-from gsy_e.models.strategy.predefined_load import DefinedLoadStrategy # NOQA
-from gsy_e.models.strategy.predefined_pv import PVPredefinedStrategy, PVUserProfileStrategy  # NOQA
-from gsy_e.models.strategy.finite_power_plant import FinitePowerPlant # NOQA
 
 from gsy_e.models.leaves import Leaf # NOQA
 from gsy_e.models.leaves import *  # NOQA  pylint: disable=wildcard-import, unused-wildcard-import
--- a/src/gsy_e/gsy_e_core/area_serializer.py
+++ b/src/gsy_e/gsy_e_core/area_serializer.py
@@ -17,10 +17,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import json
 
+from gsy_framework.utils import convert_pendulum_to_str_in_dict, key_in_dict_and_not_none
+
 from gsy_e.models.area import Area # NOQA
 from gsy_e.models.strategy import BaseStrategy
 from gsy_e.models.area.throughput_parameters import ThroughputParameters
-
+# pylint: disable=unused-import
 from gsy_e.models.strategy.market_maker_strategy import MarketMakerStrategy  # NOQA
 from gsy_e.models.strategy.commercial_producer import CommercialStrategy  # NOQA
 from gsy_e.models.strategy.pv import PVStrategy  # NOQA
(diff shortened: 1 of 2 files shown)


src/gsy_e/gsy_e_core/area_serializer.py
```diff
@@ -17,10 +17,12 @@
 """
 import json
 
+from gsy_framework.utils import convert_pendulum_to_str_in_dict, key_in_dict_and_not_none
+
 from gsy_e.models.area import Area # NOQA
 from gsy_e.models.strategy import BaseStrategy
 from gsy_e.models.area.throughput_parameters import ThroughputParameters
-
+# pylint: disable=unused-import
```
fievelk, line 25: Two notes:
1. could you please clarify why we need all these unused imports below? I'm not exactly sure I understand the reasons for all these `# NOQA` here.
2. we usually put `# pylint: disable` comments before all imports (when they apply to the entire module) or on the specific line of the warning (if it's just that line).

fievelk (dismissed):
One question. LGTM in general.

src/gsy_e/gsy_e_core/area_serializer.py
spyrostz: WRT #1, I got curious and thought to try it out. It looks like the NOQA imports are needed for running the gsy-e using a json setup file: 

```
$ gsy-e -l INFO run --setup json_file -t 15s -s 15m  --seed 0 

.......


18:00:24.919 INFO     (  22) root                          : Running without Kafka connection for simulation results.
  File "/Users/spyrostzavikas/dev/gsy-e/src/gsy_e/gsy_e_core/area_serializer.py", line 110, in area_from_dict
    return _leaf_from_dict(description, config)  # Area is a Leaf
  File "/Users/spyrostzavikas/dev/gsy-e/src/gsy_e/gsy_e_core/area_serializer.py", line 97, in _leaf_from_dict
    raise ValueError("Unknown leaf type '%s'" % leaf_type)
ValueError: Unknown leaf type '<class 'NoneType'>'

```

Therefore the deserialization needs these imports. However, from my tests all these imports can be deleted because they essentially duplicate this:
`from gsy_e.models.leaves import *  # NOQA`

which follows some lines below and also duplicates the same strategy imports. Therefore the other strategy imports can be removed (also the Area import as far as I see)

src/gsy_e/gsy_e_core/area_serializer.py
```diff
@@ -31,41 +33,46 @@
...
-            return self._encode_area(obj)
-        elif isinstance(obj, Leaf):
-            return self._encode_leaf(obj)
-        elif isinstance(obj, BaseStrategy):
-            return self._encode_subobject(obj)
-
-    def _encode_area(self, area):
+    """Json encoder for Area object."""
+
+    def default(self, o):
```
spyrostz, line 42: Isn't pylint complaining about the one-char argument ?

cruky: It was complaining about "Parameter 'o' has been renamed to 'obj' in overridden 'AreaEncoder.default' method". That's why I changed it.

Hassan754 (approved):
LGTM

src/gsy_e/gsy_e_core/area_serializer.py
```diff
@@ -128,7 +128,7 @@ def optional(attr):
         if key_in_dict_and_not_none(description, 'children'):
             children = [area_from_dict(child, config) for child in description['children']]
         else:
-            children = None
+            children = []
```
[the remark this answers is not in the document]
Hassan754: Actually it is allowed to be an empty list, in [this](https://github.com/gridsingularity/gsy-e/blob/29ce17dc01792ef41ed8b70322fda62d2364ed01/src/gsy_e/models/area/__init__.py#L111) we pass the children to the `AreaChildrenList` and this will basically apply such operation: `list([])`which will return [].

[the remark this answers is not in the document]
hannesdiedrich: Indeed, the empty list will be set to the area.children. However, in the place you posted, we are checking `if children is not None` which will never be true with this change here. 
So, IMO this needs to be changed, and also every occurrence where we check for the children of an area.
`if not area.children:` also works with `children=[]` bit if there is a place were we check like this: `if area.children is None` then it does not. 
If all agree, that we do this change from children=None to children=[] for market areas that do not have children, then please adapt and check the above.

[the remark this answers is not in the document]
fievelk: I'd agree with Hannes, since this apparently would generate some issues in other parts of the code.

In general, I think `if not foo.bar` should be preferred over `if foo.bar is not None`. The exception is e.g. when we have a very specific reason to distinguish between false values (like `False`, `0`, `""`, `[]`) and `None`.
In our case, however, I think that these 2 are equivalent: 
- `area.children = None`
- `area.children = []`

If this is true, checking for `if not area.children` seems to be the best way to proceed.

fievelk (approved):

hannesdiedrich:
@cruky Is this PR still valid?

PR #1433, opened 2022-05-04
GSYE-188: Added a dictionary to map aggregated simulation result names to appropriate file names.

mrsaemir:
- In order to match the file names listed in the [Result Download](https://gridsingularity.github.io/gsy-e/results-download/) doc, and as requested by [GSYE-188](https://gridsingularity.atlassian.net/browse/GSYE-188)

## Proposed changes
- Added a dictionary to map aggregated simulation result names (keys) to appropriate file names.

### DIFF
--- a/src/gsy_e/gsy_e_core/export.py
+++ b/src/gsy_e/gsy_e_core/export.py
@@ -66,7 +66,7 @@ alternative_pricing_subdirs = {
     3: "net_metering_pricing"
 }
 
-json_data_area_fields_mapping = {
+results_field_to_json_filename_mapping = {
     "area_throughput": "area_throughput",
     "assets_info": "assets_info",
     "bills": "bills",
@@ -134,7 +134,7 @@ class ExportAndPlot:
         with open(settings_file, "w", encoding="utf-8") as outfile:
             json.dump(constsettings_to_dict(), outfile, indent=2)
         for in_key, value in self.endpoint_buffer.generate_json_report().items():
-            out_key = json_data_area_fields_mapping[in_key]
+            out_key = results_field_to_json_filename_mapping[in_key]
             json_file = os.path.join(json_dir, out_key + ".json")
             with open(json_file, "w", encoding="utf-8") as outfile:
                 json.dump(value, outfile, indent=2)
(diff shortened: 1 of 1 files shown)


src/gsy_e/gsy_e_core/export.py
```diff
@@ -66,6 +66,26 @@
     3: "net_metering_pricing"
 }
 
+json_data_area_fields_mapping = {
```
hannesdiedrich, line 69: Minor, but I propose to rename this into:
`json_results_file_name_mapping`

hannesdiedrich (commented):
One minor comment. LGTM in general.
Can you please take the opportunity to adapt the integration tests step that we have [here](https://github.com/gridsingularity/gsy-backend-integration-tests/blob/6f5b5cfcccf7e16708b1abb2ada6523ca3ca98cc/gsy_e_tests/integration_tests/steps/integration_tests.py#L480) that checks if the backend returns all expected file names?
Since we added this test, there were added some more. Can you please run a simulation locally and see which json files are expected and adapt the test step accordingly? Thanks a lot!

src/gsy_e/gsy_e_core/export.py
mrsaemir: Actually, I used the exact name as gsy-web, so people can find about this one if they try searching for `json_data_area_fields_mapping` in GitHub (This is what I always do when I want to know which repo is using which other repo).
As, I didn't write `json_data_area_fields_mapping` in gsy-web, I preferred to not change it there, too!
But, still, if you think the name is misleading people, I can change it in gsy-e, maybe?

hannesdiedrich: Sorry, I did not see it in the gsy-web PR.
Actually, in gsy-web it is missleading, too. There I would rename it to e.g. `results_field_to_json_filename_mapping`. 
I do not have a strong opinion on exactly these names, but don't you agree that these (or a version of them) would be more explanatory? I agree, they should be the same in both repos. So please change them on both places, thanks a lot!

mrsaemir: You're right. both names are misleading. I will change the name and ask for another round of review.

hannesdiedrich (approved):
Perfect. LGTM

Hassan754 (approved):

PR #1435, opened 2022-05-10
Feature/gsye 255

spyrostz:

Strategies needed to be adapted in order to support SCM. The feature requirements defined that the new strategies should not place bids or offers, nor have any price-related characteristics attached to them.

## Proposed changes

- Refactored the strategies in a way that the energy-related parameters of these strategies are in separate classes in order to be reused
- Created the new SCM strategies using the energy parameters classes of the existing strategies
- Changed deserialization of JSON dicts to Area tree, and if the coefficients trading is enabled, then the SCM strategies should be created. If not, the existing strategies should be used instead. 

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-255


src/gsy_e/models/strategy/load_hours.py
```diff
@@ -166,38 +246,39 @@ def event_activate(self, **kwargs):
...
 
-    def add_entry_in_hrs_per_day(self, overwrite=False):
-        """Add the current day (in simulation) with the mapped hrs_per_day."""
-        current_day = self._get_day_of_timestamp(self.area.spot_market.time_slot)
-        if current_day not in self.hrs_per_day or overwrite:
-            self.hrs_per_day[current_day] = self._initial_hrs_per_day
-
     def update_state(self):
         """Update the state of the strategy."""
-        self._post_first_bid()
```
hannesdiedrich, line 200: ❤️

src/gsy_e/models/strategy/storage.py
```diff
@@ -51,11 +51,18 @@
...
+        return {
+            "initial_soc": self.state.initial_soc,
+            "min_allowed_soc": self.state.min_allowed_soc_ratio * 100.0,
+            "battery_capacity_kWh": self.state.capacity,
+            "max_abs_battery_power_kW": self.state.max_abs_battery_power_kW,
+            "cap_price_strategy": self.cap_price_strategy,
+            "initial_energy_origin": self.state._initial_energy_origin,
+            "balancing_energy_ratio": self.balancing_energy_ratio,
+            **self.bid_update.serialize(),
+            **self.offer_update.serialize(),
```
hannesdiedrich, line 64: interesting! With the [old way](https://github.com/gridsingularity/gsy-e/pull/1435/commits/4a3425118d3339ad3572aae03358d271b7865dfb#diff-5d95a36b16632bceaf86321630cbeb038104519dcd21aa330498a47d41914aabL89) of getting the parameters of a strategy class, these would not be part of  the parameters (this is what you meant by `but also dispatch the serialization to the offer/bid updater objects`, right?)

hannesdiedrich: Apparently this never was a problem. I wonder why we need the serialisation of the class parameters at all then. Can you please explain this? Thanks!

src/gsy_e/models/strategy/scm/pv.py
```diff
@@ -0,0 +1,54 @@
...
+from gsy_e.models.strategy.predefined_pv import (
+    PVPredefinedEnergyParameters, PVUserProfileEnergyParameters)
+from gsy_e.models.strategy.pv import PVEnergyParameters
+from gsy_e.models.strategy.scm import SCMStrategy
+
+
+class SCMPVStrategy(SCMStrategy):
+    """PV SCM strategy with gaussian power production."""
+
+    def __init__(self, panel_count: int = 1, capacity_kW: float = None):
```
hannesdiedrich, line 18: IRRC, panel_count will no longer be advertised. The produced energy of a PV should only be scaled with the `capacity_kW`. The deprecation of `panel_count` will probably a big change and does not need to be done soon, but IMO we could omit it here already. Or is this a requirement to have this parameter also in the SCM PV? What do you think?

src/gsy_e/models/strategy/storage.py
spyrostz: This actually was a problem in the first place :D we just never used the serialized price values of the strategies that's why we never observed it. In the former implementation, no matter the values of `update_interval`, `initial_buying_rate`, `initial_selling_rate` ... serialization code never fetched these values correctly and always reported None for these. Admittedly we never used it, however the problem was there in the first place.

src/gsy_e/models/strategy/scm/pv.py
spyrostz: I deliberately left it here in case the business requirements change, however you are right, YAGNI. I will remove the panel_count from the constructor arguments, and hardcode the value in the energy_params to 1, thanks!

src/gsy_e/models/strategy/scm/load.py
```diff
@@ -0,0 +1,58 @@
...
+
+from gsy_e.models.strategy.load_hours import LoadHoursEnergyParameters
+from gsy_e.models.strategy.predefined_load import DefinedLoadEnergyParameters
+from gsy_e.models.strategy.scm import SCMStrategy
+
+
+class SCMLoadHoursStrategy(SCMStrategy):
+    """Load SCM strategy with constant power production."""
+    def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None):
+        self._energy_params = LoadHoursEnergyParameters(
```
hannesdiedrich, line 16: Also here. How about simplifying the interface to the loadhours strategy too (of course if the stakeholders agree, but I do not see specific requirements for the loadhours strategy in the design). We could only accept an `hrs_of_day` List (which is more explicit and easier to understand also for the user). Of course, this would mean some re-implementation, but also some deprecation of class methods of the `LoadHoursEnergyParameters`, but might be worth it, what do you think?

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -50,12 +50,123 @@
 BalancingRatio = namedtuple("BalancingRatio", ("demand", "supply"))
 
 
+class LoadHoursEnergyParameters:
+    """Energy related parameters of the Load Hours Strategy."""
+    def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None, area=None):
```
Hassan754, line 55: It seems you are expecting an area object in the constructor here, is there a particular reason for that? I did not find any initializations of this class apart from within the `LoadHoursStrategy`, and we're not passing an area object there

src/gsy_e/models/strategy/scm/load.py
spyrostz: Also here I was wondering if I should expose the LoadHours strategy at all in the SCM, and I decided to expose it as it. You are right though, the hrs_per_day is a confusing parameter and it was rarely used anyways in the maps / d3a UIs. I'll disable it, thanks!

src/gsy_e/gsy_e_core/area_serializer.py
```diff
@@ -41,12 +42,16 @@
 class AreaEncoder(json.JSONEncoder):
     """Convert the Area class hierarchy to json dict."""
     def default(self, o):
-        if isinstance(o, AreaBase):
+        # Leaf classes are Areas too, therefore the Area/AreaBase classes need to be handled
+        # separately.
+        if type(o) in [Area, AreaBase]:
```
hannesdiedrich, line 47: Just had a look into `isinstance` and I think you can also write:
```
isinstance(o, (AreaBase, Area)):
```
to achieve the same.

src/gsy_e/gsy_e_core/area_serializer.py
```diff
@@ -41,12 +42,16 @@
...
             return self._encode_area(o)
         if isinstance(o, Leaf):
             return self._encode_leaf(o)
         if isinstance(o, BaseStrategy):
             return self._encode_subobject(o)
+        if isinstance(o, Duration):
+            return None
```
hannesdiedrich, line 54: Can you please explain this change? Don't we want to use `.format()` or something similar?

src/gsy_e/models/strategy/predefined_pv.py
```diff
@@ -90,7 +90,7 @@ def set_produced_energy_forecast_in_state(
 
     def reconfigure(self, **kwargs):
         """Reconfigure the device properties at runtime using the provided arguments."""
-        if key_in_dict_and_not_none(kwargs, "cloud_coverage"):
+        if "cloud_coverage" in kwargs:
```
hannesdiedrich, line 93: We can only do this, because we validate the kwargs before (in the json schema validator) and check if `cloud_coverage ` is not None, right?

hannesdiedrich (commented):
Left some comments, will review again after addressed.
I like the move of the energy related functionality of the strategies into separate classes even without the SCM strategies. Thanks!

src/gsy_e/models/strategy/load_hours.py
spyrostz: Indeed this member is populated during the activate method, therefore no need for the constructor to also include it. Removing it from here, thanks!

src/gsy_e/gsy_e_core/area_serializer.py
spyrostz: It is unfortunately not the same, `isinstance(Leaf, AreaBase)` will return True which is not what we want here. Instead, we want the Leaf classes to be handled by a different method from the pure Area / AreaBase classes.

src/gsy_e/gsy_e_core/area_serializer.py
spyrostz: I explicitly did it this way in order to match the previous behavior. Up until now, all objects that were not able to be serialized were omitted. I changed this behavior to raise an exception in these cases, and the exception was not raised for any other objects apart from a duration object. Therefore to mimic the old behavior I omitted it. If you want though, I can try to serialize this too.
WRT the serialization itself, the problem is that format() cannot be used on Duration objects, therefore I will do a Duration.seconds instead.

src/gsy_e/gsy_e_core/area_serializer.py
hannesdiedrich: Ahh. ok. Thanks!

src/gsy_e/gsy_e_core/area_serializer.py
hannesdiedrich: In this case, I agree, leave the current behaviour. Thanks!

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -50,6 +50,100 @@
...
+        # Energy consumed per market slot
+        self.energy_per_slot_Wh = None
+
+        # Maps each simulation day to the number of active hours in that day
+        self.hrs_per_day: Dict[int, int] = {}
+
+        # List of active hours of day (values range from 0 to 23)
+        self.hrs_of_day: Optional[List[int]] = None
+        self._initial_hrs_per_day: Optional[int] = None
+        self.assign_hours_of_per_day(hrs_of_day, hrs_per_day)
```
Hassan754: Nit, do you think we can call `.reset()` here?

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -491,37 +553,18 @@ def _calculate_active_markets(self):
...
-
-        if len(hrs_of_day) < hrs_per_day:
-            raise ValueError("Length of list 'hrs_of_day' must be greater equal 'hrs_per_day'")
-
     def _update_energy_requirement_future_markets(self):
         if not GlobalConfig.FUTURE_MARKET_DURATION_HOURS:
             return
         for time_slot in self.area.future_market_time_slots:
-            if self._allowed_operating_hours(time_slot):
+            if self._energy_params._allowed_operating_hours(time_slot):
```
Hassan754: Nit, It seems we're reaching out to a private member of the `_energy_params` object, do you think making it public makes sense?

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -190,6 +189,10 @@ def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None,
         self._cycled_market = set()
         self._simulation_start_timestamp = None
 
+    @classmethod
+    def _create_energy_parameters(cls, *args):
```
Hassan754: Is there a reason for encapsulating it into a class method? AFAIS we can initialize the `LoadHoursEnergyParameters` on the spot in the constructor

src/gsy_e/models/strategy/predefined_load.py
```diff
@@ -25,9 +25,56 @@
...
+        :return: None
+        """
+        self._read_or_rotate_profiles()
+
+    def reconfigure(self, **kwargs):
+        if key_in_dict_and_not_none(kwargs, "daily_load_profile"):
+            self._load_profile_input = kwargs["daily_load_profile"]
+            self._read_or_rotate_profiles(reconfigure=True)
+
+    def _read_or_rotate_profiles(self, reconfigure=False):
```
Hassan754: Nit, this seems to be used as an interface rather than an internal method

src/gsy_e/gsy_e_core/area_serializer.py
```diff
@@ -31,41 +36,48 @@
 from gsy_e.models.strategy.finite_power_plant import FinitePowerPlant # NOQA
 
 from gsy_e.models.leaves import Leaf # NOQA
-from gsy_e.models.leaves import *  # NOQA
-from gsy_framework.utils import convert_pendulum_to_str_in_dict, key_in_dict_and_not_none
+from gsy_e.models.leaves import *  # NOQA  # pylint: disable=wildcard-import
```
Hassan754, line 39: Nit, this pylint warning is defined twice in this module.
Regardless, It would be great if we can get rid of this wildcard import here, or at least encapsulate it
```import gsy_e.models.leaves as leaves```

File: src/gsy_e/models/strategy/predefined_load.py [hunk 14]
```diff
@@ -25,18 +25,85 @@
...
+    def event_activate_energy(self, area):
+        """
+        Runs on activate event.
+        :return: None
+        """
+        self.read_or_rotate_profiles()
+        super().event_activate_energy(area)
+
+    def reset(self, current_day: int, **kwargs) -> None:
+        if key_in_dict_and_not_none(kwargs, "daily_load_profile"):
```
Hassan754, line 61: Nit
```suggestion
        if kwargs.get("daily_load_profile") is not None:
```

src/gsy_e/models/strategy/predefined_load.py
```diff
@@ -25,18 +25,85 @@
...
+            self._load_profile_input = kwargs["daily_load_profile"]
+            self.read_or_rotate_profiles(reconfigure=True)
+
+    def read_or_rotate_profiles(self, reconfigure=False):
+        """Read power profiles or rotate them, from the DB or from JSON dicts."""
+        input_profile = (self._load_profile_input
+                         if reconfigure or not self._load_profile_W
+                         else self._load_profile_W)
```
Hassan754, line 69: Nit, this could be a bit easier to understand
```suggestion
        input_profile = (self._load_profile_W
                         if reconfigure and _load_profile_W
                         else self. _load_profile_input)
```

src/gsy_e/models/strategy/load_hours.py
spyrostz: Hmm, I think that you are reviewing it per-commit that's why you do not see it yet. I have converted this to public on a later commit.

src/gsy_e/models/strategy/load_hours.py
spyrostz: Same as the previous comment, this method is deleted in a followup commit.

src/gsy_e/models/strategy/predefined_load.py
spyrostz: I have converted the method to public in a followup commit.

src/gsy_e/models/strategy/predefined_load.py
```diff
@@ -68,6 +135,9 @@ def __init__(self, daily_load_profile=None,
             update_interval = \
                 duration(minutes=ConstSettings.GeneralSettings.DEFAULT_UPDATE_INTERVAL)
 
+        if not hasattr(self, "_energy_params"):
+            self._energy_params = DefinedLoadEnergyParameters(
+                daily_load_profile, daily_load_profile_uuid)
```
Hassan754, line 140: Nit, AFAIS this if condition is here (and in the `LoadHoursStrategy` init) because you want to configure different ways of initializing the _energy_params member.
A possible alternative would be to implement a private method `_setup_energy_params` in the base class which is called from the initializer always, and can be overridden in the child classes to define different implementation.

spyrostz: As you might have noticed by the `_create_energy_parameters` I tried this approach and I unfortunately did not like it that much. Main reason was that each energy parameters class requires different arguments to be constructed, therefore I had to rely to *args / **kwargs for this method, which complicated the implementation.
If you really think that it will be better this way let me know and I will give it a try. Then we can talk again about it and keep it or revert it. I am suggesting this because it is a rather important change here so it probably is worth it to see both implementation options.

src/gsy_e/models/strategy/load_hours.py
spyrostz: Might be on a later commit than this one, but you can find it in the method `LoadHoursEnergyParameters.event_activate_energy`

Hassan754 (commented):
Left some nit comments, LGTM once Hannes' comments are addressed

src/gsy_e/models/strategy/predefined_load.py
Hassan754: Yeah unfortunately it would be a bit complicated to homogenize the interface of such method, in this case feel free to leave it as it is. Thanks!

src/gsy_e/models/strategy/predefined_load.py
spyrostz: Fixed, thanks

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -51,109 +51,140 @@
...
+        return {
+            **super().serialize(),
+            "hrs_per_day": self.hrs_per_day,
+        }
 
-    def allowed_operating_hours(self, time):
-        """Check if timeslot inside allowed operating hours."""
-        return time.hour in self.hrs_of_day
+    def _assign_hours_per_day(self, hrs_per_day: int):
+        # be a parameter on the constructor or if we want to deal in percentages
```
hannesdiedrich, line 147: Sorry, I do not get this comment. What do you mean by `if we want to deal in percentages` ?

hannesdiedrich (dismissed):
One last question, LGTM already

Hassan754 (approved):
LGTM

src/gsy_e/models/strategy/load_hours.py
spyrostz: This is probably a trail from moving code. Removing this comment, thanks for catching it!

hannesdiedrich (dismissed):
LGTM

Hassan754 (dismissed):

src/gsy_e/gsy_e_core/area_serializer.py
```diff
@@ -31,41 +36,48 @@
...
 
 
 class AreaEncoder(json.JSONEncoder):
```
fievelk, line 42: Nit: is this class only used for testing? I see it only being used in `area_to_string`, which is only called in tests.

fievelk: If this is the case, maybe it would make sense to move it inside the test module.

src/gsy_e/models/leaves.py
```diff
@@ -61,7 +76,14 @@ class Leaf(Area):
     strategy_type = None
 
     def __init__(self, name, config, uuid=None, **kwargs):
-        if config.external_connection_enabled:
+        if ConstSettings.MASettings.MARKET_TYPE == SpotMarketTypeEnum.COEFFICIENTS.value:
+            # For the SCM only use the SCM-enabled strategies and ignore the rest.
+            try:
+                self.strategy_type = scm_strategy_mapping[self.strategy_type]
+            except KeyError as e:
+                logging.error(f"Strategy {self.strategy_type} not supported in SCM.")
```
fievelk, line 84: Nit: this will raise a warning because of the lazy formatting.
```suggestion
                logging.error("Strategy %s not supported in SCM.", self.strategy_type)
```

src/gsy_e/gsy_e_core/area_serializer.py
spyrostz: I hesitate to move this in the test module because in the past it was used for serializing a running simulation. And I am fairly certain that we might need this functionality soon, therefore instead of hiding it inside the test module it makes more sense to have it together with the deserializer.

src/gsy_e/models/leaves.py
```diff
@@ -74,20 +96,32 @@ def __init__(self, name, config, uuid=None, **kwargs):
...
+        for strategy_base in self.strategy_type.__mro__:
+            if '__init__' in strategy_base.__dict__:
+                allowed_arguments += inspect.getfullargspec(strategy_base).args[1:]
+
+        # Filter out the arguments that are not accepted by any constructor in the strategy
+        # class hierarchy and log them.
+        not_accepted_args = set(kwargs.keys()).difference(allowed_arguments)
+        if not_accepted_args:
+            logging.warning(f"Trying to construct area strategy {name} with not allowed "
+                            f"arguments {not_accepted_args}")
```
fievelk, line 111: Could you please describe why this is needed? It seems a bit "hacky" so I'd find it good to have some documentation that described why this is needed.

(Nit on a second note: these lines are not using `"` and lazy formatting)

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -50,12 +50,154 @@
 BalancingRatio = namedtuple("BalancingRatio", ("demand", "supply"))
 
 
+class LoadHoursEnergyParameters:
```
fievelk, line 53: Nit: I'd go even further and create a module structure. For example:
```
some-parent-folder
├── parameters
│   ├── load_hours_parameters.py
│   └── pv_parameters.py
└── strategies
    ├── full_gsye  # I'm not sure about naming
    │   ├── load_hours_strategy.py
    │   └── pv_strategy.py
    └── scm
        ├── load_hours_strategy.py
        └── pv_strategy.py
```
What do you think? (Just a nit anyway, no need to do anything here)

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -84,23 +226,10 @@ def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None,
...
-        self.avg_power_W = avg_power_W
-
-        # consolidated_cycle is KWh energy consumed for the entire year
-        self.daily_energy_required = None
-        # Energy consumed during the day ideally should not exceed daily_energy_required
-        self.energy_per_slot_Wh = None
-
-        # Maps each simulation day to the number of active hours in that day
-        self.hrs_per_day: Dict[int, int] = {}
+        if not hasattr(self, "_energy_params"):
```
fievelk, line 229: What's the case in which a strategy doesn't have the `self._energy_params` attribute?

src/gsy_e/models/leaves.py
spyrostz: The reason for its existence is the deprecation of the Strategy.parameters class member. In the previous implementation, all the accepted parameters of a strategy were defined by the `parameters` tuple / list that included the input parameters in string format. I personally find this very bad practice, because the constructor arguments need to also be duplicated in the parameters tuple in order to be taken into account to the serializer / deserializer. Therefore I opted to remove the `parameters` tuple from all strategies and try to read the constructor arguments in order to generate the constructor argument list.

Unfortunately, inheritance makes this a more daunting task than expected. To provide the exact example where this fails, is for the ExternalStrategies. These do not define constructors and they rely on the default constructor, which uses as arguments *args and **kwargs and relays them to the base class constructors. Therefore this makes it impossible to find out the constructor arguments just by inspecting this class. The solution that I relied on was MRO: essentially traversing the class hierarchy in order to gather all the constructor arguments from all the subclasses. This works, and the constructor argument list is created, from which we can exclude the arguments that cannot be used for the selected strategy.

I admit it is much less straightforward than explicitly defining the parameters tuple, the advantage of it is that we do not have to maintain these tuples and keep them synced with the constructors. 

WRT the second note, I think I forgot to run pylint in this module and it is a good opportunity to do so now, thanks!

src/gsy_e/models/strategy/load_hours.py
fievelk: I just found this conversation: https://github.com/gridsingularity/gsy-e/pull/1435/files#r870054300

src/gsy_e/models/leaves.py
spyrostz: Fixed, same for the other pylint warnings in leaves.py

src/gsy_e/models/strategy/scm/load.py
```diff
@@ -0,0 +1,63 @@
...
+
+    def market_cycle(self, area: "AreaBase") -> None:
+        """Update the load forecast and measurements for the next/previous market slot."""
+        # self._update_energy_requirement_in_state()
+        # # Provide energy values for the past market slot, to be used in the settlement market
+        # self._set_energy_measurement_of_last_market()
+
+    def get_available_energy_kWh(self, time_slot: DateTime) -> float:
+        """Get the available energy for consumption for the specified time slot."""
+        return self._energy_params.state.get_energy_requirement_Wh(time_slot) / 1000.0
```
fievelk, line 38: I'm a bit confused by this, probably because of the naming. The method is about getting the "available energy", but it's returning the "required energy" of the load. Was this made to simplify the interface? I find it a bit "conceptually" convoluted.

src/gsy_e/models/strategy/load_hours.py
spyrostz: Constructing the strategy itself and not its derived classes. For example, if the DefinedLoadStrategy class (which inherits from LoadHoursStrategy) is constructed, then the self._energy_params will be populated in the constructor of DefinedLoadStrategy class and should not be populated also from the constructor of the LoadHoursStrategy class, therefore in this case the self._energy_params will be populated. When constructing directly the LoadHoursStrategy though, the self._energy_params has not been populated by a derived class and needs to be populated from the constructor of the LoadHoursStrategy class.

src/gsy_e/models/strategy/scm/load.py
fievelk: (This is not currently called anywhere, right?)

spyrostz: Indeed this was an effort to simplify the interface. Ideally I want in the followup story for the SCM clearing algorithm, to harmonize the interface so that I can do from the Area class this: `self.strategy.get_energy_to_buy_kWh(time_slot)`

no matter the strategy class of the area. get_available_energy_kWh is probably a bad name for this though, so I would recommend either the following:
```
def get_energy_to_sell_kWh()
def get_energy_to_buy_kWh()
```

or this:
```
def get_energy_production_kWh()
def get_energy_consumption_kWh()
```

I tend to like the former more, however let me know if you prefer the latter and I will harmonise the interface even now.

src/gsy_e/models/strategy/scm/load.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None):
+        self._energy_params = LoadHoursPerDayEnergyParameters(avg_power_W, hrs_per_day, hrs_of_day)
+        self._simulation_start_timestamp = None
+
+    def activate(self, area: "AreaBase") -> None:
+        """Activate the strategy."""
+        self._energy_params.event_activate_energy(area)
+        self._simulation_start_timestamp = area.now
+
+    def _get_day_of_timestamp(self, time_slot: DateTime) -> int:
```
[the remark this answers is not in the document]
spyrostz: No problem, let me quickly do it for this one.

src/gsy_e/models/strategy/scm/pv.py
```diff
@@ -0,0 +1,61 @@
...
+        return self._energy_params._state
+
+    def activate(self, area: "AreaBase") -> None:
+        """Activate the strategy."""
+        self._energy_params.activate(area.simulation_config)
+        self._energy_params.set_produced_energy_forecast(
+            area.current_market_time_slot, area.simulation_config.slot_length)
+
+    def market_cycle(self, area: "AreaBase") -> None:
+        """Update the load forecast and measurements for the next/previous market slot."""
```
fievelk, line 38: Nit:
```suggestion

    def market_cycle(self, area: "AreaBase") -> None:
        """Update the PV forecast and measurements for the next/previous market slot."""
```

spyrostz: Upsi, fixing

src/gsy_e/models/strategy/scm/load.py
spyrostz: And WRT your second comment you are right, this is more or less preparation for the followup ticket where the actual matching algorithm will be created.

src/gsy_e/models/strategy/load_hours.py
spyrostz: I agree, but I will create a followup PR for this in order to not further complicate this one, thanks!

src/gsy_e/models/strategy/load_hours.py
fievelk: I'm not sure I understand. Isn't this how it normally works?
```py
class LoadHoursStrategy:
    def __init__(self):
        self._energy_params = "parent value"


class DefinedLoadStrategy(LoadHoursStrategy):
    def __init__(self):
        super().__init__()
        self._energy_params = "child value not overwritten"


>>> DefinedLoadStrategy()._energy_params
# "child value not overwritten"
```
It depends on the position of the `super()` call I suppose. Anyway, if you've already checked the possible available options it's all good for me 👍

fievelk:
Cool refactoring! Thanks! I've left some comments, mostly nits.

src/gsy_e/models/leaves.py
spyrostz: > Update: we're actually using this code just once, so it's not strictly necessary. Also, if the SCM strategies don't share the same base class with the "normal" strategies we would need to duplicate this code, and it's probably not worth it.

Thanks for saving me a comment to denote exactly the latter, that the SCM strategies have a different base class and the method needs to be duplicated. 

Also adding a long comment to explain the rationale behind this, thanks!

src/gsy_e/models/strategy/load_hours.py
spyrostz: I was under the impression that some of the EnergyParameters classes might not be able to be constructed due to the dummy parameters that we pass to the superclass, but you are right, I just tested it and overwriting the _energy_params member seems to work as expected! Adding this fix and thanks for the suggestion!

fievelk (dismissed):
Thanks for addressing the comments! LGTM

src/gsy_e/models/strategy/smart_meter.py
```diff
@@ -41,20 +42,103 @@
...
+        """
+        self._read_or_rotate_profiles(reconfigure=reconfigure)
+
+        if not self.profile:
+            raise GSyException(
+                f"Smart Meter {self._area.name} tries to set its required energy forecast without "
+                "a profile.")
+
+        # for market in self.area.all_markets:
+        #     slot_time = market.time_slot
```
hannesdiedrich, line 86: This is not needed any more, right?

src/gsy_e/models/strategy/scm/load.py
```diff
@@ -0,0 +1,63 @@
...
+        """Return the number of days passed from the simulation start date to the time slot."""
+        if self._simulation_start_timestamp is None:
+            return 0
+        return (time_slot - self._simulation_start_timestamp).days
+
+    def market_cycle(self, area: "AreaBase") -> None:
+        """Update the load forecast and measurements for the next/previous market slot."""
+        # self._update_energy_requirement_in_state()
+        # # Provide energy values for the past market slot, to be used in the settlement market
+        # self._set_energy_measurement_of_last_market()
```
hannesdiedrich, line 34: Just only saw this. So you are overwriting the `market_cycle` method and do nothing here, right? Please remove the commented code then. thanks!

hannesdiedrich (commented):
Two last minor comments. LGTM once the tests pass

src/gsy_e/models/strategy/scm/load.py
spyrostz: Fixed

src/gsy_e/models/strategy/smart_meter.py
spyrostz: Not needed anymore indeed, fixed

hannesdiedrich (approved):
LGTM

fievelk (approved):

PR #1450, opened 2022-06-14
Bug/gsye 317

cruky:

Offers and bids for future market are updated according to the constant minimum/maximum rates, and not according to the provided in setup file.

## Proposed changes

Before `future_market_strategy_factory` used an early binding to evaluate rate attributes, which prevented the user from modifying them through setup files. Now the `FutureTemplateStrategiesConstants` are evaluated in a function body.


src/gsy_e/models/strategy/future/strategy.py
```diff
@@ -263,10 +265,10 @@ def event_tick(self, strategy: "BaseStrategy") -> None:
...
 def future_market_strategy_factory(
         asset_type: AssetType,
-        initial_buying_rate: float = FutureTemplateStrategiesConstants.INITIAL_BUYING_RATE,
-        final_buying_rate: float = FutureTemplateStrategiesConstants.FINAL_BUYING_RATE,
-        initial_selling_rate: float = FutureTemplateStrategiesConstants.INITIAL_SELLING_RATE,
-        final_selling_rate: float = FutureTemplateStrategiesConstants.FINAL_SELLING_RATE
+        initial_buying_rate: float = None,
+        final_buying_rate: float = None,
+        initial_selling_rate: float = None,
+        final_selling_rate: float = None
```
Hassan754, line 271: I couldn't see where a user will be able to pass these variables, the calling code [here](https://github.com/gridsingularity/gsy-e/blob/7a286edadd2b336bd4b1381eaa430a916b7725c2/src/gsy_e/models/strategy/load_hours.py#L249), [here](https://github.com/gridsingularity/gsy-e/blob/7a286edadd2b336bd4b1381eaa430a916b7725c2/src/gsy_e/models/strategy/pv.py#L153) and [here](https://github.com/gridsingularity/gsy-e/blob/7a286edadd2b336bd4b1381eaa430a916b7725c2/src/gsy_e/models/strategy/storage.py#L137) are not passing them

cruky: You are right, I noticed that as well, but thought that we are gonna need it... I will remove it! Thanks

cruky: But for now the future strategy behaviour can only be controlled globally for all future strategies by these `constants.FutureTemplateStrategiesConstants`, which could be defined in the setup file. There is no option to set these values other way, or to define them for individual asset. So we can get rid of these attributes for now?

Hassan754: I see, yes sounds good to me. Sorry for the confusion and thanks again!

tests/strategies/future/test_future_strategy.py
```diff
@@ -51,10 +51,20 @@ def setup_method(self) -> None:
...
+        self._original_final_selling_rate = FutureTemplateStrategiesConstants.FINAL_SELLING_RATE
 
     def teardown_method(self) -> None:
         """Test cleanup"""
         GlobalConfig.FUTURE_MARKET_DURATION_HOURS = self._original_future_markets_duration
+        FutureTemplateStrategiesConstants.INITIAL_BUYING_RATE = self._original_initial_buying_rate
+        FutureTemplateStrategiesConstants.FINAL_BUYING_RATE = self._original_final_buying_rate
+        FutureTemplateStrategiesConstants.INITIAL_SELLING_RATE = (
+            self._original_initial_selling_rate)
+        FutureTemplateStrategiesConstants.FINAL_SELLING_RATE = self._original_final_selling_rate
```
hannesdiedrich, line 67: why is this needed? AFAIS, you are not mutating these Constants

src/gsy_e/models/strategy/future/strategy.py
hannesdiedrich, line 271: NIT:
```suggestion
        initial_buying_rate: Optional[float] = None,
        final_buying_rate: Optional[float] = None,
        initial_selling_rate: Optional[float] = None,
        final_selling_rate: Optional[float] = None
```

hannesdiedrich (approved):
LGTM

Hassan754 (approved):

src/gsy_e/models/strategy/future/strategy.py
cruky: Right! Thank you, I will keep this in mind. Although I've just removed these arguments completely after this disscussion https://github.com/gridsingularity/gsy-e/pull/1450#discussion_r897735593.

hannesdiedrich: Ah, ok. I musst have looked at a wrong commit again. Sorry. I think the goal behind having these as construction parameters, in oder to be prepared for the future, when the user wants to adapt these per area and not globally. This is why the former version made more sense to me. But as this is just a minor change, we can alway change it back later.

hannesdiedrich (approved):
LGTM

PR #1452, opened 2022-06-15
GSYE-276: Set the clock updater in the activate_future_markets

cruky:

Since the Offer and Bid gets `creation_time` from the Market `now` attribute, it should be populated before the market cycle, and not only in `execute_actions_after_tick_event`.

## Proposed changes

Add `update_clock` in `activate_future_markets` to populate `now` attribute for future markets.

### DIFF
--- a/src/gsy_e/models/area/markets.py
+++ b/src/gsy_e/models/area/markets.py
@@ -86,6 +86,7 @@ class AreaMarkets:
                               grid_fee_const=area.grid_fee_constant),
             name=area.name)
         self.future_markets = market
+        self.future_markets.update_clock(area.now)
         area.dispatcher.create_market_agents_for_future_markets(market)
 
     def activate_market_rotators(self):


src/gsy_e/models/area/markets.py
```diff
@@ -86,6 +86,7 @@ def activate_future_markets(self, area: "Area") -> None:
                               grid_fee_const=area.grid_fee_constant),
             name=area.name)
         self.future_markets = market
+        self.future_markets.update_clock(area.now)
```
hannesdiedrich, line 89: Is this really fixing the issue? From [Eduardos comment](https://gridsingularity.atlassian.net/browse/GSYE-276?focusedCommentId=26561) Is see that the creation time is missing for more than just the first offer.

Hassan754 (approved):
LGTM

src/gsy_e/models/area/markets.py
cruky: Yes, in this particular example we have 4 offers in the first tick from [H2 Storage1], in slots: 00:15:00+00:00,00:30:00+00:00,00:45:00+00:00,01:00:00+00:00 (setup in [GSYE-290 ticket](https://gridsingularity.atlassian.net/browse/GSYE-290)) . Before, these had no `creation_time` because none of the future markets had their `now` attribute updated. Now it works, all these 4 offers have their `creation_time`.

PR #1453, opened 2022-06-16
Change log level of simulation stopping to INFO

fievelk:

We're seeing logs for stopped canary networks as errors, while they can be due to admin's choice.

## Proposed changes

- Change the log level for simulations being stopped to `INFO`.

---


hannesdiedrich (approved):
LGTM

Hassan754 (dismissed):

fievelk:
I've talked to Spyros about these logs. The reason why we were using `error` level even if they're not errors is that this makes it possible to log these messages on Kibana. We decided to change the severity to `warning` instead, so that they'll still be logged by Kibana but won't be considered errors. I've pushed these changes in the last commit.

PR #1454, opened 2022-06-16
Feature/gsye 256: SCM static coefficient algorithm

spyrostz:

Deals with the addition of the static coefficient algorithm, and also partially handles the calculation of the home and community energy bills. 

## Proposed changes

-


src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -399,6 +402,9 @@ def update_and_send_results(self, current_state: dict, progress_info: Simulation
 
     @classmethod
     def _update_area_stats(cls, area: "Area", endpoint_buffer: "SimulationEndpointBuffer") -> None:
+        # TODO: Fix in the context of GSYE-258
```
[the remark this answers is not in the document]
hannesdiedrich: ognore me, wrong ticket number

src/gsy_e/models/area/scm_manager.py
```diff
@@ -0,0 +1,249 @@
...
+    consumption_kWh: float = 0.
+    production_kWh: float = 0.
+    self_consumed_energy_kWh: float = 0.
+    energy_surplus_kWh: float = 0.
+    energy_need_kWh: float = 0.
+    community_production_kWh: float = 0.
+    trades = None
+
+    def __post_init__(self):
+        self.self_consumed_energy_kWh = min(self.consumption_kWh, self.production_kWh)
```
hannesdiedrich, line 31: Can you please explain this step? Why would the production influence the consumed energy?

src/gsy_e/models/area/scm_manager.py
```diff
@@ -0,0 +1,249 @@
...
+from typing import Dict
+from uuid import uuid4
+
+from gsy_framework.data_classes import Trade
+from pendulum import DateTime
+
+from gsy_e.constants import DEFAULT_GRID_SELLER_STRING, DEFAULT_SCM_SELLER_STRING
+
+
+@dataclass
```
hannesdiedrich, line 12: Can you please explain what the benefit is here for declaring a dataclass? If I look at the logic it does a lot of things on top of only holding data.

src/gsy_e/models/area/__init__.py
```diff
@@ -240,17 +242,86 @@ def update_descendants_strategy_prices(self):
...
-            return self.strategy.state.get_results_dict(self._current_time_slot)
+            return self.strategy.state.get_results_dict(self._current_market_time_slot)
         return {}
 
 
+class CoefficientArea(AreaBase):
+    """Area class for the coefficient matching mechanism."""
+    def __init__(self, name: str = None, children: List["CoefficientArea"] = None,
+                 uuid: str = None,
+                 strategy: BaseStrategy = None,
```
hannesdiedrich, line 256:
```suggestion
                 strategy: SCMStrategy = None,
```

src/gsy_e/models/area/scm_manager.py
```diff
@@ -0,0 +1,249 @@
...
+        self.community_data.energy_bought_from_community_kWh = sum(
+            data.energy_bought_from_community_kWh for data in self._home_data.values())
+        self.community_data.energy_sold_to_grid_kWh = sum(
+            data.energy_sold_to_grid_kWh for data in self._home_data.values())
+
+    def calculate_home_energy_bills(
+            self, home_uuid: str) -> None:
+        """Calculate energy bills for one home."""
+        if home_uuid not in self._home_data:
+            return
```
hannesdiedrich, line 166: Do you think, that it makes sense to add an assertion instead here, because AFAIU, there should be only homes in the community, right?

src/gsy_e/models/area/scm_manager.py
```diff
@@ -0,0 +1,249 @@
...
+
+        for home_data in self._home_data.values():
+            home_data.set_community_production(self.community_data.energy_surplus_kWh)
+
+        self.community_data.energy_bought_from_community_kWh = sum(
+            data.energy_bought_from_community_kWh for data in self._home_data.values())
+        self.community_data.energy_sold_to_grid_kWh = sum(
+            data.energy_sold_to_grid_kWh for data in self._home_data.values())
+
+    def calculate_home_energy_bills(
```
hannesdiedrich, line 162: I will look more closely into this central method tomorrow, but I have the feeling, that it could be divided into multiple methods, also to make it more readable. DO you agree?

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -801,7 +844,7 @@ def _execute_simulation(
                 self._simulation_finish_actions(slot_count)
                 return
 
-            self._results.update_csv_on_market_cycle(slot_no, self.area)
+            # self._results.update_csv_on_market_cycle(slot_no, self.area)
```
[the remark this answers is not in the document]
spyrostz: Yes, because the simulation results are not adapted yet (will do it in the context of this https://gridsingularity.atlassian.net/browse/GSYE-258), generating the CSV files caused the simulation to crash thus not allowing me to test whether the simulation is running correctly. I'll reintroduce it at the next PR where the simulation results are dealt with.

src/gsy_e/models/area/scm_manager.py
spyrostz: Heads up on the double self :D The member variable is named `self_consumed_energy_kWh` , which reflects the amount of energy that the house would self-consume (in other words, consumption that would be covered by production of the house without external trading). Self consumption can never be greater than the production of the house, if the house produces more than it can consume it will sell the remaining energy to the grid, and the self_consumed_energy_kWh will be equal to the consumption_kWh

src/gsy_e/models/area/scm_manager.py
spyrostz: Admittedly there is not much of a benefit here. I was originally planning to have this as a container of data, but shortly after it was evident that these data would benefit from methods that would contain logic too. Of course though, it does not hurt for this to be a dataclass, but if you would prefer it to be a normal class ping me here and I will adapt it, thanks!

src/gsy_e/models/area/__init__.py
spyrostz: Fixing, thanks!

src/gsy_e/models/area/scm_manager.py
spyrostz: You are right, it is better to limit it only for houses in order to avoid any surprises. Fixing it, thanks!

src/gsy_e/models/area/scm_manager.py
spyrostz: I have already done some fixes in this method for my next task in this epic (I have essentially moved large parts of it in separate methods of the HomeEnergyBills class as part of fully calculating the energy bills) therefore if you can tolerate this to be dealt with at the results PR it would be great! If you think though that it is cleaner to do it now, then let me know and I'll change it here, no problem at all! Thanks in advance

src/gsy_e/setup/scm/default_2a.py
```diff
@@ -0,0 +1,69 @@
...
+from gsy_e.models.strategy.scm.pv import SCMPVStrategy
+from gsy_e.models.strategy.scm.storage import SCMStorageStrategy
+from gsy_e.models.strategy.scm.smart_meter import SCMSmartMeterStrategy
+
+ConstSettings.MASettings.MARKET_TYPE = SpotMarketTypeEnum.COEFFICIENTS.value
+
+
+def get_setup(config):
+    area = CoefficientArea(
+        "Grid",
```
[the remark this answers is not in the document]
spyrostz: Fixing, thanks!

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -766,6 +775,32 @@ def restore_global_state(self, saved_state: dict) -> None:
 
 class CoefficientSimulation(Simulation):
     """Start and control a simulation with coefficient trading."""
+
+    def _init(self, redis_job_id: str) -> None:
+        # has to be called before get_setup():
```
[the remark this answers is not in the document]
spyrostz: Precisely, the method in the comment should be renamed, since get_setup is now called by the load_setup_module(). Fixing it, thanks!

src/gsy_e/models/strategy/scm/load.py
```diff
@@ -29,11 +30,30 @@ def _get_day_of_timestamp(self, time_slot: DateTime) -> int:
 
     def market_cycle(self, area: "AreaBase") -> None:
         """Update the load forecast and measurements for the next/previous market slot."""
+        self._energy_params.add_entry_in_hrs_per_day(area._current_market_time_slot)
+        self._update_energy_requirement_and_measurement(area)
+
+    def _update_energy_requirement_and_measurement(self, area: "AreaBase"):
+        self._energy_params.update_energy_requirement(area._current_market_time_slot)
+
+        if not self._energy_params.allowed_operating_hours(area._current_market_time_slot):
```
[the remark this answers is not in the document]
spyrostz: Well, this specific method (allowed_operating_hours) is still needed because in the former discussion we decided to maintain the hrs_of_day parameter and to remove support for the hrs_per_day, however I reviewed the code again and you are right, by mistake I added the "EnergyParameters" class that includes the hrs_per_day instead of the one that does not! Adapting the SCM strategy to use the simpler interface, thanks again for bringing this up!

spyrostz:
> I started reviewing, but need more time to understand. I requested access to the google doc that (hopefully) explains the design of the mechanism more detailed than the confluence page). Thanks!

I fixed all the open comments, and I invited you to the google sheet that includes the formulas for these calculations, thanks!

src/gsy_e/models/area/scm_manager.py
hannesdiedrich: OK, thanks! Understood. Leave it as is then.

src/gsy_e/models/area/scm_manager.py
hannesdiedrich: Ah no. I can wait for the results PR.

src/gsy_e/models/area/__init__.py
```diff
@@ -240,17 +242,88 @@ def update_descendants_strategy_prices(self):
...
+        # pylint: disable=too-many-arguments
+        super().__init__(name, children, uuid, strategy, config, grid_fee_percentage,
+                         grid_fee_constant)
+        self._coefficient_percent = coefficient_percent
+        self._market_maker_rate_eur = market_maker_rate_eur
+        self._feed_in_tariff_eur = feed_in_tariff_eur
+        self._trade_rate = trade_rate
+        self.past_market_time_slot = None
+
+    def activate_coefficients(self, current_time_slot: DateTime) -> None:
```
hannesdiedrich, line 275: I was confused by the method name. AFAIU, the `SCMStrategy.activate()` is not dealing with any coefficients, but rather collects the energy settings of the asset. 
I know, that there is now bidding strategy for SCM assets, but how about to be consistent, as we use the term "strategy" even in the SCM case and call this method: `activate_strategies` or alternatively: `activate_energy_parameters`?

src/gsy_e/models/area/scm_manager.py
```diff
@@ -0,0 +1,249 @@
...
+        market_maker_rate_decreased_fees = (
+                market_maker_rate_eur + grid_fees_eur * self._grid_fees_reduction)
+        market_maker_rate_normal_fees = market_maker_rate_eur + grid_fees_eur
+
+        base_energy_bill = (
+                home_data.energy_need_kWh * market_maker_rate_normal_fees -
+                home_data.energy_surplus_kWh * feed_in_tariff_eur)
+
+        if home_data.allocated_community_energy_kWh > home_data.energy_need_kWh:
+            if home_data.energy_surplus_kWh > 0.0:
```
hannesdiedrich, line 182: I am confused here. How can the home have a surplus `home_data.energy_surplus_kWh > 0.0` and also need energy `home_data.allocated_community_energy_kWh > 0`?

tests/area/test_coefficient_area.py
```diff
@@ -0,0 +1,161 @@
...
+        assert isclose(trades[1].trade_rate, 0.3)
+        assert isclose(trades[1].traded_energy, 0.14)
+        assert trades[1].seller == "Grid"
+        assert trades[1].buyer == "House 1"
+        assert len(scm._home_data[house2.uuid].trades) == 2
+        trades = scm._home_data[house2.uuid].trades
+        assert isclose(trades[0].trade_rate, 0.24)
+        assert isclose(trades[0].traded_energy, 0.06)
+        assert trades[0].seller == "Community"
+        assert trades[0].buyer == "House 2"
```
hannesdiedrich, line 157: Why is the "House 2" buying energy? AFAIU, it does not need to sell energy, but needs to sell the energy, that it received from the Community (0.04 kWh) to the grid again.

hannesdiedrich (commented):
I reviewed again and tried to understand the energy flow better by using the google sheet you provided and the unit tests. But still I have some doubts and but them into comments. Haven't reviewed the pricing yet in order not confuse myself more. Would be great if you first answer the questions I have. Thanks a lot in advance!

src/gsy_e/models/area/scm_manager.py
```diff
@@ -0,0 +1,249 @@
...
+                    )
+                if self.community_data.energy_sold_to_grid_kWh > 0.:
+                    home_data.create_sell_trade(
+                        self._time_slot, DEFAULT_GRID_SELLER_STRING,
+                        self.community_data.energy_sold_to_grid_kWh,
+                        self.community_data.energy_sold_to_grid_kWh * feed_in_tariff_eur)
+            else:
+                gsy_energy_bill = home_data.energy_need_kWh * market_maker_rate_decreased_fees
+                home_data.create_buy_trade(
+                    self._time_slot, DEFAULT_SCM_SELLER_STRING, home_data.energy_need_kWh,
```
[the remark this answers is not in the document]
spyrostz: Will do so, thanks!

src/gsy_e/models/area/scm_manager.py
spyrostz: This is a tricky one indeed. The energy need of the house is the variable energy_need_kWh. The allocated energy to a house is not related to this though. Essentially, the house can be allocated more energy than it needs, hence the case that you describe: the house has energy surplus, but at the same time the community has allocated some energy to it.

hannesdiedrich: Ah! Thanks! This made me understand the whole if block here. A inline comment would be helpful. But you already promised to move parts of the logic into sub methods right? This will work too. Thanks!

tests/area/test_coefficient_area.py
hannesdiedrich: Ah yes, of course! Thanks!

src/gsy_e/models/area/scm_manager.py
```diff
@@ -0,0 +1,248 @@
...
+
+
+class SCMManager:
+    """Handle the community manager coefficient trade."""
+    def __init__(self, community_uuid: str, time_slot: DateTime):
+        self._home_data: Dict[str, HomeAfterMeterData] = {}
+        self.community_data = CommunityData(community_uuid)
+        self._time_slot = time_slot
+        self._bills: Dict[str, HomeEnergyBills] = {}
+        self._grid_fees_reduction = 0.28
```
hannesdiedrich, line 123: So I understand that this parameter controls the reduction in grid fees as a benefit for the homes to trade on the community and have less interaction with the grid right?
Who came up with this percentage? I guess it will be the same for all homes right? Or will each Home have a separate contract with the DSO? If the first is true, I suggest to add this value in a central entity (e.g. ConstSettings). If the latter is true, we probably need to expose this to the CoefficientArea class, right?

src/gsy_e/models/strategy/scm/__init__.py
```diff
@@ -28,3 +28,13 @@ def get_energy_to_sell_kWh(self, time_slot: DateTime) -> float:
...
+
+    def decrease_energy_to_buy(
+            self, traded_energy_kWh: float, time_slot: DateTime, area: "CoefficientArea"):
+        """Decrease traded energy from the state and the strategy parameters."""
+        pass
+
+    def decrease_energy_to_sell(
+            self, traded_energy_kWh: float, time_slot: DateTime, area: "CoefficientArea"):
+        """Decrease traded energy from the state and the strategy parameters."""
+        pass
```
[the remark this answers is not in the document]
hannesdiedrich: Sorry, your answer confuses me a bid. I understand, that now all the energy that is available or that is needed gets "traded" right away and in total, but in the future there might be intermediate steps where partial trading will be done even in the SCM in which case we will have to decrease the `energy_to_sell` or `energy_to_buy` ?

src/gsy_e/models/area/scm_manager.py
spyrostz: Yes, I will move it to individual methods (have more or less done it in GSYE-258 branch) and ping you in the next PR, thanks!

src/gsy_e/models/area/scm_manager.py
spyrostz: > So I understand that this parameter controls the reduction in grid fees as a benefit for the homes to trade on the community and have less interaction with the grid right?

More or less yes, it is a way the DSOs use to incentivise loads to participate in the SCM, therefore the benefit for the loads is not a decrease in energy price like in a "normal" peer-to-peer market, but in a decrease in grid fees. 

> Who came up with this percentage?
This comes from Austrian regulation as far as I know

 > I guess it will be the same for all homes right?
It is the same for all homes, and the same for all SCMs in Austria

> If the first is true, I suggest to add this value in a central entity (e.g. ConstSettings).
I agree and I will adapt it now, thanks!

src/gsy_e/models/area/scm_manager.py
spyrostz: Exactly!

hannesdiedrich (dismissed):
LGTM

src/gsy_e/models/strategy/scm/__init__.py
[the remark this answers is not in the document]
Hassan754: Just to double check, if we know that for loads and PVs these methods aren't to be called, why did we actually implement them in the first place there? (I mean we could've just left them as empty methods)

Hassan754 (approved):
Left some comments, LGTM in general

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -357,7 +358,9 @@ def update_and_send_results(self, current_state: dict, progress_info: Simulation
         """
         assert self._endpoint_buffer is not None
 
-        if ConstSettings.MASettings.MARKET_TYPE == SpotMarketTypeEnum.COEFFICIENTS:
+        if ConstSettings.MASettings.MARKET_TYPE == SpotMarketTypeEnum.COEFFICIENTS.value:
+            if not self._should_send_results_to_broker:
+                return
```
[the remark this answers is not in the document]
spyrostz: I will omit this because I have significantly refactored the simulation results for the coefficients, and remove this if block completely as part of it (I instead used factory classes to either construct an instance of SimulationResultsManager or CoefficientSImulationResultsManager to avoid the need for these if blocks). Therefore if you can be patient till the followup PR would be great!

src/gsy_e/models/area/__init__.py
```diff
@@ -240,17 +242,88 @@ def update_descendants_strategy_prices(self):
...
+                 coefficient_percent: float = 0.0,
+                 market_maker_rate_eur: float = (
+                         ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE / 100.),
+                 feed_in_tariff_eur: float = GlobalConfig.FEED_IN_TARIFF / 100.,
+                 trade_rate: float = 0.0
+                 ):
+        # pylint: disable=too-many-arguments
+        super().__init__(name, children, uuid, strategy, config, grid_fee_percentage,
+                         grid_fee_constant)
+        self._coefficient_percent = coefficient_percent
```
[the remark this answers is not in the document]
spyrostz: This is a good point, I think that a validator that handles grid structure and such dependencies is already under consideration by @fievelk , however it does not hurt to add such a validator exclusively for gsy-e that handles only this. I will add it in the context of this PR, and if we do not need it in the future then we can remove it.

src/gsy_e/models/area/__init__.py
```diff
@@ -240,17 +242,88 @@ def update_descendants_strategy_prices(self):
...
+        if self.strategy:
+            self.strategy.market_cycle(self)
+
+        for child in self.children:
+            child.cycle_coefficients_trading(current_time_slot)
+
+    def _is_home_area(self):
+        return self.children and all(
+            child.strategy and isinstance(child.strategy, SCMStrategy)
+            for child in self.children)
```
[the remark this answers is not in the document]
spyrostz: Fixed, thanks!

spyrostz:
Just to reply to @Hassan754 's comment here https://github.com/gridsingularity/gsy-e/pull/1454#discussion_r903548516, this was done due to uncertainty over what we might need or not. It was proven though that for the model that the SCM implements, there will never be any imbalances. As for removing those, since these are one-liner methods (they only relay the respective call to the state classes) I do not think that they are that big of a maintenance effort, nor that having empty implementations will increase the code readability.

fievelk (commented):
Left some comments. Thanks for working on this!

src/gsy_e/models/area/__init__.py
```diff
@@ -240,17 +242,88 @@ def update_descendants_strategy_prices(self):
...
+    def __init__(self, name: str = None, children: List["CoefficientArea"] = None,
+                 uuid: str = None,
+                 strategy: SCMStrategy = None,
+                 config: SimulationConfig = None,
+                 grid_fee_percentage: float = None,
+                 grid_fee_constant: float = None,
+                 coefficient_percent: float = 0.0,
+                 market_maker_rate_eur: float = (
+                         ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE / 100.),
+                 feed_in_tariff_eur: float = GlobalConfig.FEED_IN_TARIFF / 100.,
```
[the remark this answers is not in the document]
spyrostz: Since this is on the datasheet then I will change it, however to my knowledge this regulation is only for European countries and will not apply to any of the other currencies (might be for Switzerland but I am not certain even for this). Nevertheless, does not hurt to remove euro from the name. I will also add comments to indicate that these are in the currency of each respective country that is used.

src/gsy_e/models/area/__init__.py
```diff
@@ -240,17 +242,88 @@ def update_descendants_strategy_prices(self):
...
+    def activate_energy_parameters(self, current_time_slot: DateTime) -> None:
+        """Activate the coefficient-based area parameters."""
+        self._current_market_time_slot = current_time_slot
+
+        if self.strategy:
+            self.strategy.activate(self)
+        for child in self.children:
+            child.activate_energy_parameters(current_time_slot)
+
+    def cycle_coefficients_trading(self, current_time_slot: DateTime) -> None:
```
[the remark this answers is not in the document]
spyrostz: Precisely.

src/gsy_e/models/area/__init__.py
```diff
@@ -240,17 +242,88 @@ def update_descendants_strategy_prices(self):
...
+
+        if self.strategy:
+            self.strategy.market_cycle(self)
+
+        for child in self.children:
+            child.cycle_coefficients_trading(current_time_slot)
+
+    def _is_home_area(self):
+        return self.children and all(
+            child.strategy and isinstance(child.strategy, SCMStrategy)
```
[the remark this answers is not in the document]
spyrostz: Frankly it is mostly validation indeed. If you agree then, I will move it to the validation of the SCMManager class that also validates the Area coefficients, in order for both to be centralized.

spyrostz:
Unfortunately github did not allow me (again) to reply to one of the comments, therefore doing it here:

> I was having the same question before reading this conversation, so I'd suggest adding a comment here to clarify in the code too :)

I added a rather detailed comment in order to clarify it :)

fievelk (approved):

hannesdiedrich (approved):
LGTM

src/gsy_e/models/area/__init__.py
```diff
@@ -240,17 +242,86 @@ def update_descendants_strategy_prices(self):
...
+                 feed_in_tariff: float = GlobalConfig.FEED_IN_TARIFF / 100.,
+                 trade_rate: float = 0.0
+                 ):
+        # pylint: disable=too-many-arguments
+        super().__init__(name, children, uuid, strategy, config, grid_fee_percentage,
+                         grid_fee_constant)
+        self._coefficient_percent = coefficient_percent
+        self._market_maker_rate = market_maker_rate
+        self._feed_in_tariff = feed_in_tariff
+        self._trade_rate = trade_rate
```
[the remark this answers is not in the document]
spyrostz: Unfortunately I merged either shortly before or shortly after you wrote the comment, and I did not manage to deal with it :( It is a remnant of an old implementation and should be removed. I will remove it in the context of GSYE-258

PR #1455, opened 2022-06-16
GSYE-315: Fix the future market rates updater

Hassan754:

This PR fixes the following bug:

The incrementation of the update counter in the future markets
In the update_frequency, we used to calculate the elapsed_seconds of simulations (in a slot).
For the future markets however, this can't be calculated easily based on the tick in slot, because we may have several future markets with different starting time (as the future markets window moves right and creates new slots in place of the ones that became spot, there could be two consequent future markets that expect a different elapsed seconds taking into consideration that the elapsed seconds is since the time it was first created)

Basically, for newly added future markets, the `time_for_price_update` was always returning true for several consequent ticks (because in the original time_for_price_update method we check whether the elapsed seconds duration is greater than self.update_interval.seconds * self.update_counter[time_slot]) , and this is the bug, we were considering that the elapsed seconds for new and old future markets are the same)
## Proposed changes

Take into consideration the time of creation of the future market when calculating the elapsed seconds


hannesdiedrich:
I spent some time trying to understand your reasoning here, but I still not understand it, sorry.

AFAIU, `_elapsed_seconds` is already taking into account the `current_tick_number` (which is the current number of ticks in the market_slot). So, even if the future market started 1 day in the future the `_elapsed_seconds` should return the time inside the market slot. We have access to the global `current_tick` (which is the current tick of the whole simulation). This should increase the tick count in every future or past market evenly and simultaneously. So, if 5 minutes elapsed in the current spot market, the same applies to all future markets and they should only update the price then.
Can you please explain some more, why the reduction of the creation time of the market is needed [here](https://github.com/gridsingularity/gsy-e/pull/1455/files#diff-bf32dee2ce4eaa87404d14ed3015f3a70bf17ee8d4d87a2132ae68cbae7e0793R54)?  Thanks a lot for bering with me. 

While investigating and writing the above I realise that we only have one `update_counter` per future strategy. Shouldn't we have one update counter per strategy AND per future market? . This way the update in the first future market will not update the counter of the next future market. Please correct me if I oversaw this. Thanks!

Hassan754:
@hannesdiedrich To showcase the bug fixed in context of this PR, let's take an example of a simulation rotation with future market activated.

Assuming tick length = 15 seconds, slot length = 15 minutes, future market duration = 1 hour (3600 seconds), future market update duration = 5 min -> 240 ticks per the future markets window and update once per 20 ticks (300 seconds).

Screenshot 2022-06-21 at 14 36 24

The blue rectangle represents a spot market, and the rest represent future markets, you can find the time at which the slots were created underneath.
As you may know, before updating the rates of spot markets we check how many seconds have passed since the beginning of this same slot, so basically for each slot we need to check its age, but for the future markets, we currently check the age of the earliest future slot, whereas this doesn't make sense as there are several future slots that are created in different times.
Screenshot 2022-06-21 at 14 37 45

As you can see, when we rotated the markets, we created a new future market (relatively younger than the others)

When the simulation runs, the future markets will try to update the rate at each tick event ([code](https://github.com/gridsingularity/gsy-e/blob/4e984c3660c88d0cef239f8e259a091066ca985a/src/gsy_e/models/strategy/future/strategy.py#L245)).
Before doing the actual update, we will validate whether the time is due for an update or not ([code](https://github.com/gridsingularity/gsy-e/blob/4e984c3660c88d0cef239f8e259a091066ca985a/src/gsy_e/models/strategy/future/strategy.py#L81)), the corresponding function will calculate how many second have already passed since the beginning of the first future slot.
Let's say that the tick count is currently 50, when we create a new future market, this [function](https://github.com/gridsingularity/gsy-e/blob/4e984c3660c88d0cef239f8e259a091066ca985a/src/gsy_e/models/strategy/update_frequency.py#L239) will always return true (because it will take into account the elapsed seconds of the oldest future markets = 50 ticks = 3000 seconds )
In other words, from the first future market slot perspective, the elapsed seconds is 3000 seconds, but for the newly created market it's less than that.

I hope that it's a bit clearer now

src/gsy_e/models/strategy/future/strategy.py
```diff
@@ -48,6 +47,13 @@ def _get_all_time_slots(area: "Area") -> List[DateTime]:
             return []
         return area.future_markets.market_time_slots
 
+    def time_for_price_update(self, strategy: "BaseStrategy", time_slot: DateTime) -> bool:
+        """Check if the prices of bids/offers should be updated."""
+        return (
+                self._elapsed_seconds(strategy.area)
+                - self.market_slot_added_time_mapping[time_slot] >= (
+                        self.update_interval.seconds * self.update_counter[time_slot]))
+
```
Hassan754, line 56: If the elapsed seconds = 3000 seconds and the slot is created at 1000 seconds, then the elapsed seconds = 3000 - 1000 = 2000 seconds (age of the market)

hannesdiedrich: OK, this makes sense now. It took me a while to understand. Thanks a lot for the explanation (also the verbose one!)

src/gsy_e/models/strategy/update_frequency.py
```diff
@@ -159,11 +165,17 @@ def _populate_profiles(self, area: "Area") -> None:
                 final_rate = find_object_of_same_weekday_and_time(
                     self.final_rate_profile_buffer, time_slot)
 
+            # Hackathon TODO: get rid of self.initial_rate, self.final_rate, self.update_counter
+            # and self.market_slot_added_time_mapping in favor of one object
+            # that keeps track of time_slot: attributes
```
hannesdiedrich, line 170: 👍

src/gsy_e/models/strategy/update_frequency.py
```diff
@@ -159,11 +165,17 @@ def _populate_profiles(self, area: "Area") -> None:
...
             self.initial_rate[time_slot] = initial_rate
             self.final_rate[time_slot] = final_rate
 
             self._set_or_update_energy_rate_change_per_update(time_slot)
             write_default_to_dict(self.update_counter, time_slot, 0)
+            if time_slot not in self.market_slot_added_time_mapping:
+                elapsed_seconds = self._elapsed_seconds(area)
+                self.market_slot_added_time_mapping[time_slot] = elapsed_seconds
```
hannesdiedrich, line 178: IMO this is worth another private method with a docstring that is called in `update_and_populate_price_settings`(because it is not really fitting to the method name here `_populate_profiles`)

hannesdiedrich (commented):
LGTM for the general fix and thanks for bering with my slow head!
I have two comments for improvement.

src/gsy_e/models/strategy/update_frequency.py
Hassan754: That definitely makes sense.
I believe that we will soon refactor this module so that both spot and future markets share the same way of calculation,  if we plan on visiting it in the upcoming hackathon, I'll leave it as it, otherwise I will encapsulate this logic.

hannesdiedrich: Ok, agreed. Leave it as is for now and we will refactor the whole module later.

hannesdiedrich: But can you please do

> IMO this is worth another private method with a docstring that is called in update_and_populate_price_settings(because it is not really fitting to the method name here _populate_profiles)
?
This way we do not oversee this when refactoring. What do you think?

hannesdiedrich (approved):
LGTM

src/gsy_e/models/strategy/update_frequency.py
```diff
@@ -123,6 +128,7 @@ def delete_past_state_values(self, current_market_time_slot: DateTime) -> None:
             self.final_rate.pop(market_slot, None)
             self.energy_rate_change_per_update.pop(market_slot, None)
             self.update_counter.pop(market_slot, None)
+            self.market_slot_added_time_mapping.pop(market_slot, None)
```
cruky, line 131: I am not sure if `delete_past_state_values` is even triggered for future markets. At least what I see is that dicts like `update_counter`, `initial_rate`, `market_slot_added_time_mapping` are growing in items endlessly. Do we miss `_delete_past_state` for future markets in `event_market_cycle` ? Or maybe there is other reason for such behavior?

Hassan754: I have checked and you are right, the cleaning up of the past state seems to be missing.
I believe that the `_delete_past_state` should be overridden and adapted for the future markets.
Unless this issue is affecting the DoD of the bug ticket, I'll create a ticket to resolve this problem if you agree.

cruky (approved):
LGTM, really great that you solved the puzzle of this bug. Thanks for all clarification!

PR #1459, opened 2022-06-24
GSYE-334: Delete the spot market state from the future market strategy

Hassan754:
The future market strategies's state isn't cleared for markets that become spots (they are no longer future markets).

## Proposed changes
Add a call to the delete_past_state method in the future strategies that clears both bid and offer rate updaters' states.


spyrostz (approved):
LGTM

hannesdiedrich (approved):
LGTM

PR #1461, opened 2022-06-24
GSYE-324: Add unittest for launch_simulation_from_rq_job

mrsaemir:

As `launch_simulation_from_rq_job` is the only entrypoint to running a simulation from the rq. It seems fair to have a test for this function ([GSYE-324](https://gridsingularity.atlassian.net/browse/GSYE-324)).

## Proposed changes

- Add `TestRQJobUtils` class
- Minor refactor to `launch_simulation_from_rq_job`

### DIFF
--- a/src/gsy_e/gsy_e_core/rq_job_handler.py
+++ b/src/gsy_e/gsy_e_core/rq_job_handler.py
@@ -16,7 +16,8 @@ from gsy_e.gsy_e_core.simulation import run_simulation
 from gsy_e.gsy_e_core.util import available_simulation_scenarios, update_advanced_settings
 from gsy_e.models.config import SimulationConfig
 
-logger = logging.getLogger()
+logger = logging.getLogger(__name__)
+logger.setLevel(logging.INFO)
 
 
 def decompress_and_decode_queued_strings(queued_string: bytes) -> Dict:
--- a/src/gsy_e/gsy_e_core/rq_job_handler.py
+++ b/src/gsy_e/gsy_e_core/rq_job_handler.py
@@ -26,19 +26,21 @@ def decompress_and_decode_queued_strings(queued_string: bytes) -> Dict:
 
 def launch_simulation_from_rq_job(scenario: bytes, settings: Optional[Dict],
                                   events: Optional[str], aggregator_device_mapping: str,
-                                  saved_state: bytes, job_id: str):
+                                  saved_state: bytes, job_id: str,
+                                  connect_to_profiles_db: bool = True):
     # pylint: disable=too-many-arguments, too-many-locals
     """Launch simulation from rq job."""
     logging.getLogger().setLevel(logging.ERROR)
     scenario = decompress_and_decode_queued_strings(scenario)
-    gsy_e.constants.CONFIGURATION_ID = scenario.pop("configuration_uuid")
-    if "collaboration_uuid" in scenario:
-        gsy_e.constants.EXTERNAL_CONNECTION_WEB = True
-        GlobalConfig.IS_CANARY_NETWORK = scenario.pop("is_canary_network", False)
-        gsy_e.constants.RUN_IN_REALTIME = GlobalConfig.IS_CANARY_NETWORK
+    if isinstance(scenario, dict):
+        gsy_e.constants.CONFIGURATION_ID = scenario.pop("configuration_uuid")
+        if "collaboration_uuid" in scenario:
+            gsy_e.constants.EXTERNAL_CONNECTION_WEB = True
+            GlobalConfig.IS_CANARY_NETWORK = scenario.pop("is_canary_network", False)
+            gsy_e.constants.RUN_IN_REALTIME = GlobalConfig.IS_CANARY_NETWORK
     saved_state = decompress_and_decode_queued_strings(saved_state)
-    log.error("Starting simulation with job_id: %s and configuration id: %s",
-              job_id, gsy_e.constants.CONFIGURATION_ID)
+    log.info("Starting simulation with job_id: %s and configuration id: %s",
+             job_id, gsy_e.constants.CONFIGURATION_ID)
 
     try:
         if settings is None:
@@ -122,8 +124,7 @@ def launch_simulation_from_rq_job(scenario: bytes, settings: Optional[Dict],
         kwargs = {"no_export": True,
                   "seed": settings.get("random_seed", 0)}
 
-        gsy_e.constants.CONNECT_TO_PROFILES_DB = True
-
+        gsy_e.constants.CONNECT_TO_PROFILES_DB = connect_to_profiles_db
         run_simulation(setup_module_name=scenario_name,
                        simulation_config=config,
                        simulation_events=events,
@@ -132,8 +133,8 @@ def launch_simulation_from_rq_job(scenario: bytes, settings: Optional[Dict],
                        slot_length_realtime=slot_length_realtime,
                        kwargs=kwargs)
 
-        log.error("Finishing simulation with job_id: %s and configuration id: %s",
-                  job_id, gsy_e.constants.CONFIGURATION_ID)
+        log.info("Finishing simulation with job_id: %s and configuration id: %s",
+                 job_id, gsy_e.constants.CONFIGURATION_ID)
 
     # pylint: disable=broad-except
     except Exception:
@@ -142,3 +143,4 @@ def launch_simulation_from_rq_job(scenario: bytes, settings: Optional[Dict],
         publish_job_error_output(job_id, traceback.format_exc())
         logging.getLogger().exception("Error on jobId, %s, configuration id: %s",
                                       job_id, gsy_e.constants.CONFIGURATION_ID)
+        raise
--- a/src/gsy_e/gsy_e_core/rq_job_handler.py
+++ b/src/gsy_e/gsy_e_core/rq_job_handler.py
@@ -17,7 +17,6 @@ from gsy_e.gsy_e_core.util import available_simulation_scenarios, update_advance
 from gsy_e.models.config import SimulationConfig
 
 logger = logging.getLogger()
-logger.setLevel(logging.INFO)
 
 
 def decompress_and_decode_queued_strings(queued_string: bytes) -> Dict:
--- a/src/gsy_e/gsy_e_core/rq_job_handler.py
+++ b/src/gsy_e/gsy_e_core/rq_job_handler.py
@@ -16,7 +16,8 @@ from gsy_e.gsy_e_core.simulation import run_simulation
 from gsy_e.gsy_e_core.util import available_simulation_scenarios, update_advanced_settings
 from gsy_e.models.config import SimulationConfig
 
-log = logging.getLogger()
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
 
 
 def decompress_and_decode_queued_strings(queued_string: bytes) -> Dict:
@@ -30,7 +31,6 @@ def launch_simulation_from_rq_job(scenario: bytes, settings: Optional[Dict],
                                   connect_to_profiles_db: bool = True):
     # pylint: disable=too-many-arguments, too-many-locals
     """Launch simulation from rq job."""
-    logging.getLogger().setLevel(logging.ERROR)
     scenario = decompress_and_decode_queued_strings(scenario)
     if isinstance(scenario, dict):
         gsy_e.constants.CONFIGURATION_ID = scenario.pop("configuration_uuid")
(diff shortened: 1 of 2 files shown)


src/gsy_e/gsy_e_core/rq_job_handler.py
```diff
@@ -26,19 +26,21 @@ def decompress_and_decode_queued_strings(queued_string: bytes) -> Dict:
...
+    if isinstance(scenario, dict):
+        gsy_e.constants.CONFIGURATION_ID = scenario.pop("configuration_uuid")
+        if "collaboration_uuid" in scenario:
+            gsy_e.constants.EXTERNAL_CONNECTION_WEB = True
+            GlobalConfig.IS_CANARY_NETWORK = scenario.pop("is_canary_network", False)
+            gsy_e.constants.RUN_IN_REALTIME = GlobalConfig.IS_CANARY_NETWORK
     saved_state = decompress_and_decode_queued_strings(saved_state)
-    log.error("Starting simulation with job_id: %s and configuration id: %s",
-              job_id, gsy_e.constants.CONFIGURATION_ID)
+    log.info("Starting simulation with job_id: %s and configuration id: %s",
```
hannesdiedrich, line 42: This will clatter our logs on the simulation pod. I remember explicitly changing the log level to error.

spyrostz (approved):
LGTM

src/gsy_e/gsy_e_core/rq_job_handler.py
hannesdiedrich, line 42: Sorry, I deleted my old comment. What I really meant is that it is indeed useful to see when a simulation was started even on production, for debugging purposes.

src/gsy_e/gsy_e_core/rq_job_handler.py
```diff
@@ -132,8 +133,8 @@ def launch_simulation_from_rq_job(scenario: bytes, settings: Optional[Dict],
                        slot_length_realtime=slot_length_realtime,
                        kwargs=kwargs)
 
-        log.error("Finishing simulation with job_id: %s and configuration id: %s",
-                  job_id, gsy_e.constants.CONFIGURATION_ID)
+        log.info("Finishing simulation with job_id: %s and configuration id: %s",
```
hannesdiedrich, line 136: The same as [here](https://github.com/gridsingularity/gsy-e/pull/1461/files#r905921874) applies here too

src/gsy_e/gsy_e_core/rq_job_handler.py
```diff
@@ -26,19 +26,21 @@ def decompress_and_decode_queued_strings(queued_string: bytes) -> Dict:
...
     # pylint: disable=too-many-arguments, too-many-locals
     """Launch simulation from rq job."""
     logging.getLogger().setLevel(logging.ERROR)
     scenario = decompress_and_decode_queued_strings(scenario)
-    gsy_e.constants.CONFIGURATION_ID = scenario.pop("configuration_uuid")
-    if "collaboration_uuid" in scenario:
-        gsy_e.constants.EXTERNAL_CONNECTION_WEB = True
-        GlobalConfig.IS_CANARY_NETWORK = scenario.pop("is_canary_network", False)
-        gsy_e.constants.RUN_IN_REALTIME = GlobalConfig.IS_CANARY_NETWORK
+    if isinstance(scenario, dict):
```
hannesdiedrich, line 35: Can you please explain why this is needed? `scenario` should be a dict after `decompress_and_decode_queued_strings`

Hassan754, line 35: What was the reason for this line? And what happens when the scenario isn't a dict, shouldn't the function break?

tests/test_rq_job_utils.py
```diff
@@ -0,0 +1,20 @@
+import pickle
+import zlib
+
+from gsy_e.gsy_e_core.rq_job_handler import launch_simulation_from_rq_job
+
+
+class TestRQJobUtils:
+    """Test RQ-related utilities."""
+    @staticmethod
+    def test_launch_simulation_from_rq_job():
```
Hassan754, line 10: Sorry I do not see any assertions here, can you help me find what exactly we're testing here

src/gsy_e/gsy_e_core/rq_job_handler.py
fievelk: Maybe we can change the log level just for this logger? I.e. we show `info` logs (so that we don't need to use `error` to see them on kibana). Not sure if there is some other filter on kibana.

src/gsy_e/gsy_e_core/rq_job_handler.py
```diff
@@ -122,8 +124,7 @@ def launch_simulation_from_rq_job(scenario: bytes, settings: Optional[Dict],
         kwargs = {"no_export": True,
                   "seed": settings.get("random_seed", 0)}
 
-        gsy_e.constants.CONNECT_TO_PROFILES_DB = True
-
+        gsy_e.constants.CONNECT_TO_PROFILES_DB = connect_to_profiles_db
```
fievelk, line 127: Why do you think we need to make this dynamic? Did you find issues with the `True` value?

tests/test_rq_job_utils.py
```diff
@@ -0,0 +1,20 @@
...
+        """Assure the launch_simulation_from_rq_job can successfully launch a simulation."""
+        launch_simulation_from_rq_job(
+            scenario=zlib.compress(pickle.dumps("default_2a")),
```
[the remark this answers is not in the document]
mrsaemir: Thanks
I'll create a ticket for that.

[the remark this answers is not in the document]
mrsaemir: @spyrostz 
https://gridsingularity.atlassian.net/browse/GSYE-337

src/gsy_e/gsy_e_core/rq_job_handler.py
mrsaemir: This is because a scenario can be `None`, a dict or a string regarding [this](https://github.com/gridsingularity/gsy-e/blob/6217ead527f630fb2082a5dc1bb0a9a9b180babd/src/gsy_e/gsy_e_core/rq_job_handler.py#L116-L122) part of the code.

mrsaemir: https://github.com/gridsingularity/gsy-e/pull/1461#discussion_r905970078

src/gsy_e/gsy_e_core/rq_job_handler.py
mrsaemir: This test is running the function without a DB. So, it fails to connect to postgres.

tests/test_rq_job_utils.py
mrsaemir: It should work without any assertions, but I modified it to explicitly fail in case of a problem.
Thanks

src/gsy_e/gsy_e_core/rq_job_handler.py
hannesdiedrich: > Maybe we can change the log level just for this logger

I guess for this we need to create a separate logger (and do not call `getLogger()`) and change the log level in a central config that does not exist ATM. An overkill IMO. 

> Not sure if there is some other filter on kibana. 
I do not think so.

src/gsy_e/gsy_e_core/rq_job_handler.py
hannesdiedrich: Thanks! Understood. I think we can remove the functionality of being able to pass a scenario name at some point. Nobody is using this. Maybe worth another hackathon ticket!?

src/gsy_e/gsy_e_core/rq_job_handler.py
fievelk: Thanks for the clarification!
I personally don't like using global constants, so I see your changes as one step in a direction I personally like :D 
At the same time, we're still using and altering a global constant. This means that (AFAIK) the value we set in this test will affect all tests that will be run after this one. If this is confirmed, it's not a good thing because it might have unexpected effects. We should find a way to restore the default value after this test has finished running. Do you confirm this makes sense?

src/gsy_e/gsy_e_core/rq_job_handler.py
hannesdiedrich: @fievelk I see your PR [here](https://github.com/gridsingularity/gsy-e/pull/1453) now and also see that I gave my LGTM 😅 . I stand by my comment above, that this info to be logged every time is valuable for debugging. There are not that many simulation started AFAIK so the danger of spamming the logs is low.

fievelk: @hannesdiedrich the PR you linked is also changing the log level of the logger:
```py
logging.getLogger().setLevel(logging.WARNING)
```
so it would still log all those messages.
The reason why I didn't merge it, in the end, is that this will spam our logs even more than now.
What I'm suggesting here is a bit different. Instead of changing the log level of the root logger (`getLogger()`) we can change it just for this module:
```py
logger = logging.getLogger(__name__)
logger.setLevel(logging.INFO)  # Or warning
```
AFAIK, this should only change the log level of this logger, not of all loggers. But we should try it.

hannesdiedrich: > What I'm suggesting here is a bit different. Instead of changing the log level of the root logger (`getLogger()`) we can change it just for this module:
> 
> ```python
> logger = logging.getLogger(__name__)
> logger.setLevel(logging.INFO)  # Or warning
> ```

Ok, sounds good to me (if it indeed works).

src/gsy_e/gsy_e_core/rq_job_handler.py
mrsaemir: Exactly as you said. Tests are failing because global variables are being changed. I'm thinking about a solution ...

src/gsy_e/gsy_e_core/rq_job_handler.py
mrsaemir: @hannesdiedrich 
https://gridsingularity.atlassian.net/browse/GSYE-339

src/gsy_e/gsy_e_core/rq_job_handler.py
mrsaemir: @fievelk , It seems that setting level on one logger with a specific name changes all loggers with the name in the program.

tests/test_rq_job_utils.py
hannesdiedrich: I also do not understand how an issue, that is linked in the ticket description will be caught by this test. Can you please explain?

mrsaemir: The `launch_simulation_from_rq_job` is responsible to start a simulation using the arguments that it receives. The important note is that, if this function fails to start a simulation, then no simulations can be launched from gsy-web.
I felt that writing a simple tests that assures this function runs without any exceptions can be helpful. What do you think?

hannesdiedrich: Yeah, I thought the issue was, that there was an unsupported keyword provided gy the gsy-web. But I found the PR that was the origin for this unittest ticket (i think): https://github.com/gridsingularity/gsy-e/pull/1451/files
So indeed, an internal bug in the `launch_simulation_from_rq_job` will be caught by this test. Thanks!

hannesdiedrich (approved):
LGTM

src/gsy_e/gsy_e_core/rq_job_handler.py
fievelk: @mrsaemir could you make an example? The `__name__` variable would contain the full module name, including namespace (e.g. `simulation.graphql.configuration_objects`) so there shouldn't be any other logger with the same name in other modules.

mrsaemir: @fievelk , I used to think that you mean something like this:
```python
logger = logging.getLogger()
logger.setLevel(logging.INFO)
```

But, then I realized you mean:

```python
logger = logging.getLogger(__name__)
logger.setLevel(logging.INFO)  # Or warning
```

Now, it's fixed as you suggested. Thanks

fievelk: Great! Thanks :) 👍

tests/test_rq_job_utils.py
```diff
@@ -11,22 +12,39 @@
...
-                aggregator_device_mapping="null",
-                saved_state=zlib.compress(pickle.dumps(None)),
-                job_id="TEST_SIM_RUNS",
-                connect_to_profiles_db=False
-            )
-        except Exception as exc: # noqa
-            pytest.fail(exc)
+        """Assure the launch_simulation_from_rq_job can successfully launch a simulation.
+        The launch_simulation_from_rq_job is ran in another process to ensure that its changes on
+        constant settings will not cause other tests to fail.
```
fievelk, line 20: I'm sort of sorry that we didn't manage to find an easier way to avoid this. I think that in the future we would benefit from avoiding globals as much as possible, to avoid convoluted situations like this.
Thanks for finding this way to overcome the testing issue 👍

fievelk (approved):

PR #1462, opened 2022-06-24
GSYE-334: Fix the delete_past_state_values of the future market strategies

Hassan754:
The bug is raised when we instantiate the future updater as TemplateStrategyUpdaterInterface because it doesn't have all the attributes of the TemplateStrategyUpdaterBase

Another solution to this problem would be to add these missing attributes to the interface, if you favor it let me know. 

### DIFF
--- a/src/gsy_e/models/strategy/future/strategy.py
+++ b/src/gsy_e/models/strategy/future/strategy.py
@@ -63,11 +63,7 @@ class FutureTemplateStrategyBidUpdater(TemplateStrategyBidUpdater):
 
     def delete_past_state_values(self, current_market_time_slot: DateTime) -> None:
         """Delete irrelevant values from buffers for unneeded markets."""
-        self.initial_rate.pop(current_market_time_slot, None)
-        self.final_rate.pop(current_market_time_slot, None)
-        self.energy_rate_change_per_update.pop(current_market_time_slot, None)
-        self.update_counter.pop(current_market_time_slot, None)
-        self.market_slot_added_time_mapping.pop(current_market_time_slot, None)
+        self._delete_market_slot_data(current_market_time_slot)
 
 
 class FutureTemplateStrategyOfferUpdater(TemplateStrategyOfferUpdater):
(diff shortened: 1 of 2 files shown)


spyrostz (commented):
One clarification, will LGTM once resolved.

src/gsy_e/models/strategy/future/strategy.py
```diff
@@ -61,6 +61,14 @@ def update(self, market: "FutureMarkets", strategy: "BaseStrategy") -> None:
                 if strategy.are_bids_posted(market.id, time_slot):
                     strategy.update_bid_rates(market, self.get_updated_rate(time_slot), time_slot)
 
+    def delete_past_state_values(self, current_market_time_slot: DateTime) -> None:
+        """Delete irrelevant values from buffers for unneeded markets."""
+        self.initial_rate.pop(current_market_time_slot, None)
+        self.final_rate.pop(current_market_time_slot, None)
+        self.energy_rate_change_per_update.pop(current_market_time_slot, None)
+        self.update_counter.pop(current_market_time_slot, None)
+        self.market_slot_added_time_mapping.pop(current_market_time_slot, None)
```
spyrostz, line 70: Sorry for nitpicking a bit here, but in order to avoid code duplication one more method could be created, called `delete_market_slot_data` , better located in `TemplateStrategyUpdaterBase`, that will include all these `pop()` calls and will be responsible for popping one market slot from all dicts. That way you will be able to reuse it in `FutureTemplateStrategyOfferUpdater.delete_past_state_values`, `FutureTemplateStrategyBidUpdater.delete_past_state_values`, but also in `TemplateStrategyUpdaterBase.delete_past_state_values` (which also could benefit from this). Thanks for bearing with me, I think that this will improve the code reuse a lot!

Hassan754: Thanks for the proposal! it should now be addressed and looking much better.

spyrostz (approved):
LGTM, thanks for the fix!

hannesdiedrich (approved):
LGTM

PR #1463, opened 2022-06-27
Feature/gsye 295 Forward markets

hannesdiedrich:

We want to add forward markets that enable trading ahead of the future markets time frame.
Copy paste from the requirements in the tickets:

exchange | start | length | slot_length | matching algorithm | clearing interval
-- | -- | -- | -- | -- | --
yearly | one year in advance | 5 years | 1 year | 2-sided pay as clear | once per half-year (1.1./ 1.7.)
monthly | one month in advance | 2 years | 1 month | 2-sided pay as clear | every month (first day)
weekly | one week in advance | 1 year | 1 week | 2-sided pay as clear | every week (Monday)
daily | one day in advance | 1 day | 1 hour | 2-sided pay as bid | every hour
intraday | 15 mins in advance | 1 day | 15 mins | 2-sided pay as bid | continuously (every tick)

## Reproduction steps

- Create a new setup (e.g. `setups/forward_market.py`). I personally copied `default_2a.py` and just added the following lines at the top:
    ```py
    from gsy_framework.constants_limits import ConstSettings, SpotMarketTypeEnum
    def get_setup():
        ConstSettings.ForwardMarketSettings.ENABLE_FORWARD_MARKETS = True
        ConstSettings.MASettings.MARKET_TYPE = SpotMarketTypeEnum.TWO_SIDED.value
        # The rest stays the same
    ```
- Run gsy-e. We explicitly set a duration of 6 years (or more than 5 years anyway) to make sure that the forward markets work correctly.
    ```sh
    gsy-e run --setup forward_market --seed 0 --no-export --duration 2190d
    ```

## Proposed changes

- Adapte FutureMarket implementation so that the time frame could be set at creation: 76b43d5845c39827f86edb5d5a48ae81552860f1
- Refactor Future market settings in ConstSettings (change in gsy-framework) 9430f440b6d655e9abec9cd7cf11eb45ecc671cb
- Move AvailableMarketTypes into gsy_e_core package in order to circumvent circular imports when implementing forward markets e46341af5950db041bb8f2dcac4b8dc28c801f49
-  Improve communication between myco and markets by useing Available market types: 563ec541233b51fb68b8b1855cfa0cbe80468ce0
- Add forward markets, rotators and adaption in the internal myco matchers e5dc6b371b72b3c9eef92c619be2004d160dfb96
- Add first tests for the creation and rotation of the forward markets: 9b9c83cef81e4e81e437526cdeefddb817ef7a90
- Adapt implementation to changed requirements (mainly clearing frequency and matching algorithms): last commits

TODO:
- add more unit tests, especially for the MarketCounters 
- add a cron-alike configuration in `constants.py` that enables us to easily re-configure the rotation and clearing frequency of the forward markets (should be done in a separate ticket)

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-295


src/gsy_e/gsy_e_core/market_counters.py
```diff
@@ -0,0 +1,113 @@
...
+
+class MarketCounter:
+    """Base class for market counters"""
+
+    def __init__(self, clearing_interval_min: int):
+        self._last_clearing_time = None
+        self.clearing_interval = clearing_interval_min
+
+    def is_time_for_clearing(self, current_time: DateTime) -> bool:
+        """Return if it is time for clearing according to self.clearing_interval."""
```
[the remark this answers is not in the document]
hannesdiedrich: > Good that you kept the default implementation here for the existing markets! I think that we will revisit this approach in the future since the cron-based approach that you used for the other markets is much more easy to reason about, and it probably makes sense to make it the default behavior for the existing markets too. I would not do it now though, but at a later stage, in order to resolve any issues that might pop up with the new approach. I opened already an enhancement ticket that deals with this:
> https://gridsingularity.atlassian.net/browse/GSYE-342
> and leave it in the backlog for now.

Thanks! This will also make https://gridsingularity.atlassian.net/browse/GSYE-322 possible. I will linkt the two tickets if you do not mind.

> One more remark, we could also use this approach for the update_frequency module, in order to solve the bugs that we have encountered in the price update of the strategies by using this approach. IMO it will lead to far less uncertainties and a more robust and performant algorithm than the tick-counting approach that we use now.

👍  I will also link this ticket: https://gridsingularity.atlassian.net/browse/GSYE-332

src/gsy_e/models/market/future.py
```diff
@@ -178,33 +178,46 @@ def _expire_orders(self, orders: "FutureOrders", current_market_time_slot: DateT
...
+                               config: "SimulationConfig"):
+        future_time_slot = start_time
+        while future_time_slot <= end_time:
             if (future_time_slot not in self.slot_bid_mapping and
                     is_time_slot_in_simulation_duration(future_time_slot, config)):
                 self.bids.slot_order_mapping[future_time_slot] = []
                 self.offers.slot_order_mapping[future_time_slot] = []
-            future_time_slot = future_time_slot.add(minutes=slot_length.total_minutes())
+            future_time_slot = future_time_slot.add(
+                minutes=self._adapt_market_length(market_length, future_time_slot).total_minutes())
```
spyrostz, line 208: Since I went here from the leap year calculation, can you explain this a bit more? As far as I see, the time slot duration should be strictly one month and it is not possible by converting the time slot to minutes. Is it really important to have a constant slot length here, and is it guaranteed that this will fail if the slot length changes from one slot to another (e.g. from one month that has 30 days to another one that has 31)?

hannesdiedrich: Yes, you are absolutely right, it is not important, that the slot length is constant e.g. for the leapyears.
I need to overwrite the `_adapt_market_length` for the MonthForwardMarket, too, in order to account for the different month length. Thanks for catching this!

hannesdiedrich: I removed the `_adapt_market_length` and put the logic in the `_get_market_slot_duration` here f242c73c5b699e83477ef635188acdafda3eb4f0

Also found some things to adapt along the way, e.g. the weekahead market should only start at a Monday.

spyrostz (approved):
LGTM

src/gsy_e/models/myco_matcher/myco_internal_matcher.py
```diff
@@ -32,16 +35,29 @@ class MycoInternalMatcher(MycoMatcherInterface):
...
+            AvailableMarketTypes.WEEK_FORWARD: PayAsClearMatchingAlgorithm,
+            AvailableMarketTypes.MONTH_FORWARD: PayAsClearMatchingAlgorithm,
+            AvailableMarketTypes.YEAR_FORWARD: PayAsClearMatchingAlgorithm,
+        }
+        self._forward_market_counters = {
+            AvailableMarketTypes.DAY_FORWARD: DayForwardMarketCounter,
+            AvailableMarketTypes.WEEK_FORWARD: WeekForwardMarketCounter,
+            AvailableMarketTypes.MONTH_FORWARD: MonthForwardMarketCounter,
+            AvailableMarketTypes.YEAR_FORWARD: YearForwardMarketCounter
+        }
```
fievelk, line 50: If I get it right, the complication here is that the Myco matcher doesn't apply just one algorithm to all situations anymore (as it did before). Now it can instead use different algorithms depending on the type of exchange (spot/future/day-forward/month-forward/etc.). Is this correct?

As I mentioned in one of the previous PRs, I still think that this kind of change complicates too much the interface. Before, we only had `matcher.match_algorithm`, while now we've branched the logic of this class for two different scenarios ("normal" vs "forward" markets). This branching makes the class much more complex to understand and use in my opinion.
In fact, you can already see how the `match_recommendations()` method became way more complicated: we're passing a callable to it, and sometimes this callable is inside the class itself, while some other times it comes from a different class/algorithm.

I think that we should unify these two logics. Either we always use `self.match_algorithm` or we completely replace that and integrate it with the new logic you're implementing here. Ideally, there should be one way to deal with all types of "market types" and matching algorithms, without creating ad-hoc separate variables and logic for each of them.

I'm sorry if I can't be much more specific than this: as you see, there are several parts of the code involved and it's challenging to get a grasp of it without working on the code. But maybe there's another class emerging here and asking to be implemented.

P.S.: The current solution might also complicate the situations in which we're currently calling `matcher.match_algorithm.state`, because this doesn't work for future/forward matching algorithms.

fievelk (commented):
I've left some comments. There's especially a long comment regarding the Myco matcher, which I think should be addressed. In general, however, I think the changes are good and well structured. Thanks for working on this!

src/gsy_e/models/market/forward.py
```diff
@@ -0,0 +1,134 @@
...
+    @staticmethod
+    def _get_market_slot_duration(current_time: DateTime, _config) -> duration:
+        """Return duration of market slots inside the market block."""
+
+    def create_future_markets(self, current_market_time_slot: DateTime,
+                              config: "SimulationConfig") -> None:
+        if not ConstSettings.ForwardMarketSettings.ENABLE_FORWARD_MARKETS:
+            return
+        self._create_future_markets(self._get_start_time(current_market_time_slot),
+                                    self._get_end_time(current_market_time_slot), config)
```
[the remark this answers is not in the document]
hannesdiedrich: :-) It is not particularly inaccurate and I will rename as you suggested, if it makes reading the code easier.

src/gsy_e/models/myco_matcher/myco_internal_matcher.py
```diff
@@ -59,34 +75,54 @@ def get_matching_algorithm():
...
-                            time_slot: {**orders_data, "current_time": area_data["current_time"]}
-                            for time_slot, orders_data in orders.items()}}
-                    bid_offer_pairs = self._get_matches_recommendations(data)
-                    if not bid_offer_pairs:
-                        break
-                    trades_occurred = market.match_recommendations(bid_offer_pairs)
-                    if not trades_occurred:
-                        break
+                markets.append(area_data[AvailableMarketTypes.FUTURE])
+            self._match_recommendations(area_uuid, area_data, markets,
```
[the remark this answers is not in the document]
hannesdiedrich: Hace a look 2 lines below, where `_match_forward_recommendations` is called. In there I handle the passing the markets into `self._match_recommendations` differently.

src/gsy_e/models/myco_matcher/myco_internal_matcher.py
hannesdiedrich: > If I get it right, the complication here is that the Myco matcher doesn't apply just one algorithm to all situations anymore (as it did before). Now it can instead use different algorithms depending on the type of exchange (spot/future/day-forward/month-forward/etc.). Is this correct?

Correct!

For the rest of your comment, I hear you and was also thinking about refactoring. But I first wanted to present a first implementation. IMO the "new class" approach makes sense. I will try to incorporate the `matcher.match_algorithm.state` issue, thangs for pointing me to it.

src/gsy_e/models/market/forward.py
```diff
@@ -0,0 +1,134 @@
...
+                         readonly=readonly, grid_fee_type=grid_fee_type,
+                         grid_fees=grid_fees, name=name)
+
+    @staticmethod
+    def _get_start_time(current_time: DateTime) -> DateTime:
+        """Return time when the market block starts."""
+
+    @staticmethod
+    def _get_end_time(current_time: DateTime) -> DateTime:
+        """Return time when the market block ends."""
```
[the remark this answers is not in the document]
hannesdiedrich: https://github.com/gridsingularity/gsy-e/pull/1463/commits/ffb9033579471ab318f6ac8b83283ebc90b9100a

src/gsy_e/models/market/forward.py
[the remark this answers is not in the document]
hannesdiedrich: https://github.com/gridsingularity/gsy-e/pull/1463/commits/265b36d581a9995f0263c8c87835be66cb981d90

src/gsy_e/models/myco_matcher/myco_internal_matcher.py
hannesdiedrich: I moved the forward market matching in its own class, that is initiated by the global singleton `bid_offer_matcher`. This will also only provided with the orders from the forward markets from the area, too.
I am not sure whether this is the most pretty solution, but it divides responsibilities better than before.
https://github.com/gridsingularity/gsy-e/pull/1463/commits/d079dc44ffc6167aa9b54c152b1d18726d00e6dc

We only have to deal with the `matcher.match_algorithm.state` once we want to plot the supply and demand curves for the future markets, so I suggest to deal with this later. For now there is no urge AFAIS.

fievelk (approved):
Thank you! 🎉

fievelk (dismissed):

src/gsy_e/models/market/forward.py
```diff
@@ -0,0 +1,138 @@
...
+    def _get_start_time(current_time: DateTime) -> DateTime:
+        return current_time.set(month=1, day=1, hour=0, minute=0).add(years=2)
+
+    @staticmethod
+    def _get_end_time(current_time: DateTime) -> DateTime:
+        return current_time.set(month=1, day=1, hour=0, minute=0).add(years=6)
+
+    @staticmethod
+    def _get_market_slot_duration(current_time: DateTime, _config) -> duration:
+        return duration(years=1, days=1) if current_time.is_leap_year() else duration(years=1)
```
mrsaemir, line 138: The DateTime object is absolutely able to handle leap years on its own. There's definitely no need for us to handle them or the result will be unexpected:

Screenshot from 2022-07-13 08-05-33

src/gsy_e/models/market/forward.py
```diff
@@ -0,0 +1,138 @@
...
+        return current_time.set(day=1, hour=0, minute=0).add(months=1)
+
+    @staticmethod
+    def _get_end_time(current_time: DateTime) -> DateTime:
+        return current_time.set(day=1, hour=0, minute=0).add(years=2)
+
+    @staticmethod
+    def _get_market_slot_duration(current_time: DateTime, _config) -> duration:
+        number_of_days_month = monthrange(current_time.year, current_time.month)[1]
+        return duration(days=number_of_days_month)
```
mrsaemir, line 122: pendulum.DateTime is smart enough to know about number of days in each month, thus:
```python
return duration(months=1)
```

src/gsy_e/models/market/forward.py
```diff
@@ -0,0 +1,138 @@
...
+class MonthForwardMarket(ForwardMarketBase):
+    """Month forward market implementation"""
+
+    @staticmethod
+    def _get_start_time(current_time: DateTime) -> DateTime:
+        return current_time.set(day=1, hour=0, minute=0).add(months=1)
+
+    @staticmethod
+    def _get_end_time(current_time: DateTime) -> DateTime:
+        return current_time.set(day=1, hour=0, minute=0).add(years=2)
```
mrsaemir, line 117: Change the order of `.set` and `.add` as discussed below in `YearForwardMarket`. Thanks

src/gsy_e/models/market/forward.py
```diff
@@ -0,0 +1,138 @@
...
+    def _get_market_slot_duration(_current_time: DateTime, _config) -> duration:
+        return duration(weeks=1)
+
+
+class MonthForwardMarket(ForwardMarketBase):
+    """Month forward market implementation"""
+
+    @staticmethod
+    def _get_start_time(current_time: DateTime) -> DateTime:
+        return current_time.set(day=1, hour=0, minute=0).add(months=1)
```
mrsaemir, line 113: Change the order of `.set` and `.add` as discussed below in `YearForwardMarket`. Thanks

src/gsy_e/models/market/forward.py
```diff
@@ -0,0 +1,138 @@
...
+    """Week forward market implementation"""
+
+    @staticmethod
+    def _get_start_time(current_time: DateTime) -> DateTime:
+        days_until_next_monday = 7 - (current_time.day_of_week - 1)
+        return current_time.set(hour=0, minute=0).add(days=days_until_next_monday)
+
+    @staticmethod
+    def _get_end_time(current_time: DateTime) -> DateTime:
+        return current_time.set(hour=0, minute=0).add(years=1)
```
mrsaemir, line 101: Change the order of `.set` and `.add` as discussed below in `YearForwardMarket`. Thanks

src/gsy_e/models/market/forward.py
```diff
@@ -0,0 +1,138 @@
...
+        self._create_future_market_slots(self._get_start_time(current_market_time_slot),
+                                         self._get_end_time(current_market_time_slot), config)
+
+
+class DayForwardMarket(ForwardMarketBase):
+    """Day forward market implementation"""
+
+    @staticmethod
+    def _get_start_time(current_time: DateTime) -> DateTime:
+        return current_time.set(hour=0, minute=0).add(days=1)
```
mrsaemir, line 80: Change the order of `.set` and `.add` as discussed below in `YearForwardMarket`. Thanks

src/gsy_e/models/market/forward.py
```diff
@@ -0,0 +1,138 @@
...
+
+    @staticmethod
+    def _get_end_time(current_time: DateTime) -> DateTime:
+        return current_time.set(hour=0, minute=0).add(weeks=1, days=1, hours=-1)
```
mrsaemir, line 84: Change the order of `.set` and `.add` as discussed below in `YearForwardMarket`.
Also I believe the end time should be like this:
```python
return current_time.add(days=2).set(hour=0, minute=0)
````

**Review by mrsaemir — COMMENTED** [review]
Beautifully implemented! Thanks.
There are just a couple of minor issues in `models/market/forward.py`. 
Also it seems that [ExternalMixin._get_market_from_time_slot](https://github.com/gridsingularity/gsy-e/blob/07a0c4500d626ba5b07cfa6404378216f44323ef/src/gsy_e/models/strategy/external_strategies/__init__.py#L307-L319) should be updated to adapt forward markets.

File: src/gsy_e/models/market/future.py [hunk 14]
```diff
@@ -178,33 +178,44 @@ def _expire_orders(self, orders: "FutureOrders", current_market_time_slot: DateT
...
+    @staticmethod
+    def _get_market_slot_duration(_current_time: DateTime, config: "SimulationConfig") -> duration:
+        return config.slot_length
+
+    def _create_future_market_slots(
+            self, start_time: DateTime, end_time: DateTime, config: "SimulationConfig") -> None:
+        future_time_slot = start_time
+        while future_time_slot <= end_time:
             if (future_time_slot not in self.slot_bid_mapping and
                     is_time_slot_in_simulation_duration(future_time_slot, config)):
```
fievelk, line 203: At the moment, it looks like this check never returns `True`. This is the method in gsy-framework: https://github.com/gridsingularity/gsy-framework/blob/bd4851c95241904e2c30788dbb0e34e78903dd2e/gsy_framework/utils.py#L552-L570

For our simulation, this check returns `False` because `config.end_date` is just 1 day in the future (it uses the default `GlobalConfig.sim_duration` value). So the following check is always false:
```py
start_date <= time_slot < end_date
```

src/gsy_e/models/market/forward.py
```diff
@@ -0,0 +1,138 @@
...
+        number_of_days_month = monthrange(current_time.year, current_time.month)[1]
+        return duration(days=number_of_days_month)
+
+
+class YearForwardMarket(ForwardMarketBase):
+    """Year forward market implementation"""
+
+    @staticmethod
+    def _get_start_time(current_time: DateTime) -> DateTime:
+        return current_time.set(month=1, day=1, hour=0, minute=0).add(years=2)
```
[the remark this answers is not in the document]
fievelk: @spyrostz can we have your feedback on this?
If today is `2021-01-01 00:00`, when should the "year-forward" market start? From what I read in the docs, I'd say it should start on `2027-01-01 00:00` (1 year gap + 5 years duration). Is this correct?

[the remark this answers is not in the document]
spyrostz: > Based on the description of the task, the yearly market should start 1 year in advance and should last for 5 years. With these in mind, Given the current time is: 2021-01-01 00:00, the _get_start_time function will return 2023-01-01 00:00 which is more than 1 years ahead.

This is the correct implementation as far as I see. The yearly market does not operate on a rolling-window fashion, but on calendar years. Therefore the start time of a market, no matter when it opens, should be at 01-01-202*. As for why there is no open market for 01-01-2022, I am copying from the requirements document:

`For a yearly product, the market opens 5 years in advance and closes 1 year before 1st day of the year of T0. This means that if you are to sell a product now, you can only sell for 2024-2028 because the yearly period starts on the 1st of the calendar year.`

Therefore I think that there is no problem with this part of the code and it adheres to the requirements.

src/gsy_e/models/market/forward.py
spyrostz: I do not really believe that, because different months have different number of days. What is probably misleading you here is the `current_time`, which is not really the wall clock time, but the `current_simulation_time`, which will coincide with the wall clock for CNs, but will not for simulations (e.g. I can start a simulation now for February 2019, and the function will return 31 days, even though the intended result of the method would be 28/29 days)

src/gsy_e/models/market/forward.py
spyrostz: For the latter it looks correct IMO. Again copying from the documentation:

`For hourly products, the market opens one week in advance and closes one hour in advance.`

src/gsy_e/models/market/forward.py
spyrostz: @mrsaemir I am not sure about that, check out this output from ipython:

```
In [1]: from pendulum import duration

In [2]: d = duration(months=1)

In [3]: d.days
Out[3]: 30
```

fievelk: I think that using `__add__` (`+`) between `DateTime` and `duration` takes days into account:
```py
> january
DateTime(2021, 1, 1, 0, 0, 0)
> january + duration(months=1)
DateTime(2021, 2, 1, 0, 0, 0)
> february = DateTime(2021, 2, 1)
> february.days_in_month
28
> february + duration(months=1)
DateTime(2021, 3, 1, 0, 0, 0)
>
DateTime(2021, 3, 1, 0, 0, 0)
```

mrsaemir: Also, this is part of the implementation for .add() method for pendulum.DateTime objects:
```python
    if months:
        month += months
        if month > 12:
            year += 1
            month -= 12
        elif month < 1:
            year -= 1
            month += 12

    day = min(DAYS_PER_MONTHS[int(is_leap(year))][month], dt.day)
    ```

Comment (spyrostz) [reply src/gsy_e/models/market/forward.py:122]: What you suggest implies though that this duration will only be used in adding / subtracting to datetime values, which I am not sure is the case here. Should be the case indeed, however it needs to be checked whether it indeed works the same way as with days (which we know that works even if the duration is converted to days int - which I suspect is the reason that it is implemented like this here). 

If it indeed is the case then all good, however I wonder whether you noticed an issue somewhere due to this calculation. Regardless of that I do not mind if this discussion is theoretical, or more focused in order to use pendulum here too, however if this really causes an issue it would be great to also point it out to see what causes it.

Comment (fievelk) [reply src/gsy_e/models/market/forward.py:122]: > Edit: meanwhile Pierpaolo confirmed my suspicion:
> 
> ```
> future_time_slot.add(minutes=self._get_market_slot_duration(future_time_slot, config).total_minutes())
> ```
> 
> this will not work with `duration(month=1)` because this will always return 30 days and will not take into account the future_time_slot.

Regarding this, my suggested solution would be to avoid using `.add()` and use
```py
future_time_slot + self._get_market_slot_duration(future_time_slot, config)
```
This will take the actual date of `future_time_slot` and add the correct duration to it (e.g. 1 month, automatically adjusting the number of days).

src/gsy_e/models/market/forward.py
fievelk: @spyrostz How did we decide the `.add(days=1)` here? 

AFAIU, if the daily market is on the `2022-07-28` (all day), it should open on the  `2022-07-21T00:00` (one week in advance) and it should close on the `2022-07-27T23:00:00` (one hour before the first market slot). But how did we decide to create these future market slots exactly 1 day in advance?

src/gsy_e/models/market/forward.py
mrsaemir: So, to recap and resolve this conversation, are we going to return `duration` objects in `_get_market_slot_duration` methods and use the `+` operator instead of the `.add` method?
@spyrostz @fievelk

src/gsy_e/models/market/forward.py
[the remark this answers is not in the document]
mrsaemir: @spyrostz , I got your point but what the code returns doesn't match the years you mentioned:
> Hence, the market responsible for the 2026 energy product remains open for 2021, 2022, 2023 and 2024
Screenshot from 2022-07-20 11-31-20

fievelk:
I would vote for this solution, yes (using pendulum duration and +).

Il mer 20 lug 2022, 11:09 Amirhossein Saemi ***@***.***> ha
scritto:

src/gsy_e/models/market/forward.py
fievelk: I've just pushed one commit that changes that logic: 54ec1320d. If this is not optimal, please feel free to revert it in the next days.

src/gsy_e/models/market/forward.py
spyrostz: @fievelk I think that we have more or less discussed the same thing in other comments too, but IMO this implementation is correct. All day-ahead (hourly) markets for one day (2022-07-28 in your example) should open 7 days in advance and close one hour before the day. If we did not have the `add(days=1)` here, each hourly market for the same day would start at a different time and not at exactly 00:00 as the requirements dictate.

src/gsy_e/models/area/markets.py
```diff
@@ -98,9 +98,7 @@ def update_area_market_id_lists(self) -> None:
 
     def activate_future_markets(self, area: "Area") -> None:
         """Wrapper for activation methods for all future market types."""
-        if ConstSettings.ForwardMarketSettings.ENABLE_FORWARD_MARKETS:
-            self._activate_forward_markets(area)
```
spyrostz: Is this check applied somewhere else that I do not recall ? Otherwise I would recommend to reinstate it since the forward markets should not be activated by default yet

spyrostz (approved):
LGTM

mrsaemir (approved):
LGTM

src/gsy_e/models/area/markets.py
spyrostz: I see that the _activate_forward_markets includes a similar check therefore this is indeed redundant here.

src/gsy_e/models/market/forward.py
fievelk: @spyrostz I'm not sure we're on the same topic here. My question would basically be: why do we use `.add(days=1)` and not (for example) `.add(days=5)` (just random)? This should just decide how much in advance we create the market slots, but AFAIU it doesn't affect the duration of the open market. So I just wanted to double check: if the behaviour is the same, this `.add(days=1)` could be anything >=1, right?

spyrostz: @fievelk now I got your question, in theory you are right, we could even have `.add(days=12243)` and it would still work as expected.

PR #1475, opened 2022-07-07
Feature/gsye 258

spyrostz:

Added support for generating simulation results for the SCM. Backend csv files and plots have been adapted (the ones that are compatible) to be generated after an SCM simulation run. Also endpoint buffer and the framework simulation results have been adapted.

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-258


src/gsy_e/models/area/scm_manager.py
```diff
@@ -196,81 +304,122 @@ def calculate_home_energy_bills(
...
+                home_data.energy_need_kWh * market_maker_rate_normal_fees + marketplace_fee +
+                fixed_fee - home_data.energy_surplus_kWh * feed_in_tariff)
+        home_bill.base_energy_bill = base_energy_bill
 
         if home_data.allocated_community_energy_kWh > home_data.energy_need_kWh:
             if home_data.energy_surplus_kWh > 0.0:
-                gsy_energy_bill = -(
-                    home_data.energy_bought_from_community_kWh * market_maker_rate_decreased_fees +
-                    self.community_data.energy_sold_to_grid_kWh * feed_in_tariff)
+                home_bill.set_bought_from_community(
```
cruky, line 328: I know it's not ready yet, but I cannot understand the concept of calculating the gsy bill for houses which have surplus and sell energy to community. Here in the loop I don't think so that we will ever have `energy_bought_from_community_kWh`, because house has energy surplus and was allocated with more energy than it needed. The another thing is that I do not see the method like `set_sold_to_community` here, which I suppose should be introduced to pay this house for it's contribution to community, and not only what it gave to the grid. But I am not sure if I get it right in the first place, so help me here.

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
```diff
@@ -245,3 +252,121 @@ def _update_plot_stats(self, area: Area) -> None:
...
+        if isinstance(self._area.strategy, SCMStorageStrategy):
+            s = self._area.strategy.state
+            return [[slot,
+                    s.charge_history_kWh[slot],
+                    s.offered_history[slot],
+                    s.charge_history[slot]]]
+        elif isinstance(self._area.strategy, (SCMLoadHoursStrategy, SCMLoadProfile)):
+            desired = self._area.strategy.state.get_desired_energy_Wh(slot) / 1000
+            # All energy is traded in SCM
+            return [[self._area.past_market_time_slot, desired, 0.]]
```
hannesdiedrich, line 367: Why adding the `deficit` column then? As you have a separate class, I suggest to remove it then. 
The same applies to the `not_sold` in the PV strategy, right?

src/gsy_e/models/area/scm_manager.py
```diff
@@ -196,81 +304,122 @@ def calculate_home_energy_bills(
...
-
-        self._bills[home_uuid] = HomeEnergyBills(
-            base_energy_bill=base_energy_bill,
-            gsy_energy_bill=gsy_energy_bill,
-            savings=(base_energy_bill - gsy_energy_bill),
-            savings_percent=savings_percent)
+        self._bills[home_uuid] = home_bill
+
+    def get_area_results(self, area_uuid: str, serializable: bool = False) -> Dict:
+        """Return the SCM results for one area (and one time slot)."""
```
hannesdiedrich, line 384: `and one time slot` means that these are the accumulated results until the current time slot. Right? 
(maybe worth to mention in the docstring IMO)

hannesdiedrich (commented):
Left some minor comments, LGTM in general. Thanks!

src/gsy_e/models/area/scm_manager.py
spyrostz: > Here in the loop I don't think so that we will ever have energy_bought_from_community_kWh, because house has energy surplus and was allocated with more energy than it needed.

This is not true actually. In the community there will be homes for which the supply is not enough and also homes that might have no supply at all (homes without PVs). These homes will still benefit by participating in the community by buying energy with decreased grid fees. 

> The another thing is that I do not see the method like set_sold_to_community here, which I suppose should be introduced to pay this house for it's contribution to community, and not only what it gave to the grid.

Admittedly the set_sold_to_community method is not there, the reason being because it is not used in the example calculations that we had from the research team https://docs.google.com/spreadsheets/d/1CS3jFzxv_QLw6aXr-TvwLqoe9LKqoUZb5ckiECHs0kk/edit#gid=0 . And it makes sense, because there will be no revenue at all from the "sold to community" trades for the producers.

spyrostz: > > Here in the loop I don't think so that we will ever have energy_bought_from_community_kWh, because house has energy surplus and was allocated with more energy than it needed.
> 
> This is not true actually. In the community there will be homes for which the supply is not enough and also homes that might have no supply at all (homes without PVs). These homes will still benefit by participating in the community by buying energy with decreased grid fees.
> 
> > The another thing is that I do not see the method like set_sold_to_community here, which I suppose should be introduced to pay this house for it's contribution to community, and not only what it gave to the grid.
> 
> Admittedly the set_sold_to_community method is not there, the reason being because it is not used in the example calculations that we had from the research team https://docs.google.com/spreadsheets/d/1CS3jFzxv_QLw6aXr-TvwLqoe9LKqoUZb5ckiECHs0kk/edit#gid=0 . And it makes sense, because there will be no revenue at all from the "sold to community" trades for the producers, as far as I understood from the regulation at least. The only benefit of the community is in the form of decreased grid fees.

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -776,14 +837,17 @@ def restore_global_state(self, saved_state: dict) -> None:
 class CoefficientSimulation(Simulation):
     """Start and control a simulation with coefficient trading."""
 
+    area: "CoefficientArea"
+    _results: CoefficientSimulationResultsManager
```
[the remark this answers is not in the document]
spyrostz: Yes, but also in order to be able for methods that use the self._results to detect the type hints correctly.

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -776,14 +837,17 @@ def restore_global_state(self, saved_state: dict) -> None:
...
+
     def _init(self, redis_job_id: str) -> None:
         # has to be called before load_setup_module():
         global_objects.profiles_handler.activate()
 
         self.area = self._setup.load_setup_module()
 
         self._results.init_results(redis_job_id, self.area, self._setup)
-        self._results.update_and_send_results(
+        self._results.update_send_coefficient_results(
```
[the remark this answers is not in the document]
spyrostz: The problem is that they do not have exactly the same interface. `update_send_coefficient_results` also requires the scm_manager object in order to calculate the results. Due to the fact that the 2 methods have different signatures / number of arguments, method overriding would not be a possibility (it would be, but pylint complains about overriding a method with different number of arguments). So, and in order to avoid using *args and **kwargs as arguments (which would make the method overriding pylint-compatible, but it is much worse IMO) I decided to use another method name, and deal with code duplication. Hope the above is clear let me know if it does not make sense somewhere.

src/gsy_e/gsy_e_core/sim_results/results_plots.py
```diff
@@ -0,0 +1,610 @@
+from collections import namedtuple
```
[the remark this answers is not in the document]
spyrostz: Will do and thanks for the reminder!

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
spyrostz: Admittedly I added both in order to have exactly the same generated files for both SCM and normal exchanges, however since you prefer to omit this I will do so, thanks!

src/gsy_e/models/area/scm_manager.py
spyrostz: No, these are results just from one time slot. Accumulation happens in gsy-framework (essentially before the simulation results are written to the database)

src/gsy_e/gsy_e_core/simulation.py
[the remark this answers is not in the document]
hannesdiedrich: 🤦  Sorry, I over-read passing the `scm_manager`. Ok, understood. Thanks!

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
hannesdiedrich: Ok, makes sense. 
Maybe others beg to differ, so maybe you can hold off this for now and wait for other opinions?

src/gsy_e/models/area/scm_manager.py
hannesdiedrich: Ah yes, the ScmManager is re-instantiated every market slot. Thanks! for the explanation!

src/gsy_e/models/area/scm_manager.py
cruky: Thanks. I am aware that there will be different homes, but I was referring to the scenario where `allocated_community_energy_kWh` > `energy_need_kWh` and `energy_surplus_kWh` > 0.0 , so the homes with no supply or with not enough supply won't be evaluated in this if statement. And that's why I think that there is no need for `set_bought_from_community`. The second issue, I found in the [doc](https://docs.google.com/document/d/1TeykhSzA952U8BtqvpqbVfpH5-_dzIQhCKxSykBv5G4/edit#) that the formula for the `GSy energy bill = - (total energy sold to the community * Community price + total energy sold to the grid * grid price) + PENALTY`. Which in my opinion makes sense. And according to spreadsheet, there is **GSy energy bill**  `R15`, which is calculated with use of `I21` and `J21` (`-(($I$21-I18)*(K18+M18*O18)+$J$21*L18)` it works only because there is only one house who is selling to the community, probably with more such houses, the formula should be revisit). So AFAIU the bill considers revenue for the energy sold to the community. But of course I could be totally wrong.

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -419,16 +402,94 @@ def update_csv_on_market_cycle(self, slot_no: int, area: "Area") -> None:
...
+                        progress_info.current_slot_str,
+                        self._endpoint_buffer.flattened_area_core_stats_dict
+                    )
+
+                self._export.file_stats_endpoint(area, scm_manager)
+
+
+def simulation_results_manager_factory():
+    if ConstSettings.MASettings.MARKET_TYPE == SpotMarketTypeEnum.COEFFICIENTS.value:
+        return CoefficientSimulationResultsManager
```
cruky, line 489: AFAIS we set the `MARKET_TYPE` variable in the setup file, which will not be evaluated until the `_init` method with `load_setup_module`. The current order is:
1. `run_simulation` with `simulation_class_factory` 
2.  `Simulation`.`__init__` with `simulation_results_manager_factory`
3.  `_init` with `load_setup_module` when we assign MARKET_TYPE within the setup file
So we probably need to introduce a new parameter for cli to include the coefficient type or use the --settings-file instead, which probably will be less work than the first option.

spyrostz: I've been thinking about this a lot in the past too, that maybe adding the market type in the CLI arguments is probably the most convenient choice here. Let me quickly add it.

src/gsy_e/models/area/scm_manager.py
spyrostz: > but I was referring to the scenario where allocated_community_energy_kWh > energy_need_kWh and energy_surplus_kWh > 0.0 , so the homes with no supply or with not enough supply won't be evaluated in this if statement. And that's why I think that there is no need for set_bought_from_community.

This case is not exactly true. If a house has energy_surplus > 0, it is guaranteed that its energy_need is 0. This is justified because the energy_need does not include self-consumed energy, meaning that each house first covers its own consumption with its own production and then evaluates whether it has a need or a surplus. The energy assignment to each house is independent from this mechanism, therefore it really needs to be there no matter if a house has energy need or energy surplus. 

> And according to spreadsheet, there is GSy energy bill R15, which is calculated with use of I21 and J21 (-(($I$21-I18)*(K18+M18*O18)+$J$21*L18) it works only because there is only one house who is selling to the community, probably with more such houses, the formula should be revisit). So AFAIU the bill considers revenue for the energy sold to the community. 

Here you do have a point though as far as I see. This formula does not make sense IMO too (because as you said, I21 is the total energy bought from the community , therefore I21 - I18 is the energy bought from the community from every house other than this one. Therefore for multiple houses, it will pay them more than the energy that they have sold to the grid). Same goes for the energy excess sold to the grid, with this formula each house is paid for the whole community excess (J21). I think that the best solution here is to change the formula to this `(-(I18*(K18+M18*O18) + J18*L18)`. If you agree I will reflect this change to the code too.

cruky: I think that I get it now, thank you for your clarification. Yes, I agree to change the formula. Nevertheless I still miss the `total energy sold to the community * Community price` in our calculations, something like `sold_to_community` will work IMO, don't we need it ?

cruky: This is true, that requirements are not clear here. I only see that energy sold to the community is in this [doc](https://docs.google.com/document/d/1TeykhSzA952U8BtqvpqbVfpH5-_dzIQhCKxSykBv5G4/edit#), but you are right, there is no algorithm distinguishing from grid vs community energy. Your proposition looks great for me, and is probably fair for the parties. Sorry for dragging this discussion, but I think that we are on the same page now. Thank you!

spyrostz: Great to hear that we are in sync now ! I adapted the code to support what was discussed here

mrsaemir (approved):
LGTM, Thanks.

PR #1476, opened 2022-07-08
Bump mkdocs version and related packages

fievelk:

Running mkdocs currently raises the following error:
```sh
AttributeError: module 'jinja2' has no attribute 'contextfilter'
```

This happens because the old mkdocs version is incompatible with jinja2 3.1.0 release. See:
- https://github.com/mkdocs/mkdocs/issues/2799

## Proposed changes

- Update mkdocs.


Hassan754 (approved):

hannesdiedrich (approved):
LGTM

PR #1477, opened 2022-07-13
Remove unused methods for settlement orders

fievelk:

These methods were introduced in https://github.com/gridsingularity/gsy-e/pull/1214/ and they were already unused. In one comment ([see comment](https://github.com/gridsingularity/gsy-e/pull/1214/files#r698586657)) it was explained that these methods would have been used in the context of **D3ASIM-3567**. However, the methods were still not used in the relative PR (https://github.com/gridsingularity/gsy-e/pull/1217/files) so I would assume that they are not needed.

Please let me know if you think these can still be useful.
 
## Proposed changes

- Remove unused methods.

### DIFF
--- a/src/gsy_e/models/strategy/__init__.py
+++ b/src/gsy_e/models/strategy/__init__.py
@@ -653,25 +653,6 @@ class BaseStrategy(EventMixin, AreaBehaviorBase, ABC):
             offer_energy, offer_price, available_energy, market, time_slot=time_slot,
             replace_existing=replace_existing)
 
-    def can_settlement_offer_be_posted(
-            self, offer_energy: float, offer_price: float,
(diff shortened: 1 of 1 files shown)


spyrostz (commented):
I think that these still have to be there and be called from other parts of the code too. The template settlement strategy somehow circumvents the need for this, due to the fact that it accesses the state directly (uses the get_forecast_measurement_deviation_kWh), and by the fact that it is the only entity that posts bids and offers. Where this is really important to have is in the external strategies, where multiple bids / offers can be posted for the same settlement market. 
Seeing that the settlement market has diminished in importance lately I am not sure whether it makes sense to spend resources to add these calls to the external strategies. I will add a ticket though so that it remains in our backlog at least.

PR #1478, opened 2022-07-13
Feature/gsye 257

cruky:

Add support for dynamic change of allocated energy (`coefficient_percentage`) for the SCM.

## Proposed changes

`energy_need_kWh` was used to calculate the future coefficient percentage, the main reason is that there will be more energy distributed across the community, than sold to a grid when using this factor.

Assumption:
- The calculation is triggered after `calculate_community_after_meter_data`
- The new `coefficient_percentage` will be included in the next slot
- The null values for sharing coefficients are allowed
- Skip calculating new coefficient percentages when there is no total `energy_need_kWh` in community*

*Whenever there is no `energy_need_kWh` in community the `coefficient percentage` will stay the same, which could make houses to pay penalties not reflecting their current consumption.

## Running the setup file

To enable the Coefficient market and the dynamic algorithm you need to use the setup file with the json settings file ([the reason why](https://github.com/gridsingularity/gsy-e/pull/1475#discussion_r917950349)).

1. Modify the gsy_e_settings.json with these values:
- advanced_settings.MASettings."`MARKET_TYPE`": 3
- advanced_settings.SCMSettings."`MARKET_ALGORITHM`": 2
2. Run gsy-e with the scm default setup file and modified settings file:
`gsy-e run --setup scm.default_2a --settings-file /GSY_PATH/gsy-e/src/gsy_e/setup/gsy_e_settings.json
`

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-257


tests/area/test_coefficient_area.py
```diff
@@ -118,9 +119,10 @@ def test_calculate_community_after_meter_data(cls):
         assert isclose(scm._home_data[house2.uuid].energy_bought_from_community_kWh, 0.00)
         assert isclose(scm._home_data[house2.uuid].energy_sold_to_grid_kWh, 0.04)
 
-    @classmethod
-    def test_trigger_energy_trades(cls):
-        grid_area = cls._create_2_house_grid()
+    @staticmethod
+    @pytest.mark.skip("calculate_home_energy_bills should be revisit")
+    def test_trigger_energy_trades(_create_2_house_grid):
```
cruky, line 124: This test was failing before this PR. For now it is marked as to skip and the new bug ticket was created to deal with it.

spyrostz (approved):
LGTM

tests/area/test_coefficient_area.py
spyrostz: Already fixed here https://github.com/gridsingularity/gsy-e/pull/1485

mrsaemir (approved):
LGTM, Special thanks!

PR #1479, opened 2022-07-18
refactor/GSYE-12: Refactor read/rotate of strategy profiles

mrsaemir:
Read/Rotate of profiles was done individually for each strategy, resulting in duplicated code among different strategies.

## Proposed changes
- Introducing `EnergyProfile` class as a way to handle read/rotate of profiles in one place.
- Adapt different strategies to use `EnergyProfile` class.


spyrostz (approved):
Great refactoring IMO, LGMT. Let's see how the changes after the refactoring perform and later on we can even move the `EnergyProfile` class to the framework. Not for now though, better to see whether the changes introduce a bug or unintended changes.

PR #1480, opened 2022-07-20
Feature/gsye 280

spyrostz:

The energy benchmark needs to be added to the energy bills of each home. This is indicating how well did the home perform savings-wise compared to the other homes in the community.

## Proposed changes

The formula for calculating the energy benchmark is the following:
```
benchmark = (home_savings_percentage - min_savings_percentage_in_community) / (max_savings_percentage_in_community - min_savings_percentage_in_community) * 100.0
```


src/gsy_e/models/area/scm_manager.py
```diff
@@ -441,7 +465,7 @@ def _validate_coefficients(community: "CoefficientArea") -> None:
                 f"Home {home.name} has assets with non-SCM strategies."
 
     @staticmethod
-    def _validate_market_maker_rate(community):
+    def _validate_market_maker_rate(community: "CoefficientArea") -> None:
         for home in community.children:
-            assert home.get("market_maker_rate") is not None, \
+            assert home._market_maker_rate is not None, \
```
mrsaemir, line 470: Just out of curiosity, how it is guaranteed that _market_maker_rate is always an attribute of each home?

mrsaemir (commented):
Just one question, LGTM once resolved.
Thanks

src/gsy_e/models/area/scm_manager.py
spyrostz: The type hint is supposed to protect us from that. It explicitly defines that a CoefficientArea is passed in this method, and the _market_maker_rate is explicitly defined in the constructor of the CoefficientArea class. 
We can make it even more ironclad by adding an assertion that the area class passed in this method is indeed a CoefficientArea, however I thought that it was an overkill while implementing it. If you insist though I will add it.

mrsaemir (approved):
LGMT, Thanks.

mrsaemir (approved):
LGTM

PR #1484, opened 2022-07-22
GSYE-341: Increment update counter always with first order

cruky:

The issue in current code is due to the storage strategy behavior, and that it distributes the energy to sell uneven to the market slots. It could create the future markets, and not post any offer/bid at the beginning of the slot because of there is no energy. 

## Proposed changes

Constraint that `update_counter` is 0 for the first order should be lifted.

## What is not handled by this PR

The rates are still incorrect. Since in future market the `fit_to_limit` is always true, we should update the `energy_rate_change_per_update` and `number_of_available_updates` accordingly when the first offer/bid is not at the beginning of the slot.


mrsaemir:
```python
df = pd.read_csv('house-1-future-bids.csv')
df2 = df[df["slot"] == "2022-08-03 02:15:00+00:00"]
df2.plot.scatter(x='creation_time', y='rate [ct./kWh]', figsize=(20, 15))
```
Screenshot from 2022-08-03 08-24-15

Screenshot from 2022-08-03 08-23-12

spyrostz (approved):
LGTM

ERafaelMartinez (approved):
LGTM

PR #1491, opened 2022-07-29
GSYE-263: Added splitting of base energy bill to find portion of the …

spyrostz:
…bill that was originated by revenue and expenditure accordingly.

Required for the Savings KPI, in order to calculate these the base energy cost needed to be split in revenue / non revenue parts, in order to use the cost of the energy without the revenue for these calculations. 

## Proposed changes

-


mrsaemir (approved):
LGTM

PR #1492, opened 2022-08-01
GSYE-370: Ignore the KeyError due to a strategy not being supported f…

spyrostz:
…or the operating mode or unknown, and do not construct the area object at all.

View the surrounding logs here https://kibana.production.gridsingularity.com/app/discover#/doc/62ff47d0-4520-11ec-a9a4-0b7ad4b10190/kubernetes_cluster-2022.07.29?id=-qi9SYIBjq9RY1BENQgj for more info about the root cause of the bug.

## Proposed changes

-


spyrostz:
> LGTM, Though I have a question: How will the end user know about his invalid area being ignored? Isn't this a bit implicit?

The end user will not care, since the end user anyways does not add an Infinite Bus here. The problem lies on the frontend since it automatically adds one Infinite Bus area without it being defined in the community datasheet. 

Reason why I fixed this, is in order to accelerate the frontend development. It really should be fixed on the frontend side too.

mrsaemir (approved):

fievelk (approved):
This LGTM but I'd rather see it as a temporary fix. For example, I think we shouldn't use `globals` to store the available leaf types here: https://github.com/gridsingularity/gsy-e/blob/531423af3a14d69c47493021f1b137135ae6e063/src/gsy_e/gsy_e_core/area_serializer.py#L116

I'd suggest getting rid of this logic and replacing it with something stricter. If we try to create an SCM collaboration passing it unsupported areas, I think that raising an error is actually appropriate and shouldn't be silenced. What do you think?

PR #1493, opened 2022-08-01
GSYE-276: Set the clock updater in the _activate_future_markets

cruky:

Since the Offer and Bid gets `creation_time` from the Market now attribute, it should be populated before the market cycle, and not only in `execute_actions_after_tick_event`.

## Proposed changes

Add `update_clock` in `_activate_future_markets` to populate `now` attribute for future markets.

### DIFF
--- a/src/gsy_e/models/area/markets.py
+++ b/src/gsy_e/models/area/markets.py
@@ -117,6 +117,7 @@ class AreaMarkets:
                               grid_fee_const=area.grid_fee_constant),
             name=area.name)
         self.future_markets = market
+        self.future_markets.update_clock(area.now)
         area.dispatcher.create_market_agents_for_future_markets(market)
 
     def _activate_forward_markets(self, area: "Area") -> None:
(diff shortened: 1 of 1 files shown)


src/gsy_e/models/area/markets.py
```diff
@@ -117,6 +117,7 @@ def _activate_future_markets(self, area: "Area") -> None:
                               grid_fee_const=area.grid_fee_constant),
             name=area.name)
         self.future_markets = market
+        self.future_markets.update_clock(area.now)
         area.dispatcher.create_market_agents_for_future_markets(market)
 
     def _activate_forward_markets(self, area: "Area") -> None:
```
cruky, line 123: Should we add similar change for _activate_forward_markets? I was not sure.

spyrostz: I think so, since the forward markets inherit from the future market, therefore in theory if one of these markets posts an offer on the first tick it will not populate the creation time correctly.

spyrostz (commented):
LGTM once added to the _activate_forward_markets

mrsaemir (approved):
LGTM

spyrostz (approved):
LGTM

PR #1495, opened 2022-08-02
GSYE-370: Omit null children from the Area constructor.

spyrostz:

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/area/__init__.py
+++ b/src/gsy_e/models/area/__init__.py
@@ -112,6 +112,7 @@ class AreaBase:
         self.uuid = uuid if uuid is not None else str(uuid4())
         self.slug = slugify(name, to_lower=True)
         self.parent = None
+        children = [child for child in children if child is not None]
         self.children = (
             AreaChildrenList(self, children)
             if children is not None
--- a/src/gsy_e/models/area/__init__.py
+++ b/src/gsy_e/models/area/__init__.py
@@ -112,6 +112,8 @@ class AreaBase:
         self.uuid = uuid if uuid is not None else str(uuid4())
         self.slug = slugify(name, to_lower=True)
         self.parent = None
+        if not children:
+            children = []
         children = [child for child in children if child is not None]
         self.children = (
             AreaChildrenList(self, children)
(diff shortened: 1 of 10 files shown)


Maheen12 (dismissed):
LGTM

src/gsy_e/models/area/__init__.py
```diff
@@ -112,6 +112,9 @@ def __init__(self, name: str = None,
         self.uuid = uuid if uuid is not None else str(uuid4())
         self.slug = slugify(name, to_lower=True)
         self.parent = None
+        if not children:
+            children = []
+        children = [child for child in children if child is not None]
         self.children = (
             AreaChildrenList(self, children)
             if children is not None
```
cruky, line 120: I think that we should change this to align with the change:
```suggestion
            if children
```
There is also a children set to None in the [area_serializer](https://github.com/gridsingularity/gsy-e/blob/6774e3819af1b17405beca31a1751eff45cfcbd5/src/gsy_e/gsy_e_core/area_serializer.py#L148) , so maybe that part should be aligned as well (not sure about it).

spyrostz: I wanted to catch all cases (children == None, children == [], children == {}) in order to harmonize all cases and to not fail, hence the use of the `not`. In a better implementation, it should be inside the AreaChildrenList class, but since we are in a hurry I wanted to go the easy way out.

src/gsy_e/models/leaves.py
```diff
@@ -64,30 +62,23 @@
 
 scm_strategy_mapping = {
     LoadHoursStrategy: SCMLoadHoursStrategy,
-    DefinedLoadStrategy: SCMLoadProfile,
+    DefinedLoadStrategy: SCMLoadProfileStrategy,
     PVStrategy: SCMPVStrategy,
     PVPredefinedStrategy: SCMPVPredefinedStrategy,
     PVUserProfileStrategy: SCMPVUserProfile,
     StorageStrategy: SCMStorageStrategy
 }
```
cruky, line 70: If I'm not wrong then it can be deleted, since it's obsolete.

spyrostz: Unfortunately not this one, the one that can probably be removed since there are no plans to add it in the community datasheet is the SCMLoadHoursStrategy. Still though for the latter, I would like to maintain this because it allows the CLI user to set a constant consumption in a convenient way, without being obliged to use a profile.

PR #1497, opened 2022-08-04
Added the is_scm parameter in the ResultsHandler constructor for the …

spyrostz:
…case that the coefficients market are enabled, in which case the CoefficientEndpointBuffer will be used instead of the normal SimulationEndpointBuffer.

## Proposed changes

-


mrsaemir (approved):
LGTM

cruky (approved):
Great, now the results_mapping for SCM is correct. Thank you!

mrsaemir (approved):
LGTM

PR #1498, opened 2022-08-04
GSYE-370: Corrected calculation of the minutes that the slot length c…

spyrostz:
…omprises of by using the total_minutes() method and not the minutes member, the latter just returns the minutes of the hour. The outcome of this was that for the 60 min slot length, the simulation results were not generated because the current_market_time_slot_str was always the start time (due to the fact that slot_length.minutes was always 0 for the case of slot length == 1 hour).

## Proposed changes

-


mrsaemir (dismissed):
Thanks

mrsaemir (commented):
LGTM

PR #1500, opened 2022-08-08
PH-1217: Calculated correctly the community area to either be the top…

spyrostz:
… level area or the one below the top, depending on the area name. Accumulated the trades from all homes in the community data too, in order to be available in the generation of the community results.

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/area/scm_manager.py
+++ b/src/gsy_e/models/area/scm_manager.py
@@ -1,5 +1,7 @@
 import logging
 from dataclasses import dataclass, asdict
+
+import gsy_e.constants
 from math import isclose
 from typing import Dict, TYPE_CHECKING, List, Optional
 from uuid import uuid4
@@ -165,10 +167,21 @@ class CommunityData:
     energy_need_kWh: float = 0.
     energy_bought_from_community_kWh: float = 0.
     energy_sold_to_grid_kWh: float = 0.
+    trades: List[Trade] = None
 
     def to_dict(self) -> Dict:
         """Dict representation of the community energy data."""
         return asdict(self)
+
+    def serializable_dict(self) -> Dict:
+        """Dict representation that can be serialized."""
+        output_dict = self.to_dict()
+        output_dict["trades"] = [trade.serializable_dict() for trade in self.trades]
+        return output_dict
+
+    def __post_init__(self):
+        if self.trades is None:
+            self.trades = []
 
 
 @dataclass
@@ -301,12 +314,26 @@ class SCMManager:
         SCMCommunityValidator.validate(community=area)
 
         self._home_data: Dict[str, HomeAfterMeterData] = {}
-        # Community is always the root area in the context of SCM.
-        self._community_uuid = area.uuid
+
+        self._community_uuid = self._get_community_uuid_from_area(area)
         self.community_data = CommunityData(self._community_uuid)
         self._time_slot = time_slot
         self._bills: Dict[str, AreaEnergyBills] = {}
         self._grid_fees_reduction = ConstSettings.SCMSettings.GRID_FEES_REDUCTION
+
+    def _get_community_uuid_from_area(self, area):
+        # Community is always the root area in the context of SCM.
+        # The following hack in order to support the UI which defines the Community one level
+        # lower than the Grid area.
+        if "Community" in area.name:
+            return area.uuid
+        else:
+            for child in area.children:
+                if "Community" in child.name:
+                    return child.uuid
+        assert False, f"Should not reach here, configuration {gsy_e.constants.CONFIGURATION_ID}" \
+                      f"should have an area with name 'Community' either on the top level or " \
+                      f"on the second level after the top."
 
     def add_home_data(self, home_uuid: str, home_name: str,
                       grid_fees: float, coefficient_percentage: float,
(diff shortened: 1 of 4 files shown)


src/gsy_e/models/area/scm_manager.py
```diff
@@ -165,11 +167,22 @@ class CommunityData:
     energy_need_kWh: float = 0.
     energy_bought_from_community_kWh: float = 0.
     energy_sold_to_grid_kWh: float = 0.
+    trades: List[Trade] = None
```
fievelk, line 170: Did you use `__post_init__` to avoid issues with the list mutable type? If so, dataclasses have a specific approach for it:
```suggestion
    trades: List[Trade] = field(default_factory=list)
```

- [Docs: Mutable default values](https://docs.python.org/3/library/dataclasses.html#mutable-default-values)

src/gsy_e/models/area/scm_manager.py
```diff
@@ -301,13 +314,27 @@ def __init__(self, area: "CoefficientArea", time_slot: DateTime):
...
+        self._community_uuid = self._get_community_uuid_from_area(area)
         self.community_data = CommunityData(self._community_uuid)
         self._time_slot = time_slot
         self._bills: Dict[str, AreaEnergyBills] = {}
         self._grid_fees_reduction = ConstSettings.SCMSettings.GRID_FEES_REDUCTION
 
+    def _get_community_uuid_from_area(self, area):
+        # Community is always the root area in the context of SCM.
+        # The following hack in order to support the UI which defines the Community one level
+        # lower than the Grid area.
```
fievelk, line 327: Nit: I now understand the rationale of this PR. However, a few weeks ago we had a conversation with Arslan and we agreed that the FE would have changed the 4-hierarchy logic as soon as possible. I would personally do this rather than implementing hacks in the backend, because they are likely to become technical debt as soon as we move on to implementing new features. Also, the ticket that the FE agree to create will now have 2 dependencies on the backend too ([gsy-framework](https://github.com/gridsingularity/gsy-framework/blob/c2804443c4a3f6f108311f60a4065e4c43feaa91/gsy_framework/community_datasheet/community_datasheet_parser.py#L114) and gsy-e here)

src/gsy_e/models/area/scm_manager.py
```diff
@@ -445,12 +472,24 @@ def calculate_home_energy_bills(
...
         """Return the SCM results for one area (and one time slot)."""
         if area_uuid == self._community_uuid:
             return {
                 "bills": self.community_bills,
-                "after_meter_data": {}
+                "after_meter_data": self.community_data.serializable_dict(),
+                "trades": [
+                    trade.serializable_dict()
+                    for trade in self.community_data.trades
+                ]
```
fievelk, line 492: Is this needed? Isn't it just a duplicate of `["after_meter_data"]["trades"]`?

spyrostz: It is. However the community area needs to have the community trades in order to generate the results. Keep in mind that all results handlers operate on individual areas / markets, therefore if there are no trades in the community then simply no results will be generated. This is the actual root cause of the "no results are generated for communities" and the main fix here

src/gsy_e/models/area/scm_manager.py
spyrostz: I agree with you, however we need to deploy. If you want I can add a TODO / FIXME comment to remember. The UI has already a ticket open to deal with this IIRC , so it will not be neglected

src/gsy_e/models/area/scm_manager.py
spyrostz: Let me try this, thanks!

src/gsy_e/models/area/scm_manager.py
fievelk: I couldn't find the ticket you mention. Can you share a link? Otherwise I'll ask the FE team to create it.

src/gsy_e/models/area/scm_manager.py
fievelk: I'm not sure I understood correctly. As far as I see, all "area results" (for all areas) will contain an `after_meter_data` key. Is this correct? 
AFAIU, `area["after_meter_data"]["trades"]` will be present for community areas, even if we don't duplicate it in `area["trades"]`. Could you please share a link of where this would break?

spyrostz: https://github.com/gridsingularity/gsy-framework/blob/master/gsy_framework/sim_results/all_results.py#L68 have a look at this method. This iterates over all areas, passes the trade objects associated with each area to each simulation result handler, and calculates the results / updates each handler according to the trades that are associated with each area. If the community area does not have associated trades, then its handlers will not be updated because they assume that no trades happened for this area. 
Keep in mind, that this needs to be like this in order to support our "market" model, where the trades from different areas are represented as chain trades. Therefore in order to reuse this logic, we have to duplicate the home trades, and simulate that a "chain trade" happened on the community level too. That way the calculations are only per-area and decoupled. 
Let me know if this is not clear enough

fievelk (approved):

src/gsy_e/models/area/scm_manager.py
spyrostz: You are right, I also cannot find the ticket so I might be wrong. If you can remind Arslan to add one then it would be great, thanks in advance!

PR #1501, opened 2022-08-08
Changed serialization of the community data to maintain the Python ob…

spyrostz:
…jects in case of CLI mode, and to use the serialized JSON on the deployed services mode. That way the simulation results will be generated correctly in both cases.

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/area/scm_manager.py
+++ b/src/gsy_e/models/area/scm_manager.py
@@ -1,16 +1,15 @@
 import logging
+from calendar import monthrange
 from dataclasses import dataclass, asdict, field
-
-import gsy_e.constants
 from math import isclose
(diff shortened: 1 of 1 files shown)


src/gsy_e/models/area/scm_manager.py
```diff
@@ -481,7 +481,10 @@ def get_area_results(self, area_uuid: str, serializable: bool = False) -> Dict:
         if area_uuid == self._community_uuid:
             return {
                 "bills": self.community_bills,
-                "after_meter_data": self.community_data.serializable_dict(),
+                "after_meter_data": (
```
spyrostz, line 484: This is the main fix, rest are pylint warning fixes.

fievelk (approved):

cruky (approved):
LGTM

PR #1506, opened 2022-08-10
Unit tests for forward markets

cruky:

These are the unit tests for this [PR](https://github.com/gridsingularity/gsy-e/pull/1503).
## Proposed changes

- test the update_clock in future/forward markets
- test the creation of `forward_markets_mapping` in `_update_myco_matcher`, as it should be aligned with the `FORWARD_MARKET_TYPES`.

### DIFF
--- a/tests/area/test_area.py
+++ b/tests/area/test_area.py
@@ -15,7 +15,7 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-# pylint: disable=missing-function-docstring
+# pylint: disable=missing-function-docstring,protected-access
 from unittest.mock import MagicMock, patch, Mock, call
 
 import pytest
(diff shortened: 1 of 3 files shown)


mrsaemir (approved):

tests/area/test_area.py
```diff
@@ -190,6 +191,20 @@ def test_tick(mock_match_recommendations, config):
         area.tick()
         assert manager.mock_calls == [call.match(), call.update_matcher()]
 
+    @staticmethod
+    @patch("gsy_e.models.market.forward.ConstSettings.ForwardMarketSettings."
+           "ENABLE_FORWARD_MARKETS", True)
+    def test_forward_market_mapping_for_matcher():
```
fievelk, line 197: Nit: unit tests usually target one specific public method. That's why we usually use names like `test_<method-name>` or similar for tests. Because of this, I tried to understand what this method is testing, and I have the feeling that this test is conflating tests for `Area._update_myco_matcher()` and `MycoMatcher.update_area_uuid_forward_markets_mapping()` together. In this case I don't think it's a big deal, but I'd suggest you to check ["The magic tricks of testing" by Sandi Metz](https://youtu.be/URSWYvyc42M) and the unit-testing workshop slides (that are just a short version of her talk).

tests/market/test_forward.py
```diff
@@ -122,3 +124,23 @@ def test_market_rotation(self, market_class, rotator_class, expected_market_coun
         # Market should be deleted if the rotation time has been reached
         rotator.rotate(rotation_time)
         count_orders_in_buffers(forward_markets, expected_market_count - 1)
+
+    @staticmethod
+    @patch("gsy_e.models.market.forward.ConstSettings.ForwardMarketSettings."
+           "ENABLE_FORWARD_MARKETS", True)
+    def test_clock_updates():
```
fievelk, line 131: The same comment as [above](https://github.com/gridsingularity/gsy-e/pull/1506/files#r943238217) applies here: which method are we testing? I would say that we're basically testing the `market.update_clock()` part of `Area.execute_actions_after_tick_event()`. One possible alternative is to create `test_execute_actions_after_tick_event()`, make sure it's correctly calling `market.update_clock()`, and then test `market.update_clock()` separately.

fievelk (approved):
I've left some comments about unit testing approaches. I'm giving my LGTM. I suggest you check the resources I linked because they're great and they might help us create more flexible and thorough unit tests :)

PR #1507, opened 2022-08-12
Bug/ph 1285

spyrostz:

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/user_profile_handler.py
+++ b/src/gsy_e/gsy_e_core/user_profile_handler.py
@@ -154,15 +154,16 @@ class ProfileDBConnectionHandler:
 
     @db_session
     def _buffer_profile_types(self):
-        """ Buffers list of the profile_uuids that correspond to this simulation into
-        self._profile_uuids"""
+        """
+        Buffers profile types for the profiles of this simulation.
+        """
         profile_selection = select(
             (datapoint.profile_uuid, datapoint.profile_type)
             for datapoint in self.Profile_Database_ConfigurationAreaProfileUuids
             if datapoint.configuration_uuid == uuid.UUID(gsy_e.constants.CONFIGURATION_ID))
 
-        for p in profile_selection:
-            self._profile_types[p[0]] = InputProfileTypes(p[1])
+        for profile in profile_selection:
+            self._profile_types[profile[0]] = InputProfileTypes(profile[1])
 
     @db_session
     def _buffer_all_profiles(self, current_timestamp: DateTime):
--- a/src/gsy_e/gsy_e_core/user_profile_handler.py
+++ b/src/gsy_e/gsy_e_core/user_profile_handler.py
@@ -53,6 +53,7 @@ class ProfileDBConnectionHandler:
 
     def __init__(self):
         self._user_profiles = {}
+        self._profile_types = {}
         self._buffered_times = []
         self._profile_uuids = None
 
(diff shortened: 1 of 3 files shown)


src/gsy_e/gsy_e_core/user_profile_handler.py
```diff
@@ -151,6 +152,18 @@ def _buffer_profile_uuid_list(self):
 
         self._profile_uuids = list(profile_selection)
 
+    @db_session
+    def _buffer_profile_types(self):
+        """ Buffers list of the profile_uuids that correspond to this simulation into
+        self._profile_uuids"""
```
fievelk, line 158: Nit (but since it's in the PEP and in our style conventions):
```suggestion
        """
        Buffer a list of the profile_uuids that correspond to this simulation into
        self._profile_uuids.
        """
```

src/gsy_e/gsy_e_core/user_profile_handler.py
```diff
@@ -151,6 +152,18 @@ def _buffer_profile_uuid_list(self):
...
+        profile_selection = select(
+            (datapoint.profile_uuid, datapoint.profile_type)
+            for datapoint in self.Profile_Database_ConfigurationAreaProfileUuids
+            if datapoint.configuration_uuid == uuid.UUID(gsy_e.constants.CONFIGURATION_ID))
+
+        for p in profile_selection:
+            self._profile_types[p[0]] = InputProfileTypes(p[1])
```
fievelk, line 165: This should have raised a pylint warning, maybe your linter was disabled for this file?
```suggestion
        for profile in profile_selection:
            self._profile_types[profile[0]] = InputProfileTypes(profile[1])
```

fievelk (approved):
I left two small comments, you can still merge if it's urgent.

AFAIU, this PR is to fix the case in which we have multiple profile, each one with a different unit. Is this correct?

src/gsy_e/gsy_e_core/user_profile_handler.py
spyrostz: I'll fix this, however it looks like pylint does not complain at all about one-char variables in for loops: 
```
(gsy-e) 192:gsy-e spyrostzavikas$ pylint --version 
pylint 2.11.1
astroid 2.8.4
Python 3.8.9 (default, May 17 2022, 12:55:41) 
[Clang 13.1.6 (clang-1316.0.21.2.5)]
(gsy-e) 192:gsy-e spyrostzavikas$ pylint src/gsy_e/gsy_e_core/user_profile_handler.py 

--------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 10.00/10, +0.00)

```

Maybe you are using a different pylint version? Because this is the one defined by the project requirements. 
Regardless of whether it is a pylint warning or not, let me fix this.

src/gsy_e/gsy_e_core/user_profile_handler.py
spyrostz: I'll also fix the docstring, thanks

spyrostz:
> I left two small comments, you can still merge if it's urgent.
> 
> AFAIU, this PR is to fix the case in which we have multiple profile, each one with a different unit. Is this correct?

The purpose of this was in order to use the profile types from the database, and not let the strategy dictate the type of the profile. When we upload the profile we also associate the type (which essentially is the unit / conversion type of the profile values). Up until now it was ignored, and the strategies demanded specific profile types. However with the datasheet it is not possible anymore, since the PV / Load profiles are in kWh, while in GSy Maps / Rebase it is obligatory to upload power profiles. Therefore it was necessary to read the profile types from the database, so that the strategy can convert the profile to kWh or not depending on whether it is needed.

src/gsy_e/gsy_e_core/user_profile_handler.py
fievelk:
```bash
$ pylint --version
pylint 2.11.1
astroid 2.8.4
Python 3.8.6 (default, Nov  8 2021, 17:05:07)
[Clang 12.0.5 (clang-1205.0.22.9)]
```
It looks like the version is the same, I only have a slightly different python version but I don't think it should matter.

fievelk (approved):

fabriciocarboni (approved):
LGTM

PR #1508, opened 2022-08-15
GSYE-373-2: Use AvailableMarketTypes.value instead of AvailableMarketTypes

mrsaemir:
AvailableMarketTypes objects are not json serializable.

## Proposed changes

- Use AvailableMarketTypes.value


fievelk (approved):

fievelk:
LGTM. I would have expected this change to cause test failures, though. Do you think it's possible to test this somehow, so that we will avoid regressions?

mrsaemir:
> LGTM. I would have expected this change to cause test failures, though. Do you think it's possible to test this somehow, so that we will avoid regressions?

@fievelk 
I definitely agree. Actually the whole forward market lacks tests. For now, I'm trying to unblock Kamil's task on saving aggregated forward market results and I'll come back to write unit tests ASAP.

cruky (approved):

cruky (approved):

PR #1511, opened 2022-08-19
GSYE-377: Split the LoadForecastExternalStrategy to 2 classes, LoadPr…

spyrostz:
…ofileForecastExternalStrategy and LoadHoursForecastExternalStrategy in order to separate the default behavior for the case that no forecast stream is available and the strategy has to fall back to default strategy profiles / energy values.

## Proposed changes

-


fabriciocarboni (approved):
LGTM

mclrch (approved):
LGTM

PR #1513, opened 2022-08-22
feature/GSYE-373-3: Filter bids/offers/trades of forward market based on current time_slot

mrsaemir:

We want to transfer only bids/offers/trades that have happened in the current simulation time of forward markets. 

## Proposed changes
- Add `_get_current_forward_orders_from_timeslot`

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -177,7 +177,7 @@ class SimulationEndpointBuffer:
         return [order.serializable_dict()
                 for order in forward_orders
                 if order.time_slot == market_time_slot and
-                last_time_slot < order.creation_time <= current_time_slot]
+                last_time_slot <= order.creation_time < current_time_slot]
 
     def _read_future_markets_stats_to_dict(self, area: "Area") -> Dict[str, Dict]:
         """Read future markets and return market_stats in a dict."""
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -166,6 +166,19 @@ class SimulationEndpointBuffer:
                 for order in future_orders
                 if order.time_slot == time_slot]
 
+    @staticmethod
+    def _get_current_forward_orders_from_timeslot(forward_orders: List,
+                                                  market_time_slot: DateTime,
+                                                  area: "Area") -> List:
+        """Filter orders that have happened in the current simulation time for
+        the specified market time slot."""
+        current_time_slot = area.now
+        last_time_slot = area.now - area.config.slot_length
+        return [order.serializable_dict()
+                for order in forward_orders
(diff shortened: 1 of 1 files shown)


src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -166,6 +166,19 @@ def _get_future_orders_from_timeslot(future_orders: List, time_slot: DateTime) -
...
                 if order.time_slot == time_slot]
 
+    @staticmethod
+    def _get_current_forward_orders_from_timeslot(forward_orders: List,
+                                                  market_time_slot: DateTime,
+                                                  area: "Area") -> List:
+        """Filter orders that have happened in the current simulation time for
+        the specified market time slot."""
+        current_time_slot = area.now
+        last_time_slot = area.now - area.config.slot_length
```
fievelk, line 176: Nit:
```suggestion
        previous_time_slot = area.now - area.config.slot_length
```
I'm not sure why, but my brain finds this easier to understand. I would intuitively think that `area.now` is actually the last time slot.
Just a nit anyway.

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -166,6 +166,19 @@ def _get_future_orders_from_timeslot(future_orders: List, time_slot: DateTime) -
                 for order in future_orders
                 if order.time_slot == time_slot]
 
+    @staticmethod
+    def _get_current_forward_orders_from_timeslot(forward_orders: List,
+                                                  market_time_slot: DateTime,
+                                                  area: "Area") -> List:
```
fievelk, line 172: Nit:
```suggestion
    def _get_current_forward_orders_from_timeslot(
        forward_orders: List,
        market_time_slot: DateTime,
        area: "Area") -> List:
```
I think we had this conversation on Slack once, and we sort of agreed that this was our favourite type of indentation. Or do I remember wrong?

fievelk (approved):
LGTM, just left two nits.

As usual: it would be great to have tests for this change to avoid regressions.

cruky (approved):
LGTM

spyrostz (approved):

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
mrsaemir: I've actually though of that, but IMO, the favorite word in gsy-e is `last` and not `previous`. I'm telling this because if you search gsy-e or gsy-framwork for the word `previous`, you won't get so many results in code but it is full of `last_*` variables.

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
mrsaemir: As this part of the EndpointBuffer has changed several times during the last weeks. I'm unwilling to write tests for it now. I will definitely write tests as soon as the design is settled. Thanks for reminding that.

PR #1515, opened 2022-08-22
GSYE-297: Parse SSP and maintain asset states for forward markets

fievelk:

We want to
- Parse the Standard Solar Profile
- Extend the state variables of the assets based on the maximum duration of the forward markets (when they are enabled).

### Example

Assets that use `LoadHoursStrategy` have the following two state variables:
- `_desired_energy_Wh`
- `_energy_requirement_Wh`

These values currently hold the energy requirements for all **settlement**, **spot** and **future** markets. These should be extended using **15 mins lots** for the duration of the `YearForwardMarket` (currently 6 years into the future).

---


src/gsy_e/models/strategy/load_hours.py
```diff
@@ -242,6 +240,10 @@ def __init__(self, avg_power_W, hrs_per_day=None, hrs_of_day=None,
     def state(self) -> LoadState:
         return self._energy_params.state
 
+    @property
+    def asset_type(self):
+        return AssetType.CONSUMER
```
fievelk, line 245: I just moved this up closer to the class declaration. 
It's usually good to have the interface (public methods/properties) declared at the beginning of the class, so that the purpose of the class is immediately clear.

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -449,7 +451,7 @@ def _one_sided_market_event_tick(self, market, offer=None):
                 self._energy_params.decrease_hours_per_day(time_slot, energy_Wh)
 
         except MarketException:
-            self.log.exception("An Error occurred while buying an offer")
```
fievelk, line 452: I believe that this was never reached, because AFAIS there is no `self.log` variable/property (so this should have raised an exception if the flow ever got to this line). I replaced it with a module logger.

src/gsy_e/models/state.py
```diff
@@ -400,10 +402,6 @@ def total_energy_demanded_Wh(self) -> float:
         """Return the total energy demanded in Wh."""
         return self._total_energy_demanded_Wh
 
-    def get_desired_energy(self, time_slot: DateTime) -> float:
-        """Return the desired energy (based on profile data)."""
-        return self._desired_energy_Wh[time_slot]
```
fievelk, line 405: This was just a duplication of `get_desired_energy_Wh()`, so I removed and replaced it.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -0,0 +1,182 @@
...
+            "capacity_kW": self.capacity_kW
+        }
+
+    def event_activate_energy(self, area):
+        """Update energy requirement upon the activation event."""
+        self._area = area
+        self._profile_generator = ForwardTradeProfileGenerator(
+            peak_kWh=convert_kW_to_kWh(self.capacity_kW, self._area.config.slot_length))
+
+    def event_traded_energy(self, energy_kWh, market_slot: pendulum.DateTime, product_type):
```
fievelk, line 42: This method will change based on the actual actions that we want to take on trades. I've just used this name as a way to show how the profile generator would be called with traded energy values.

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -833,6 +830,13 @@ def restore_global_state(self, saved_state: dict) -> None:
         self._time.slot_length_realtime = duration(
             seconds=saved_state["slot_length_realtime_s"])
 
+    @staticmethod
+    def _compute_memory_info():
+        gc.collect()
+        process = psutil.Process(os.getpid())
+        mbs_used = process.memory_info().rss / 1000000.0
+        log.error("Used %s MBs.", mbs_used)
```
spyrostz, line 836: Please revert the log to debug in order to not spam the dev environment logs.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -1,16 +1,157 @@
...
+        """
+        When a trade happens, we want to split the energy through the entire year following a
+        standard solar profile.
+        """
+        assert self._profile_generator is not None
+        # Create a new profile that spreads the trade energy across multiple slots. The values
+        # of this new profile are obtained by scaling the values of the standard solar profile
+        return self._profile_generator.generate_trade_profile(
+            energy_kWh=energy_kWh,
```
spyrostz, line 51: Interesting, therefore in this case you are using the peak value in order to scale the whole profile, am I correct ? I will implement it first and then maybe ask some potential optimization here, because the first thing that comes to mind is that finding this peak value might be more performance-intensive than necessary. I will come back to it if I see that the performance here becomes a problem, thanks!

spyrostz (approved):
Some minor comments, LGTM though

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
fievelk: The steps I'm following are these:
1. Find the scaling factor as:
    ```py
    # `new_peak` is the `capacity_kWh` of the asset
    scaling_factor = new_peak / peak_of_SSP
    ```
2. Scale the original SSP using the scaling factor. We obtain the **"capacity profile"**.
3. Compute a new scaling factor for the traded energy: 
    ```py
    # `new_peak` is the energy traded by the asset
    scaling_factor = new_peak / peak_of_SSP
    ```
4. Scale the original SSP using the "trade" scaling factor. In this way, we obtain the profile we need to subtract from the energy that remains to be traded.
5. Subtract this profile from the **capacity profile** in order to know the residual energy that can be bid in follow-up trades.

Right now I'm a bit confused about the last two steps (4 and 5). Shouldn't (4) already represent the energy we want to trade? Why do we want to get the residual dictionary (5)?

spyrostz (approved):
LGTM

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -0,0 +1,133 @@
...
+            product_type=product_type)
+
+        for time_slot, energy_value_kWh in trade_profile.items():
+            self._state.decrement_energy_requirement(
+                purchased_energy_Wh=energy_value_kWh * 1000,
+                time_slot=time_slot,
+                area_name=self._area.name)
+
+
+class ProductionStandardProfileEnergyParameters:
```
fievelk, line 84: This class is basically identical to the `ConsumptionStandardProfileEnergyParameters` above. The only thing that changes is that the `event_traded_energy()` method calls `self._state.decrement_available_energy()` (instead of `decrement_required_energy()`) with kWh units.
I decided to keep the duplication because I don't know if/how these two classes might change after comments and after Spyros' template strategy PR.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -0,0 +1,133 @@
...
+        assert self._profile_generator is not None
+        # Create a new profile that spreads the trade energy across multiple slots. The values
+        # of this new profile are obtained by scaling the values of the standard solar profile
+        trade_profile = self._profile_generator.generate_trade_profile(
+            energy_kWh=energy_kWh,
+            market_slot=market_slot,
+            product_type=product_type)
+
+        for time_slot, energy_value_kWh in trade_profile.items():
+            self._state.decrement_energy_requirement(
```
fievelk, line 78: NOTE: This will not work unless there are already values in the state dictionary (it will raise a KeyError).

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -0,0 +1,33 @@
...
+        self._original_profile: Dict[pendulum.DateTime, float] = profile
+
+    def scale_by_peak(self, peak_kWh: float) -> Dict[pendulum.DateTime, float]:
+        """Return a profile obtained by scaling the original one using a new peak capacity."""
+        scaling_factor = self._compute_scaling_factor(peak_kWh)
+        return self._scale_by_factor(scaling_factor)
+
+    @property
+    def _original_peak_kWh(self) -> float:
+        return max(self._original_profile.values())
```
spyrostz: This property will be evaluated every time that it is called. It will probably be much more performant to calculate the _original_peak_kWh in the constructor and maintain this as a class member variable, because that way it will be calculated only once.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -1,16 +1,157 @@
...
+        self._scaled_capacity_profile = self._scaler.scale_by_peak(peak_kWh=peak_kWh)
+
+    def generate_trade_profile(
+            self, energy_kWh, market_slot, product_type):
+        """Generate a profile based on the market slot and the product type."""
+        profile_scaled_by_traded_energy = self._scaler.scale_by_peak(peak_kWh=energy_kWh)
+
+        # This subtraction is done _before_ expanding the slots to improve performance
+        residual_energy_profile = self._subtract_profiles(
+            self._scaled_capacity_profile, profile_scaled_by_traded_energy)
```
spyrostz: Minor and can be skipped in order to avoid premature optimization, but it could be further optimized if we maintained the peak of the scaled_capacity_profile. That way, we would subtract only 2 floating point values, and then use only once the scale_by_peak method without any need for subtracting the 2 profiles. As mentioned, no need to deal with it now though but only if this becomes a problem

spyrostz (approved):
Some minor comments, LGTM otherwise. Thanks for fixing this!

src/gsy_e/models/strategy/energy_parameters/forward_profile.py
```diff
@@ -0,0 +1,185 @@
...
+    _STANDARD_SOLAR_PROFILE = StandardProfileParser.parse()
+
+    def __init__(self, peak_kWh: float):
+        self._scaler = ProfileScaler(self._STANDARD_SOLAR_PROFILE)
+        self._scaled_capacity_profile = self._scaler.scale_by_peak(peak_kWh=peak_kWh)
+
+    def generate_trade_profile(
+            self, energy_kWh, market_slot, product_type):
+        """Generate a profile based on the market slot and the product type."""
+        trade_profile = self._scaler.scale_by_peak(peak_kWh=energy_kWh)
```
mrsaemir, line 74: I'd suggest to assert `energy_kWh` is a positive float.

mrsaemir (dismissed):
Looks promising.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
fievelk: Done in [73fc176](https://github.com/gridsingularity/gsy-e/pull/1515/commits/73fc1761571e174bfcd4025216adb5ea7abfbdf0)

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
fievelk: I noticed that it looks like you commented on an outdated version of the code. I commented out these lines because I'm not dealing with the residual energy, but only with the traded energy.

tests/strategy/models/energy_parameters/test_energy_params_eb.py
```diff
@@ -0,0 +1,473 @@
...
+        energy_params.event_traded_energy(
+            energy_kWh=50,
+            market_slot=market_slot,
+            product_type=product_type)
+
+        assert state_mock.decrement_available_energy.call_count == 96
+
+        area_name = "PV"
+        # We want to make sure that all the expected calls to the state class are correctly issued
+        calls = [
```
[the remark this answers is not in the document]
fievelk: What kind of `for` loop would you recommend using? Each call is different: the first N calls have a different datetime but always `0` energy. The following calls (the middle of the profile) all have different energy values. Finally, the tail of the profile again has the same energy (`0`) but different datetimes.

I could in theory create a list of dicts for each call (like `[{date: ..., energy: ...}, {date: ..., energy: ...} ...]` and the use a `for` loop to create the `call` objects, but IMO it wouldn't change much. So I think this is not what you were thinking of.

src/gsy_e/models/strategy/energy_parameters/forward_profile.py
fievelk: Done in [e75be3b](https://github.com/gridsingularity/gsy-e/pull/1515/commits/e75be3b218ac0605d6fa4af5ee1bd0b2180795ed)!

spyrostz (approved):
LGTM

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: Thanks for pointing it out, I forgot to reply to this comment after I saw the later commits! Looks good as is!

tests/strategy/models/energy_parameters/test_energy_params_eb.py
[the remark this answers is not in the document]
mrsaemir: I got it. However, the energy value seems to be a normal curve and I was thinking of generating the curve at run time instead of hard coding it there. Is it possible (easy) to implement?

src/gsy_e/models/strategy/energy_parameters/forward_profile.py
```diff
@@ -0,0 +1,192 @@
...
+                    <time_slot>: value,
+                    ...
+                }
+        """
+        assert product_type in ALLOWED_MARKET_TYPES
+
+        period = None
+        if product_type == AvailableMarketTypes.YEAR_FORWARD:
+            period = pendulum.period(
+                start=market_slot.start_of("year"), end=market_slot.end_of("year"))
```
hannesdiedrich, line 107: I had a look into the documentation of pendulum and AFAIU, you need to have a datetime object in order for the `start_of` and `end_of` methods to work properly. Think about leap years or the length of the months. You somehow have to tell pendulum which month you want the end of for. Or did I mis a step?

src/gsy_e/models/strategy/energy_parameters/load.py
```diff
@@ -73,8 +73,6 @@ def update_energy_requirement(self, time_slot, overwrite=False):
             self.avg_power_W, self._area.config.slot_length)
         if self.allowed_operating_hours(time_slot):
             desired_energy_Wh = self.energy_per_slot_Wh
-            if not self.allowed_operating_hours(time_slot):
-                desired_energy_Wh = 0.0
```
hannesdiedrich, line 77: 🎉

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
hannesdiedrich: >I'd like to refactor the common parts into an abstract/interface class.

👍

tests/strategy/models/energy_parameters/test_energy_params_eb.py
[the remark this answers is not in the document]
hannesdiedrich: AFAIU, the energy values are from standard_solar_profile.csv
If we want to change the file, we need to also change every line here. So yeah, a list comprehension that reads the SSP for the selected time_slot could work. However, then the test is influenced by the CSV parser, which is not what we want right?

hannesdiedrich (commented):
One question regarding pendulum. If this is answered, LGTM
Thanks!

src/gsy_e/models/strategy/energy_parameters/forward_profile.py
spyrostz: I think that market_slot is a DateTime object, that's why it works that way here.

src/gsy_e/models/state.py
```diff
@@ -400,10 +400,6 @@ def total_energy_demanded_Wh(self) -> float:
         """Return the total energy demanded in Wh."""
         return self._total_energy_demanded_Wh
 
-    def get_desired_energy(self, time_slot: DateTime) -> float:
```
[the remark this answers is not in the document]
spyrostz: The get_desired_energy_Wh of the ConsumptionState base class is used instead.

src/gsy_e/models/strategy/energy_parameters/forward_profile.py
hannesdiedrich: Ahh, I oversaw this. My attention was towards `pendulum.period`. This should work. Thanks!

hannesdiedrich (approved):
LGTM

tests/strategy/models/energy_parameters/test_energy_params_eb.py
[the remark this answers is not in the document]
fievelk: @hannesdiedrich exactly. Basically we don't want to implement logic in tests, because that logic should then be tested as well. Specifically: if we create a function that generates a gaussian curve (that replaces the SSP) we would then also need to be sure that that function works properly. So the straightforward way is to just hardcode the values we expect to see as an output for the current query we're testing. If the SSP changes, tests will fail. This might or might not be desired, but it's better than testing something else by mistake and seeing tests passing by accident (without catching bugs).

PR #1520, opened 2022-08-30
Feature/gsye 333 forward market params

spyrostz:

## Proposed changes

-


src/gsy_e/models/market/forward.py
```diff
@@ -31,6 +32,15 @@
     from gsy_e.models.area.event_dispatcher import AreaDispatcher
 
 
+@dataclass(frozen=True)
+class ForwardMarketSlotParameters:
+    """Parameters that describe a forward market slot."""
+    open_timestamp: DateTime
+    close_timestamp: DateTime
+    delivery_start_timestamp: DateTime
+    delivery_end_timestamp: DateTime
```
fievelk, line 41: A few points about this. First just a small nit: I think that there's a difference between timestamp and datetime. To mitigate this maybe we could just use something like:
```suggestion
    opening_time: DateTime
    closing_time: DateTime
    delivery_start_time: DateTime
    delivery_end_time: DateTime
```
(time is also a bit different, but maybe it's more generic).

fievelk: Other two more relevant (IMO) points:
1. The "delivery times" are basically the same thing as just `market.time_slot`. What I mean is that, in the case of a yearly market (for example), the slot would be `2024-01-01T00:00:00` and it would already represent on its own the entire delivery time (because its duration is 1 year).
I find this a bit confusing: the market's time slot represents a **period** of time, while the delivery start (or end) time represents a precise **point** in time. IMO this difference is not immediately evident from the code now and it might lead to issues and confusion if we introduce both concepts. One way to make this more obvious IMO would be by creating a specific primitive class for market slots.
2. Since these are the parameters of a market slot, why don't we create a `MarketSlot` class and add these parameters inside of it? This would remove the need for all the `get_market_parameters_for_market_slot` logic, because each market slot would already have its parameters inside of it. I know this would take some time to implement (and it wouldn't be in the context of these PRs), but I think it would help (also given what I said in point 1). We could use slots if you're worried about performance. What do you think?

src/gsy_e/models/market/forward.py
```diff
@@ -58,17 +69,50 @@ def _get_start_time(current_time: DateTime) -> DateTime:
...
+
+    def _set_open_market_slot_parameters(
+            self, current_market_slot: DateTime, created_market_slots: List[DateTime]):
+        """Update the parameters of the newly opened market slots."""
+        for market_slot in created_market_slots:
+            if market_slot in self._open_market_slot_parameters:
+                continue
+
+            self._open_market_slot_parameters[market_slot] = ForwardMarketSlotParameters(
+                delivery_start_timestamp=market_slot,
```
fievelk, line 110: I'm not sure I understand this. AFAIU, `created_market_slots` contains the slots in which users can post orders for a forward exchange, but these slots aren't the delivery slots of that exchange. So here we're iterating over open slots (not delivery slots) of a forward exchange.
For example, in the case of a `YearForwardMarket` the `_create_future_market_slots()` method will generate something like this:
```py
['2024-01-01T00:00', '2025-01-01T00:00', '2026-01-01T00:00', '2027-01-01T00:00', '2028-01-01T00:00']
```
For each one of these (open) slots, we're now setting:
```py
delivery_start_timestamp=market_slot
```
But this seems odd to me. Shouldn't the delivery start time be the time of the forward exchange slot?
Also: since we're iterating over open slots (not delivery slots), what do the open/close timestamps in the next lines represent?

Maybe I missed some part or I was wrong in one of my assumptions. Let me know if this doesn't make much sense :)

src/gsy_e/models/market/forward.py
```diff
@@ -58,17 +69,50 @@ def _get_start_time(current_time: DateTime) -> DateTime:
...
+                delivery_end_timestamp=(
+                        market_slot + self._get_market_slot_duration(market_slot, None)),
+                open_timestamp=current_market_slot,
+                close_timestamp=market_slot - self._time_from_market_close_till_delivery
```
fievelk, line 114: Shouldn't this just be the same as `_get_end_time()`?

src/gsy_e/models/market/forward.py
```diff
@@ -86,6 +130,14 @@ def _get_end_time(current_time: DateTime) -> DateTime:
     def _get_market_slot_duration(_current_time: DateTime, _config) -> duration:
         return duration(minutes=15)
 
+    @property
+    def _time_from_market_close_till_delivery(self) -> duration:
+        return duration(minutes=15)
```
fievelk, line 135: I have the feeling that this indirectly replicates the logic of `_get_end_time()`.

src/gsy_e/models/market/forward.py
spyrostz: I was also confused at first, that's why it took me that long to prepare this draft. The created_market_slots are actually the delivery times. Using your example, the current date is 31-08-2022. This means that the yearly markets that should be open now are the following:
delivery time: 01-01-2024 - > 31-12-2024 . Should be open and close at 31-12-2022. 
delivery time: 01-01-2025 - > 31-12-2025 . Should be open and close at 31-12-2023. 
delivery time: 01-01-2026 - > 31-12-2026 . Should be open and close at 31-12-2024. 
delivery time: 01-01-2027 - > 31-12-2027 . Should be open and close at 31-12-2025. 

Therefore the created_market_slots indeed represent the start of the delivery time.
The confusing part in this is the naming of start / end time IMO. They are not the start and the end time of the market slot, but the time that the rotator should start opening new markets based on a current time, and the time that the rotator should end opening new markets. Hard to name this properly admittedly though. 
One last remark, is that the market for 2028 is opened. This should not be the case since IIRC we only need 4 open yearly markets and not 5. I will also fix it in the context of this PR.

src/gsy_e/models/market/forward.py
spyrostz: As I mentioned in the previous comment, this is not the closing time of the market with a specific market slot, but rather the time that the rotator should stop creating new markets in the future, based on the current time slot.

src/gsy_e/models/market/forward.py
spyrostz: I do not think so. The _get_end_time sets the time horizon over which new markets will be created. However the closing time of these markets is implicitly defined by the ForwardMarketRotator (which I am also in the process of updating) , and more specifically by its abstract method `_get_last_slot_to_be_deleted`. This is set for the yearly markets to this `current_time.add(years=1)` which is correct for the yearly market (essentially we want to close all markets before 2023, for the example that you presented above).

src/gsy_e/models/market/forward.py
spyrostz: Some remarks on the points:
1. Actually the time_slot (the way we use it at least) is just a timestamp that identifies the market. If you just check the type of the time_slot it is a DateTime and not a Duration. In order to represent the duration aspect of this, each market has to implement this method `_get_market_slot_duration` which is indeed representing the length of the delivery period. We could indeed wrap all these in a separate class, however I am afraid of the changes that we will have to make to our legacy code, and I personally think it is not worth doing this for the old markets at the very end, since the method I mentioned makes its intent quite clear. Having said that, let me try to see if it is feasible to implement something like you mention in point #2 only for the forward markets, where we will have more use of it.

spyrostz: Ah and of course I am fine with the nits. After all this is a draft and I wanted your opinion on the naming too, thanks

src/gsy_e/models/market/forward.py
fievelk: > One last remark, is that the market for 2028 is opened. This should not be the case since IIRC we only need 4 open yearly markets and not 5. I will also fix it in the context of this PR.

I'm still a bit confused. From the product document:

> For a yearly product, the market opens 5 years in advance and closes 1 year before 1st day of the year of T0. This means that if you are to sell a product now, you can only sell for 2024-2028 because the yearly period starts on the 1st of the calendar year. The market for 2028 will in this case close on the 1st Jan of 2027 to ensure a yearly gap for the other products to be valid and participate in the market.  For the purposes of visualisation on the UI we will aggregate the data for whole months; the graphs will be yearly viewed with monthly intervals.

From this, it looks like the 2028 slot should be "open" and it will only "close" on the 1st Jan 2027.

I think the terms "market", "slot" and "exchange" create some confusion. If we say that `market == exchange`, then we can say that our market lasts 5 years (this is different from a market slot, which lasts 1 year for Yearly exchanges).

Suppose today is August 2022. 
There must be a gap of 1 year between "now" and the first day of the year/slot when I want to trade. So the exchange can't start in 2023 (less than 1 year from now). The first good start date for the yearly exchange will then be 2024.01.01.

This exchange lasts 5 years, so the slots of this exchange should be `[2024, 2025, 2026, 2027, 2028]` (5 slots).
However, since there must be a gap of 1 year between my order and the first day of the year, the last good moment for me to make an order on the 2028 slot will be the `2027.01.01` (or maybe it should actually be the `2026.12.31`).

Could you please check if this sounds right?

spyrostz: > From this, it looks like the 2028 slot should be "open" and it will only "close" on the 1st Jan 2027.

Indeed, I did not recall the product document correctly. Since it is 5 years that will be open and not 4 as I recalled then the implementation is correct. Thanks for checking it out ! 

> I think the terms "market", "slot" and "exchange" create some confusion. If we say that market == exchange, then we can say that our market lasts 5 years (this is different from a market slot, which lasts 1 year for Yearly exchanges).

This is the reason why we use the term "product" more and more these days. I would refrain from using the term exchange, since this is appropriate for gsy-e as a whole and not for individual "markets" / "products". "market" is the term that we use in order to describe traded products that have a specific delivery duration. E.g. the yearly market has a delivery duration of one year. The exact time that the energy will be delivered is determined by the slot, in order to be able to trade for the same product (e.g. yearly) on different timeframes.  

> Suppose today is August 2022.
There must be a gap of 1 year between "now" and the first day of the year/slot when I want to trade. So the exchange can't start in 2023 (less than 1 year from now). The first good start date for the yearly exchange will then be 2024.01.01.

I would urge again to not use the terms start / end, but delivery start / end and open / close in order to not fall into the same pitfall as before. Under these terms, I am rephrasing this:

`There must be a gap of 1 year between "now" and the first day of the year/slot when I want the traded energy to be delivered. So the slot for the delivery year of 2024 can't remain open in 2023 (less than 1 year from now). The slot will be open from now till 31-01-2022, and I can trade till that time if I want energy delivery for 2024.`

Rephrasing the following too:
`Each slot remains open for 5 years, and opens 5+1 years in advance, so the slots of the yearly market should be [2024, 2025, 2026, 2027, 2028] (5 slots).
However, since there must be a gap of 1 year between my order and the first day of the year, the last good moment for me to make an order on the 2028 slot will be the 2026.12.31 end of the day.`

fievelk (dismissed):

src/gsy_e/models/market/forward.py
```diff
@@ -119,13 +180,24 @@ def _get_end_time(current_time: DateTime) -> DateTime:
...
+    def _calculate_closing_time(self, delivery_time: DateTime) -> DateTime:
+        """
+        Closing time of the week market is one week before the date that the energy will be
+        delivered.
+        """
+        return delivery_time.set(hour=0, minute=0).subtract(weeks=1)
+
+    @property
+    def type_name(self):
+        return "Week forward Market"
```
hannesdiedrich, line 192:
```suggestion
        return "Week Forward Market"
```

src/gsy_e/models/market/forward.py
```diff
@@ -150,3 +233,14 @@ def _get_end_time(current_time: DateTime) -> DateTime:
...
+    def _calculate_closing_time(self, delivery_time: DateTime) -> DateTime:
+        """
+        Closing time of the yearly market is one year before the date that the energy will be
+        delivered.
+        """
+        return delivery_time.set(month=1, day=1, hour=0, minute=0).subtract(years=1)
+
+    @property
+    def type_name(self):
+        return "Year Froward Market"
```
hannesdiedrich, line 246:
```suggestion
        return "Year Forward Market"
```

src/gsy_e/models/market/forward.py
```diff
@@ -86,30 +129,48 @@ def _get_end_time(current_time: DateTime) -> DateTime:
     def _get_market_slot_duration(_current_time: DateTime, _config) -> duration:
         return duration(minutes=15)
 
+    def _calculate_closing_time(self, delivery_time: DateTime) -> DateTime:
+        """Closing time of the intraday market is 15 mins before delivery."""
```
hannesdiedrich, line 133: So currently, the closing time is the end of the former market slot, and this method is there to easily change this individually for each forward market type, right?

src/gsy_e/models/market/forward.py
```diff
@@ -47,6 +57,7 @@ def __init__(self, bc: Optional[NonBlockchainInterface] = None,
         super().__init__(bc=bc, notification_listener=notification_listener,
                          readonly=readonly, grid_fee_type=grid_fee_type,
                          grid_fees=grid_fees, name=name)
+        self._open_market_slots = {}
```
hannesdiedrich, line 60: NIT, but could be helpful
```suggestion
        self._open_market_slots: Dict[DateTime, ForwardMarketSlot] = {}
```

src/gsy_e/models/area/market_rotators.py
```diff
@@ -53,31 +53,26 @@ class ForwardMarketRotatorBase:
...
 
     def rotate(self, current_time: DateTime) -> None:
         """Delete orders in expired day-ahead markets."""
         if self._is_it_time_to_rotate(current_time):
-            self.markets.delete_orders_in_old_future_markets(
-                last_slot_to_be_deleted=self._get_last_slot_to_be_deleted(
-                    current_time))
+            slots_deleted = []
+            for delivery_time, market_slot_info in self.markets.open_market_slot_info.items():
+                if market_slot_info.closing_time <= current_time:
```
hannesdiedrich, line 65: AFAIU, the delivery timeframe is after the closing time. Shouldn't we wait until the energy is delivered before we delete the orders? Just in case?

hannesdiedrich (commented):
Some nits and two questions. LGTM in general

src/gsy_e/models/area/market_rotators.py
spyrostz: The orders should be persisted in the database, therefore no need to maintain them in the simulation after the market has been closed. Main reason for that is in order to offload the exchange with orders that will not be useful (it is not possible to match open orders after the market has closed)

src/gsy_e/models/market/forward.py
spyrostz: Fixing, thanks!

src/gsy_e/models/market/forward.py
spyrostz: Exactly, instead of reusing the duration of the market slot (which is essentially the delivery duration) we define now explicitly the closing time of the slot, and indeed, the method is overridden in order to easily change it and oblige each market type to implement it.

src/gsy_e/models/market/forward.py
spyrostz: Fixing, thanks

src/gsy_e/models/market/forward.py
spyrostz: Fixing, thanks

hannesdiedrich (approved):
LGTM

cruky (approved):
LGTM

mrsaemir (approved):
LGTM

PR #1524, opened 2022-09-09
GSYE-389: Send heartbeats during the initialization of areas and strategies

fievelk:

When the `Simulation` class is initialized on gsy-e, an instance of `SimulationExternalEvents` is created. This class sends heartbeats to watchdog to notify it that the simulation is "healthy".

However, the constructor of the `SimulationExternalEvents` is only called after the simulation `_init()` method, which constructs the `Area` and `Strategy` objects used by the simulation. Since this initialization can take a long time to process, heartbeats are not being sent to watchdog for quite some time. 

We want to start sending heartbeats to watchdog before the `_init()` method is called.

## Proposed changes

- Move the constructor of the `SimulationExternalEvents` before the `_init()` method call. 
- Refactoring: simplify the interface of `SimulationExternalEvents` (and consequently modify parts of the `Simulation` class).

## Note

I recommend reviewing commit-by-commit. Only the last commit contains the actual fix, while the others are refactorings for pylint and to simplify the interface.

---

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-389


src/gsy_e/gsy_e_core/redis_connections/simulation.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, simulation_status, simulation_id, live_events, progress_info,
             del pubsub
             return
         heartbeat = RepeatingTimer(HeartBeat.RATE, self._heartbeat_tick)
-        heartbeat.setDaemon(True)
+        heartbeat.daemon = True
```
fievelk, line 77: `setDaemon()` has been deprecated in favor of the `daemon` property.

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -544,45 +521,49 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig,
...
         self.progress_info = SimulationProgressInfo()
-        self._simulation_id = redis_job_id
+        self.simulation_id = redis_job_id
 
-        self._init(redis_job_id)
-
-        self._external_events = SimulationExternalEvents(
-            redis_job_id, self._setup.config, self._status, self.progress_info, self.area)
+        self._external_events = SimulationExternalEvents(self)
+        self._init()
```
fievelk, line 529: This is the actual fix.

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -496,7 +496,7 @@ def __init__(self, setup_module_name: str, simulation_config: SimulationConfig,
                  no_export: bool = False, export_path: str = None,
                  export_subdir: str = None, redis_job_id=None, enable_bc=False,
                  slot_length_realtime: Duration = None, incremental: bool = False):
-        self._status = SimulationStatusManager(
+        self.status = SimulationStatusManager(
```
spyrostz, line 499: I assume that this was needed only in order to be accessed from the tests ? Reason I am asking is because I cannot see where this might be required, since the status is self-contained in the Simulation classes.

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -925,7 +925,7 @@ def __init__(self, simulation: Simulation) -> None:
         self.live_events = LiveEvents(simulation.config)
         self.redis_connection = RedisSimulationCommunication(
             simulation_status=simulation.status,
-            simulation_id=simulation._simulation_id,
+            simulation_id=simulation.simulation_id,
```
spyrostz, line 928: Forget my question about the _status member, I assume it is needed for the same reason as this one.

spyrostz (approved):
LGTM

src/gsy_e/gsy_e_core/simulation.py
fievelk: Exactly. For the other reviewers: this PR changes the interface of `SimulationExternalEvents` so that it now accepts just one parameter (`simulation`). This is a refactoring to make the code more OOP oriented, which is our paradigm; if you're curious, you can find some more info in the "Refactoring" book, and a short visual description on the website ([Introduce Parameter Object](https://refactoring.com/catalog/introduceParameterObject.html) and [Preserve Whole Object](https://refactoring.com/catalog/preserveWholeObject.html)).

Because of this, we now access directly some attributes like `Simulation._status` and `Simulation_simulation_id` from another class. So these parameters shouldn't be considered private anymore, but rather public. That's why I removed their underscores.

hannesdiedrich: Makes sense, thanks!
One nit though: IMO it makes more sense that the `SimulationExternalEvents` class lives at the top among the other helper classes for the `Simulation` class (as before).

hannesdiedrich (dismissed):
One nit comment. LGTM

src/gsy_e/gsy_e_core/simulation.py
fievelk: The reason why I moved it is to avoid errors with the type hints classes (that need to be declared before they're used). I could probably change it (e.g. using a string). However, I think this class should actually live in a different module. I think you mentioned it in one of your slack messages a few days ago.

mrsaemir (approved):
LGTM

spyrostz (approved):
LGTM

PR #1528, opened 2022-09-13
GSYE-380: Add WindUserProfileStrategy to leaves.py in order to be sel…

hannesdiedrich:
…ected by gsy-web.

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-380


src/gsy_e/models/leaves.py
```diff
@@ -76,6 +77,7 @@ class LeafBase:
     instantiated and serialized in a more compact format
     """
     strategy_type = None
+    strategy = None
```
mrsaemir, line 80: Can you please explain why this is needed?

mrsaemir (commented):
Just a small question.

src/gsy_e/models/leaves.py
hannesdiedrich: In order to make pylint happy.

mrsaemir (approved):
LGTM

hannesdiedrich:
> LGTM, I will relay it to the respective Forward Strategy in the context of the EB template strategies ticket.

Sorry, I do not follow here. Can you please explain?

spyrostz:
> > LGTM, I will relay it to the respective Forward Strategy in the context of the EB template strategies ticket.
> 
> Sorry, I do not follow here. Can you please explain?

Nothing to do here for this PR, to explain a bit more what I meant, the purpose of this is to be the connection of the new Wind Forward strategy, which is not yet deployed on dev / master. That's what I meant here, that since I am currently developing this (essentially the LoadForwardStrategy and PVForwardStrategy) in gsy-e, I will adapt the area_serializer.py / leaves.py on my end too in order to handle a future WindForwardStrategy. Hope that it is more clear now, my comment was indeed very vague.

PR #1530, opened 2022-09-14
GSYE-375: Fix timeslot formatting in _read_forward_markets_stats_to_dict

mrsaemir:
Formatted time_slot should have the same schema as time_slot in Bid/Offer/Trade objects.

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -24,7 +24,8 @@ from gsy_framework.constants_limits import (DATE_TIME_FORMAT, DATE_TIME_UI_FORMA
 from gsy_framework.enums import AvailableMarketTypes
 from gsy_framework.results_validator import results_validator
 from gsy_framework.sim_results.all_results import ResultsHandler
-from gsy_framework.utils import get_json_dict_memory_allocation_size
+from gsy_framework.utils import (datetime_to_string_incl_seconds,
+                                 get_json_dict_memory_allocation_size)
 from pendulum import DateTime
 
 from gsy_e.gsy_e_core.sim_results.offer_bids_trades_hr_stats import OfferBidTradeGraphStats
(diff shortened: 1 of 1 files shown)


src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -217,7 +218,7 @@ def _read_forward_markets_stats_to_dict(
 
         for market_type, market in area.forward_markets.items():
             for time_slot in market.market_time_slots:
-                time_slot_str = time_slot.format(DATE_TIME_FORMAT)
+                time_slot_str = datetime_to_string_incl_seconds(time_slot)
```
hannesdiedrich, line 221: Lets stick to using pendulum on the gsy-e side and use the DATE_TIME_FORMAT_SECONDS format:
```suggestion
                time_slot_str = time_slot.format(DATE_TIME_FORMAT_SECONDS)
```

hannesdiedrich: Why are the seconds needed BTW? Is the a requirement of the B2B?

mrsaemir: The format for time_slot here and in the Bid/Offer/Trade object should be the same. Otherwise my [assertions](https://github.com/gridsingularity/gsy-framework/blob/86001fcfe3cc8862f8f9c2c57c50639e12d5c2cf/gsy_framework/sim_results/electric_blue/aggregate_results.py#L80), fail. I could have written my assertions in some other way to get along with this difference, but the cost is higher that way.

hannesdiedrich: This has to be decided collectively, because until now, a time_slot_str never included the seconds because we did not see the time when there will be time_slots that are shorter than one minute. 
This is why I vote for leaving the seconds out, even if it means some more work for you, sorry.

mrsaemir: There's no problem on my side. I just wanted to avoid unnecessary conversions on gsy-framework.

PR #1532, opened 2022-09-15
GSYE-388: Add SimulationTimeManagerScm in order to handle slowdown an…

hannesdiedrich:
…d realtime SCM simulations
-

### DIFF
--- a/src/gsy_e/gsy_e_core/simulation.py
+++ b/src/gsy_e/gsy_e_core/simulation.py
@@ -247,16 +247,88 @@ class SimulationTimeManager:
         """
         if gsy_e.constants.RUN_IN_REALTIME:
             tick_runtime_s = time() - self.tick_time_counter
-            sleep(abs(config.tick_length.seconds - tick_runtime_s))
+            sleep_time_s = config.tick_length.seconds - tick_runtime_s
         elif self.slot_length_realtime:
             current_expected_tick_time = self.tick_time_counter + self.tick_length_realtime_s
             sleep_time_s = current_expected_tick_time - now().timestamp()
-            if sleep_time_s > 0:
-                sleep(sleep_time_s)
-                log.debug("Tick %s/%s: Sleep time of %s s was applied",
-                          tick_no + 1, config.ticks_per_slot, sleep_time_s)
+        else:
+            return
+
+        if sleep_time_s > 0:
+            sleep(sleep_time_s)
+            log.debug("Tick %s/%s: Sleep time of %s s was applied",
+                      tick_no + 1, config.ticks_per_slot, sleep_time_s)
 
         self.tick_time_counter = time()
+
+
+@dataclass
+class SimulationTimeManagerScm:
+    """Handles simulation time management."""
+    start_time: DateTime = now(tz=TIME_ZONE)
+    paused_time: int = 0  # Time spent in paused state, in seconds
+    slot_length_realtime: duration = None
+    slot_time_counter: float = time()
+
+    def reset(self, not_restored_from_state: bool = True) -> None:
+        """
+        Restore time-related parameters of the simulation to their default values.
+        Mainly useful when resetting the simulation.
+        """
+        self.slot_time_counter = time()
+        if not_restored_from_state:
+            self.start_time = now(tz=TIME_ZONE)
+            self.paused_time = 0
+
+    def handle_slowdown_and_realtime_scm(self, slot_no: int, slot_count: int,
+                                         config: SimulationConfig) -> None:
+        """
+        Handle simulation slowdown and simulation realtime mode, and sleep the simulation
+        accordingly for SCM simulations.
+        """
+
+        slot_length_realtime_s = self.slot_length_realtime.total_seconds()
+
+        if gsy_e.constants.RUN_IN_REALTIME:
+            slot_runtime_s = time() - self.slot_time_counter
+            sleep_time_s = config.slot_length.total_seconds() - slot_runtime_s
+        elif slot_length_realtime_s:
+            current_expected_tick_time = self.slot_time_counter + slot_length_realtime_s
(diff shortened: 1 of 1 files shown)


src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -867,11 +945,11 @@ def _deactivate_areas(self, area: "AreaBase"):
         pass
 
     def _execute_simulation(
-            self, slot_resume: int, tick_resume: int, console: NonBlockingConsole = None) -> None:
-
-        slot_count, slot_resume, tick_resume = (
-            self._time.calculate_total_initial_ticks_slots(
-                self._setup.config, slot_resume, tick_resume, self.area))
+            self, slot_resume: int, _tick_resume: int, console: NonBlockingConsole = None) -> None:
+        gsy_e.constants.RUN_IN_REALTIME = True
```
cruky, line 949: It looks like this line makes every simulation to run in a real time.
```suggestion

```

hannesdiedrich: Nice catch! I forgot to remove this after testing. Will do it now. Thanks!!!

src/gsy_e/gsy_e_core/simulation.py
```diff
@@ -422,7 +498,7 @@ def init_results(self, redis_job_id: str, area: "AreaBase",
             self._export = CoefficientExportAndPlot(
                 area, self.export_path, self.export_subdir, self._endpoint_buffer)
 
-    def update_results(
+    def update_and_send_results(
```
hannesdiedrich, line 501: @spyrostz I found this thanks to pylint. But I think we should just overwrite the `update_and_send_results` in the `CoefficientSimulation` in order to also handle the pause functionality.
Otherwise the following error will be shown:
```
  File "/Users/hannesd/Envs/gsy-e/lib/python3.8/site-packages/rq/worker.py", line 1043, in perform_job
    rv = job.perform()
  File "/Users/hannesd/Envs/gsy-e/lib/python3.8/site-packages/rq/job.py", line 814, in perform
    self._result = self._execute()
  File "/Users/hannesd/Envs/gsy-e/lib/python3.8/site-packages/rq/job.py", line 837, in _execute
    result = self.func(*self.args, **self.kwargs)
  File "/Users/hannesd/gsy/gsy-e/src/gsy_e/gsy_e_core/exchange_jobs.py", line 40, in start
    launch_simulation_from_rq_job(**payload, job_id=current_job.id)
  File "/Users/hannesd/gsy/gsy-e/src/gsy_e/gsy_e_core/rq_job_handler.py", line 124, in launch_simulation_from_rq_job
    run_simulation(setup_module_name=scenario_name,
  File "/Users/hannesd/gsy/gsy-e/src/gsy_e/gsy_e_core/simulation.py", line 1054, in run_simulation
    simulation.run()
  File "/Users/hannesd/gsy/gsy-e/src/gsy_e/gsy_e_core/simulation.py", line 685, in run
    self._execute_simulation(initial_slot, tick_resume)
  File "/Users/hannesd/gsy/gsy-e/src/gsy_e/gsy_e_core/simulation.py", line 955, in _execute_simulation
    self._handle_paused(console)
  File "/Users/hannesd/gsy/gsy-e/src/gsy_e/gsy_e_core/simulation.py", line 831, in _handle_paused
    self._results.update_and_send_results(
  File "/Users/hannesd/gsy/gsy-e/src/gsy_e/gsy_e_core/simulation.py", line 504, in update_and_send_results
    raise NotImplementedError
```
Or did I misunderstand the reason for raising the `NotImplementedError` ?

hannesdiedrich: Ah I see now why overwriting is not possible: Because the interface is different.
The I need to handle it in the `_handle_paused` method and decide there which `SimulationResultsManager` is used

hannesdiedrich: Please have a look at my latest commit: https://github.com/gridsingularity/gsy-e/pull/1532/commits/bacf0e257a56dfee659bf6f64f44f819ccab9338

spyrostz: Hmmm this will definitely work but I am thinking whether it makes sense to harmonize the interface if it is possible. I would recommend a separate PR though in order to be dealt with in isolation if possible. Or a TODO comment, even though it is something outside our usual workflow it can be added so that we notice it later along the way.

hannesdiedrich: I have an idea: One could add the scm manager instance to the `CoefficientSimulationResultsManager` via an extra step (method) and not pass it in the `update_send_coefficient_results`. This is easily done. If you agree, i can quickly add it.

spyrostz: Interesting, therefore in order to update there will need to be 2 distinct method calls: one that updates the scm_manager, and one that actually calls the update_send_results (which would be overridden in the CoefficientSimulationResultsManager class, and would replace the update_send_coefficient_results). I like this and support this, thanks a lot for dealing with it !

spyrostz (dismissed):
LGTM

spyrostz (approved):
LGTM

cruky (approved):
LGTM

mrsaemir (approved):
LGTM, but I just wanted to know wasn't it possible to inherit the SimulationTimeManager rather than hard copying all of it for SimulationTimeManagerSCM?

hannesdiedrich:
> LGTM, but I just wanted to know wasn't it possible to inherit from the SimulationTimeManager rather than hard copying all of it for SimulationTimeManagerSCM?

It was not possible. Please have a look at the members of the `SimulationTimeManagerScm` , e.g. there is no `tick_time_counter` but `slot_time_counter`. Also, I am overwriting every method with a different logic. So there is no benefit from inheriting.

PR #1533, opened 2022-09-15
Refactor simulation.py and move all utility classes into a separte pa…

hannesdiedrich:
…ckage

## Proposed changes

-


spyrostz (commented):
Minor and mostly cosmetic, however I keep on thinking that a better structure would be this: 
- simulation/
    - `__init__.py` (should include the former simulation.py module)
    - external_events.py
    - progress_info.py
    - results_manager.py
    - setup.py
    - status_manager.py
    - time_manager.py

This will avoid the `simulation_` prefix, will be significantly shorter when importing (therefore not requiring that many line breaks) and avoids changing the imports of the Simulation class (since it will be the same as now, `from gsy_e.gsy_e_core.simulation import Simulation`) . Let me know what you think, if you prefer the existing structure then it also works for me.

hannesdiedrich:
> Minor and mostly cosmetic, however I keep on thinking that a better structure would be this:
> 
> * simulation/
>   
>   * `__init__.py` (should include the current simulation.py module)
>   * external_events.py
>   * progress_info.py
>   * results_manager.py
>   * setup.py
>   * status_manager.py
>   * time_manager.py
> 
> This will avoid the `simulation_` prefix, will be significantly shorter when importing (therefore not requiring that many line breaks) and avoids changing the imports of the Simulation class (since it will be the same as now, `from gsy_e.gsy_e_core.simulation import Simulation`) . Let me know what you think, if you prefer the existing structure then it also works for me.

You convinced me. Let me try it out.

src/gsy_e/gsy_e_core/simulation/__init__.py
```diff
@@ -0,0 +1,521 @@
+"""
```
fievelk, line 1: We've previously talked about avoiding putting code inside `__init__.py` modules. I'd suggest moving all this code into a `simulation.py` module for clarity and to follow on that.

src/gsy_e/gsy_e_core/simulation/external_events.py
```diff
@@ -0,0 +1,51 @@
...
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program.  If not, see <http://www.gnu.org/licenses/>.
+"""
```
fievelk, line 17: Nit: I found out that it's usually better to use comments for licences, and not docstrings. This is because license is not code documentation, and docstrings end up being part of the module documentation. You can see it for example like this:
```py
from gsy_e.gsy_e_core import simulation

print(simulation.__doc__)
```
This should print something that describes the content of the `simulation` module. Instead, right now it prints the license of GSy (and it will do the same for all modules, which makes this documentation unusable).
What I'd suggest for now and for the future is to use normal comments with `#`.

src/gsy_e/gsy_e_core/simulation/results_manager.py
```diff
@@ -0,0 +1,218 @@
+"""
```
fievelk, line 1: We have a package called `gsy_e_core.sim_results`. Why don't we put this module (`results_manager`) inside of that package instead? I think it would be a more suitable place for it. What do you think?

src/gsy_e/gsy_e_core/simulation/setup.py
```diff
@@ -0,0 +1,95 @@
+"""
```
fievelk, line 1: Nit: I'm not sure, but maybe calling this module `simulation_setup.py` would be better to respect the naming of the class (making it easier to find). I know this leads to the repetition of the namespace (i.e. `simulation.simulation_setup`) and that's why I'm not completely sure this is a good practice. But I couldn't find anything about it. What's your opinion?

fievelk (commented):
Left some comments/nits

src/gsy_e/gsy_e_core/simulation/__init__.py
fievelk: I just noticed @spyrostz's [comment here](https://github.com/gridsingularity/gsy-e/pull/1533#pullrequestreview-1109126282):
> [...] avoids changing the imports of the Simulation class (since it will be the same as now, from `gsy_e.gsy_e_core.simulation import Simulation`) [...]

This can be achieved by importing inside of `__init__.py` just the classes you want to expose on the outside. For example:
```py
# __init__.py
from gsy_e.gsy_e_core.simulation.simulation import Simulation

__all__ = [
    "Simulation"
]
```
So this will preserve the possibility to just use
```py
from gsy_e.gsy_e_core.simulation import Simulation
```
as before.

src/gsy_e/gsy_e_core/simulation/setup.py
fievelk: I just saw the comment from spyros about the modules' names ([here](https://github.com/gridsingularity/gsy-e/pull/1533#pullrequestreview-1109126282)). If you've already agreed on that, that's good to go.

src/gsy_e/gsy_e_core/simulation/__init__.py
hannesdiedrich: OK, will do.

src/gsy_e/gsy_e_core/simulation/external_events.py
hannesdiedrich: OK, thanks for the info. We should discuss and decide about this in the team. Does not make sense, that I change this here in the few modules. It should be changed globally at once IMO. So How about a short discussion on slack and a ticket that come out of it?

src/gsy_e/gsy_e_core/simulation/results_manager.py
hannesdiedrich: I do not agree, because the results manager is "only" the caller code of the sim_results modules (amongst others) and belongs into the simulation package. All this code used to be in the `Simulation` class until a few months back.

src/gsy_e/gsy_e_core/simulation/external_events.py
fievelk: Sure, I can bring this on slack!

fievelk (approved):

spyrostz (approved):
LGTM

PR #1534, opened 2022-09-20
Feature/gsye 333

spyrostz:

gsy-e changes for the forward market strategies. The changes consist of the following:

- allow a CLI user to run a simulation using the forward markets
- new OrderUpdater class that is responsible for calculating the energy rate of the updated bid and offer in a market slot. Works for all forward markets
- added load and PV strategies dedicated for the forward markets 
- adapted energy parameters for the forward markets to retrieve the available energy of the asset, and also the energy that has been already posted in bids / offers. That  way the orders in one market will be aware of energy posted in other markets, and not include it in the orders that they are about to post. 
- Added parameter for the percentage of the capacity that needs to be posted per-market. 
- fixed the forward market creation / rotation in order to not create markets whose close timestamp is in the past
- Unit tests 

## Proposed changes

In order to test this locally, please use the newly-created setup file named forward.py:
`gsy-e -l INFO run -t 60s -s 15m -d 365d --setup forward `


src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -77,31 +243,70 @@ def event_traded_energy(
         for time_slot, energy_value_kWh in trade_profile.items():
             self._state.decrement_energy_requirement(
                 purchased_energy_Wh=energy_value_kWh * 1000,
-                time_slot=time_slot,
+                time_slot=time_slot.set(tz=pendulum.UTC),
```
hannesdiedrich, line 246: This alarms me. IMO this should ne already done in `generate_trade_profile` or more specific in `_expand_profile_slots`. What do you think? 
Of course, this can be done in a separate PR.

src/gsy_e/models/strategy/forward/order_updater.py
```diff
@@ -0,0 +1,68 @@
...
+    update_interval: duration
+    initial_rate: float
+    final_rate: float
+    capacity_percent: float
+
+    def __post_init__(self):
+        assert 0.0 <= self.capacity_percent <= 100.0
+
+
+class OrderUpdater:
```
hannesdiedrich, line 23: Finally an understandable and easy to maintain implementation of the update of the energy rate! Thanks a lot!

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -119,6 +324,21 @@ def event_traded_energy(
             product_type: One of the available market types.
         """
         assert self._profile_generator is not None
+
+        if product_type == AvailableMarketTypes.INTRADAY:
```
hannesdiedrich, line 328: Minor, and just see this no: Why not naming it `market_type`?

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -21,37 +21,190 @@
...
+        """Retrieve the already posted energy on this market slot."""
+        if market_type == AvailableMarketTypes.INTRADAY:
+            if market_slot not in self._posted_energy_kWh:
+                self._posted_energy_kWh[market_slot] = 0.
+
+            return self._posted_energy_kWh[market_slot]
+        if market_type == AvailableMarketTypes.DAY_FORWARD:
+            for slot in self.day_forward_slots(market_slot):
+                if slot not in self._posted_energy_kWh:
+                    self._posted_energy_kWh[slot] = 0.
```
hannesdiedrich, line 55: What do you think about using `defaultdict` for `self._posted_energy_kWh` and save this for loop here?
```suggestion
from collections import defaultdict
self._posted_energy_kWh: Dict =defaultdict(lambda: 0.0)
```

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -77,31 +243,70 @@ def event_traded_energy(
...
+        reference_slot = market_slot.set(hour=12, minute=0)
+
+        if self._state.get_energy_production_forecast_kWh(
+                reference_slot) <= FLOATING_POINT_TOLERANCE:
+            scaling_factor = 0.
+        else:
+            scaling_factor = (
+                    self._state.get_available_energy_kWh(reference_slot) /
+                    self._state.get_energy_production_forecast_kWh(reference_slot))
+        return abs(scaling_factor) * self.capacity_kWh
```
hannesdiedrich, line 297: I know, I think you explained this once already to me, but I still need to ask again.
AFAIU, the capacity of the PV is scaled with the available_energy.
So, if the 
production_forecast_kWh = 2
available_energy_kWh = 1 (1 was already traded)
the scaling factor is 0.5
What is the capacity_kWh here? It is the integral over production curve, right? So it should be grater that 2 kWh, right?
Thanks a lot for bearing with me!

src/gsy_e/models/strategy/forward/pv.py
```diff
@@ -0,0 +1,85 @@
...
+                  for offer in market.slot_offer_mapping[market_slot]
+                  if offer.seller == self.owner.name]
+        for offer in offers:
+            market.delete_offer(offer)
+
+    def post_order(self, market: "ForwardMarketBase", market_slot: DateTime):
+        order_rate = self._order_updaters[market][market_slot].get_energy_rate(
+            self.area.now)
+
+        capacity_percent = self._order_updaters[market][market_slot].capacity_percent / 100.0
```
hannesdiedrich, line 47: So I understand that the `capacity_percent` contains information about how much of the capacity should be placed in which kind of market, right?

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: This algorithm is probably an overkill, however let me elaborate a bit further on the rationale behind this. First of all, all these are peak values and not integral values, including capacity_kWh. The reason for adding this part of the code, was because I needed to be able to calculate the scaled available peak energy for a duration of time (depending on the market type). In order to calculate this, I wanted to have one time slot as reference, which I knew that had non-zero production / consumption, and for all our profiles the most straightforward option is to use 12:00 as a reference point. However, it might be that the peak / max energy of the profile (essentially the capacity_kWh) is not at 12:00, but at some other time during the day. In order to get the scaled peak value (according to the already traded energy) I am calculating the scaling factor at 12:00 (essentially the ratio of available energy / production forecast at 12:00), and using this scaling factor, I calculate the peak of the available energy profile by multiplying with the peak value of the production profile. 

If any of the above does not make sense or is not clear let me know to further clarify this, thanks!

src/gsy_e/models/strategy/forward/pv.py
```diff
@@ -0,0 +1,85 @@
...
+        max_energy_kWh = self._energy_params.capacity_kWh * capacity_percent
+        energy_kWh = self._energy_params.get_available_energy_kWh(
+            market_slot, market.market_type)
+
+        posted_energy_kWh = self._energy_params.get_posted_energy_kWh(
+            market_slot, market.market_type)
+        energy_kWh -= posted_energy_kWh
+
+        energy_kWh = min(energy_kWh, max_energy_kWh)
```
hannesdiedrich, line 56: Minor, but the following way of implementation would look clearer to my eyes. 
```suggestion
        available_energy_kWh = self._energy_params.get_available_energy_kWh(
            market_slot, market.market_type)

        posted_energy_kWh = self._energy_params.get_posted_energy_kWh(
            market_slot, market.market_type)

        energy_kWh = min(available_energy_kWh-posted_energy_kWh, max_energy_kWh)
```

src/gsy_e/models/strategy/forward/pv.py
spyrostz: Precisely. And the reason for its existence is a UI input field in the design which is supposed to control that.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: I think that the code works even without the timezone assignment, let me give it a try by removing all these. I will also adapt your suggestion to strip the timezone information from the point that this profile is generated. Thanks!

src/gsy_e/models/area/market_rotators.py
```diff
@@ -53,26 +53,31 @@ class ForwardMarketRotatorBase:
...
+                    current_time))
 
 
 class IntradayMarketRotator(ForwardMarketRotatorBase):
     """Handles market rotation for the intraday market block"""
 
+    @staticmethod
+    def _get_last_slot_to_be_deleted(current_time: DateTime) -> DateTime:
+        return current_time + duration(minutes=15)
+
```
hannesdiedrich, line 80: I thought this was there from the first implementation. Or do I remember incorrectly? Was it removed due to a change in design?

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -77,31 +243,70 @@ def event_traded_energy(
...
+    @property
+    def capacity_kWh(self):
+        """Capacity of the PV in kWh."""
+        return convert_kW_to_kWh(self.capacity_kW, self._area.config.slot_length)
+
+    def get_available_energy_kWh(
+            self, market_slot: pendulum.DateTime, market_type: AvailableMarketTypes):
+        """Get the available offer energy of the PV."""
+        if market_type == AvailableMarketTypes.INTRADAY:
+            self._state.get_available_energy_kWh(market_slot)
```
hannesdiedrich, line 281: This line dies not do anything AFAIU. Did you want to return the value early?

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
hannesdiedrich: Ok, so if the peak of the production curve is at noon, the following applies:

```python
self._state.get_energy_production_forecast_kWh("12:00") = self.capacity_kWh
```
?
I think I start to understand now. You want to get the available_energy_kWh for any market slot (e.g. 10:15) from the one at noon, because of the mentioned reasons. You do not care about the other market slots on the day, you only want to read and write to the reference_slot. Right?

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -21,37 +21,190 @@
...
+        """Retrieve the internal state of the asset."""
+
+    @staticmethod
+    def day_forward_slots(market_slot: pendulum.DateTime):
+        """Get the market slots for the day forward market."""
+        return [
+            market_slot,
+            market_slot + pendulum.duration(minutes=15),
+            market_slot + pendulum.duration(minutes=30),
+            market_slot + pendulum.duration(minutes=45)]
```
hannesdiedrich, line 145: So the DayForward is really only a HourForward any more right?

tests/strategies/forward/test_strategies.py
```diff
@@ -0,0 +1,196 @@
...
+            list(strategy._order_updaters[market_object].keys()) ==
+            market_object.market_time_slots)
+        assert (
+            len(strategy._order_updaters[market_object].keys()) ==
+            expected_order_updater_count)
+
+        with patch("gsy_e.models.area.Area.now", new_callable=PropertyMock) as now_mock:
+            now_mock.return_value = next_slot_timestamp
+            area.cycle_markets()
+            # strategy.event_market_cycle()
```
hannesdiedrich, line 104: this is not needed, because you are doing it in line 92 already right? Otherwise I would expect this to be needed.

tests/strategies/forward/test_strategies.py
```diff
@@ -0,0 +1,196 @@
...
+    AvailableMarketTypes.DAY_FORWARD: OrderUpdaterParameters(duration(minutes=30), 42, 22, 20),
+    AvailableMarketTypes.WEEK_FORWARD: OrderUpdaterParameters(duration(days=1), 53, 33, 20),
+    AvailableMarketTypes.MONTH_FORWARD: OrderUpdaterParameters(duration(weeks=1), 66, 39, 20),
+    AvailableMarketTypes.YEAR_FORWARD: OrderUpdaterParameters(duration(months=1), 72, 56, 20)
+}
+
+
+@pytest.fixture(name="forward_strategy_fixture", params=[
+    (ForwardLoadStrategy, load_parameters,),
+    (ForwardPVStrategy, pv_parameters, )])
```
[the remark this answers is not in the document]
hannesdiedrich: Ah no. How does it work. Are all test run with both params?

hannesdiedrich (commented):
Thanks a lot.
Have some comments and questions and will follow up on the answers.
Can you please update the wiki page for the forward market rotation?
https://gridsingularity.atlassian.net/wiki/spaces/D3A/pages/3242885121/Forward+Markets+and+Rotation
It seems that there are some updates. Thanks!

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -21,37 +21,190 @@
...
+
+        reference_slot = market_slot.set(hour=12, minute=0, tz=pendulum.UTC)
+
+        if self._state.get_desired_energy_Wh(reference_slot) <= FLOATING_POINT_TOLERANCE:
+            scaling_factor = 0.
+        else:
+            scaling_factor = (
+                    self._state.get_energy_requirement_Wh(reference_slot) /
+                    self._state.get_desired_energy_Wh(reference_slot))
+        return abs(scaling_factor) * self.capacity_kWh
```
fievelk, line 194: Could you please explain this part? What's the "available energy" for a load? The docstring says:
> Get the available bid energy of the load

but it sounds a bit counterintuitive to me to grasp. Is the "available bid energy" the same thing as "required energy"? If so, could we use the same "energy requirement" naming we used in the state?

Also, regarding the scaling factor. AFAIU, this means something like:

$\displaystyle scaling\ factor = \frac{energy\ still\ needed}{original\ total\ energy\ needed}$

Is this correct? What do we obtain by multiplying this for the capacity?

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -21,37 +21,190 @@
...
+
     def event_activate_energy(self, area):
         """Initialize values that are required to compute the energy values of the asset."""
         self._area = area
         self._profile_generator = ForwardTradeProfileGenerator(
-            peak_kWh=convert_kW_to_kWh(self.capacity_kW, self._area.config.slot_length))
+            peak_kWh=self.capacity_kWh)
+        for i in range(4):
```
fievelk, line 201: This `event_activate_energy()` method doesn't distinguish between intraday/daily/etc. market types. How does this work in the case of an intraday market?

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
fievelk: pendulum's default timezone is already UTC when creating new DateTime objects:
```py
>>> pendulum.datetime(2022, 1, 1)
DateTime(2022, 1, 1, 0, 0, 0, tzinfo=Timezone('UTC'))
```
so (in theory) we should be already fine unless some part of the code explicitly sets this to something different, or in cases where we don't use pendulum objects.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
fievelk, line 281: Nit:
```suggestion
            self.state.get_available_energy_kWh(market_slot)
```
(same in the rest of the module)

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
fievelk: @hannesdiedrich I think this was a decision that changed a few times (it was "market type" initially, then "exchange", then we settled on "product type"). I think the important thing would be to always use the same naming though, so to replace the occurrences of `market_type` with `product_type`.

fievelk (commented):
I finished reviewing the area, rotators, markets and energy parameters part. I still need to check the strategy part but I'll send my current comments now.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
hannesdiedrich: :-( ok, Thanks! I really wonder why `market type` was not suitable any more. Anyway. Will name it `product_type` from now on then.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: > Ok, so if the peak of the production curve is at noon, the following applies:

> self._state.get_energy_production_forecast_kWh("12:00") = self.capacity_kWh

Exactly! I think that we have reached a common understanding, however just to make it 100% clear: the reason for this calculation is that the peak of the profile (essentially the capacity_kWh) might not be at 12:00 but at 10:15 to use your example. Therefore I am calculating the "scaling factor" using an arbitrary timeslot which I know will have non-zero energy value (12:00 typically) and then multiply the self.capacity_kWh (which is really the profile peak).

As an aftermath of our discussion, I think that the `capacity_kWh` naming is a bit problematic here. If you agree I will rename it to `peak_energy_kWh` in order to be more understandable for the reader. Thanks!

fievelk: I think it would be useful for our future selves to add part of this conversation to the code in form of comments/docstrings!

tests/strategies/forward/test_strategies.py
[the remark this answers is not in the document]
fievelk: @hannesdiedrich yes: all tests that use this fixture will be run twice (once for each param).
- [Pytest docs: parameterizing fixtures](https://docs.pytest.org/en/6.2.x/fixture.html#parametrizing-fixtures)

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: Fixing this, thanks!

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: Will do so thanks!

src/gsy_e/models/strategy/forward/pv.py
spyrostz: Fixed, thanks

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: Indeed, thanks for catching that!

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: Yes, in order to be similar to the day-ahead EPEX spot market. Essentially the market slots are hourly, but expand over the next day. Therefore there are 24 slots from 00:00 till 23:45 of the next day.

tests/strategies/forward/test_strategies.py
spyrostz: Missed that, it is indeed not needed. I initially wanted to cycle only the strategy, but during the implementation it was proven that some parts of the Area class need to be cycled too. Will change this.

src/gsy_e/models/area/market_rotators.py
spyrostz: Yes, there was an attempt to use the closing time of each market for recycling (essentially to compare the expected closing time of the market with the current time and only close markets when the current time is greater than or equal to the closing time) however due to complications in the unit tests we ditched it again in favor of the old approach.

src/gsy_e/models/area/market_rotators.py
```diff
@@ -53,26 +53,31 @@ class ForwardMarketRotatorBase:
     def __init__(self, markets: ForwardMarketBase):
         self.markets = markets
 
+    @staticmethod
+    def _get_last_slot_to_be_deleted(current_time: DateTime) -> DateTime:
```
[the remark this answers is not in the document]
spyrostz: Will do so thanks!

src/gsy_e/models/area/market_rotators.py
```diff
@@ -53,26 +53,31 @@ class ForwardMarketRotatorBase:
...
         if self._is_it_time_to_rotate(current_time):
-            slots_deleted = []
-            for delivery_time, market_slot_info in self.markets.open_market_slot_info.items():
-                if market_slot_info.closing_time <= current_time:
-                    self.markets.delete_orders_in_old_future_markets(
-                        last_slot_to_be_deleted=delivery_time)
-                    slots_deleted.append(delivery_time)
-            for slot_deleted in slots_deleted:
-                self.markets.open_market_slot_info.pop(slot_deleted)
+            self.markets.delete_orders_in_old_future_markets(
```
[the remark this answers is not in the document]
spyrostz: You got it right, and the comment is wrong here. time_slot is referring to the delivery time of the energy. Therefore it really is not the "market slot of creation" (we have another member of the BaseBidOffer called creation_time that reflects the time of the creation of the order). Let me adapt the comment accordingly in gsy-framework

src/gsy_e/models/market/forward.py
```diff
@@ -129,44 +142,58 @@ def _get_end_time(current_time: DateTime) -> DateTime:
     def _get_market_slot_duration(_current_time: DateTime, _config) -> duration:
         return duration(minutes=15)
 
-    def _calculate_closing_time(self, delivery_time: DateTime) -> DateTime:
-        """Closing time of the intraday market is 15 mins before delivery."""
-        return delivery_time - duration(minutes=15)
+    @property
+    def _time_from_market_close_till_delivery(self) -> duration:
+        return duration(minutes=15)
```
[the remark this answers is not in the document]
spyrostz: Since the matter was mentioned once more, let me retry the approach that we had before my PR, where rotation was controlled by the closing time of each market (therefore implicitly by the time_from_market_close_till_delivery). I think this will simplify the process

src/gsy_e/models/market/future.py
```diff
@@ -195,14 +195,25 @@ def delete_orders_in_old_future_markets(self, last_slot_to_be_deleted: DateTime
...
+        """
+        Retrieves the time duration from the time that the market closes till the time that the
+        traded energy should be delivered.
+        """
+        return duration(minutes=0)
+
     def _create_future_market_slots(
             self, start_time: DateTime, end_time: DateTime,
-            config: "SimulationConfig") -> List[DateTime]:
+            config: "SimulationConfig", current_market_time_slot: DateTime) -> List[DateTime]:
```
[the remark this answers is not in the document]
spyrostz: Unfortunately these are not implemented for the future markets but only for the forward markets. I'll implement these methods in the future market class in order for them to be overridden by the forward markets, and that way we will be able to use them.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -21,37 +21,190 @@
...
+
+class ForwardEnergyParams(ABC):
+    """Common abstract base class for the energy parameters of the forward strategies."""
+    def __init__(self):
+        self._posted_energy_kWh: Dict = {}
+
+    def get_posted_energy_kWh(
+            self, market_slot: pendulum.DateTime, market_type: AvailableMarketTypes) -> float:
+        """Retrieve the already posted energy on this market slot."""
+        if market_type == AvailableMarketTypes.INTRADAY:
```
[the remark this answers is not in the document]
spyrostz: I thought about it originally too, however it looks like an overkill to my eyes, because there are only 2 methods that share this behavior (get_available_energy and event_traded_energy). If you insist and think that it is worth it let me know and I will give it a try.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -21,37 +21,190 @@
...
+    @property
+    def capacity_kWh(self):
+        """Capacity of the load in kWh."""
+        return convert_kW_to_kWh(self.capacity_kW, self._area.config.slot_length)
+
+    def get_available_energy_kWh(
+            self, market_slot: pendulum.DateTime, market_type: AvailableMarketTypes):
+        """Get the available bid energy of the load."""
+        if market_type == AvailableMarketTypes.INTRADAY:
+            return self._state.get_energy_requirement_Wh(market_slot) / 1000.0
```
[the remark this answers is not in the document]
spyrostz: I am actually not sure whether in these cases it is better to access the private or the public property because there are (dis)advantages to both. Since you prefer the public property though I will replace it.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: I added a general comment that explains it already, however let me also amend the docstring in order to better reflect the purpose of this method. To recap, available energy == energy that the asset can generate / consumed , that has not been traded yet. Therefore the formula for the scaling factor is the following:
`scaling factor = not-traded-energy-at-12:00 / total-energy-needed-at-12:00` 

As for why it is needed, the reason is that we always use the peak energy as a unit for orders. The peak energy though is not necessarily at 12:00 on every profile. In order to avoid iterating through the profile and finding the peak energy slot, we instead use a timeslot that we know has nonzero energy (12:00 is one of these), we calculate the scaling factor at 12:00, and then we use this scaling factor in order to scale the peak value of the profile (which we know is self.peak_energy_kWh), which we can then use to post orders (since the scaling factor has removed from the peak the energy that has already been traded). Hope that the above is clear, let me know if further explanations are needed here!

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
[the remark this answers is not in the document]
fievelk: AFAIS the conditionals are also in `increment_posted_energy()`, `decrement_posted_energy()` and `get_posted_energy_kWh()` (in the parent class). In my opinion it would make sense to find a way to "simplify" these by moving the conditional logic into different classes. We can also wait for other opinions of course

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: This method is responsible for setting up the production / consumption forecast profile of an asset. It does not distinguish different markets, since the energy values that can be provided at each time period is constant and does not depend on the market type. Therefore it has to be done only once for the yearly market, since this is the market with the longest duration, in order to accomodate it. The other markets will operate on the same slots that were created for the yearly market. 
Having said that, I noticed that 4 years will not be enough since the yearly market spans across 6 years in the future. I'll change it with a constant that will control how far in the future can the forward markets reach.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: Also replied above, fixed

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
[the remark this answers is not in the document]
fievelk: I'm not sure why the public `state` property was created (rather than converting the `_state` variable into a public one). Maybe it was to avoid setting `state` directly, since the property will raise an `AttributeError` in that case?
In any case, if setters/getters are defined for a variable, that's usually done because setting/getting that variable should automatically trigger some other action. For example:
```py
@property
def state(self):
    self.do_something_else()
    return self._state
```
In our case, this doesn't happen. But since we define the property I think it would be better to use that.
What are the disadvantages you see in using the public variable?

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
hannesdiedrich: I do not get it. I see here a list of 4 market slots, not 24.

spyrostz: Probably the comment is misleading here. What you see here is the 15-min times that one DayForward market slot includes. Since the duration of a DayForward slot is 1 hour, and the market slot is e.g. today 12:00, then there will be 4 entries in the profile affected by this market: 12:00, 12:15, 12:30 and 12:45. I struggle to find a better explanation though for this in order to reflect it in a comment.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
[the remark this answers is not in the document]
spyrostz: Indeed, I did not check the base class methods ! I have refactored it already then, please have a look and let me know what you think, thanks!

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -36,24 +36,273 @@
...
+    def get_available_load_energy_kWh(
+            self, market_slot: pendulum.DateTime, state: "LoadState",
+            peak_energy_kWh: float) -> float:
+        """Get the available (not traded) energy of a load asset."""
+
+    @abstractmethod
+    def get_available_pv_energy_kWh(
+            self, market_slot: pendulum.DateTime, state: "PVState",
+            peak_energy_kWh: float) -> float:
+        """Get the available (not traded) energy of a PV asset."""
```
fievelk, line 81: I still find the `available_load_energy` wording somehow confusing. The reason is that we've always used "require/desired energy" for loads' energy, while the "available energy" naming was always related to production assets. Do you think there can be a different way to name this?

On another side: we had previously structured the energy parameter classes to be grouped into consumption vs production. Now we're mixing these two, by adding both load and PV methods in the same class. Despite the name, the new `*EnergyParams` classes don't have the same interface as the previous `*EnergyParams` classes, and `state` isn't an instance variable anymore. So I'm not sure about how these new classes will be used. 
Are they only supposed to be used in the `ForwardEnergyParams._forward_energy_params` dictionary (so we will just keep using `Production/ConsumptionStandardProfileEnergyParameters` as before)? Or are they also going to be used "directly"?
In the first case, I'd suggest adding a note in their docstrings stating something like "This class should not be instantiated on its own. Please only use `Production/ConsumptionStandardProfileEnergyParameters`". What do you think?

spyrostz: > I still find the available_load_energy wording somehow confusing. The reason is that we've always used "require/desired energy" for loads' energy, while the "available energy" naming was always related to production assets. Do you think there can be a different way to name this?

We can harmonise the names to something like `required_load_energy` and `available_pv_energy` in order to be consistent with the state classes, but I frankly would prefer using a common name for both, because they are both going to be used for the same purpose: get the energy that is available for trading. Another possibility would be to name both `non_traded_energy`, do you think that this would make it more easily readable ?

src/gsy_e/models/market/forward.py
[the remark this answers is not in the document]
fievelk: @spyrostz I think this wasn't addressed in the recent commits, right? Do you plan on working on it or is the PR complete?

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -265,7 +265,9 @@ def _populate_core_stats_and_sim_state(self, area: "Area") -> None:
         core_stats_dict = {"bids": [], "offers": [], "trades": [], "market_fee": 0.0}
 
         if area.current_market:
-            core_stats_dict.update(self._read_market_stats_to_dict(area.current_market))
+            if not ConstSettings.ForwardMarketSettings.ENABLE_FORWARD_MARKETS:
+                # Spot market cannot operate in parallel with the forward markets
+                core_stats_dict.update(self._read_market_stats_to_dict(area.current_market))
```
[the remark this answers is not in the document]
fievelk: I've actually already created a ticket (GSYE-406) after a conversation with Amir in a different PR :) You can check if that can also work, otherwise feel free to edit it and add more info!

src/gsy_e/models/market/forward.py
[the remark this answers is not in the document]
spyrostz: @fievelk thanks for the reminder, I'll try to address it now so that it is not forgotten!

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
hannesdiedrich: We decided to always explicitly set the time zone because I remember having issues with running tests locally when the timezone is set differently on you computer. Also, in order to change it globally if it needs to be change in the future.

src/gsy_e/models/strategy/forward/load.py
```diff
@@ -0,0 +1,83 @@
...
+            self.area.now)
+        capacity_percent = self._order_updaters[market][market_slot].capacity_percent / 100.0
+        max_energy_kWh = self._energy_params.peak_energy_kWh * capacity_percent
+        available_energy_kWh = self._energy_params.get_available_energy_kWh(
+            market_slot, market.market_type)
+
+        posted_energy_kWh = self._energy_params.get_posted_energy_kWh(
+            market_slot, market.market_type)
+
+        available_energy_kWh = min(available_energy_kWh - posted_energy_kWh, max_energy_kWh)
```
hannesdiedrich, line 54: I propose not to re-use the `available_energy_kWh` but to name it `order_energy` or similar. What do you think?

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -21,37 +21,407 @@
...
+
+    def get_posted_energy_kWh(self, market_slot: pendulum.DateTime) -> float:
+        return max(
+            self._posted_energy_kWh[slot]
+            for slot in self._day_forward_slots(market_slot)
+        )
+
+    def increment_posted_energy(self, market_slot: pendulum.DateTime, posted_energy_kWh: float):
+        slots = self._day_forward_slots(market_slot)
+        for slot in slots:
```
hannesdiedrich, line 161: NIT:
```suggestion
        for slot in self._day_forward_slots(market_slot):
```

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
```diff
@@ -21,37 +21,407 @@
...
+                self._posted_energy_kWh, AvailableMarketTypes.MONTH_FORWARD),
+            AvailableMarketTypes.YEAR_FORWARD: _LongForwardEnergyParameters(
+                self._posted_energy_kWh, AvailableMarketTypes.YEAR_FORWARD)
+        }
+
+        self._area = None
+        self._profile_generator: Optional[ForwardTradeProfileGenerator] = None
+
+    def get_posted_energy_kWh(
+            self, market_slot: pendulum.DateTime, market_type: AvailableMarketTypes) -> float:
```
hannesdiedrich, line 316: Although I do not like the name, we should be consistent, I think [it was decided](https://github.com/gridsingularity/gsy-e/pull/1534#discussion_r979841354) to use `product_type` from now on right?

hannesdiedrich (approved):
I left some minor comments. Thanks a lot for addressing the others! 
Already giving my LGTM.

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: Let me fix this here too, thanks!

src/gsy_e/models/strategy/energy_parameters/energy_params_eb.py
spyrostz: Fixing, thanks!

src/gsy_e/models/strategy/forward/load.py
spyrostz: Agreed, fixing

hannesdiedrich (approved):
LGTM

fievelk (approved):

PR #1535, opened 2022-09-20
GSYE-390: Add tests for simulation endpoint buffers

fievelk:

Add more tests for the `SimulationEndpointBuffer` and `CoefficientEndpointBuffer` classes.

---

### DIFF
--- a/tests/test_endpoint_buffers.py
+++ b/tests/test_endpoint_buffers.py
@@ -192,6 +192,17 @@ class TestSimulationEndpointBuffer:
             time_slot=DateTime(2022, 10, 30),
             time_slot_str="2021-10-30T00:00:00+00:00")
 
+        # Popoulate strategy and children to update the result_area_uuids dictionary
+        child_1 = MagicMock(uuid="child-uuid-1")
+        child_1.name = "child_1"
+        child_1.parent = area
+        child_2 = MagicMock(uuid="child-uuid-2")
+        child_2.name = "child_2"
+        child_2.parent = area
+
+        area.children = [child_1, child_2]
+        area.strategy = MagicMock(name="some-strategy")
+
         endpoint_buffer = SimulationEndpointBuffer(
             job_id="JOB_1",
             random_seed=41,
@@ -222,7 +233,22 @@ class TestSimulationEndpointBuffer:
             "uuid": "AREA",
             "parent_uuid": "",
             "type": "MagicMock",
-            "children": []
+            "children": [
+                {
+                    "children": [],
+                    "name": "child_1",
+                    "parent_uuid": "AREA",
+                    "type": "MagicMock",
+                    "uuid": "child-uuid-1",
+                },
+                {
+                    "children": [],
+                    "name": "child_2",
+                    "parent_uuid": "AREA",
+                    "type": "MagicMock",
+                    "uuid": "child-uuid-2",
+                },
+            ],
         }
         assert endpoint_buffer.status == "some-state"
         assert endpoint_buffer.simulation_state["general"] == sim_state_mock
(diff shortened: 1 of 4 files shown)


tests/test_endpoint_buffers.py
```diff
@@ -93,3 +105,143 @@ def test_forward_results_are_generated(   # pylint: disable-msg=too-many-locals
...
+            "current_market": "",
+            "current_market_ui_time_slot_str": "",
+            "random_seed": 41,
+            "status": "",
+            "progress_info": {
+                "eta_seconds": 0,
+                "elapsed_time_seconds": 0,
+                "percentage_completed": 0,
+            },
+            "bids_offers_trades": {},
```
mrsaemir, line 130: Shouldn't we also write tests to check `bids_offers_trade` or `simulation_raw_data` attributes contain useful data?

mrsaemir (approved):
LGTM. Just a single question.

tests/test_endpoint_buffers.py
fievelk: The reason why this is currently left out is that `bids_offers_trades` is populated using `self.flattened_area_core_stats_dict`. In turn, this is populated inside of `_populate_core_stats_and_sim_state()`. 
As I said somewhere, testing this method is tricky and takes quite long, because it has too many conditional branches and it's also overridden by the child `CoefficientEndpointBuffer` class. Testing all the `if` branches would create a lot of clutter in tests and make them difficult to understand and modify. That's why I think it would be better to refactor this method into a class first, and only then write tests for it.

fievelk: If you think it works, I might create a new ticket for this. Or we can keep this PR open and get back to it later on (since right now this is not the top priority because of EB tasks).

mrsaemir: I definitely agree. Let's have a new ticket for that. Thanks a lot for dealing with this.

tests/test_endpoint_buffers.py
```diff
@@ -139,10 +140,9 @@ def test_prepare_results_for_publish(forward_setup):
             }
         }
 
-    @staticmethod
     @patch("gsy_e.gsy_e_core.sim_results.endpoint_buffer.get_json_dict_memory_allocation_size")
     def test_prepare_results_for_publish_output_too_big(
-            get_json_dict_memory_allocation_size_mock, forward_setup, caplog):
+            self, get_json_dict_memory_allocation_size_mock, forward_setup, caplog):
```
hannesdiedrich, line 145: Why is this change needed?

tests/test_endpoint_buffers.py
```diff
@@ -229,7 +229,6 @@ def test_update_stats(self, forward_setup):
 
         assert endpoint_buffer.current_market_time_slot_str == "2021-10-30T00:00:00+00:00"
         assert endpoint_buffer.current_market_ui_time_slot_str == "October 30 2022, 00:00 h"
-        assert endpoint_buffer.current_market_time_slot_unix == 1667080800.0
```
hannesdiedrich: Why is this failing?

hannesdiedrich (commented):
Two questions. Will LGTM after answered.
Also, we will have some merge conflicts with 388. Lets see who pushes first :-)

tests/test_endpoint_buffers.py
fievelk: This is from some days ago, but AFAIR the results were different between my local environment and the CI. I'm not sure this value is being used anywhere, so I just removed the check to move faster.

tests/test_endpoint_buffers.py
fievelk: That's something that gave me some headaches. I wanted to create a class (`TestCoefficientEndpointBuffer`) that inherited from `TestSimulationEndpointBuffer` in order to automatically run all the tests that are declared in the parent class.
Unfortunately, if you try to do that when the parent class has static methods, this will break up stuff. For some reason (I suppose something internal to pytest) inherited methods that are declared as `@staticmethod` will not work (they raise an error). That seems to be related to how pytest works. I'm not sure whether this has been fixed in recent releases.

hannesdiedrich: Interesting. Thanks!

PR #1537, opened 2022-10-06
GSYE-353: Fix post orders to future markets by external strategies an…

hannesdiedrich:
…d add unit tests

The ticket seeks to add unit tests for placing order on future market via the external strategies.

## Proposed changes

- enabled posting orders via the external strategies (load, pv, smart_meter) to the future markets. This included passing the time_slot also in the external command as well as to circumvent the fact that the FutureMarkets class does not have a time_slot member. For finding out what the required or available energy is, the time_slot argument is used now. 
- I skipped adding this for the storage because it is not clear whether it will take part in the future markets. Or will it?
- Added unit tests for placing and deleting orders via the aggregator


src/gsy_e/models/config.py
```diff
@@ -128,6 +128,10 @@ def read_market_maker_rate(self, market_maker_rate):
         """
         Reads market_maker_rate from arbitrary input types
         """
+        print("GlobalConfig.market_maker_rate", GlobalConfig.market_maker_rate)
+        print("ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE",
+              ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE)
+        print("read_market_maker_rate", market_maker_rate)
```
fievelk, line 134: I think this is a leftover, right?

src/gsy_e/models/strategy/external_strategies/load.py
```diff
@@ -314,9 +315,14 @@ def _bid_aggregator(self, arguments: Dict) -> Dict:
...
                 required_energy_kWh = self.state.get_unsettled_deviation_kWh(market.time_slot)
-            else:
+            elif self.area.is_market_future(market.id):
+                required_energy_kWh = self.state.get_energy_requirement_Wh(
+                    str_to_pendulum_datetime(arguments["time_slot"])) / 1000.
+            elif self.area.is_market_spot(market.id):
                 required_energy_kWh = (
                         self.state.get_energy_requirement_Wh(market.time_slot) / 1000.)
+            else:
+                raise OrderCanNotBePosted("Market can not be posted to.")
```
fievelk, line 325: Nit:
```suggestion
                raise OrderCanNotBePosted("The order cannot be posted on the market.")
```
It might even help to add a debug log, for example:
```suggestion
                logger.debug("The order cannot be posted on the market. (arguments: %s, market_id: %s"), arguments, market.id)
                raise OrderCanNotBePosted("The order cannot be posted on the market.")
```
This was we would be able to debug production if needed.

src/gsy_e/models/strategy/external_strategies/smart_meter.py
```diff
@@ -374,8 +381,14 @@ def _offer_aggregator(self, arguments: Dict) -> Dict:
                     raise OrderCanNotBePosted("The smart meter did not produce enough energy, ",
                                               "settlement offer can not be posted.")
                 available_energy = self.state.get_unsettled_deviation_kWh(market.time_slot)
-            else:
+            elif self.area.is_market_future(market.id):
+                available_energy = self.state.get_available_energy_kWh(
+                    str_to_pendulum_datetime(arguments["time_slot"]))
+            elif self.area.is_market_spot(market.id):
```
fievelk, line 387: Nit: the same comment as https://github.com/gridsingularity/gsy-e/pull/1537/files#r988908565 can be applied here.

src/gsy_e/models/strategy/external_strategies/pv.py
```diff
@@ -342,8 +343,13 @@ def _offer_aggregator(self, arguments: Dict) -> Dict:
...
                                               "settlement offer can not be posted.")
                 available_energy_kWh = self.state.get_unsettled_deviation_kWh(market.time_slot)
-            else:
+            elif self.area.is_market_future(market.id):
+                available_energy_kWh = self.state.get_available_energy_kWh(
+                    str_to_pendulum_datetime(arguments["time_slot"]))
+            elif self.area.is_market_spot(market.id):
                 available_energy_kWh = self.state.get_available_energy_kWh(market.time_slot)
+            else:
+                raise OrderCanNotBePosted("Market can not be posted to.")
```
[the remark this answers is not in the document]
hannesdiedrich: I had a similar idea, but please have a look at the settlement market branch of the if block. There `get_available_energy_kWh` is not called.

src/gsy_e/models/strategy/external_strategies/load.py
hannesdiedrich: Great Thanks! This error message was a placeholder and I wanted to improve it before pushing, I forgot unfortunately. Will go for your suggestion.

src/gsy_e/models/config.py
hannesdiedrich: for sure. Thanks for catching this!

src/gsy_e/models/strategy/external_strategies/__init__.py
```diff
@@ -349,11 +349,10 @@ def _offer_aggregator_impl(
 
             offer_arguments = {k: v
                                for k, v in arguments.items()
-                               if k not in ["transaction_id", "type", "time_slot"]}
+                               if k not in ["transaction_id", "type"]}
```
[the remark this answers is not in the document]
fievelk: I think that it would be better to pick one of the two options:
1. We keep the code in gsy-e, so we also maintain it and keep it working
2. We do not maintain this code, so we remove it completely.

I think that if we keep a code that is "wrong" and doesn't work, this won't help in the future and it will actually confuse readers. Even just the fact that the code is here would raise the question "so are we supporting this or not?". In my opinion, unused code should (almost) always be deleted. If we want to put it back where it was, we can check the git history (given we use proper commit messages).

What do you think?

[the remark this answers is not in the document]
hannesdiedrich: We had multiple long discussions about this before and I remember us deciding to keep it. The code can still be used, just not with the sdk. If you write your own client, you could still post direct orders via the gsy-web or via a redis connection. Only our sdk is not exposing this possibility. 
So, let me go for the first option and add the possibility to post on the future market. I am not sure, but I think that also the settlement market is not supported there. So, this might be even worth a separate story. 
Let me start and if I think it is too much, I will create a ticket (could also be an onboarding task)

fievelk (approved):

fievelk (approved):

tests/strategies/external/test_pv.py
```diff
@@ -41,6 +43,22 @@ def settlement_market_fixture():
                             time_slot=datetime(2021, 12, 7, 12, 00))
 
 
+@pytest.fixture(name="future_markets")
```
mrsaemir, line 46: Is it possible to reuse the fixture among different strategies and do not repeat it everywhere?

tests/strategies/external/test_pv.py
```diff
@@ -132,6 +150,27 @@ def test_offer_aggregator(external_pv):
         )
         assert_bid_offer_aggregator_commands_return_value(return_value, True)
 
+    @staticmethod
+    def test_offer_aggregator_places_future_offer(external_pv, future_markets):
```
mrsaemir, line 154: I see tests for placing bids/offers and deleting them. How about updating of bids/offers? Isn't it needed?

tests/strategies/external/test_smart_meter.py
```diff
@@ -250,6 +312,27 @@ def test_delete_offer_aggregator(external_smart_meter: SmartMeterExternalStrateg
         assert return_value["command"] == "offer_delete"
         assert return_value["deleted_offers"] == [offer.id]
 
+    @staticmethod
+    def test_delete_offer_aggregator_deletes_offer_from_future_market(
```
mrsaemir, line 316: I think `test_delete_bid_aggregator_deletes_bid_from_future_market` is missing.

mrsaemir (commented):
Just a few questions. LGTM.

tests/strategies/external/test_pv.py
hannesdiedrich: I was also thinking about this but remembered something from @fievelk who said that each test module should be self-sufficient. Is that correct?

tests/strategies/external/test_pv.py
hannesdiedrich: An sort-of update can be achieved by setting `replace_existing=True`. With this, the existing order is deleted and the updated values are posted in a new order. We need to do this in order to broadcast to all listeners, that the order was updated, only with the existing implementation, two events are broadcasted: DELETE_OFFER and NEW_OFFER. 
I hope it makes sense.

tests/strategies/external/test_smart_meter.py
hannesdiedrich: Indeed, it is missing. I thought I added it, but apparently not. let me add it now.

tests/strategies/external/test_pv.py
fievelk: It really depends on the case. Re-using code in tests can be fine. A fixture can be shared, if we know that many of our tests require that object (that's also one of the reasons why we use factories in gsy-web). Also, fixtures are made exactly with that purpose in mind: being shared.

It's often better to hardcode values inside of tests, rather than create a method that generates them in a complex way. That's because otherwise we would need to write tests for that method too!

What should be avoided is adding logic inside tests. For example, tests should not contain `if` statements: each test should only check one "flow", and it should do it in an obvious way (without behaving differently depending on conditionals).

Not sure if this answers the current situation :)

hannesdiedrich: Great! I moved the fixtures in their own module here: [d720f58](https://github.com/gridsingularity/gsy-e/pull/1537/commits/d720f58d73e5b921a9c4454cbe913b2f4fd8def9)

mrsaemir (approved):

fievelk (approved):

PR #1542, opened 2022-10-07
GSYE-388-2: Fix SCM time-related failures

mrsaemir:
*NOTE: The duplicated key exception was never spotted again after Hannes' fix. This PR resolves only some other minor issues.*

- `slot_time_counter` increases irregularly because of accumulated miliseconds.
- Status of simulation is not set to "running" before the initial sleep of `calculate_total_initial_ticks_slots`.

### Before this PR:
```
slot_no start_of_slot
506 2022-10-07T16:52:00.022406+02:00
507 2022-10-07T16:53:35.357447+02:00
508 2022-10-07T16:55:35.457563+02:00
509 2022-10-07T16:57:35.557277+02:00
510 2022-10-07T16:59:35.657843+02:00
511 2022-10-07T17:01:35.741415+02:00
512 2022-10-07T17:03:35.840907+02:00
513 2022-10-07T17:05:35.942096+02:00
514 2022-10-07T17:07:36.042252+02:00
515 2022-10-07T17:09:36.140363+02:00
516 2022-10-07T17:11:36.223580+02:00
517 2022-10-07T17:13:36.297965+02:00
518 2022-10-07T17:15:36.398065+02:00
...
```

### After this PR:
```
slot_no start_of_slot
538 2022-10-07T17:56:00.097372+02:00
539 2022-10-07T17:58:00.098123+02:00
540 2022-10-07T18:00:00.100255+02:00
541 2022-10-07T18:02:00.100579+02:00
542 2022-10-07T18:04:00.097163+02:00
543 2022-10-07T18:06:00.097132+02:00
544 2022-10-07T18:08:00.097409+02:00
545 2022-10-07T18:10:00.097453+02:00
546 2022-10-07T18:12:00.100899+02:00
547 2022-10-07T18:14:00.097614+02:00
548 2022-10-07T18:16:00.097326+02:00
...
```

**INTEGRATION_TESTS_BRANCH**=bug/GSYE-388

### DIFF
--- a/src/gsy_e/gsy_e_core/simulation/simulation.py
+++ b/src/gsy_e/gsy_e_core/simulation/simulation.py
@@ -152,6 +152,8 @@ class Simulation:
             if self._setup.started_from_cli:
                 self._run_cli_execute_cycle(initial_slot, tick_resume)
             else:
+                # update status of the simulation before executing it.
+                self._results.update_and_send_results(simulation=self)
                 self._execute_simulation(initial_slot, tick_resume)
         except (KeyboardInterrupt, SimulationResetException):
             pass
--- a/src/gsy_e/gsy_e_core/simulation/simulation.py
+++ b/src/gsy_e/gsy_e_core/simulation/simulation.py
@@ -419,6 +419,8 @@ class CoefficientSimulation(Simulation):
             self._time.calc_resume_slot_and_count_realtime(
                 self.config, slot_resume))
 
+        self._time.reset(not_restored_from_state=(slot_resume == 0))
+
         for slot_no in range(slot_resume, slot_count):
             self._handle_paused(console)
 
--- a/src/gsy_e/gsy_e_core/simulation/time_manager.py
+++ b/src/gsy_e/gsy_e_core/simulation/time_manager.py
@@ -155,11 +155,10 @@ class SimulationTimeManagerScm:
             log.debug("Slot %s/%s: Sleep time of %s s was applied",
                       slot_no, slot_count, sleep_time_s)
 
-        self.slot_time_counter = time()
+        self.slot_time_counter = int(time())
 
-    @staticmethod
     def calc_resume_slot_and_count_realtime(
-            config: "SimulationConfig", slot_resume: int) -> Tuple[int, int]:
+            self, config: "SimulationConfig", slot_resume: int) -> Tuple[int, int]:
         """Calculate total slot count and the slot where to resume the realtime simulation."""
         slot_count = int(config.sim_duration / config.slot_length)
 
@@ -173,6 +172,7 @@ class SimulationTimeManagerScm:
             sleep_time_s = config.slot_length.total_seconds() - seconds_elapsed_in_slot
             sleep(sleep_time_s)
 
+        self.slot_time_counter = int(time())
         return slot_count, slot_resume
 
 
--- a/src/gsy_e/gsy_e_core/simulation/time_manager.py
+++ b/src/gsy_e/gsy_e_core/simulation/time_manager.py
@@ -155,7 +155,7 @@ class SimulationTimeManagerScm:
             log.debug("Slot %s/%s: Sleep time of %s s was applied",
                       slot_no, slot_count, sleep_time_s)
 
-        self.reset(not_restored_from_state=False)
+        self.slot_time_counter = int(time())
 
     def calc_resume_slot_and_count_realtime(
             self, config: "SimulationConfig", slot_resume: int) -> Tuple[int, int]:
@@ -172,7 +172,6 @@ class SimulationTimeManagerScm:
             sleep_time_s = config.slot_length.total_seconds() - seconds_elapsed_in_slot
             sleep(sleep_time_s)
 
-        self.reset(not_restored_from_state=False)
         return slot_count, slot_resume
 
 
--- a/src/gsy_e/gsy_e_core/simulation/time_manager.py
+++ b/src/gsy_e/gsy_e_core/simulation/time_manager.py
@@ -127,7 +127,7 @@ class SimulationTimeManagerScm:
         Restore time-related parameters of the simulation to their default values.
         Mainly useful when resetting the simulation.
         """
-        self.slot_time_counter = time()
+        self.slot_time_counter = int(time())
         if not_restored_from_state:
             self.start_time = now(tz=TIME_ZONE)
             self.paused_time = 0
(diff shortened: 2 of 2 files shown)


hannesdiedrich:
> NOTE: The duplicated key exception was never spotted again after Hannes' fix. This PR resolves only some other minor issues.

🥳

src/gsy_e/gsy_e_core/simulation/time_manager.py
```diff
@@ -155,11 +155,10 @@ def handle_slowdown_and_realtime_scm(self, slot_no: int, slot_count: int,
             log.debug("Slot %s/%s: Sleep time of %s s was applied",
                       slot_no, slot_count, sleep_time_s)
 
-        self.slot_time_counter = time()
+        self.slot_time_counter = int(time())
```
hannesdiedrich, line 158: I do not really get the reason for this change. AFAIK, `time()` return the current time in unix seconds as float, right? Adding `int()` will just strip of the milliseconds. This is not the actual fix right? So I propose to revert this here in oder to als take into account the milliseconds (this difference could add up to some seconds when the simulation has progressed a bit).

src/gsy_e/gsy_e_core/simulation/simulation.py
```diff
@@ -152,6 +152,8 @@ def run(self, initial_slot: int = 0) -> None:
             if self._setup.started_from_cli:
                 self._run_cli_execute_cycle(initial_slot, tick_resume)
             else:
+                # update status of the simulation before executing it.
+                self._results.update_and_send_results(simulation=self)
```
hannesdiedrich, line 156: This is the actual fix, right?

src/gsy_e/gsy_e_core/simulation/time_manager.py
mrsaemir: Since the smallest amount of time in gsy-e is one second, fractions of seconds can add up and eventually change the slot time after some slots are passed. Please notice how it can affect slot time in the description of the PR.

src/gsy_e/gsy_e_core/simulation/simulation.py
mrsaemir: Yes. This will fix the `running` status for SCM and non-SCM simulations.

src/gsy_e/gsy_e_core/simulation/time_manager.py
hannesdiedrich: I just wrote a long comment, but had to remove it again, as I only see now, why it works.
The actual fix is that you added line 175 here, in order to re-set the `self.slot_time_counter` in `calc_resume_slot_and_count_realtime`. Great! 
However, for this we actually have the `reset` method in the `SimulationTimeManagerScm`. When I implemented it, I forgot to call it in the `CoefficientSimulation`. So, I propose to remove this line from `calc_resume_slot_and_count_realtime` and actually call reset in the `_execute_simulation`: 
Screenshot 2022-10-10 at 11 25 12

hannesdiedrich: I still not agree on cutting off the fractions of seconds. IMO they are needed to really set the actual end of the timeslot. The indeed add up, and if you always omit a fraction of a seconds, the start of the slots will be off when you run the simulation for longer.

hannesdiedrich:
Can you please add the following change that would make the code a little consistent (I should have added this in my former PR). Thanks a lot!
Screenshot 2022-10-10 at 11 12 42

src/gsy_e/gsy_e_core/simulation/simulation.py
fievelk: Wait, why is this the actual fix? Isn't the actual fix the one on this line? https://github.com/gridsingularity/gsy-e/pull/1542/files#diff-dd05e0495c9ab88ead3805c4dc846ef58cd1d882f667c952e12e721812338f2aR158
Or maybe I didn't get exactly what the GSYE-388 ticket is asking?

Regarding this specific line. My first question would have been "If the simulation hasn't started yet, why do we need to update the status?".
AFAIU, this fix only applies to simulations that were stopped and that we're now restarting. Is this correct?

fievelk (dismissed):

src/gsy_e/gsy_e_core/simulation/time_manager.py
mrsaemir: IMO, fractions of second are only because of the time used for calculating methods of simulation time manager. I think in an ideal world, the amount time needed to calculate time should be zero. That's why I suggest to get rid of second fractions.

src/gsy_e/gsy_e_core/simulation/simulation.py
mrsaemir: It depends on how we define start of a simulation. That line of code is running just before the real start of simulation loop.

mrsaemir: And for the second question, you're right. It only runs once and does not run in case of a paused simulation.

spyrostz (dismissed):
LGTM, once the discussion with the float part of the slot time counter settles. I do not have a personal preference since on the one hand, we only support second precision till now at least, but on the other hand would not hurt to maintain the float value.

src/gsy_e/gsy_e_core/simulation/time_manager.py
hannesdiedrich: These fraction can add up do large amounts (when you accumulate them a lot of times, like we did once in the "normal" simulation where we track each tick). 
Here it does not seem to get added up along the way (judging from the PR description) which I do not understand why now. I give up though. Please leave the `int()` in and we can remove it when we actually have long simulations running and see a slight difference. We probably will not..
Thanks for bering with me!

mrsaemir: Thanks a lot

src/gsy_e/gsy_e_core/simulation/time_manager.py
```diff
@@ -172,7 +172,7 @@ def calc_resume_slot_and_count_realtime(
             sleep_time_s = config.slot_length.total_seconds() - seconds_elapsed_in_slot
             sleep(sleep_time_s)
 
-        self.slot_time_counter = int(time())
+        self.reset(not_restored_from_state=False)
```
hannesdiedrich, line 175: Sorry, this is not what I meant [here](https://github.com/gridsingularity/gsy-e/pull/1542#discussion_r991090307).
Please remove this line here and add it in the simulation.py in order to follow the single responsibility principle.

hannesdiedrich (approved):
LGTM

fievelk (approved):

PR #1543, opened 2022-10-10
GSYE-417: Add raise of exception when profile in DB is empty.

hannesdiedrich:

If there is an empty profile in the DB, the code will fail with an IndexError that is not very understandable.

## Proposed changes

Add a custom exception that is raised if there is an empty profile in the DB that is about to be read. 

### DIFF
--- a/src/gsy_e/gsy_e_core/user_profile_handler.py
+++ b/src/gsy_e/gsy_e_core/user_profile_handler.py
@@ -32,6 +32,10 @@ import gsy_e.constants
 from gsy_e.gsy_e_core.util import should_read_profile_from_db
 
 
+class ProfileDBConnectionException(Exception):
+    """Exception that is raised inside ProfileDBConnectionHandler."""
+
+
 class ProfileDBConnectionHandler:
     """
     Handles connection and interaction with the user-profiles postgres DB via pony ORM
@@ -98,10 +102,13 @@ class ProfileDBConnectionHandler:
         """
         if not isinstance(profile_uuid, uuid.UUID):
             profile_uuid = uuid.UUID(profile_uuid)
-        first_datapoint_time = select(
+        first_datapoint = select(
             datapoint for datapoint in self.Profile_Database_ProfileTimeSeries
             if datapoint.profile_uuid == profile_uuid
-        ).order_by(lambda d: d.time).limit(1)[0].time
+        ).order_by(lambda d: d.time).limit(1)
+        if len(first_datapoint) == 0:
+            raise ProfileDBConnectionException("Profile in DB is empty.")
+        first_datapoint_time = first_datapoint[0].time
 
         datapoints = select(
             datapoint for datapoint in self.Profile_Database_ProfileTimeSeries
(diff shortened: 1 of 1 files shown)


src/gsy_e/gsy_e_core/user_profile_handler.py
```diff
@@ -98,10 +102,13 @@ def get_first_week_from_profile(self, profile_uuid, current_timestamp) -> dict:
...
         if not isinstance(profile_uuid, uuid.UUID):
             profile_uuid = uuid.UUID(profile_uuid)
-        first_datapoint_time = select(
+        first_datapoint = select(
             datapoint for datapoint in self.Profile_Database_ProfileTimeSeries
             if datapoint.profile_uuid == profile_uuid
-        ).order_by(lambda d: d.time).limit(1)[0].time
+        ).order_by(lambda d: d.time).limit(1)
+        if len(first_datapoint) == 0:
+            raise ProfileDBConnectionException("Profile in DB is empty.")
```
spyrostz, line 110: I would also add the configuration uuid and / or at least the profile uuid. The latter would actually be the only thing necessary since we can deduce the configuration uuid from the profile uuid, however if it is possible to also add the config uuid for convenience would be great !

hannesdiedrich: I added the profile_uuid here: [e14b37c](https://github.com/gridsingularity/gsy-e/pull/1543/commits/e14b37c8e0195bdc8723f2657fd10a4765542a39)
The config_uuid is not that easy to inject into there. But you are right, it is easily acquired from the DB from the profile_uuid.

fievelk (approved):

src/gsy_e/gsy_e_core/user_profile_handler.py
```diff
@@ -98,10 +102,14 @@ def get_first_week_from_profile(self, profile_uuid, current_timestamp) -> dict:
...
         if not isinstance(profile_uuid, uuid.UUID):
             profile_uuid = uuid.UUID(profile_uuid)
-        first_datapoint_time = select(
+        first_datapoint = select(
             datapoint for datapoint in self.Profile_Database_ProfileTimeSeries
             if datapoint.profile_uuid == profile_uuid
-        ).order_by(lambda d: d.time).limit(1)[0].time
+        ).order_by(lambda d: d.time).limit(1)
+        if len(first_datapoint) == 0:
+            raise ProfileDBConnectionException(
```
mrsaemir, line 110: Solving this will definitely help us understand further issues better. But I was thinking, when a simulation fails because of an exception, why the UI is stuck in the initializing/running phase? Shouldn't it be updated to failed immediately? How can we achieve that? I'm suggesting to catch the exception on a higher layer and set the status of the simulation to failed and prevent watchdog from restarting it as well. What do you think?

mrsaemir (approved):

src/gsy_e/gsy_e_core/user_profile_handler.py
hannesdiedrich: This is handled. If there is an error in the simulation, the error is send to the DB on gsy-web (`SimulationResults.error_output`, I already debugged a lot of issues using this entry) and the status is set to failed.
There are rare occasions when the error appears when the setup and the simulation is initialized. When this happens, the simulation is indeed stuck in initialising. I hope it makes sense.

PR #1544, opened 2022-10-11
GSYE-416: Add assertions for CoefficientArea settings

hannesdiedrich:

Settings for the SCM areas should not be set to None

## Proposed changes

Add validation method that checks for all scm settings to be not-None
This PR should only be merged after GSYE-415


mrsaemir (approved):

src/gsy_e/models/area/__init__.py
```diff
@@ -252,6 +252,18 @@ def get_results_dict(self):
...
 
 
+class CoefficientAreaException(Exception):
+    """Exception that is raised when serializing an area."""
+
+
+def assert_coefficient_area_settings(
+        strategy: Optional[SCMStrategy], setting: Optional[float], setting_name: str) -> float:
+    """Check if coefficient area that is not an asset provided SCM setting."""
+    if strategy is None and setting is None:
```
spyrostz, line 262: I think that this would still fail ATM because of 2 areas: the "Community" area (which should be there) and the "Grid" area (which should not be there, but the UI is probably in the process of removing this). Therefore I would suggest to pass the area instead and check if it is a home area (essentially if all the children of this area have non-none strategy members). Thanks!

spyrostz (commented):
One comment, LGTM once fixed

src/gsy_e/models/area/__init__.py
hannesdiedrich: I thought this check also applies to the Grid and community area. But you are right, all the settings are only used in the SCM homes. Will change this. Thanks!

hannesdiedrich: [a6cadd9](https://github.com/gridsingularity/gsy-e/pull/1544/commits/a6cadd97b72f56f44f60b5367179d6460cd9a882)

src/gsy_e/models/area/__init__.py
```diff
@@ -252,6 +252,19 @@ def get_results_dict(self):
...
 
 
+class CoefficientAreaException(Exception):
+    """Exception that is raised when serializing an area."""
+
+
+def assert_coefficient_area_settings(
+        children: Optional[List], setting: Optional[float], setting_name: str) -> float:
+    """Check if coefficient area that is not an asset provided SCM setting."""
+    if children and len(children) > 0:
```
fievelk, line 262:
```suggestion
    if children:
```
If a list is empty its truth value is `False`.

src/gsy_e/models/area/__init__.py
```diff
@@ -252,6 +252,19 @@ def get_results_dict(self):
         return {}
 
 
+class CoefficientAreaException(Exception):
+    """Exception that is raised when serializing an area."""
+
+
+def assert_coefficient_area_settings(
```
fievelk, line 259: I'd suggest to use a different naming in place of "assert". Since `assert` has a very specific meaning in Python, this might be confusing. Maybe `validate_coefficient_area_setting()` could be an option?

src/gsy_e/models/area/__init__.py
```diff
@@ -252,6 +252,19 @@ def get_results_dict(self):
...
+        children: Optional[List], setting: Optional[float], setting_name: str) -> float:
```
fievelk, line 260: I'm not very convinced by the fact that here we had to add `setting_name` just to report it as part of the exception message. This is easily prone to mistakes (e.g. if one changes the `setting` parameter without updating the `setting_name`). Do you think there can be some other way around it?

src/gsy_e/models/area/__init__.py
hannesdiedrich: It could also be set to None.

fievelk (commented):

src/gsy_e/models/area/__init__.py
hannesdiedrich: Hmm. I really do not have an idea how to solve this. IMO this is a case of YAGNI, or rather: If it happens, it happens, we will deal with this then. But I rather return the name of the misconfigured setting. Do you agree?

src/gsy_e/models/area/__init__.py
hannesdiedrich: Ok, will do

src/gsy_e/models/area/__init__.py
fievelk: `None` if a false-y value too, so `if children` is `False` if `children == None`.

fievelk (dismissed):

src/gsy_e/models/area/__init__.py
hannesdiedrich: Ah yes, and if it is an empty list, if will also not go into the if block. My bad! 
Applied here: [7ba425d](https://github.com/gridsingularity/gsy-e/pull/1544/commits/7ba425d05ff0cd3cb2fb9c35f02d8515af156048)

fievelk (approved):

spyrostz (approved):
LGTM

PR #1545, opened 2022-10-11
GSYE-416-refactor: Move area classes into separate modules

hannesdiedrich:

## Proposed changes

-


mrsaemir (dismissed):
Apart from moving things, have you done any addition/deletion to the classes? I'm asking this because the diff is so large and I couldn't review it in details.
Thanks

hannesdiedrich:
> Apart from moving things, have you done any addition/deletion to the classes? I'm asking this because the diff is so large and I couldn't review it in details. Thanks

I only moved things around on the code side. I had to adapt the tests though. Things had to get mocked that did not have to before (Thanks Spyros, who helped me).

spyrostz (approved):
LGTM. I do not like that much the fact that the string "area" is repeated in the individual modules but it is tough to name it otherwise. TLDR I am good as it is

spyrostz (approved):
LGTM

fievelk (approved):

hannesdiedrich:
> > LGTM. I do not like that much the fact that the string "area" is repeated in the individual modules but it is tough to name it otherwise. TLDR I am good as it is
> 
> @spyrostz @hannesdiedrich if you mean that the word "area" is repeated in the imports, this can be fixed by using the same strategy we've employed in `simulation/__init__.py` (importing the relevant "root" classes and using `__all__` to make them safely available).
> 
> https://github.com/gridsingularity/gsy-e/blob/8df13e08286e097e181fdbb0192f0d571554ac45/src/gsy_e/gsy_e_core/simulation/__init__.py#L15-L24

I did that:
https://github.com/gridsingularity/gsy-e/pull/1545/files#diff-5c02562bac27318ba374fac2bdc5d43cafe5cf176c7ada2a500071713d48a637R23

where do you see the area.area . Only in the test files, right? Will change it.

PR #1552, opened 2022-10-19
Adapted CONFIGURATION_ID parameter to be configurable via an env vari…

spyrostz:
…able. This will allow an SDK to connect to a specific simulation run when running via Redis.

## Proposed changes

-

### DIFF
--- a/src/gsy_e/constants.py
+++ b/src/gsy_e/constants.py
@@ -18,9 +18,10 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 # Need to import required settings from gsy-framework in order to be available in d3a,
 # thus avoiding accessing the gsy-framework constants.
 # pylint: disable=unused-import
-from gsy_framework.constants_limits import TIME_FORMAT, DATE_FORMAT, GlobalConfig # NOQA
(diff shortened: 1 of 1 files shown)


mrsaemir (approved):

mrsaemir (commented):
Shouldn't we also add some sort of help docs to the CLI?

spyrostz:
> Shouldn't we also add some sort of help docs to the CLI?

Not on the CLI since this is not a feature that we plan to advertise, but a Confluence page with detailed instructions on how to run multiple parallel simulations will follow this.

src/gsy_e/constants.py
```diff
@@ -37,7 +38,7 @@
 DISPATCH_EVENT_TICK_FREQUENCY_PERCENT = 10
 DISPATCH_MYCO_EVENT_TICK_FREQUENCY_PERCENT = 2
 
-CONFIGURATION_ID = ""
+CONFIGURATION_ID = os.environ.get("CONFIGURATION_ID", "")
```
hannesdiedrich, line 41: I was worried at first that the `rq_job_handler` would be influenced by this in the rare case when this variable is set in the environment and we start the launcher. But AFAIU, this defined at startup and the rq_job_handler will overwrite this value. 
Nothing to do here, just thinking out loud and hope that I am right. Thanks!

hannesdiedrich (approved):
LGTM

PR #1553, opened 2022-10-19
GSYE-431: Adapt scm strategies to also have profile uuids as memebers…

hannesdiedrich:
… in order for the profile handler to retrieve the used profiles before buffering them

## Proposed changes

-


spyrostz (approved):
LGTM, but to understand it a bit better, why did this happen only for this SCM simulation and not for the ones that we have tried till now?

hannesdiedrich:
> LGTM, but to understand it a bit better, why did this happen only for this SCM simulation and not for the ones that we have tried till now?

Because I added this in the original PR: https://github.com/gridsingularity/gsy-e/pull/1550/files

cruky (approved):

PR #1556, opened 2022-11-03
Feature/gsye 333 serialization

spyrostz:

## Proposed changes

-


src/gsy_e/gsy_e_core/area_serializer.py
```diff
@@ -127,9 +137,10 @@ def _leaf_from_dict(description, config):
     display_type = description.pop("display_type", None)
     try:
         leaf_object = leaf_type(**description, config=config)
-    except KeyError:
+    except KeyError as exc:
         # If the strategy is not supported in SCM or normal operating mode, do
         # not create the area at all.
+        logging.error("Failed to create leaf %s %s %s", leaf_type, description, exc)
```
hannesdiedrich, line 143: Thanks a lot! I was about to ad this in one of my last PRs, but forgot. Took me a long time to find this spot. Debugging of the init of the simulation will be easier now.

src/gsy_e/gsy_e_core/rq_job_handler.py
```diff
@@ -77,6 +72,20 @@ def launch_simulation_from_rq_job(scenario: Dict,
...
+                ConfigurationType.CANARY_NETWORK.value, ConfigurationType.B2B.value]:
+            gsy_e.constants.EXTERNAL_CONNECTION_WEB = True
+            GlobalConfig.IS_CANARY_NETWORK = scenario.pop("is_canary_network", False)
+            gsy_e.constants.RUN_IN_REALTIME = GlobalConfig.IS_CANARY_NETWORK
+
+            if settings["type"] == ConfigurationType.B2B.value:
+                ConstSettings.ForwardMarketSettings.ENABLE_FORWARD_MARKETS = True
+                # Disable fully automatic trading mode for the template strategies in favor of
+                # UI manual and auto modes.
+                ConstSettings.ForwardMarketSettings.FULLY_AUTO_TRADING = False
```
hannesdiedrich, line 87: I understand that in the case of `ConstSettings.ForwardMarketSettings.FULLY_AUTO_TRADING = True`, the template strategies will trade automatically (like they do also in the "normal" simulations. However, what is the UI auto mode? I thought that this is exactly this case? Or is the UI actually sending orders automatically?

src/gsy_e/gsy_e_core/area_serializer.py
hannesdiedrich: Although, let's follow the approach to first call 'getLogger' and do not use `logging` directly, ok?

src/gsy_e/models/leaves.py
```diff
@@ -118,6 +120,8 @@ def __init__(self, name, config, uuid=None, **kwargs):
                 uuid=uuid
             )
         except TypeError as ex:
+            print(self.strategy_type)
+            print(kwargs)
```
hannesdiedrich, line 124: Could be remove, right?

src/gsy_e/models/strategy/forward/__init__.py
```diff
@@ -25,10 +28,26 @@ def __init__(self,
             100.0)
         super().__init__()
         self.enabled = True
```
hannesdiedrich, line 30: Can you please explain the reason for this switch? I do not see this being toggled in any part of the code. AFAIU, you enable and disable the trading with `enable_trading` and `disable_trading` events. Why is this switch needed?

src/gsy_e/models/strategy/forward/__init__.py
```diff
@@ -94,14 +119,122 @@ def event_offer_traded(self, *, market_id, trade):
...
+            self._remove_order_event(event.get("args"))
+
+    def _auto_trading_event(self, args):
+        """
+        Enable energy trading, and create order updater for all open markets between the start and
+        end time.
+        """
+        if not LiveEventArgsValidator(
+                self._strategy.log.error).validate_start_trading_event_args(args):
+            return
```
hannesdiedrich, line 175: What do you think about adding a decorator that deals with this?

src/gsy_e/models/strategy/forward/pv.py
```diff
@@ -40,11 +52,24 @@ def remove_open_orders(self, market: "ForwardMarketBase", market_slot: DateTime)
...
 
-    def post_order(self, market: "ForwardMarketBase", market_slot: DateTime):
-        order_rate = self._order_updaters[market][market_slot].get_energy_rate(
-            self.area.now)
-
-        capacity_percent = self._order_updaters[market][market_slot].capacity_percent / 100.0
+    def remove_order(self, market: "ForwardMarketBase", market_slot: DateTime, order_uuid: str):
+        bids = [bid
+                for bid in market.slot_bid_mapping[market_slot]
+                if bid.buyer == self.owner.name and bid.id == order_uuid]
```
hannesdiedrich, line 58: This list will always have the length 1 right? Just a minor comment, 
``` python
        for bid in bids:
            market.delete_bid(bid)
```
suggests that there are actually more bids to be deleted. 
The following approach would also work right?
```python
market.delete(bids[0])
```

hannesdiedrich: Also, this is the PV, I guess that this is a copy paste issue, but I propose to rename to `offer`.

src/gsy_e/models/strategy/forward/load.py
```diff
@@ -33,17 +44,31 @@ def state(self) -> "LoadState":
...
+                           order_uuid, market.market_type, market_slot)
+        for bid in bids:
+            market.delete_bid(bid)
+
     def remove_open_orders(self, market: "ForwardMarketBase", market_slot: DateTime):
         bids = [bid
                 for bid in market.slot_bid_mapping[market_slot]
                 if bid.buyer == self.owner.name]
         for bid in bids:
             market.delete_bid(bid)
```
hannesdiedrich, line 63: Maybe you could add an optional parameter `oder_uuid` here and re-use this method in `remove_order`

src/gsy_e/models/strategy/forward/__init__.py
```diff
@@ -94,14 +119,122 @@ def event_offer_traded(self, *, market_id, trade):
...
+            order_updater_params.initial_rate = min(
+                order_updater_params.initial_rate, energy_rate)
+
+            self._strategy._order_updaters[market][slot] = OrderUpdater(
+                order_updater_params, market_parameters)
+            self._strategy.post_order(market, slot)
+
+    def _stop_auto_trading_event(self, event: Dict):
+        """Apply stop automatic trading event to the strategy."""
+        self._remove_order_event(event)
```
hannesdiedrich, line 208: This confused me at first. Do you agree on the following change?
```suggestion
    def _stop_auto_trading_event(self, args: Dict):
        """Apply stop automatic trading event to the strategy."""
        self._remove_order_event(args)
```

src/gsy_e/models/strategy/forward/__init__.py
```diff
@@ -94,14 +119,122 @@ def event_offer_traded(self, *, market_id, trade):
...
-        self._post_orders_to_new_markets()
+        if self.fully_automated_trading:
+            self._post_orders_to_new_markets()
 
     @property
     def state(self) -> "StateInterface":
         """Get the state class of the strategy. Needs to be implemented by all strategies"""
         raise NotImplementedError
+
+    def apply_live_event(self, event: Dict):
```
hannesdiedrich, line 141: I am missing the instance, that is actually calling this method, so the connection between gsy-web and here. Did I miss this?

src/gsy_e/models/strategy/forward/__init__.py
```diff
@@ -94,14 +119,122 @@ def event_offer_traded(self, *, market_id, trade):
...
+            if updater:
+                market.remove_open_orders(slot)
+
+            market_parameters = market.get_market_parameters_for_market_slot(slot)
+
+            order_updater_params = deepcopy(self._strategy._order_updater_params[market_type])
+            order_updater_params.capacity_percent = capacity_percent
+            order_updater_params.final_rate = energy_rate
+            order_updater_params.initial_rate = min(
+                order_updater_params.initial_rate, energy_rate)
```
hannesdiedrich, line 200: Doesn't this depend on whether you have a PV or a Load?

src/gsy_e/gsy_e_core/area_serializer.py
spyrostz: Agreed! Will fix this

src/gsy_e/gsy_e_core/rq_job_handler.py
spyrostz: This is one thing that was not clear from the requirements. The FULL_AUTO_TRADING is essentially only used in the CLI (where the template strategies are posting automatically bids and offers on all open market slots). This does not apply to the UI, since both auto and manual trading are only posting orders on specific market slots (according to the start / end time). The difference between the manual and the auto mode, is that in the auto mode, the order is updating its energy rate during the market slot, contrary to manual mode, where the order never updates its energy rate.

src/gsy_e/models/leaves.py
spyrostz: Indeed, thanks!

src/gsy_e/models/strategy/forward/__init__.py
spyrostz: This is still a remnant from the old "disable" toggle in the UI. I do not think that the UI will need it in the future, so I will remove it, thanks!

src/gsy_e/models/strategy/forward/__init__.py
spyrostz: Agreed, will rename this!

src/gsy_e/models/strategy/forward/pv.py
spyrostz: Agreed on both, will fix them, thanks!

src/gsy_e/models/strategy/forward/__init__.py
spyrostz: Thanks for the notification, I unfortunately missed committing my changes to the live_events.py file :( 
Having a second look at these changes though, I noticed that I had handled the live events for the forward markets similarly to the other live events - essentially processing them on every market slot. This will not be good for these though, since they involve trading commands. For this purpose, I will also change the event handling specifically for these events, to be processed on every tick and not on every market slot. Let me know if you do not agree, otherwise I will add this fix too. Thanks!

src/gsy_e/models/strategy/forward/__init__.py
```diff
@@ -94,14 +119,122 @@ def event_offer_traded(self, *, market_id, trade):
...
     def state(self) -> "StateInterface":
         """Get the state class of the strategy. Needs to be implemented by all strategies"""
         raise NotImplementedError
+
+    def apply_live_event(self, event: Dict):
+        """Apply the incoming live event to the strategy."""
+        self._live_event_handler.dispatch(event)
+
+
+class ForwardLiveEvents:
```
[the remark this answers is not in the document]
spyrostz: Agreed, moving it to its own module, thanks!

src/gsy_e/models/strategy/forward/__init__.py
spyrostz: This will not be that reusable, but I will give it a try to implement it as a decorator. Thanks!

src/gsy_e/models/strategy/forward/load.py
spyrostz: This was a mistake on my part, since it is not that reusable at the end. The PV needs to check the offer.seller and the Load the bid.buyer, and use the market.slot_offer_mapping and market.slot_bid_mapping respectively. I have adapted the tests already in one of the previous comments in order to catch this error. Thanks!

src/gsy_e/models/strategy/forward/__init__.py
spyrostz: Great catch! In order to not split the flow with if blocks for producer and consumer strategies, I will clamp the value between initial and final rate. That way it will never fall out of bounds of the predefined OrderUpdaterParams for each market type. Thanks again!

src/gsy_e/gsy_e_core/rq_job_handler.py
hannesdiedrich: I think I understand, let me summarise:
1. cli: FULLY_AUTO_TRADING = True: (default from gsy-framework), orders are posted on all markets automatically (handled [here](https://github.com/gridsingularity/gsy-e/pull/1556/files#diff-808df61501ae0753a81cf36ebb1d3ced3f1cacf3a0f07b870b1f8b0c7c42a784R123))
2. UI auto mode: orders are updated in a specific time frame (implemented [here](https://github.com/gridsingularity/gsy-e/pull/1556/files#diff-b79d23882e80bccfa3f0e1810199e0825122845183b2e1ade1bb7d9d9fea325dR63) )
3. UI manual mode: orders are updated in a specific time frame ones (implemented [here](https://github.com/gridsingularity/gsy-e/pull/1556/files#diff-b79d23882e80bccfa3f0e1810199e0825122845183b2e1ade1bb7d9d9fea325dR108))

src/gsy_e/models/strategy/forward/__init__.py
hannesdiedrich: I see. In retrospect, I agree, your solution was fine. 😅  But keep the decorator. Thanks a lot for dealing with this!

src/gsy_e/models/strategy/forward/load.py
```diff
@@ -33,17 +44,31 @@ def state(self) -> "LoadState":
...
 
+    def remove_order(self, market: "ForwardMarketBase", market_slot: DateTime, order_uuid: str):
+        bids = [bid
+                for bid in market.slot_bid_mapping[market_slot]
+                if bid.buyer == self.owner.name and bid.id == order_uuid]
+        if not bids:
+            self.log.error("Bid with id %s does not exist on the market %s %s.",
+                           order_uuid, market.market_type, market_slot)
+        for bid in bids:
+            market.delete_bid(bid)
```
hannesdiedrich, line 55: You probably forgot to adapt here:
```suggestion
      market.delete_bid(bids[0])
```

src/gsy_e/models/strategy/forward/__init__.py
hannesdiedrich: Your implementation makes sense. Thanks!

src/gsy_e/models/strategy/forward/__init__.py
hannesdiedrich: Great, but don't we also clamp the final rate, too?

hannesdiedrich (commented):
Two comments left. Thanks a lot for dealing with former.

src/gsy_e/gsy_e_core/rq_job_handler.py
spyrostz: You got it, but I would rephrase #3 to "orders are only posted once and never updated."

src/gsy_e/models/strategy/forward/__init__.py
spyrostz: The thing is that the final rate is configurable by the user. If we clamp it, then we will only allow the exchange to operate in a specific energy rate range, which is not something that we want. We only want to limit the starting price, not violate the desire of the user.

src/gsy_e/models/strategy/forward/load.py
spyrostz: Good catch, thanks!

src/gsy_e/models/strategy/forward/__init__.py
hannesdiedrich: Ok, so the `energy_rate` is rather a `not_buy_higher`/`not_sell_lower`- `rate` ?
Who is deciding about the initial values of the initial and final rates in the workflow. They are set when creating the setup AFAIS, right? So the owner of a PV power plant has to provide a bidding range when they join the exchange, right?
What if the like to exceed the initial/final range in the auto or manual trading mode? They just can not atm, right? Is there a motivation behind this that I do not see?

spyrostz: First of all, this is only applicable for the auto trading mode, since the manual trading mode dictates a specific price and no updates then there is no concept of initial / final update prices. WRT auto mode, there is only one field in the UI actually, only to denote the final price that is acceptable for the customer. The initial price is left uncontrolled, therefore the customer does not define the initial / final range, only the upper / lower rate limit that he is willing to accept. The initial rate is defined by us, and at the moment is set when creating the setup (will typically have the default value from the OrderUpdaterParameter class). 
Let's try to rethink this using all possible cases. For instance, think of these values for a load strategy:
```
initial_rate = 10
final_rate = 30
energy_rate = 50
```
In this case, the correct behavior is to not change the energy rate of the user, but instead to move only the final rate to 50. Therefore the values will be:
```
initial_rate = 10
final_rate = 30
energy_rate = 50
```
Second cornercase scenario is this:
```
energy_rate = 5
initial_rate = 10
final_rate = 30
```
In this case the final_rate will be equal to energy_rate, and the initial_rate should again drop to 5:
```
energy_rate = 5
initial_rate = 5
final_rate = 5
```
With the clamping, this one will not work as far as I see (the original implementation would, but only for the load and not the PV). Similar observations occur for the PV too. Let me adapt the code in order to handle both cases. Thanks for insisting here!

hannesdiedrich: Thanks a lot for the verbose answer! It and the implementation makes sense now. 

>  The initial rate is defined by us,

So how would we find numbers that are suitable for "intelligent" trading and also allow a wide range for the user?
This is no question for this PR, but for a general discussion when we revise this feature. So, feel free to skip answering it. :-) 
Thanks again!

hannesdiedrich (approved):
LGTM

PR #1559, opened 2022-11-17
Feature/gsye 434

hannesdiedrich:

The SCM strategies should be able to receive forecasts and measurements from the gsy-e-sdk

## Proposed changes

- Add ability to digest forecast and measurements sent by the gsy-e-sdk 
- fix pylint warning (you better review this commit separately) 

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-434

### DIFF
--- a/src/gsy_e/gsy_e_core/global_objects_singleton.py
+++ b/src/gsy_e/gsy_e_core/global_objects_singleton.py
@@ -16,7 +16,8 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from gsy_e.gsy_e_core.user_profile_handler import ProfilesHandler
-from gsy_e.gsy_e_core.global_stats import ExternalConnectionGlobalStatistics
+from gsy_e.gsy_e_core.global_stats import (
+    ExternalConnectionGlobalStatistics, SCMExternalConnectionGlobalStatistics)
 from gsy_e.gsy_e_core.market_counters import FutureMarketCounter
 
 
@@ -26,6 +27,7 @@ class GlobalObjects:
     profiles_handler = ProfilesHandler()
     external_global_stats = ExternalConnectionGlobalStatistics()
     future_market_counter = FutureMarketCounter()
+    scm_external_global_stats = SCMExternalConnectionGlobalStatistics()
 
 
 global_objects = GlobalObjects()
--- a/src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
+++ b/src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
@@ -16,6 +16,7 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 from abc import ABC, abstractmethod
+from statistics import mean
 from typing import TYPE_CHECKING, Dict, List
 
 from gsy_framework.constants_limits import ConstSettings
@@ -307,7 +308,7 @@ class CoefficientDataExporter(BaseDataExporter):
 
     @property
     def rows(self) -> List:
-        from statistics import mean
+
         if not self._scm:
             return []
         area_results = self._scm.get_area_results(self._area.uuid, serializable=False)
@@ -325,7 +326,7 @@ class CoefficientDataExporter(BaseDataExporter):
             t.traded_energy
             for t in area_results["after_meter_data"]["trades"]
         ]
-        return [[self._area.past_market_time_slot,
+        return [[self._area.current_market_time_slot,
                 mean(trade_rates) if trade_rates else 0.,
                 min(trade_rates) if trade_rates else 0.,
                 max(trade_rates) if trade_rates else 0.,
@@ -350,7 +351,7 @@ class CoefficientLeafDataExporter(BaseDataExporter):
 
     @property
     def rows(self):
-        slot = self._area.past_market_time_slot
+        slot = self._area.current_market_time_slot
         if slot is None:
             return []
         if isinstance(self._area.strategy, SCMStorageStrategy):
@@ -362,9 +363,9 @@ class CoefficientLeafDataExporter(BaseDataExporter):
         elif isinstance(self._area.strategy, (SCMLoadHoursStrategy, SCMLoadProfileStrategy)):
             desired = self._area.strategy.state.get_desired_energy_Wh(slot) / 1000
             # All energy is traded in SCM
-            return [[self._area.past_market_time_slot, desired, 0.]]
+            return [[slot, desired, 0.]]
         elif isinstance(self._area.strategy, SCMPVStrategy):
             not_sold = self._area.strategy.state.get_available_energy_kWh(slot)
             produced = self._area.strategy.state.get_energy_production_forecast_kWh(slot, 0.0)
-            return [[self._area.past_market_time_slot, produced, not_sold]]
+            return [[slot, produced, not_sold]]
         return []
--- a/src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
+++ b/src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
@@ -357,7 +357,7 @@ class CoefficientDataExporter(BaseDataExporter):
 
 
 class CoefficientLeafDataExporter(BaseDataExporter):
-    """LeafDataExporter for SCm simulations."""
+    """LeafDataExporter for SCM simulations."""
 
     def __init__(self, area: "CoefficientArea"):
         self._area = area
--- a/src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
+++ b/src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
@@ -37,6 +37,7 @@ if TYPE_CHECKING:
 
 
 class BaseDataExporter(ABC):
+    """Base class for data exporter."""
 
     @property
     @abstractmethod
@@ -50,6 +51,7 @@ class BaseDataExporter(ABC):
 
 
 class UpperLevelDataExporter(BaseDataExporter):
+    """Data exporter for areas higher in the grid tree."""
     def __init__(self, past_markets):
         self.past_markets = past_markets
 
@@ -67,7 +69,8 @@ class UpperLevelDataExporter(BaseDataExporter):
     def rows(self):
         return [self._row(m.time_slot, m) for m in self.past_markets]
 
-    def _row(self, slot, market):
+    @staticmethod
+    def _row(slot, market):
         return [slot,
                 market.avg_trade_price,
                 market.min_trade_price,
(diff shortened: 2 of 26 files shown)


src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
```diff
@@ -335,22 +357,24 @@ def rows(self) -> List:
 
 
 class CoefficientLeafDataExporter(BaseDataExporter):
+    """LeafDataExporter for SCm simulations."""
```
spyrostz, line 360: Minor typo, SCM with uppercase

src/gsy_e/gsy_e_core/simulation/results_manager.py
```diff
@@ -157,20 +157,20 @@ def update_scm_manager(self, scm_manager: SCMManager) -> None:
...
         """Update the csv results on market cycle."""
         if self.export_results_on_finish:
             self._export.data_to_csv(area, current_time_slot, slot_no == 0, scm_manager)
 
-    def save_csv_results(self, area: "Area") -> None:
+    def save_csv_results(self, area: "CoefficientArea") -> None:
         """Update the CSV results on finish, and write the CSV files."""
         if self.export_results_on_finish:
             log.info("Exporting simulation data.")
-            self._export.data_to_csv(area, False, None)
```
spyrostz, line 173: How come? Was it due to an error on the generation of the CSV files ?

hannesdiedrich: yes, the last market slot was written twice to the CSV files.

src/gsy_e/gsy_e_core/simulation/simulation.py
```diff
@@ -438,10 +448,24 @@ def _execute_simulation(
             self.progress_info.update(slot_no, slot_count, self._time, self.config)
 
             self.area.cycle_coefficients_trading(self.progress_info.current_slot_time)
+            log.error(self.area.current_market_time_slot)
```
spyrostz, line 451: I think that this can be removed, right ?

src/gsy_e/gsy_e_core/simulation/simulation.py
```diff
@@ -438,10 +448,24 @@ def _execute_simulation(
...
 
+            if self.config.external_connection_enabled:
+
+                self.config.external_redis_communicator.approve_aggregator_commands()
+
+                self.area.market_cycle_external()
+
+                self.config.external_redis_communicator.\
+                    publish_aggregator_commands_responses_events()
+
```
spyrostz, line 468: Hmmm, I do not like it that much that external connection handling is in the main simulation method. Can these if blocks not reside in a separate class? For instance, I think that all these can be wrapped around a class that we will only have to call once and execute all these methods in the class body instead of the _execute_simulation.

src/gsy_e/gsy_e_core/simulation/simulation.py
```diff
@@ -422,14 +425,21 @@ def _time_since_start(self) -> Duration:
...
     def _execute_simulation(
             self, slot_resume: int, _tick_resume: int, console: NonBlockingConsole = None) -> None:
         slot_count, slot_resume = (
             self._time.calc_resume_slot_and_count_realtime(
                 self.config, slot_resume))
 
+        if self.config.external_connection_enabled:
+            self.config.external_redis_communicator.sub_to_aggregator()
+            self.config.external_redis_communicator.start_communication()
+
```
spyrostz, line 442: I think that this is also not needed, if these were performed in the constructor of the external_redis_communicator.

src/gsy_e/gsy_e_core/simulation/simulation.py
```diff
@@ -411,6 +411,9 @@ def _init(self) -> None:
 
         self.area.activate_energy_parameters(self.config.start_date)
 
+        if self.config.external_connection_enabled:
+            global_objects.scm_external_global_stats(self.area)
```
spyrostz, line 415: Same comment as below, I think that it is better to embed the if block inside the body of the `scm_external_global_stats.__call__` in order to not have the if block in the simulation class

src/gsy_e/gsy_e_core/global_objects_singleton.py
```diff
@@ -26,6 +27,7 @@ class GlobalObjects:
     profiles_handler = ProfilesHandler()
     external_global_stats = ExternalConnectionGlobalStatistics()
     future_market_counter = FutureMarketCounter()
+    scm_external_global_stats = SCMExternalConnectionGlobalStatistics()
```
spyrostz, line 30: I think that the singleton can be avoided by dependency injection on the simulation class. Essentially the simulation class could potentially keep a member with an instance of this class, and pass it to the classes that need it. The complex part is that external_redis_communicator (that requires this) is part of SimulationConfig, therefore it would also need to be passed in the SimulationConfig class. However I think that it would be much better WRT code grouping than the singleton, and it can be avoided

spyrostz (commented):
Some comments, most about code structure. Functionality-wise it looks good!

src/gsy_e/gsy_e_core/global_objects_singleton.py
hannesdiedrich: This comment caught me off guard. I remember discussion in length how to make the `grid_stats_tree` available in multiple places of the code and decided for a singleton. Now I spend some time to investigate and indeed it turns out, that we only need the grid_stats_tree only here, as you said: 
`Simulation->SimulationConfig->ExternalConnectionCommunicator->AggregatorHandler`
However, in order to create and update the `grid_stats_tree`, we need to pass the root area like done [here](https://github.com/gridsingularity/gsy-e/blob/36d8a6fe0fe93b69fd5b88ddb44281c1de3d9699/src/gsy_e/gsy_e_core/simulation/simulation.py#L112). 
The class will probably need an `update_area` method then. Let me try this (just fearing that I already tried this in the past which was not successful and lead to the singleton)

src/gsy_e/gsy_e_core/simulation/simulation.py
hannesdiedrich: Let me try this. I only copied this from the main simulation class. There must be a reason why we did not put these in the constructor of the `ExternalConnectionCommunicator` class.

src/gsy_e/gsy_e_core/simulation/simulation.py
hannesdiedrich: I played around a lot with the order of these (also mixed them up with the the other commands in this method). After finding the right order i wanted to refactor this into a separate method, but forgot. I agree, lets even add a class that deals with the external connections on the SCM simulations.

src/gsy_e/gsy_e_core/simulation/simulation.py
hannesdiedrich: yes, remnant of testing (normally I use print() which is why this got overlooked).

src/gsy_e/gsy_e_core/global_objects_singleton.py
hannesdiedrich: @spyrostz I made good progress on this, but unfortunately, I found out, that we also need access to the `ExternalConnectionGlobalStatistics` in the external mixing [here](https://github.com/gridsingularity/gsy-e/blob/36d8a6fe0fe93b69fd5b88ddb44281c1de3d9699/src/gsy_e/models/strategy/external_strategies/__init__.py#L472) and [here](https://github.com/gridsingularity/gsy-e/blob/36d8a6fe0fe93b69fd5b88ddb44281c1de3d9699/src/gsy_e/models/strategy/external_strategies/__init__.py#L542) [whereas the second usage seems to be avoidable]
The only solution I see is to create an `EXTERNAL_TICK` event and to controll the dispatching from the simulation class, which I think is an overkill. What do you think.

src/gsy_e/gsy_e_core/simulation/simulation.py
hannesdiedrich: As the external communication is done a bit differently in the normal simulation I decided to only move these steps into a separate method in the SCMSimulation: [0d2cbfc](https://github.com/gridsingularity/gsy-e/pull/1559/commits/0d2cbfcb410817f89992821aed22687ae80b80af)
Please let me know what you think. Thanks

src/gsy_e/gsy_e_core/simulation/simulation.py
hannesdiedrich: I fixed this in here [0d2cbfc](https://github.com/gridsingularity/gsy-e/pull/1559/commits/0d2cbfcb410817f89992821aed22687ae80b80af)

src/gsy_e/gsy_e_core/simulation/simulation.py
```diff
@@ -430,16 +427,26 @@ def _deactivate_areas(self, area: "AreaBase"):
         for child in area.children:
             self._deactivate_areas(child)
 
+    def _handle_external_communication(self):
```
spyrostz, line 430: I like a lot that the external connection stuff are at least extracted to their own method! However, is it possible to not even call these methods using an early-return? For instance:

```suggestion
    def _handle_external_communication(self):
        if not self.config.external_connection_enabled:
            return
```
Would that work out ?

spyrostz (dismissed):
Looks much better now, and left the LGTM already even though I have one minor comment! 

As a general refactoring step for the future though (and frankly the root cause of my former comments), I have to suggest to take out the external_redis_communicator from SimulationConfig class. I know that this is a long-standing issue and we have discussed it again in the past, however I believe that this will become even less manageable in the future if we add more external connections (in the sense that we will be obliged to use more singletons).
To summarise my thoughts, if the external_redis_communicator was not hidden inside config, then we could pass the area object to it and encapsulate this whole logic inside the communicator, without having singletons and it being completely decoupled. This is outside the scope of this PR though, and I will probably open a ticket in order to explicitly refactor this. Thanks !

src/gsy_e/gsy_e_core/simulation/simulation.py
hannesdiedrich: That should work. All the methods inside check again if the aggregator is enabled, but why not returning early. Will do. Thanks!

spyrostz (dismissed):

src/gsy_e/gsy_e_core/simulation/simulation.py
hannesdiedrich: @spyrostz we have a problem with this change. The connection to redis might not be initiated in this when calling these methods in the init of the external_redis_communicator [here](https://github.com/gridsingularity/gsy-e/pull/1559/commits/0d2cbfcb410817f89992821aed22687ae80b80af#diff-dacdd5512566d09ead79d7866dc73d81fffa3907abad2f0880ddb06c8af5d343R158) (have a look at the failing tests, you can reproduce these locally by turning off the redis server)
This is probably why it was pulled out of there into the simulation class. I propose to add an `activate` method that is called after the init (from the simulation class) which is the same implementation as before, only grouped into a method. What do you thin?

spyrostz: @hannesdiedrich too bad to hear! I agree with your suggestion, and I think that it is good because it will separate the parts of the code that need to be part of the activate process. Thanks again!

hannesdiedrich: I did that [cee9417](https://github.com/gridsingularity/gsy-e/pull/1559/commits/cee9417511d93ec337c3e32943889dc3aa19c5a4)

PR #1561, opened 2022-11-18
GSYE-450: Pass daily_load_profile_uuid to super class in LoadProfileF…

hannesdiedrich:
…orecastExternalStrategy in order to be saved as attribute to the class and read by the profile handler

In CNs profiles where not buffered by the profiles handler because it was not aware of the profiles that were used by the `LoadProfileForecastExternalStrategy` . This Strategy is automatically selected, because `forecast_stream_enabled=True` is default for al CN assets.  

## Proposed changes

The fix was easy. We need to pass the `daily_load_profile_uuid` to the super class `DefinedLoadStrategy` that is saving it as an class attribute to be read by `ProfilesHandler._get_profile_uuids_from_setup`.


spyrostz (approved):
LGTM

PR #1564, opened 2022-11-23
GSYE-447: Test for normal simulation of endpoint buffer.

Maheen12:

## Proposed changes

-

### DIFF
--- a/tests/test_endpoint_buffers.py
+++ b/tests/test_endpoint_buffers.py
@@ -38,6 +38,7 @@ def forward_setup_fixture():
     GlobalConfig.market_maker_rate = original_market_marker_rate
     ConstSettings.ForwardMarketSettings.ENABLE_FORWARD_MARKETS = original_enable_forward_markets
 
+
 @pytest.fixture(name="general_setup")
 def general_setup_fixture():
     """Create area with spot market"""
@@ -178,8 +179,8 @@ class TestSimulationEndpointBuffer:
 
                 market.trades.extend([
                     self._generate_order(
-                    creation_time=current_time - slot_length,
-                    time_slot=f"TIME_SLOT_{i}") for i in range(2)])
+                        creation_time=current_time - slot_length,
+                        time_slot=f"TIME_SLOT_{i}") for i in range(2)])
             current_time += slot_length
 
             # check results are generated correctly.
@@ -201,6 +202,7 @@ class TestSimulationEndpointBuffer:
                         assert (current_time - slot_length <=
                                 orders[0]["creation_time"] < current_time)
                 current_time += slot_length
+
 
 class TestSimulationEndpointBufferForward:
     """Tests for the SimulationEndpointBuffer class."""
--- a/tests/test_endpoint_buffers.py
+++ b/tests/test_endpoint_buffers.py
@@ -4,11 +4,13 @@ from uuid import uuid4
 
 import pytest
 from gsy_framework.constants_limits import ConstSettings, GlobalConfig
-from pendulum import DateTime, duration
+from pendulum  import datetime
+import pendulum
 
-from gsy_framework.enums import AvailableMarketTypes
+from gsy_framework.enums import AvailableMarketTypes, SpotMarketTypeEnum
 from gsy_e.gsy_e_core.enums import FORWARD_MARKET_TYPES
-from gsy_e.gsy_e_core.sim_results.endpoint_buffer import SimulationEndpointBuffer
+from gsy_e.gsy_e_core.sim_results.endpoint_buffer import SimulationEndpointBuffer, CoefficientEndpointBuffer
+from gsy_e.models.area.scm_manager import SCMManager
 
 logger = logging.getLogger(__name__)
 
@@ -20,7 +22,7 @@ def forward_setup_fixture():
     original_enable_forward_markets = ConstSettings.ForwardMarketSettings.ENABLE_FORWARD_MARKETS
     GlobalConfig.market_maker_rate = 30
     ConstSettings.ForwardMarketSettings.ENABLE_FORWARD_MARKETS = True
-    slot_length = duration(minutes=15)
+    slot_length = pendulum.duration(minutes=15)
     forward_markets = {  # each forward market will have 5 timeslots.
         market_type: MagicMock(market_time_slots=[f"TIME_SLOT_{i}" for i in range(5)])
         for market_type in FORWARD_MARKET_TYPES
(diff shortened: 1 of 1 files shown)


tests/test_endpoint_buffers.py
```diff
@@ -38,7 +39,172 @@ def forward_setup_fixture():
...
+        """Test results are being correctly generated with respect to
+        spot market"""
+        area, slot_length = general_setup
+
+        progress_info = MagicMock()
+        endpoint_buffer = SimulationEndpointBuffer("JOB_1", 41, area, False)
+
+        # add bids/offers/trades
+        start_time = DateTime(2020, 1, 1, 0, 15)
+        for market in area.spot_market():
```
hannesdiedrich, line 164: Why is this for loop needed?

tests/test_endpoint_buffers.py
```diff
@@ -38,7 +39,172 @@ def forward_setup_fixture():
...
+
+            # check results are generated correctly.
+            current_time = start_time
+            area.now = current_time
+            endpoint_buffer.update_stats(area, "running", progress_info, {}, False)
+            # pylint: disable=protected-access
+            raw_results = endpoint_buffer._generate_result_report()["simulation_raw_data"]
+            area_general_stats = raw_results[area.uuid]["general_market_stats"]
+
+            for market_type in AvailableMarketTypes.SPOT:
```
hannesdiedrich, line 194: Also here, if you are only testing the SPOT market, no loop is needed, right?

tests/test_endpoint_buffers.py
```diff
@@ -38,7 +39,172 @@ def forward_setup_fixture():
...
+
+                market.offers[uuid4()] = self._generate_order(
+                    creation_time=current_time - slot_length,
+                    time_slot=f"TIME_SLOT_{i}")
+
+                market.trades.extend([
+                    self._generate_order(
+                        creation_time=current_time - slot_length,
+                        time_slot=f"TIME_SLOT_{i}") for i in range(2)])
+            current_time += slot_length
```
hannesdiedrich, line 184: Why you are adding to the `current_time` here, but overwrite this variable in line 187?

tests/test_endpoint_buffers.py
```diff
@@ -38,7 +39,172 @@ def forward_setup_fixture():
...
+
+            # check results are generated correctly.
+            current_time = start_time
+            area.now = current_time
+            endpoint_buffer.update_stats(area, "running", progress_info, {}, False)
+            # pylint: disable=protected-access
+            raw_results = endpoint_buffer._generate_result_report()["simulation_raw_data"]
+            area_general_stats = raw_results[area.uuid]["general_market_stats"]
```
hannesdiedrich, line 192: Were are the `general_market_stats` added and what are they? I do not see it in the code.

tests/test_endpoint_buffers.py
```diff
@@ -38,7 +39,172 @@ def forward_setup_fixture():
...
+                "percentage_completed": 0
+            },
+            "simulation_state": {"general": {}, "areas": {}},
+            "mocked-results": "some-results"
+        }
+
+    def test_general_results_are_generated(  # pylint: disable-msg=too-many-locals
+            self, general_setup):
+        """Test results are being correctly generated with respect to
+        spot market"""
```
hannesdiedrich, line 156: I do not understand which unit you are really testing here. Can you please explain?

hannesdiedrich (commented):
Left some comments. 
Also, I would expect that you test if update_stats performs all the stuff that is part of the method (e.g. creating the `area_tree_dict`) or what do you think?

Maheen12:
> Left some comments. Also, I would expect that you test if update_stats performs all the stuff that is part of the method (e.g. creating the `area_tree_dict`) or what do you think?

Yes I think you are right, update_stats should be tested because its part of SimulationEndpointBuffer. I will work on it. Thanks

tests/test_endpoint_buffers.py
Maheen12: Actually some tests are already written in correspondence to Forward Market. Most of the testing units are common for both Forward and normal simulation. I understood that there will be a corresponding unit for normal results being generated. I think I might got it wrong.
But I will have a detailed look once again,  otherwise I will remove it and add the test for update stats of normal simulation.

tests/test_endpoint_buffers.py
```diff
@@ -38,7 +42,217 @@ def forward_setup_fixture():
     ConstSettings.ForwardMarketSettings.ENABLE_FORWARD_MARKETS = original_enable_forward_markets
 
 
+@pytest.fixture(name="general_setup")
+def general_setup_fixture():
+    """Create area with spot market"""
```
hannesdiedrich, line 47: I like to link a tip from Spyros here: https://github.com/gridsingularity/gsy-e/pull/1568#discussion_r1043071363

tests/test_endpoint_buffers.py
```diff
@@ -272,8 +486,37 @@ def test_update_stats_forward_markets(self, forward_setup):
 
 
 class TestCoefficientEndpointBuffer(TestSimulationEndpointBuffer):
-    """Tests for the CoefficientEndpointBuffer class.
+    """Tests for the CoefficientEndpointBuffer class."""
+    def test_update_stats_scm(self, scm_setup):
+        CoefficientArea, _ = scm_setup
```
hannesdiedrich, line 491: CamelCase should only be used for [classes](https://peps.python.org/pep-0008/#class-names). For normal variables and fucntions we use [lowercase](https://peps.python.org/pep-0008/#function-and-variable-names)
```suggestion
        coefficient_area, _ = scm_setup
```

hannesdiedrich (commented):
Left some comments regarding cosmetics, LGTM in general

Maheen12:
> Left some comments regarding cosmetics, LGTM in general

I will fix them! Thank you

tests/test_endpoint_buffers.py
```diff
@@ -38,7 +42,217 @@ def forward_setup_fixture():
...
+
+        get_json_dict_memory_allocation_size_mock.return_value = 128000
+        with caplog.at_level(logging.WARNING):
+            output = endpoint_buffer.prepare_results_for_publish()
+            assert "Do not publish message bigger than 64 MB, current message size 128.0 MB." \
+                in caplog.text
+
+        assert output == {}
+
+    def test_generate_json_report(self, general_setup):
```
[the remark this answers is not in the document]
Maheen12: Can I name it something like this: `def test_generate_json_report_created_successfully`. and this should be done to all the names of the unit tests?

tests/test_endpoint_buffers.py
Maheen12: got it!

hannesdiedrich (approved):
LGTM

PR #1565, opened 2022-11-24
GSYE-427-pre: Move OrderUpdater out of forward package in order to be…

hannesdiedrich:
… used by normal strategies, too

For the heat pump strategy we like to re-use classes that were implemented for the forward markets.

## Proposed changes

- Moved OrdeUpdater out of the forward package [323a165](https://github.com/gridsingularity/gsy-e/pull/1565/commits/323a1652a2122a9133d3dc952b0bf1697e2fb6d8)
- Fixed a bug in the ForwardBaseStrategy [8e73427](https://github.com/gridsingularity/gsy-e/pull/1565/commits/8e73427f3d0e0b89fbcb8a21e9514f224a21eee5)
- Pulled re-usable functionality out of the ForwardStrategyBase class and moved it to the NewStrategyBase class [862bb65](https://github.com/gridsingularity/gsy-e/pull/1565/commits/862bb6509b36a43ac2fad6787bf8a85563e10f93)

### DIFF
--- a/src/gsy_e/models/market/__init__.py
+++ b/src/gsy_e/models/market/__init__.py
@@ -16,7 +16,6 @@ You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
 import uuid
-from abc import abstractmethod
 from collections import namedtuple
 from dataclasses import dataclass
 from functools import wraps
@@ -293,7 +292,6 @@ class MarketBase:  # pylint: disable=too-many-instance-attributes
         return sum(trade.trade_price for trade in self.trades if trade.seller == seller)
 
     @staticmethod
-    @abstractmethod
     def _calculate_closing_time(delivery_time: DateTime) -> DateTime:
         """
         Closing time of the market. Uses as basis the delivery time in order to calculate it.
--- a/src/gsy_e/models/market/__init__.py
+++ b/src/gsy_e/models/market/__init__.py
@@ -15,19 +15,20 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
-
 import uuid
+from abc import abstractmethod
 from collections import namedtuple
+from dataclasses import dataclass
 from functools import wraps
 from logging import getLogger
 from threading import RLock
-from typing import Dict, List, Union, Optional, Callable
+from typing import Dict, List, Union, Optional, Callable, TYPE_CHECKING
 
 from gsy_framework.constants_limits import ConstSettings, GlobalConfig
 from gsy_framework.data_classes import Offer, Trade, Bid
 from gsy_framework.enums import SpotMarketTypeEnum
 from numpy.random import random
-from pendulum import DateTime
+from pendulum import DateTime, duration
 
 from gsy_e.constants import FLOATING_POINT_TOLERANCE, DATE_TIME_FORMAT
 from gsy_e.gsy_e_core.device_registry import DeviceRegistry
@@ -37,6 +38,9 @@ from gsy_e.models.market.grid_fees.constant_grid_fees import ConstantGridFees
 from gsy_e.models.market.market_redis_connection import (
     MarketRedisEventSubscriber, MarketRedisEventPublisher,
     TwoSidedMarketRedisEventSubscriber)
+
+if TYPE_CHECKING:
+    from gsy_e.models.config import SimulationConfig
 
 log = getLogger(__name__)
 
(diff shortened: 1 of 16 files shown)


src/gsy_e/models/strategy/new_base_strategy.py
```diff
@@ -0,0 +1,96 @@
...
+from gsy_e.events import EventMixin
+from gsy_e.models.base import AreaBehaviorBase
+from gsy_e.models.strategy.order_updater import (OrderUpdater, OrderUpdaterParameters)
+
+if TYPE_CHECKING:
+    from gsy_e.models.state import StateInterface
+    from gsy_e.models.market import MarketBase
+
+
+class NewStrategyBase(EventMixin, AreaBehaviorBase, ABC):
```
hannesdiedrich, line 16: I am very open for a better naming for this class. Do you have an idea?

src/gsy_e/models/strategy/new_base_strategy.py
```diff
@@ -0,0 +1,96 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def post_order(self, market: "MarketBase", market_slot: DateTime,
+                   order_rate: float = None, **kwargs):
+        """Post orders to the markets that just opened."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def remove_order(self, market: "MarketBase", market_slot: DateTime, order_uuid: str):
```
hannesdiedrich, line 77: @spyrostz I see that this method is not being used anywhere (also not in the child classes like forward load). Do you foresee it being used in the future or do you think that `remove_open_orders` would be sufficient for now? 
In any case, I would vote to remove the `abstractmethod` decorator. Do you agree?

src/gsy_e/models/market/__init__.py
```diff
@@ -271,3 +291,42 @@ def total_earned(self, seller: str) -> float:
         """Return the aggregated money earned by the passed-in seller."""
 
         return sum(trade.trade_price for trade in self.trades if trade.seller == seller)
+
+    @staticmethod
+    @abstractmethod
```
spyrostz, line 296: Hmm, I am generally not a fan of abstractmethods with implementations, however for this case it is okay to have. The alternative would be to implement it in the OneSidedMarket class however it does not make that much of a difference here.

src/gsy_e/models/strategy/new_base_strategy.py
spyrostz: I am in between here actually. The reason why I implemented this class was my misunderstanding of a part of the design, where I assumed that a user should be able to delete a bid / offer based on the order uuid. This was not supported in the B2B design after all, that's the reason why it is not being used anywhere. Having said that though, this method deserves to be in the NewStrategyBase class, since I am 100% certain that it will be needed in the long run (e.g. I suspect that the SDK user will have to also remove the orders that were posted, in which case this method would be used). That's why I would recommend to keep it despite it not being used ATM

src/gsy_e/models/strategy/new_base_strategy.py
spyrostz: I am tempted to suggest `PlanningTradingStrategyBase` in order to indicate that this is the base class that should incorporate both planning and trading algorithms (planning algorithms belong in the EnergyParams classes, which are a member of the strategy class). Or, if we do not want to mention the planning part, `TradingStrategyBase` would also be okay for me.

spyrostz (commented):
Some comments, will LGTM once resolved. Thanks!

src/gsy_e/models/strategy/new_base_strategy.py
hannesdiedrich: Ok, so you mean that the Baseclass that we used until now `BaseStrategy` only dealt with the trading and with this we distinguish between those. 
I would still vote for TradingStrategyBase.
Eventually (or never, because it is too complicated) we could make the original strategies use this new base blass right?

src/gsy_e/models/strategy/new_base_strategy.py
hannesdiedrich: I understand and agree. Will implement it with a `pass` in the HeatPump

src/gsy_e/models/market/__init__.py
hannesdiedrich: Actually, having this as abstractmethod, I need to implement it in the OneSidedMarket and all subclasses, right?
let me remove the abstractmethod here and let the subclasses re-use it (as intended)

hannesdiedrich: [575a5ba](https://github.com/gridsingularity/gsy-e/pull/1565/commits/575a5ba109398e065ca763f6ea44371fd8e66bda)

src/gsy_e/models/strategy/new_base_strategy.py
hannesdiedrich: [46102c1](https://github.com/gridsingularity/gsy-e/pull/1565/commits/46102c1a8d2626a8dab189d1e01d0a06e3f2acdc)

spyrostz (approved):
LGTM

PR #1566, opened 2022-11-28
feature/GSYE-349: Add future market tests for ExternalStorageStrategy

mrsaemir:
## Proposed changes

### DIFF
--- a/src/gsy_e/models/strategy/__init__.py
+++ b/src/gsy_e/models/strategy/__init__.py
@@ -21,22 +21,21 @@ import sys
 from abc import ABC
 from dataclasses import dataclass
 from logging import getLogger
-from typing import List, Dict, Union, Optional, Generator, Callable, TYPE_CHECKING
+from typing import TYPE_CHECKING, Callable, Dict, Generator, List, Optional, Union
 from uuid import uuid4
 
 from gsy_framework.constants_limits import ConstSettings
-from gsy_framework.data_classes import (Offer, Bid, Trade)
+from gsy_framework.data_classes import Bid, Offer, Trade
 from gsy_framework.enums import SpotMarketTypeEnum
 from gsy_framework.utils import limit_float_precision
 from pendulum import DateTime
 
 from gsy_e import constants
-from gsy_e.constants import FLOATING_POINT_TOLERANCE
-from gsy_e.constants import REDIS_PUBLISH_RESPONSE_TIMEOUT
+from gsy_e.constants import FLOATING_POINT_TOLERANCE, REDIS_PUBLISH_RESPONSE_TIMEOUT
 from gsy_e.events import EventMixin
 from gsy_e.events.event_structures import AreaEvent, MarketEvent
 from gsy_e.gsy_e_core.device_registry import DeviceRegistry
-from gsy_e.gsy_e_core.exceptions import D3ARedisException, SimulationException, MarketException
+from gsy_e.gsy_e_core.exceptions import D3ARedisException, MarketException, SimulationException
 from gsy_e.gsy_e_core.redis_connections.area_market import BlockingCommunicator
 from gsy_e.gsy_e_core.util import append_or_create_key
 from gsy_e.models.base import AreaBehaviorBase
@@ -50,6 +49,7 @@ log = getLogger(__name__)
 
 if TYPE_CHECKING:
     from gsy_framework.data_classes import TradeBidOfferInfo
+
     from gsy_e.models.market.one_sided import OneSidedMarket
     from gsy_e.models.market.two_sided import TwoSidedMarket
 
@@ -462,7 +462,9 @@ class BaseStrategy(EventMixin, AreaBehaviorBase, ABC):
         self._settlement_market_strategy = self._create_settlement_market_strategy()
         self._future_market_strategy = self._create_future_market_strategy()
 
-    def serialize(self):
+    @staticmethod
+    def serialize():
+        """Serialize strategy status."""
         return {}
 
     @property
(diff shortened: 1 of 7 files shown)


src/gsy_e/models/strategy/__init__.py
```diff
@@ -511,7 +519,7 @@ def post_offer(self, market, replace_existing=True, **offer_kwargs) -> Offer:
         """
         if replace_existing:
             # Remove all existing offers that are still open in the market
-            self.offers.remove_offer_from_cache_and_market(market)
+            self._remove_existing_offers(market, offer_kwargs.get("time_slot") or market.time_slot)
```
hannesdiedrich, line 522: Can you please give a bid of context here? AFAIS: `self.offers.remove_offer_from_cache_and_market` also dealt with deleting all offers from the market (if `offer_id` was set to None, which was done here before).
So, either it is able to delete all orders of one asset from the market or we need to remove this functionality from this method. 
Curious about your explanation.

tests/strategies/external/test_storage.py
```diff
@@ -93,6 +127,82 @@ def test_bid_aggregator_succeeds_with_warning_if_dof_are_disabled(external_stora
...
+                    "transaction_id": str(uuid.uuid4())
+                }
+            )
+            assert return_value["status"] == "ready"
+            bid_id = json.loads(return_value["bid"])["id"]
+            assert future_markets.bids[bid_id].energy == future_energy_kWh
+        assert len(future_markets.bids.values()) == len(future_markets.market_time_slots)
+
+    @staticmethod
+    def test_bid_aggregator_fails_placing_future_bid_more_than_required_energy(
```
hannesdiedrich, line 154: NIT (which applies to all method names in this PR). We have a convention for unit test method names [here](https://gridsingularity.atlassian.net/wiki/spaces/D3A/pages/2406449155/Naming+and+Style+conventions#Unittests) . Following this, I would expect another `_fails` here. Would you agree?

hannesdiedrich (commented):
Some questions and one comment regarding our naming convention. LGTM in general

src/gsy_e/models/strategy/__init__.py
mrsaemir: I did this to avoid deletion of all bids/offers when somebody posts a new bid/offer on the future market (and sets replace_existing=True). IMO, this is a bug and is evident in existing unit tests that I have fixed. What do you think?

tests/strategies/test_strategy_base.py
```diff
@@ -449,19 +449,19 @@ def test_post_offer_with_replace_existing(market_class):
 
     # Post a first offer on the market
     offer_1_args = {
-        "price": 1, "energy": 1, "seller": "seller-name", "seller_origin": "seller-origin-name"}
+        "price": 1, "energy": 1, "seller": "FakeOwner", "seller_origin": "seller-origin-name"}
```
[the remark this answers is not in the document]
mrsaemir: Sure. The seller/buyer name should match the owner name [here](https://github.com/gridsingularity/gsy-e/blob/9fc1f45a9491cb56b99072ee79e6f5c53461b85b/src/gsy_e/models/strategy/__init__.py#L509) and [here](https://github.com/gridsingularity/gsy-e/blob/9fc1f45a9491cb56b99072ee79e6f5c53461b85b/src/gsy_e/models/strategy/__init__.py#L752), otherwise we end up having failed assertions.

tests/strategies/external/test_storage.py
```diff
@@ -101,7 +211,7 @@ def test_offer_aggregator_succeeds_with_warning_if_dof_are_disabled(external_sto
         """
         external_storage.simulation_config.enable_degrees_of_freedom = False
         external_storage.state.energy_to_sell_dict[
-            external_storage.spot_market.time_slot] = 1000.0
+            external_storage.spot_market.time_slot] = 1
```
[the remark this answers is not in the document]
mrsaemir: Just to show that the units are in kWh and not Wh.

tests/strategies/external/test_storage.py
mrsaemir: Thanks. Will fix it.

tests/strategies/test_strategy_base.py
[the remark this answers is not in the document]
hannesdiedrich: Ah, I see now. Thanks! Thanks for adding the assertions"

src/gsy_e/models/strategy/__init__.py
hannesdiedrich: > I did this to avoid deletion of all bids/offers when somebody posts a new bid/offer on the future market (and sets replace_existing=True

We are talking about bids/offers of the same asset right? AFAIS, `self.offers.remove_offer_from_cache_and_market` already took care of this. 

> is evident in existing unit tests that I have fixed 

can you please point me to it? Thanks!

mrsaemir: Sure. Lines 147-151 in [here](https://github.com/gridsingularity/gsy-e/pull/1566/files#diff-e5331d4b9154206362fa92d2e6f72dd3003a688e680626f0ec535f65af9a79f6L147). This test was posting offers for different timeslots in the future but only the last one remained because it used to flush the offers cache before posting a new one!

hannesdiedrich: Ahh, now I get it. The `post_offer(..., replace_existing=True)` was not respecting the market slot and was deleting all orders from all future markets before. Now it reads the `market_slot` and calls `_remove_existing_offers` where `get_posted_offers` finds the list of offers that are posted in the future market. Right?
(Sorry for thinking out loud)

hannesdiedrich (approved):
Thanks for bearing with me! LGTM

PR #1568, opened 2022-11-29
GSYE-427: Add HeatPumpStrategy and tests

hannesdiedrich:

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-427

### DIFF
--- a/src/gsy_e/models/strategy/heat_pump.py
+++ b/src/gsy_e/models/strategy/heat_pump.py
@@ -1,0 +1,171 @@
+from typing import Dict, TYPE_CHECKING
+
+from gsy_framework.constants_limits import ConstSettings
+from gsy_framework.data_classes import Trade
+from gsy_framework.enums import AvailableMarketTypes
+from pendulum import DateTime, duration
+
+from gsy_e.constants import FLOATING_POINT_TOLERANCE
+from gsy_e.constants import FutureTemplateStrategiesConstants
+from gsy_e.models.state import HeatPumpState
+from gsy_e.models.strategy.energy_parameters.heat_pump import HeatPumpEnergyParameters
+from gsy_e.models.strategy.new_base_strategy import NewStrategyBase
+from gsy_e.models.strategy.order_updater import OrderUpdaterParameters, OrderUpdater
+
+DEFAULT_HEAT_PUMP_ORDER_UPDATE_PARAMS = {
+    AvailableMarketTypes.SPOT: OrderUpdaterParameters(
+        initial_rate=ConstSettings.HeatPumpSettings.BUYING_RATE_RANGE.initial,
+        final_rate=ConstSettings.HeatPumpSettings.BUYING_RATE_RANGE.final,
+        update_interval=duration(
+                minutes=ConstSettings.GeneralSettings.DEFAULT_UPDATE_INTERVAL)),
+    AvailableMarketTypes.FUTURE: OrderUpdaterParameters(
+        initial_rate=FutureTemplateStrategiesConstants.INITIAL_BUYING_RATE,
+        final_rate=FutureTemplateStrategiesConstants.FINAL_BUYING_RATE,
+        update_interval=duration(
+            minutes=FutureTemplateStrategiesConstants.UPDATE_INTERVAL_MIN)),
+}
+
+if TYPE_CHECKING:
+    from gsy_e.models.market import MarketBase
+
+
+class HeatPumpStrategy(NewStrategyBase):
+    """Strategy for heat pumps with storages."""
+    # pylint: disable=too-many-arguments)
+    def __init__(self,
+                 maximum_power_rating: float = ConstSettings.HeatPumpSettings.MAX_POWER_RATING_KW,
+                 min_temperatur_C: float = ConstSettings.HeatPumpSettings.MIN_TEMP_C,
+                 max_temperatur_C: float = ConstSettings.HeatPumpSettings.MAX_TEMP_C,
+                 external_temperatur_C: float = ConstSettings.HeatPumpSettings.EXT_TEMP_C,
+                 tank_volume_l: float = ConstSettings.HeatPumpSettings.TANK_VOL_L,
+                 average_consumption: float =
+                 ConstSettings.HeatPumpSettings.AVERAGE_CONSUMPTION_KW,
+                 source_type: int = ConstSettings.HeatPumpSettings.TYPE,
+                 order_updater_parameters: Dict[
+                     AvailableMarketTypes, OrderUpdaterParameters] = None,
+                 preferred_buying_rate: float =
+                 ConstSettings.HeatPumpSettings.PREFERRED_BUYING_RATE
+                 ):
+        if not order_updater_parameters:
+            order_updater_parameters = DEFAULT_HEAT_PUMP_ORDER_UPDATE_PARAMS
+
+        super().__init__(order_updater_parameters=order_updater_parameters)
+
+        self._energy_params = HeatPumpEnergyParameters(
+            maximum_power_rating=maximum_power_rating,
+            min_temperatur_C=min_temperatur_C,
+            max_temperatur_C=max_temperatur_C,
+            external_temperatur_C=external_temperatur_C,
+            tank_volume_l=tank_volume_l,
+            average_consumption=average_consumption,
+            source_type=source_type
+        )
+
+        self.preferred_buying_rate = preferred_buying_rate
+        self._markets_buffer = []  # TODO: discuss if this buffer is really necessary
+
+    def serialize(self):
+        """Return serialised energy params."""
+        return {**self._energy_params.serialize()}
+
+    @property
+    def state(self) -> HeatPumpState:
+        return self._energy_params.state
+
+    def event_activate(self, **kwargs):
+        self._populate_markets()
+
+    def _populate_markets(self):
+        self._markets_buffer = [self.area.spot_market]
+        if ConstSettings.FutureMarketSettings.FUTURE_MARKET_DURATION_HOURS > 0:
+            self._markets_buffer.append(self.area.future_markets)
+
+    def event_market_cycle(self) -> None:
+        super().event_market_cycle()
+        self._post_orders_to_new_markets()
+
+    def event_tick(self):
+        self._update_open_orders()
+
+    def event_bid_traded(self, *, market_id: str, bid_trade: Trade) -> None:
+        """TODO: to be tested in the frame of GSYE-426"""
+        market = [market for market in self._markets_buffer if market_id == market.id]
+        if not market:
+            return
+
+        time_slot = bid_trade.time_slot
+        if bid_trade.buyer_id != self.owner.uuid:
+            return
+
+        self._energy_params.event_traded_energy(market[0], time_slot)
+        self._energy_params.decrement_posted_energy(market[0], time_slot)
+
+    def remove_order(self, market: "MarketBase", market_slot: DateTime, order_uuid: str):
+        """TODO: discuss if this is needed and decide: implement or pass"""
+
+    def post_order(
+            self, market: "MarketBase", market_slot: DateTime, order_rate: float = None, **kwargs):
+
+        if not order_rate:
+            order_rate = self._order_updaters[market][market_slot].get_energy_rate(
+                self.area.now)
+        order_energy_kWh = self._get_energy_buy_energy(order_rate, market_slot)
+
+        if order_energy_kWh <= FLOATING_POINT_TOLERANCE:
+            return
+        market.bid(
+            order_rate * order_energy_kWh, order_energy_kWh,
+            buyer=self.owner.name,
+            original_price=order_rate * order_energy_kWh,
+            buyer_origin=self.owner.name,
+            buyer_origin_id=self.owner.uuid,
+            buyer_id=self.owner.uuid,
+            time_slot=market_slot)
+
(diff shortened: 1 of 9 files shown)


src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -0,0 +1,171 @@
...
+                 external_temperatur_C: float = ConstSettings.HeatPumpSettings.EXT_TEMP_C,
+                 tank_volume_l: float = ConstSettings.HeatPumpSettings.TANK_VOL_L,
+                 average_consumption: float =
+                 ConstSettings.HeatPumpSettings.AVERAGE_CONSUMPTION_KW,
+                 source_type: int = ConstSettings.HeatPumpSettings.TYPE,
+                 order_updater_parameters: Dict[
+                     AvailableMarketTypes, OrderUpdaterParameters] = None,
+                 preferred_buying_rate: float =
+                 ConstSettings.HeatPumpSettings.PREFERRED_BUYING_RATE
+                 ):
```
hannesdiedrich, line 48: Please ignore the interface for now. It will slightly change with GSYE-426. Thanks!

mrsaemir (approved):
LGTM.

src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -0,0 +1,171 @@
...
+            min_temperatur_C=min_temperatur_C,
+            max_temperatur_C=max_temperatur_C,
+            external_temperatur_C=external_temperatur_C,
+            tank_volume_l=tank_volume_l,
+            average_consumption=average_consumption,
+            source_type=source_type
+        )
+
+        self.preferred_buying_rate = preferred_buying_rate
+        self._markets_buffer = []  # TODO: discuss if this buffer is really necessary
```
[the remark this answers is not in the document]
hannesdiedrich: I discussed with @spyrostz yesterday, the HeatPump will not be taking part in the future markets for now, because the accounting of the storage is complicated for future time slot. It also does not seem that the design includes trading in the future markets. So, we decided, that the HeatPump will only trade in the spot market, which will simplify this strategy implementation a lot. With this we do not need the `self._markets_buffer`.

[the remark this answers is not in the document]
hannesdiedrich: I already did this here: [6e248d9](https://github.com/gridsingularity/gsy-e/pull/1568/commits/6e248d9ce74f332c47d5f35108e15b1639e2ecbd)

src/gsy_e/models/strategy/trading_strategy_base.py
```diff
@@ -94,3 +94,8 @@ def event_market_cycle(self):
     def state(self) -> "StateInterface":
         """Get the state class of the strategy. Needs to be implemented by all strategies"""
         raise NotImplementedError
+
+    @property
+    def trades(self):
```
spyrostz, line 99: Not saying to remove this, because I believe that this will be very useful in the long run, but I did not spot this being used in the actual strategy, right ?

src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -0,0 +1,152 @@
...
+    def event_market_cycle(self) -> None:
+        super().event_market_cycle()
+        self._post_order_to_new_market(
+            self.area.spot_market, self.area.spot_market.time_slot)
+
+    def event_tick(self):
+        self._update_open_orders()
+
+    def event_bid_traded(self, *, _market_id: str, bid_trade: Trade) -> None:
+        """TODO: to be tested in the frame of GSYE-426"""
```
spyrostz, line 77: I see, so you want to create a single test that will test the decrement of the posted energy functionality at the same time and not mock it, right? I agree then, this can be added in the context of the energy parameters ticket.

src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -0,0 +1,152 @@
...
+                    for bid in market.slot_bid_mapping[market_slot]
+                    if bid.buyer == self.owner.name]
+
+        for bid in bids:
+            market.delete_bid(bid)
+
+    def _get_energy_buy_energy(self, buy_rate: float, market_slot: DateTime) -> float:
+        if buy_rate > self.preferred_buying_rate:
+            return self._energy_params.energy_loss_kWh.get(market_slot, 1)
+        return self._energy_params.energy_demand_kWh.get(market_slot, 1)
```
spyrostz, line 128: I think this method is important to be tested despite its small size, but again this can be deferred for the energy parameters tests.

spyrostz (dismissed):
Some comments, but LGTM ! Thanks

src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -0,0 +1,152 @@
...
+from gsy_e.constants import FLOATING_POINT_TOLERANCE
+from gsy_e.models.state import HeatPumpState
+from gsy_e.models.strategy.energy_parameters.heat_pump import HeatPumpEnergyParameters
+from gsy_e.models.strategy.trading_strategy_base import TradingStrategyBase
+from gsy_e.models.strategy.order_updater import OrderUpdaterParameters, OrderUpdater
+
+DEFAULT_HEAT_PUMP_ORDER_UPDATE_PARAMS = {
+    AvailableMarketTypes.SPOT: OrderUpdaterParameters(
+        initial_rate=ConstSettings.HeatPumpSettings.BUYING_RATE_RANGE.initial,
+        final_rate=ConstSettings.HeatPumpSettings.BUYING_RATE_RANGE.final,
```
hannesdiedrich, line 17: @spyrostz We have a requirement, that by default (if the user does not provide one):
- the final rate should be the MMR by default
- the initial rate should be the FIT by default 
But now I have a problem. The `InfiniteBusStrategy` could be initiated after this one and these global settings could not be updated in time for the init of the HP class. Also, if the MMR is variable, this is even more complicated.
I propose to allow None values in the `OrderUpdaterParameters` data class and deal with this in the `_create_order_updaters` class with something like:
```python
if not self._order_updater_params[market_type].final_rate:
   self._order_updater_params[market_type].final_rate = GlobalConfig.market_maker_rate[market_slot]
```
Or do you have a better idea?

src/gsy_e/models/strategy/heat_pump.py
hannesdiedrich: Agreed, I will add a test in the next PR.

src/gsy_e/models/strategy/heat_pump.py
hannesdiedrich: Exactly!

tests/strategies/test_heat_pump.py
```diff
@@ -0,0 +1,96 @@
...
+from gsy_e.models.strategy.heat_pump import HeatPumpStrategy
+
+if TYPE_CHECKING:
+    from gsy_e.models.strategy.trading_strategy_base import TradingStrategyBase
+
+CURRENT_MARKET_SLOT = datetime(2022, 6, 13, 0, 0)
+
+
+@pytest.fixture(name="heatpump_fixture")
+def heatpump_strategy_fixture() -> Tuple["TradingStrategyBase", "Area"]:
```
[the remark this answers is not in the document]
hannesdiedrich: Ahhh! Of course! Thanks for reminding me. Will change this.

src/gsy_e/models/strategy/trading_strategy_base.py
hannesdiedrich: I had to learn the hard way, that this is needed in the [endpoint_buffer.py](https://github.com/gridsingularity/gsy-e/blob/8a8b0a4e5e651db8eb2b1b55665b68dc2afaa145/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py#L322)

spyrostz: Thanks for pointing this out, and too bad to hear that! I am currently wondering whether it makes sense to add it as an abtractproperty on the AreaBehaviorBase class and fix the current BaseStrategy class to handle this, however since we plan to slowly migrate away from these classes please leave it as it is, no need to do this now.

src/gsy_e/models/strategy/heat_pump.py
spyrostz: I agree, however you already have default values for MMR and FIT, therefore I do not see the need to pass None values to the OrderUpdaterParameters class. This also covers the bug / cornercase where neither user parameters, nor MMR / FIT are passed to this strategy. This will make it flexible enough to work without any parameters at all IMO, no matter if we decide to force these parameters in the gsy-web validation or not.

hannesdiedrich: @spyrostz 
I am confused by the first part of your comment unfortunately.
> ... therefore I do not see the need to pass None values to the OrderUpdaterParameters class. ...

But I need to distinguish between the case when the user actually sets the initial and final rate and the case when they want to use the current FIT and MMR. Or am I missing something here?

spyrostz: Distinguishing between these 2 cases does not need to be part of the OrderUpdaterParameters class though, and IMO it should be part of the strategy (by e.g. having a boolean variable that dictates which of the 2 cases is active, the MMR/FIT or the user-defined rate). Reason that I believe that this would be preferable, is because the OrderUpdaterParameters is a dataclass that encapsulates only the parameters of the order updater, and ideally should not contain any strategy-specific logic, since this will affect other strategies too. For instance, if you add such logic to the OrderUpdaterParameters, then this logic would be imported to the forward strategies. Even if it is not used by the forward markets, it is probably better to not be part of the class in the first place.

hannesdiedrich: Ah yes, did not want to add it to the `OrderUpdaterParameters`, but to the Strategy. However, members of the `OrderUpdaterParameters` should be able to be set to None AFAIU, in order to set the `DEFAULT_HEAT_PUMP_ORDER_UPDATE_PARAMS` so that I can distinguish in the strategy if the user set them or not. Does it make sense?

spyrostz: I think this is the only minor point where we do not agree in our comments. IMO allowing the members of the OrderUpdaterParameters to be None would be implicit behavior, because you would then have to check one of the members whether it is None or not. The explicit behavior would be to have a separate boolean variable in the strategy class (e.g. _are_user_parameters_set) that can be used in a more clear way. I have 2 arguments in favor of the latter:
- it is more misleading / confusing for the unaware reader to see something like this `if self._energy_params.initial_rate is None: <use parameters from FIT/MMR>` compared to this `if not self._are_user_parameters_set: <use parameters from FIT/MMR>`. 
- in reality, in order for the OrderUpdaterParameters to be a usable object, they have to be populated, and in order to protect ourselves from the None case we would have to add if / else blocks in actions that depend on OrderUpdaterParameters. For now the only place that we have to do this is the _create_order_updaters, however if we add more methods that use the OrderUpdaterParameters this will quickly become convoluted IMO

hannesdiedrich: Thank you. The `_are_user_parameters_set` parameter makes sense. However, the user is limited to two choices:
1. Set all `OrderUpdaterParameters` that are static for the whole simulation 
2. Not set `OrderUpdaterParameters` and let the strategy use the default ones (which will be initial_rate=FIT; final_rate=MMR, update_interval=DEFAULT_UPDATE_INTERVAL)
But there is also the following case, when the user wants to set e.g. only the final_rate to a constant value.
Do you think we should support this case? We can also resume this discussion offline (and post an outcome here) as the number of comments has already exceeded our workflow limit.

src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -113,37 +117,38 @@ def post_order(
...
+                    if bid.buyer.name == self.owner.name]
 
         for bid in bids:
             market.delete_bid(bid)
 
     def _get_energy_buy_energy(self, buy_rate: float, market_slot: DateTime) -> float:
         if buy_rate > self.preferred_buying_rate:
-            return self._energy_params.energy_loss_kWh.get(market_slot, 0)
-        return self._energy_params.energy_demand_kWh.get(market_slot, 0)
+            return self._energy_params.energy_loss_kWh.get(market_slot, 1)
```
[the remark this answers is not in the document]
hannesdiedrich: @mrsaemir I think this comment is not valid any more right?
The energy parameters of the `HeatPump EnergyParamters` deal with populating the `min_energy_demand_kWh` and `max_energy_demand_kWh`. They can only be 0 if the heatpump does not consume energy. But this case is caught in the `post_order` method. 
Having both values in one profile does not work, because the price changes over the frame of the time_slot. As soon as the `preferred_buying_rate` is reached, the other energy value need to be accessed and not calculated on the fly. 
I hope this makes sense.

src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -80,7 +98,7 @@ def event_bid_traded(self, *, _market_id: str, bid_trade: Trade) -> None:
             return
 
         time_slot = bid_trade.time_slot
-        if bid_trade.buyer_id != self.owner.uuid:
+        if bid_trade.buyer.origin_uuid != self.owner.uuid:
```
spyrostz, line 101: I do not necessarily mind for this one, since for the case of the Trade in the "Home" market the origin_uuid is the same as the "uuid", however I was wondering whether you stumbled upon this issue when running the simulation.

spyrostz (approved):
LGTM

PR #1569, opened 2022-11-29
Feature/gsye 441 remove accept offer bid params

spyrostz:

Followup to this https://github.com/gridsingularity/gsy-e/pull/1567 , this PR is removing the already_tracked and trade_rate parameters from the market class method arguments, since they are not necessary for the operation of the exchange and it will simplify the class interface. 

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-441-remove-accept-offer-bid-params


src/gsy_e/models/market/one_sided.py
```diff
@@ -315,10 +318,9 @@ def accept_offer(  # pylint: disable=too-many-locals
 
         self.bc_interface.track_trade_event(self.time_slot, trade)
 
-        if already_tracked is False:
-            self._update_stats_after_trade(trade, offer)
-            log.info("%s[TRADE][OFFER] [%s] [%s] %s",
-                     self._debug_log_market_type_identifier, self.name, trade.time_slot, trade)
+        self._update_stats_after_trade(trade, offer)
+        log.info("%s[TRADE][OFFER] [%s] [%s] %s",
+                 self._debug_log_market_type_identifier, self.name, trade.time_slot, trade)
```
hannesdiedrich, line 323: How is the logging handled now? Will we see two lines
```
[TRADE][OFFER] ...
[TRADE][BID] ...
```
for the same trade now?

src/gsy_e/models/market/two_sided.py
```diff
@@ -255,15 +255,18 @@ def accept_bid(self, bid: Bid,
...
+                      residual=residual_bid,
                       offer_bid_trade_info=updated_bid_trade_info,
                       fee_price=fee_price, time_slot=bid.time_slot,
                       matching_requirements=trade_offer_info.matching_requirements
                       )
 
-        if already_tracked is False:
-            self._update_stats_after_trade(trade, bid, already_tracked)
+        if not offer:
+            # This is a chain trade, therefore needs to be tracked. For the trade on the market
```
hannesdiedrich, line 266: What do you think about setting `offer=None` in the `accept_bid` call in the engine in order to make it explicit?

hannesdiedrich (commented):
A few questions, LGTM in general.
I also like your proposal to rename `TradeBidOfferInfo` into `TradePricingDetails` (even in this PR)

src/gsy_e/models/market/two_sided.py
spyrostz: Thanks, fixing

src/gsy_e/models/market/one_sided.py
spyrostz: Exactly!

spyrostz:
> A few questions, LGTM in general. I also like your proposal to rename `TradeBidOfferInfo` into `TradePricingDetails` (even in this PR)

Thanks for the comments and agreeing to the suggestion! I will open a separate PR though in order to not overcomplicate more this chain of PRs.

hannesdiedrich (approved):
LGTM

PR #1571, opened 2022-11-30
GSYE-459: Move FutureMarketCounter from singletons to the MycoInterna…

hannesdiedrich:
…lMatcher

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/market_counters.py
+++ b/src/gsy_e/gsy_e_core/market_counters.py
@@ -48,6 +48,12 @@ class FutureMarketCounter(MarketCounter):
             clearing_interval_min=ConstSettings.FutureMarketSettings.
             FUTURE_MARKET_CLEARING_INTERVAL_MINUTES)
 
+    def is_time_for_clearing(self, current_time: DateTime) -> bool:
+        """Only return if future markets are enabled."""
+        if ConstSettings.FutureMarketSettings.FUTURE_MARKET_DURATION_HOURS > 0:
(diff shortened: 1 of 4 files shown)


src/gsy_e/gsy_e_core/market_counters.py
```diff
@@ -48,6 +48,12 @@ def __init__(self):
             clearing_interval_min=ConstSettings.FutureMarketSettings.
             FUTURE_MARKET_CLEARING_INTERVAL_MINUTES)
 
+    def is_time_for_clearing(self, current_time: DateTime) -> bool:
+        """Only return if future markets are enabled."""
+        if ConstSettings.FutureMarketSettings.FUTURE_MARKET_DURATION_HOURS > 0:
```
mrsaemir, line 53: It seems that the unit tests don't cover a scenario where `FUTURE_MARKET_DURATION_HOURS <= 0`. If you think it's not that important feel free to skip it, otherwise it would be nice to cover this scenario as well. Thanks

mrsaemir (approved):
LGTM + A suggestion.

src/gsy_e/gsy_e_core/market_counters.py
hannesdiedrich: Makes sense, I added it here: [22b903e](https://github.com/gridsingularity/gsy-e/pull/1571/commits/22b903ed3f9e0110f00264f54fc2db707e7ba118)

spyrostz (approved):
LGTM

PR #1572, opened 2022-12-01
GSYE-444: Remove last_energy_trades_high_resolution

mrsaemir:
As requested by [GSYE-444](https://gridsingularity.atlassian.net/browse/GSYE-444), we want to want to remove all traces of high-resolution graphs for the offer/bid/trade and the energy trade profile

## Proposed changes

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -69,7 +69,6 @@ class SimulationEndpointBuffer:
         }
 
         self.bids_offers_trades = {}
-        self.last_energy_trades_high_resolution = {}
         self.results_handler = self._create_endpoint_buffer(should_export_plots)
         self.simulation_state = {"general": {}, "areas": {}}
 


src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -69,7 +69,6 @@ def __init__(self, job_id, random_seed, area, should_export_plots):
         }
 
         self.bids_offers_trades = {}
-        self.last_energy_trades_high_resolution = {}
         self.results_handler = self._create_endpoint_buffer(should_export_plots)
         self.simulation_state = {"general": {}, "areas": {}}
```
Maheen12, line 74: The purpose of high resolution energy trades was to create plots of energy trades(which were of high resolution)?

mrsaemir: I'm not sure exactly. Maybe @spyrostz can help us here?

spyrostz: Exactly. The default resolution of our graphs is tied to the market slot, so with default parameters 15 minutes. These graphs intended to display graphs with lowef resolution than 15-mins, typically second resolution (IIRC in order to indicate the time that the offer / bid was posted, and that the trade was created during a market slot). However these graphs will not be needed now, especially in the gsy-e (we created these in the past in order to use them for a research project, however I doubt that we will need them in the future)

PR #1581, opened 2022-12-09
Feature/gsye 426 Add EnergyParamters for the HeatPump

hannesdiedrich:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-426

### DIFF
--- a/src/gsy_e/models/state.py
+++ b/src/gsy_e/models/state.py
@@ -913,8 +913,10 @@ class HeatPumpState(StateInterface):
     def update_storage_temp(self, current_time_slot: DateTime):
         """Update storage temperature of the given slot with the accumulated changes. """
         new_temp = (self.get_storage_temp_C(self._last_time_slot(current_time_slot))
-                    - self.get_temp_decrease_K(current_time_slot)
-                    + self.get_temp_increase_K(current_time_slot))
+                    - self.get_temp_decrease_K(self._last_time_slot(current_time_slot))
+                    + self.get_temp_increase_K(self._last_time_slot(current_time_slot)))
+        if new_temp < -FLOATING_POINT_TOLERANCE:
+            raise UnexpectedStateException("Storage of heat pump should not drop below zero.")
         self._storage_temp_C[current_time_slot] = new_temp
 
     def set_min_energy_demand_kWh(self, time_slot: DateTime, energy_kWh: float):
--- a/src/gsy_e/models/state.py
+++ b/src/gsy_e/models/state.py
@@ -985,9 +985,9 @@ class HeatPumpState(StateInterface):
 
     def get_results_dict(self, current_time_slot: DateTime) -> Dict:
         return {
-            "storage_temp_CC": self.get_storage_temp_C(current_time_slot),
-            "temp_decrease_KK": self.get_temp_decrease_K(current_time_slot),
-            "temp_increase_KK": self.get_temp_increase_K(current_time_slot)
+            "storage_temp_C": self.get_storage_temp_C(current_time_slot),
+            "temp_decrease_K": self.get_temp_decrease_K(current_time_slot),
+            "temp_increase_K": self.get_temp_increase_K(current_time_slot)
         }
 
     def _last_time_slot(self, current_market_slot: DateTime) -> DateTime:
--- a/src/gsy_e/models/state.py
+++ b/src/gsy_e/models/state.py
@@ -28,7 +28,7 @@ from gsy_framework.utils import (
     limit_float_precision)
 from pendulum import DateTime, duration
 
-from gsy_e.constants import FLOATING_POINT_TOLERANCE
+from gsy_e.constants import FLOATING_POINT_TOLERANCE, RETAIN_PAST_MARKET_STRATEGIES_STATE
 from gsy_e.gsy_e_core.util import is_time_slot_in_past_markets, write_default_to_dict
 
 StorageSettings = ConstSettings.StorageSettings
@@ -967,7 +967,7 @@ class HeatPumpState(StateInterface):
             state_dict["max_energy_demand_kWh"])
 
     def delete_past_state_values(self, current_time_slot: DateTime):
-        if not current_time_slot:
+        if not current_time_slot or RETAIN_PAST_MARKET_STRATEGIES_STATE:
             return
         self._delete_time_slots(self._min_energy_demand_kWh,
                                 self._last_time_slot(current_time_slot))
--- a/src/gsy_e/models/state.py
+++ b/src/gsy_e/models/state.py
@@ -895,63 +895,100 @@ class StorageState(StateInterface):
 class HeatPumpState(StateInterface):
     """State for the heat pump strategy."""
 
-    def __init__(self, initial_temp_C: float):
+    def __init__(self, initial_temp_C: float, slot_length: duration):
         # the defaultdict was only selected for the initial slot
         self._storage_temp_C: Dict[DateTime, float] = defaultdict(lambda: initial_temp_C)
         self._min_energy_demand_kWh: Dict[DateTime, float] = {}
         self._max_energy_demand_kWh: Dict[DateTime, float] = {}
-        # buffer for decrease of storage temp due to heat consumption
-        self._temp_decrease_K: Dict[DateTime, float] = {}
-
-    def copy_to_next_slot(self, current_market_slot: DateTime, slot_length: duration):
-        next_slot = current_market_slot.add(minutes=slot_length.total_minutes())
-        self._storage_temp_C[next_slot] = self._storage_temp_C[current_market_slot]
+        # buffers for increase and  decrease of storage
+        self._temp_decrease_K: Dict[DateTime, float] = defaultdict(lambda: 0)
+        self._temp_increase_K: Dict[DateTime, float] = defaultdict(lambda: 0)
+        self._slot_length = slot_length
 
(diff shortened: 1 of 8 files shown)


src/gsy_e/models/strategy/energy_parameters/heat_pump.py
```diff
@@ -1,48 +1,171 @@
...
+
+        self._calc_energy_demand(time_slot)
+
+    def _calc_energy_demand(self, time_slot: DateTime):
+        self.state.set_min_energy_demand_kWh(
+            time_slot, self._calc_energy_to_buy_minimum(time_slot))
+        self.state.set_max_energy_demand_kWh(
+            time_slot, self._calc_energy_to_buy_maximum(time_slot))
+
+    def _calc_Q_consumption(self, time_slot: DateTime, consumption_kWh: float) -> float:
```
mrsaemir, line 149: From Google: `Q is the net heat transferred into/out of the system. And the unit for heat is Joules.`
Since you're using this function for both `_calc_temp_decrease_K` and `_calc_temp_increase_K`, I would suggest to rename it to something else to better fit the usage (e.g. `energy_kWh_to_heat_joules`). If you agree this, it would be nice to also rename `consumption_kWh` to `energy_kWh`.

src/gsy_e/models/state.py
```diff
@@ -894,19 +896,109 @@ def get_results_dict(self, current_time_slot: DateTime) -> dict:
 class HeatPumpState(StateInterface):
     """State for the heat pump strategy."""
 
+    def __init__(self, initial_temp_C: float, slot_length: duration):
+        # the defaultdict was only selected for the initial slot
+        self._storage_temp_C: Dict[DateTime, float] = defaultdict(lambda: initial_temp_C)
+        self._min_energy_demand_kWh: Dict[DateTime, float] = {}
+        self._max_energy_demand_kWh: Dict[DateTime, float] = {}
+        # buffers for increase and  decrease of storage
+        self._temp_decrease_K: Dict[DateTime, float] = defaultdict(lambda: 0)
```
mrsaemir, line 905: Could you please describe why don't we operate on the `_storage_temp_C` directly and separate temp decrease/increase here? IMO, It would be much simpler to handle just one dictionary in `HeatPumpEnergyParameters` instead of three of them. This could also lead to removal of many setter/getter methods in `HeatPumpState` and simpler serializing/deserializing of the state.

src/gsy_e/models/state.py
```diff
@@ -894,19 +896,109 @@ def get_results_dict(self, current_time_slot: DateTime) -> dict:
...
+        self._delete_time_slots(self._storage_temp_C,
+                                self._last_time_slot(current_time_slot))
+        self._delete_time_slots(self._temp_increase_K,
+                                self._last_time_slot(current_time_slot))
+        self._delete_time_slots(self._temp_decrease_K,
+                                self._last_time_slot(current_time_slot))
+
+    def get_results_dict(self, current_time_slot: DateTime) -> Dict:
+        return {
+            "storage_temp_CC": self.get_storage_temp_C(current_time_slot),
```
mrsaemir, line 986: Could you please describe what `CC` and `KK` are? Thanks.

mrsaemir (commented):
LGTM in general. There are some suggestions to make it even better. Thanks a lot.

src/gsy_e/models/strategy/energy_parameters/heat_pump.py
```diff
@@ -1,48 +1,171 @@
...
+        return self._get_cop(time_slot) * consumption_kWh
+
+    def _get_cop(self, time_slot: DateTime) -> float:
```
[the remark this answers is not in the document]
hannesdiedrich: I explicitly divided the getting of correct temperatures from the calculation for better readability actually. 
But this seems to be subjective apparently. 
I agree on adding some documentation in the docstring of the `_calc_cop` though. Will do.

src/gsy_e/models/strategy/energy_parameters/heat_pump.py
hannesdiedrich: `1 J` can also be rewritten as` 1 Ws` ([wikipedia](https://de.wikipedia.org/wiki/Joule)) which in turn can be converted into `kWh` which is widely used in our codebase. With `Q` I was following the specs on the [requirement document](https://docs.google.com/document/d/1r_oZh6dFpz5iHdO3MI4w4JpDr_WKc0L3PnUprHLiQw4/edit#).
I am open to rename the `_Q_` into `_energy_` or '_heat_' in the whole module, but I vote for keeping `kWh` as unit. What do you think?

src/gsy_e/models/state.py
hannesdiedrich: Ah, this is remnant of testing, let me fix this.

hannesdiedrich: [574c0ce](https://github.com/gridsingularity/gsy-e/pull/1581/commits/574c0ce6174241e6c983a8740d9162618124d2e7)

src/gsy_e/models/strategy/energy_parameters/heat_pump.py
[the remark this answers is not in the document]
hannesdiedrich: Added documentation: [fb126aa](https://github.com/gridsingularity/gsy-e/pull/1581/commits/fb126aa09f0accea881bfbfc5c403ecf27aa3aa3)

src/gsy_e/models/strategy/energy_parameters/heat_pump.py
mrsaemir: I am also in your side to stick with kWh and not introduce any new concepts (heat/joules) in the code. What I say is that the HeatPump is itself very convoluted and our code should not add to its complexity.
Since the `_calc_Q_consumption` is used for both heat consumed and produced, how about we remove the trailing `_consumption` from its name? Something like:
`_calc_Q_from_energy_kWh` or `_calc_Q_from_kWh`?

src/gsy_e/models/state.py
mrsaemir: Your first reason convinced me. Thanks a lot

src/gsy_e/models/strategy/energy_parameters/heat_pump.py
hannesdiedrich: Good idea! IMO `_calc_Q_from_energy_kWh` is most fitting. let me rename it.

hannesdiedrich: How about this: [a6c376d](https://github.com/gridsingularity/gsy-e/pull/1581/commits/a6c376dc1914bf35371c1e5a12a9d94fda9711a4)

mrsaemir: I like it very much now! Thank you.

src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -41,20 +41,26 @@ class HeatPumpStrategy(TradingStrategyBase):
...
+                 consumption_profile_uuid: Optional[str] = None,
                  source_type: int = ConstSettings.HeatPumpSettings.SOURCE_TYPE,
                  order_updater_parameters: Dict[
                      AvailableMarketTypes, HeatPumpOrderUpdaterParameters] = None,
                  preferred_buying_rate: float =
                  ConstSettings.HeatPumpSettings.PREFERRED_BUYING_RATE
                  ):
 
+        if ConstSettings.MASettings.MARKET_TYPE == 1:
+            raise NotImplementedError("Heatpump has not been implemented for the OneSidedMarket")
```
spyrostz, line 62: I would even add an assertion here, to denote that this is really something that should not happen under any circumstances.

spyrostz (dismissed):
Two small comments, but in general LGTM even now. Thanks for bearing with me and dealing with this complex PR!

src/gsy_e/models/strategy/energy_parameters/heat_pump.py
```diff
@@ -1,48 +1,176 @@
...
-        self.max_temperatur_C = max_temperatur_C
-        self.external_temperatur_C = external_temperatur_C
-        self.tank_volume_l = tank_volume_l
-        self.average_consumption = average_consumption
-        self.source_type = source_type
+        self._min_temp_C = min_temp_C
+        self._max_temp_C = max_temp_C
+        self._source_type = source_type
+        self._tank_volume_l = tank_volume_l
+        self._Q_specific = SPECIFIC_HEAT_CONST_WATER * tank_volume_l * WATER_DENSITY  # [kWh / K]
```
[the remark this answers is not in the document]
hannesdiedrich: [2dbeb3b](https://github.com/gridsingularity/gsy-e/pull/1581/commits/2dbeb3b320961f7fed714268022a7f54779a6917)

src/gsy_e/models/strategy/heat_pump.py
hannesdiedrich: [2f5a617](https://github.com/gridsingularity/gsy-e/pull/1581/commits/2f5a6176aa55f7138056dc02d8793c2d17bd3b17)

[ce93cae](https://github.com/gridsingularity/gsy-e/pull/1581/commits/ce93caed8217088b8346a705f0a79634a47ed8c4)

mrsaemir (approved):
LGTM, thanks.

spyrostz (approved):
LGTM

PR #1584, opened 2022-12-19
Feature/gsye 482

spyrostz:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-482

### DIFF
--- a/src/gsy_e/models/strategy/market_agents/two_sided_agent.py
+++ b/src/gsy_e/models/strategy/market_agents/two_sided_agent.py
@@ -58,7 +58,7 @@ class TwoSidedAgent(OneSidedAgent):
     # pylint: disable=unused-argument
     def event_bid(self, *, market_id: str, bid: "Bid"):
         for engine in sorted(self.engines, key=lambda _: random()):
-            engine.event_bid()
+            engine.event_bid(bid)
 
     # pylint: disable=unused-argument
     def event_bid_traded(self, *, market_id: str, bid_trade: "Trade"):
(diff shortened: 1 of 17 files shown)


src/gsy_e/models/strategy/market_agents/two_sided_agent.py
```diff
@@ -55,6 +55,11 @@ def usable_bid(self, bid: "Bid") -> bool:
         """Prevent MAEngines from trading their counterpart's bids."""
         return all(bid.id not in engine.forwarded_bids.keys() for engine in self.engines)
 
+    # pylint: disable=unused-argument
+    def event_bid(self, *, market_id: str, bid: "Bid"):
```
hannesdiedrich, line 59: Why do we need this interface in the first place? The market_id is not used here and also not in the engine. Same applies to the event_offer in the one-sided market.

src/gsy_e/setup/kpi/d3asim_3464_family_with_pv_ess.py
```diff
@@ -29,8 +29,8 @@
 def get_setup(config):
     ConstSettings.FutureMarketSettings.FUTURE_MARKET_DURATION_HOURS = 0
     ConstSettings.MASettings.MARKET_TYPE = 2
-    ConstSettings.MASettings.MIN_OFFER_AGE = 0
-    ConstSettings.MASettings.MIN_BID_AGE = 0
+    ConstSettings.MASettings.MIN_OFFER_AGE = 1
```
hannesdiedrich, line 32: :partying_face:

hannesdiedrich (approved):
LGTM

src/gsy_e/models/strategy/market_agents/two_sided_agent.py
spyrostz: Indeed, the reason why I added this was in case these methods are used by the strategies in the future. In this case, there needs to be a way to distinguish which market emitted the offer / bid event, in order to treat differently future markets and spot markets.

PR #1587, opened 2022-12-21
Share tox env

spyrostz:

Main change involves sharing the same environment across all tox targets, thus allowing to separate the setup process to a separate target and creating dedicated targets for running the unit and integration tests that do not invoke the setup process (therefore do not try to reinstall the packages). 

The approach worked well in the terminal, however does not work well with the Pycharm tox runners. In order to configure it in Pycharm a configuration for unittests and setup needs to be created (in order to not have to invoke these via terminal but directly in your IDE):  
Screenshot 2022-12-21 at 09 49 49

It generally works, what does not work though is the ability to run individual tests through tox, because the Pycharm tox runner does not support this. 

## Proposed changes

-

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -45,12 +45,6 @@ envdir = {toxworkdir}/setup
 skip_install = true
 commands = coverage erase
 
-[testenv:check_readme]
-envdir = {toxworkdir}/setup
-skip_install = true
-deps = readme_renderer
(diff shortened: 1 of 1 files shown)


hannesdiedrich (approved):
Thanks for the explanations regarding pycharm. LGTM

PR #1591, opened 2023-01-09
Tests 112

hannesdiedrich:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=TESTS-112


mrsaemir (approved):
LGTM

src/gsy_e/setup/strategy_tests/future_market_strategy.py
```diff
@@ -0,0 +1,97 @@
...
+
+    return area
+
+
+def get_var_from_env(var_name: str, default_val: float):
+    """Read variable from env and parse as string"""
+    var = os.environ.get(var_name)
+    if var is not None:
+        return float(var)
+    return default_val
```
spyrostz, line 97: I assume that you did not use this instead:
```
def get_var_from_env(var_name: str, default_val: float):
    """Read variable from env and parse as string"""
    var = float(os.environ.get(var_name, default_val) or default_val)
```
in order to cover the case that the variable has a value of 0 (in which case you do not want to fall back to the default value)?

spyrostz (approved):
One clarification question, LGTM already

src/gsy_e/setup/strategy_tests/future_market_strategy.py
hannesdiedrich: I did not see this part until now. The reasoning you describe make sense.

PR #1595, opened 2023-01-13
Disable the triggering of the garbage collector and the debug calcula…

spyrostz:
…tion of the memory information on SCM, since it contributed a significant amount of the SCM runtime.

The garbage collector invocation and the process memory calculation had detrimental effect on the performance of the SCM simulation. Around 40% of the simulation runtime was spent on the commented-out method which significantly slowed down the SCM simulation especially for longer simulation durations. 

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/simulation/simulation.py
+++ b/src/gsy_e/gsy_e_core/simulation/simulation.py
@@ -485,7 +485,7 @@ class CoefficientSimulation(Simulation):
 
             self._external_events.update(self.area)
 
-            self._compute_memory_info()
+            # self._compute_memory_info()
 
             self._time.handle_slowdown_and_realtime_scm(slot_no, slot_count, self.config)
 


src/gsy_e/gsy_e_core/simulation/simulation.py
```diff
@@ -485,7 +485,7 @@ def _execute_simulation(
 
             self._external_events.update(self.area)
 
-            self._compute_memory_info()
+            # self._compute_memory_info()
```
mrsaemir, line 488: Do we have any plans to bring this feature back in the future?

mrsaemir (approved):
LGTM

src/gsy_e/gsy_e_core/simulation/simulation.py
spyrostz: Only while debugging memory leaks or other issues. That's why I would recommend to violate our usual rule and leave it commented out here, for convenience.

PR #1597, opened 2023-01-17
FE-280: Included the self-production for community for each home to t…

spyrostz:
…he community self-consumed energy, in order to take into account intracommunity trading.

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/area/scm_manager.py
+++ b/src/gsy_e/models/area/scm_manager.py
@@ -376,6 +376,8 @@ class SCMManager:
             self.community_data.energy_bought_from_community_kWh += (
                 home_data.energy_bought_from_community_kWh)
             self.community_data.energy_sold_to_grid_kWh += home_data.energy_sold_to_grid_kWh
+            self.community_data.self_consumed_energy_kWh += (
+                home_data.self_production_for_community_kWh)
 
     def calculate_home_energy_bills(
             self, home_uuid: str) -> None:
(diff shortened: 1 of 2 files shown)


src/gsy_e/models/area/scm_manager.py
```diff
@@ -376,6 +376,8 @@ def calculate_community_after_meter_data(self):
             self.community_data.energy_bought_from_community_kWh += (
                 home_data.energy_bought_from_community_kWh)
             self.community_data.energy_sold_to_grid_kWh += home_data.energy_sold_to_grid_kWh
+            self.community_data.self_consumed_energy_kWh += (
+                home_data.self_production_for_community_kWh)
```
hannesdiedrich, line 380: I need to write it down and think out load in order to make sure I understood correctly. 
The self_production of the community should also be added to the self_consumption, because the PV energy was used inside of the community and not sold to the grid. Before, only the self_consumption of each House was accumulated which did not include the trades between houses.
Did I get it?

spyrostz: Exactly. The self consumption of the community should also include the intracommunity trades (essentially the trades from the PVs of one house in the community to another house in the community). This is reflected in the SCM results by the self_production_for_community in each house, therefore it also needs to be added to the community self_consumed_energy.

hannesdiedrich (approved):
LGTM

PR #1598, opened 2023-01-24
GSYE-452: Integrate validation with AVRO schemas

mrsaemir:
## Proposed changes

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-452
**GSY_FRAMEWORK_BRANCH**=bug/GSYE-452

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -22,7 +22,6 @@ from typing import TYPE_CHECKING, Dict, Iterable, List
 from gsy_framework.constants_limits import (DATE_TIME_FORMAT, DATE_TIME_UI_FORMAT, ConstSettings,
                                             GlobalConfig)
 from gsy_framework.enums import AvailableMarketTypes
-from gsy_framework.results_validator import results_validator
 from gsy_framework.schema.validators import get_schema_validator
 from gsy_framework.sim_results.all_results import ResultsHandler
 from gsy_framework.utils import get_json_dict_memory_allocation_size
@@ -83,7 +82,6 @@ class SimulationEndpointBuffer:
     def prepare_results_for_publish(self) -> Dict:
         """Validate, serialise and check size of the results before sending to gsy-web."""
         result_report = self._generate_result_report()
-        self.validate_results(result_report)
 
         message_size = get_json_dict_memory_allocation_size(result_report)
         if message_size > 64000:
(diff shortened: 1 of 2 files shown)


src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -45,6 +45,9 @@
     "max": None
 }
 
+scm_simulation_raw_data_validator = get_schema_validator("scm_simulation_raw_data_v1")
+simulation_raw_data_validator = get_schema_validator("simulation_raw_data_v1")
```
hannesdiedrich, line 49: If [this discussion](https://github.com/gridsingularity/gsy-framework/pull/421#discussion_r1086319896) reveals that we keep this interface: How about moving these into the constructor of the respective class? This way the SCM validator is not initialised when we have a normal simulation.

hannesdiedrich (commented):
One comment, LGTM in general

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
mrsaemir: Regardless of the result of our other discussion, it's possible that I move this into the constructor of the EndpointBuffer class if you agree.

mrsaemir: Done.

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -45,8 +45,22 @@
     "max": None
 }
 
-scm_simulation_raw_data_validator = get_schema_validator("scm_simulation_raw_data_v1")
-simulation_raw_data_validator = get_schema_validator("simulation_raw_data_v1")
+
+class SimulationResultValidator:
```
hannesdiedrich, line 49: I do not understand the need for a separate class actually. Why not adding 
```python
self.results_validator = get_schema_validator("scm_simulation_raw_data")
```
into `CoefficientEndpointBuffer.__init__()`
and 
```python
self.results_validator = get_schema_validator("simulation_raw_data")
```
into `SimulationEndpointBuffer.__init__()`
?

mrsaemir: I did this for a couple of reasons:
1- There are more schema validators to be added in the future (for state variables, bids_offers_trades and configuration_tree). I felt that putting them all in the endpoint buffer will make the constructor of these classes even bigger.
2- I wanted to hide validation details from the endpoint buffers. This class does some sort of abstraction here.
3- (Personal pereference) I felt better to put all the real validation logic in one place to be able to manage them better in the future.

hannesdiedrich: Ok, after all this is only minor. Thanks for your explanation!

hannesdiedrich (approved):
LGTM

PR #1600, opened 2023-01-27
Bug/gsye 513 Fix callback channel for RedisSimulationCommunication._calculate_area_map_callback

hannesdiedrich:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=bug/GSYE-513

### DIFF
--- a/src/gsy_e/gsy_e_core/redis_connections/simulation.py
+++ b/src/gsy_e/gsy_e_core/redis_connections/simulation.py
@@ -30,7 +30,6 @@ from redis.exceptions import ConnectionError as RedisConnectionError
 from rq import get_current_job
 from rq.exceptions import NoSuchJobError
 
-import gsy_e.constants
 from gsy_e.gsy_e_core.exceptions import LiveEventException
 
 log = getLogger(__name__)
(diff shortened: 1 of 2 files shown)


src/gsy_e/gsy_e_core/redis_connections/simulation.py
```diff
@@ -56,7 +55,7 @@ def __init__(self, simulation_status, simulation_id, live_events, progress_info,
         self._area = area
         self._progress_info = progress_info
         self._sub_callback_dict = {
-            f"{gsy_e.constants.CONFIGURATION_ID}/area-map/": self._calculate_area_map_callback,
+            f"{self._simulation_id}/area-map/": self._calculate_area_map_callback,
```
spyrostz, line 58: Are you certain about this ? I was under the impression that this was used only by the gsy-e-sdk, and I recalled that we only used the configuration uuid and not the simulation id for the communication with the gsy-e-sdk.

spyrostz (commented):
One comment, will LGTM once discussed. On another note, I think we finally need to open a ticket that will finally remove the simulation id and use the configuration id for these occurrences.

src/gsy_e/gsy_e_core/redis_connections/simulation.py
hannesdiedrich: Actually, this channel is only used by the myco-sdk (see reference [here](https://github.com/gridsingularity/gsy-myco-sdk/blob/dbd4b1b5085ee0491dc152581a9e4f546172418a/gsy_myco_sdk/matchers/redis_base_matcher.py#L83)). There, the simulation_id is used. So I thought that this is a bug or a remnant here. 
(AFAIR, for the tests the simulation_id is None on the myco side, because we do not set them.)

spyrostz: All clear, thanks for the response!

spyrostz (approved):
LGTM

PR #1603, opened 2023-02-17
GSYE-526: Make sure that the first week of the DB profile is used whe…

hannesdiedrich:
…n the simulation time doe snot overlap with the profile time

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=bug/GSYE-526

### DIFF
--- a/src/gsy_e/gsy_e_core/user_profile_handler.py
+++ b/src/gsy_e/gsy_e_core/user_profile_handler.py
@@ -123,28 +123,21 @@ class ProfileDBConnectionHandler:
                 f"Profile in DB is empty for profile with uuid {profile_uuid}")
         first_datapoint_time = first_datapoint[0].time
 
-        datapoints = select(
+        datapoints = list(select(
             datapoint for datapoint in self.Profile_Database_ProfileTimeSeries
             if datapoint.profile_uuid == profile_uuid
             and datapoint.time >= first_datapoint_time
             and datapoint.time <= first_datapoint_time + duration(days=7)
+        ))
+        diff_current_to_db_time = (
+            current_timestamp -
+            self._strip_timezone_and_create_pendulum_instance_from_datetime(datapoints[0].time)
         )
-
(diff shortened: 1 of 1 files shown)


src/gsy_e/gsy_e_core/user_profile_handler.py
```diff
@@ -130,13 +130,17 @@ def get_first_week_from_profile(self, profile_uuid, current_timestamp) -> dict:
...
-                month=current_timestamp.month,
-                day=current_timestamp.day
-            ): datapoint.value for datapoint in datapoints
-        }
+        datapoint_dict = {}
+        diff_current_to_db_time = None
+        for datapoint in datapoints:
+            db_time = self._strip_timezone_and_create_pendulum_instance_from_datetime(
+                datapoint.time)
+            if diff_current_to_db_time is None:
```
mrsaemir, line 138: I'd suggest to move this before the `for` loop as it's calculated only once with the first element.

hannesdiedrich: I Ieft a verbose comment one line below in order to explain why it is here. Let me know if it is not clear.

mrsaemir: Yes. I do get it. However, you may be a able to evaluate the query before the for loop.
```python
datapoints = list(datapoints)
```
I guess this way you are not refetching the whole set from the DB again.

hannesdiedrich: Good idea, will change this!

hannesdiedrich: Done [890adb1](https://github.com/gridsingularity/gsy-e/pull/1603/commits/890adb1115f6135eba8843a91a4667ecb38c187e)

mrsaemir (approved):
LGTM. Thanks for dealing with this.

PR #1604, opened 2023-02-17
GSYE-428: Populate order_updater_params for HeatPump strategy

mrsaemir:
## Proposed changes

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-428
**GSY_FRAMEWORK_BRANCH**=feature/GSYE-428

### DIFF
--- a/src/gsy_e/models/strategy/heat_pump.py
+++ b/src/gsy_e/models/strategy/heat_pump.py
@@ -84,6 +84,19 @@ class HeatPumpStrategy(TradingStrategyBase):
 
         self.preferred_buying_rate = preferred_buying_rate
 
+    @staticmethod
+    def deserialize_args(constructor_args: Dict) -> Dict:
+        """Deserialize the constructor arguments for the HeatPump strategy."""
+        if "order_updater_params" not in constructor_args:
+            constructor_args["order_updater_params"] = {
+                AvailableMarketTypes.SPOT:
+                    HeatPumpOrderUpdaterParameters(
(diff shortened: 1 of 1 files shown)


hannesdiedrich (commented):
This is all? I do not see a Heapump class or any mapping that casts the "HeatPump" string from the are representation into a class method. Is it not needed?
Did you try starting a configuration including a heatpump from the gsy-web?

mrsaemir:
> 

What you did before is enough. I tried testing it e2e and it works just fine. I even watched for the storage temp and I saw it rising in value. Do we expect more? I also saw some trades happening for the heatpump.

src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -84,6 +84,19 @@ def __init__(self,
 
         self.preferred_buying_rate = preferred_buying_rate
 
+    @staticmethod
+    def deserialize_args(constructor_args: Dict) -> Dict:
+        """Deserialize the constructor arguments for the HeatPump strategy."""
```
spyrostz, line 89: I assume that the other arguments do not need to be deserialized because they either do not need to modify their type during deserialization, or they have exactly the same name in gsy-e and gsy-web right? In this case looks good to me

spyrostz (approved):
LGTM

src/gsy_e/models/strategy/heat_pump.py
mrsaemir: Yes. Everything else is in good shape.

PR #1605, opened 2023-02-21
GSYE-502: Added validation for the configuration tree in the simulati…

spyrostz:
…on results.

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -66,7 +66,7 @@ class SimulationResultValidator:
 
     def validate_configuration_tree(self, data: Dict):
         """Validate configuration_tree dict."""
-        self.simulation_configuration_tree_validator.validate(data=data, raise_exception=False)
+        self.simulation_configuration_tree_validator.validate(data=data, raise_exception=True)
 
 
 # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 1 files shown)


src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -57,11 +57,17 @@ def __init__(self, is_scm: bool):
...
+        self.simulation_configuration_tree_validator = get_schema_validator(
+            "results_configuration_tree")
 
     def validate_simulation_raw_data(self, data: Dict):
         """Validate flattened_area_core_stats_dict."""
         self.simulation_raw_data_validator.validate(data=data, raise_exception=True)
 
+    def validate_configuration_tree(self, data: Dict):
+        """Validate configuration_tree dict."""
+        self.simulation_configuration_tree_validator.validate(data=data, raise_exception=False)
```
mrsaemir, line 69: is there any reason you decided not to raise exceptions here? We are at the moment not raising any exceptions in the `validate` implementation, so you can set `raise_exception=True` without any worries.

mrsaemir (commented):
LGTM.
There's only one question to resolve. Thanks

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
spyrostz: Good point, let me bring it back to True. Thanks!

mrsaemir (approved):
LGTM

PR #1612, opened 2023-03-20
GSYE-434: Fix ExternalSCMLoadStrategy

hannesdiedrich:
We need the SCMLoadHoursStrategy in  the ExternalSCMLoadStrategy in order to initiate a profile with 0 values for the forecast SCM strategies.
I also propose to rename the ExternalSCMLoadStrategy into ForecastSCMLoadStrategy. What do you think @spyrostz ?

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-434


spyrostz (approved):
LGTM, thanks for also catching that indeed the forecast strategies and not the external strategies should be used in this case !

PR #1616, opened 2023-04-13
GSYE-581: Allow unused parameter capacity_kW in ForecastSCMPVStrategy

hannesdiedrich:

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/leaves.py
+++ b/src/gsy_e/models/leaves.py
@@ -18,6 +18,9 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 import inspect
 import logging
 
+from gsy_framework.constants_limits import ConstSettings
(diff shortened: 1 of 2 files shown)


src/gsy_e/models/leaves.py
```diff
@@ -76,7 +79,8 @@ class LeafBase:
     strategy = None
 
     def __init__(self, name, config, uuid=None, **kwargs):
-        if config.external_connection_enabled:
+        if (config.external_connection_enabled and
+                ConstSettings.MASettings.MARKET_TYPE != SpotMarketTypeEnum.COEFFICIENTS.value):
```
hannesdiedrich, line 83: This is not part of the fix, only to remove the error messages that appear when launching SCMs

PR #1617, opened 2023-04-14
Feature/gsye 552

spyrostz:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-552-2

### DIFF
--- a/src/gsy_e/models/leaves.py
+++ b/src/gsy_e/models/leaves.py
@@ -46,7 +46,7 @@ from gsy_e.models.strategy.pv import PVStrategy
 from gsy_e.models.strategy.scm.external.load import ForecastSCMLoadStrategy
 from gsy_e.models.strategy.scm.external.pv import ForecastSCMPVStrategy
 from gsy_e.models.strategy.scm.load import SCMLoadHoursStrategy, SCMLoadProfileStrategy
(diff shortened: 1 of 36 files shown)


src/gsy_e/models/leaves.py
```diff
@@ -244,16 +236,14 @@ class ForecastSCMPV(CoefficientLeaf):
     "LoadHours": SCMLoadHours,
     "LoadProfile": SCMLoadProfile,
     "Storage": SCMStorage,
-    "PV": SCMPV,
-    "PredefinedPV": SCMPredefinedPV,
```
hannesdiedrich, line 248: What if the gsy-web sends a `PredefinedPV` ? Or will this be resolved by https://github.com/gridsingularity/gsy-web/pull/1966 ?

src/gsy_e/setup/gsy_e_settings.json
hannesdiedrich, line 1: Thanks a lot for this!

src/gsy_e/models/leaves.py
spyrostz: Exactly, this will be resolved by the other PR. This should not happen with GSy Maps and SCM though, no Predefined PVs are configured.

PR #1620, opened 2023-04-19
Enable HeatPump Strategy to also take comsumption profiles

hannesdiedrich:

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/strategy/energy_parameters/heat_pump.py
+++ b/src/gsy_e/models/strategy/energy_parameters/heat_pump.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Optional, Dict, Union
 
 from gsy_framework.constants_limits import ConstSettings, GlobalConfig
 from gsy_framework.enums import HeatPumpSourceType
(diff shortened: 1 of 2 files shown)


src/gsy_e/models/strategy/energy_parameters/heat_pump.py
```diff
@@ -52,9 +52,13 @@ def __init__(self,
         self.state = HeatPumpState(initial_temp_C, self._slot_length)
 
         if not consumption_profile_uuid and not consumption_kW:
-            consumption_kW = ConstSettings.HeatPumpSettings.CONSUMPTION_KW
+            consumption_W = ConstSettings.HeatPumpSettings.CONSUMPTION_KW * 1000
+        if isinstance(consumption_kW, Dict):
+            consumption_W = {k: v * 1000 for k, v in consumption_kW.items()}
```
spyrostz, line 57: I assume that this was the actual issue here right ? That if the consumption_kW was passed as a dict the multiplication with 1000 to convert to W did not work.

spyrostz (approved):
LGTM

src/gsy_e/models/strategy/energy_parameters/heat_pump.py
hannesdiedrich: Exactly.

PR #1623, opened 2023-04-26
Feature/gsye 515

hannesdiedrich:

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/export.py
+++ b/src/gsy_e/gsy_e_core/export.py
@@ -167,7 +167,7 @@ class ExportAndPlot:
         subdirectory = pathlib.Path(self.directory, "raw_data")
         if not subdirectory.exists():
             subdirectory.mkdir(exist_ok=True, parents=True)
-        json_file = os.path.join(self.directory, "area_tree_summary.json")
+        json_file = os.path.join(subdirectory, "area_tree_summary.json")
         with open(json_file, "w", encoding="utf-8") as outfile:
             json.dump(data, outfile, indent=2)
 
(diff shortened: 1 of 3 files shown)


spyrostz (approved):
LGTM, thanks for the offline clarification!

src/gsy_e/gsy_e_core/export.py
```diff
@@ -167,7 +167,7 @@ def area_tree_summary_to_json(self, data: Dict) -> None:
         subdirectory = pathlib.Path(self.directory, "raw_data")
         if not subdirectory.exists():
             subdirectory.mkdir(exist_ok=True, parents=True)
-        json_file = os.path.join(self.directory, "area_tree_summary.json")
+        json_file = os.path.join(subdirectory, "area_tree_summary.json")
```
hannesdiedrich, line 170: @spyrostz this change will move the `area_tree_summary.json` from the root results folder into the `raw_data` folder. Before I adapt the integration tests. Is this where it supposed to be or was the creation of the `raw_data` path in this method misleading and it should indeed live in the root folder?

spyrostz: Hmmm, in this case I think that it would be better to revert the change. This file is used as a reference for grid tree information, which IMO are not considered raw data, but rather configuration. Therefore I think it would be better to keep on living on the root directory.

tests/test_endpoint_buffers.py
```diff
@@ -171,6 +171,7 @@ def test_generate_json_report_returns_successfully(self, general_setup):
             "mocked-results": "some-results"
         }
 
+    # @staticmethod
```
spyrostz, line 174: Can be removed, right  ?

spyrostz (commented):
2 comments but LGTM in general. Would be also great to give one final look just in case this violates the AVRO schema of the results in case you have not yet (I think it does not but does not hurt to check)

PR #1624, opened 2023-04-27
Production trades and production related statistics were not updated …

spyrostz:
…for homes that had only PVs and coefficient configured to 0.

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/area/scm_manager.py
+++ b/src/gsy_e/models/area/scm_manager.py
@@ -414,26 +414,29 @@ class SCMManager:
                 home_data.energy_need_kWh * market_maker_rate_normal_fees + marketplace_fee
                 + fixed_fee)
 
+        # First handle the sold energy case. This case occurs in case of energy surplus of a home.
+        if home_data.energy_surplus_kWh > 0.0:
+            home_bill.set_sold_to_community(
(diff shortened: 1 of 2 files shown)


src/gsy_e/models/area/scm_manager.py
```diff
@@ -414,26 +414,29 @@ def calculate_home_energy_bills(
...
-                home_bill.set_sold_to_grid(
-                    home_data.self_production_for_grid_kWh, feed_in_tariff)
-                if home_data.self_production_for_grid_kWh > FLOATING_POINT_TOLERANCE:
-                    home_data.create_sell_trade(
-                        self._time_slot, DEFAULT_SCM_GRID_NAME,
-                        home_data.self_production_for_grid_kWh,
-                        home_data.self_production_for_grid_kWh * feed_in_tariff)
-
-            elif home_data.energy_need_kWh > FLOATING_POINT_TOLERANCE:
+            if home_data.energy_need_kWh > FLOATING_POINT_TOLERANCE:
```
hannesdiedrich, line 439: You can push without it, but I guess that you could add this if statement to line 438 and save one level of indention, right?

hannesdiedrich (approved):
Thanks a lot for fixing and adding inline comments!

PR #1633, opened 2023-06-06
Feature/gsye 560

hannesdiedrich:

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-560


tests/market/test_matching_engine_external_matcher.py
```diff
@@ -43,19 +45,20 @@ def _populate_market_bids_offers(self):
     def test_subscribes_to_redis_channels(self):
         self.matcher.matching_engine_ext_conn.sub_to_multiple_channels.assert_called_once_with(
             {
-                "external-matching-engine/simulation-id/": self.matcher.publish_simulation_id,
-                f"{self.channel_prefix}offers-bids/":
+                MatchingEngineChannels(gsy_e.constants.CONFIGURATION_ID).simulation_id:
```
spyrostz, line 48: Minor, but you can construct the MatchingEngineChannels object once and reuse it.

tests/test_external_strategies.py
```diff
@@ -15,25 +16,27 @@
...
-from gsy_framework.constants_limits import ConstSettings, GlobalConfig
-from pendulum import now, duration
-from collections import deque
-from gsy_e.models.strategy.external_strategies import IncomingRequest
 
 
-@pytest.fixture
-def ext_strategy_fixture(request):
+@pytest.fixture(name="ext_strategy_fixture")
+def fixture_ext_strategy(request):
```
spyrostz, line 39: I assume that this is only to satisfy Pycharm right ?

tests/test_gsy_core/test_redis_area_market_communicator.py
```diff
@@ -1,41 +1,45 @@
...
+from gsy_framework.redis_channels import AggregatorChannels
+
 
 from gsy_e.gsy_e_core.redis_connections.aggregator import AggregatorHandler
 from gsy_e.gsy_e_core.redis_connections.area_market import ExternalConnectionCommunicator
 
 
-@pytest.fixture(scope="function", autouse=True)
-def strict_redis():
+@pytest.fixture(scope="function", autouse=True, name="strict_redis")
```
spyrostz, line 13: I think that name was not mandatory here, because these fixtures are not used as function arguments as far as I see, but rather applied for all test functions of the module. Does not change the functionality though as far as I recall

spyrostz (commented):
Some comments, LGTM already

tests/test_gsy_core/test_redis_area_market_communicator.py
hannesdiedrich: You are right. With `autouse=True` the name is not needed. At least in our tests. Let me remove it.

src/gsy_e/gsy_e_core/redis_connections/aggregator.py
```diff
@@ -150,7 +151,7 @@ def _select_aggregator(self, message):
                 "device_uuid": message["device_uuid"],
                 "transaction_id": message["transaction_id"]}
         self.redis_db.publish(
-            "aggregator_response", json.dumps(response_message)
+            AggregatorChannels.response(), json.dumps(response_message)
```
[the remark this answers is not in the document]
spyrostz: IIRC these are optional too right (for the case of gsy-e-sdk where the aggregator has direct access on gsy-e)? Then IMO it is fine to construct the AggregatorChannels with empty strings indeed. 
To my knowledge there is no classproperty decorator, therefore you would have to convert the classmethods to mormal Python methods. I think it is ok to do so though.

spyrostz (approved):

PR #1634, opened 2023-06-08
Feature/gsye 626

hannesdiedrich:

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-626


src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
```diff
@@ -185,9 +185,9 @@ def _specific_row(self, slot, market):
             produced = self.area.strategy.state.get_energy_production_forecast_kWh(slot, 0.0)
             return [produced, not_sold]
         if isinstance(self.area.strategy, HeatPumpStrategy):
-            return [self.area.strategy.state.get_storage_temp_C(slot),
-                    self.area.strategy.state.get_temp_decrease_K(slot),
-                    self.area.strategy.state.get_temp_increase_K(slot)
+            return [round(self.area.strategy.state.get_storage_temp_C(slot), 5),
```
spyrostz, line 188: I think that it would be better to add this as a constant, we already have FLOATING_POINT_TOLERANCE but this is not compatible with `round`, therefore I would recommend to add a new constant called `ROUND_TOLERANCE = 5`.

spyrostz (approved):
One minor comment, LGTM

PR #1689, opened 2023-09-15
GSYE-637: Corrected the SCM past slots operation mode, by passing the…

spyrostz:
… SCM simulation state from the past slots simulation run to the following Canary Network. Modified state in order to trick the subsequent Canary Network run that it is a running Canary Network that resumed operation.

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/rq_job_handler.py
+++ b/src/gsy_e/gsy_e_core/rq_job_handler.py
@@ -209,6 +209,11 @@ def _handle_scm_past_slots_simulation_run(
 
     config = _create_config_settings_object(
         scenario_copy, settings_copy, aggregator_device_mapping)
+    # We are running SCM Canary Networks with some days of delay compared to realtime in order to
+    # compensate for delays in transmission of the asset measurements.
+    # Adding 4 hours of extra time to the SCM past slots simulation duration, in order to
+    # compensate for the runtime of the SCM past slots simulation and to not have any results gaps
+    # after this simulation run and the following Canary Network launch.
     config.end_date = now(tz=gsy_e.constants.TIME_ZONE).subtract(
         days=gsy_e.constants.SCM_CN_DAYS_OF_DELAY).add(hours=4)
     config.sim_duration = config.end_date - config.start_date
--- a/src/gsy_e/gsy_e_core/rq_job_handler.py
+++ b/src/gsy_e/gsy_e_core/rq_job_handler.py
@@ -1,11 +1,10 @@
 import ast
-from copy import deepcopy
 import logging
 import traceback
+from copy import deepcopy
 from datetime import datetime, date
 from typing import Dict, Optional
 
-import pendulum
 from gsy_framework.constants_limits import GlobalConfig, ConstSettings
 from gsy_framework.enums import ConfigurationType, SpotMarketTypeEnum
 from gsy_framework.settings_validators import validate_global_settings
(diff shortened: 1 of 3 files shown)


src/gsy_e/gsy_e_core/rq_job_handler.py
```diff
@@ -182,28 +191,38 @@ def _handle_scm_past_slots_simulation_run(
...
+    # Deepcopy the scenario and settings objects, because they are mutated by the
+    # run_simulation function, and we need the original versions for the subsequent
+    # Canary Network run.
+    scenario_copy = deepcopy(scenario)
+    settings_copy = deepcopy(settings)
+
+    config = _create_config_settings_object(
+        scenario_copy, settings_copy, aggregator_device_mapping)
+    config.end_date = now(tz=gsy_e.constants.TIME_ZONE).subtract(
+        days=gsy_e.constants.SCM_CN_DAYS_OF_DELAY).add(hours=4)
```
hannesdiedrich, line 213: The `add()` is probably a remnant of testing, right?

hannesdiedrich (commented):

src/gsy_e/gsy_e_core/rq_job_handler.py
spyrostz: No, this is actually necessary in order to compensate for the runtime of the simulation. Think of the sequence of events that way:
- gsy-web triggers an SCM past slot simulation with a start time 4 months in the past
- gsy-e receives the event and starts a simulation with a start time 4 months ago, and a duration of `now() - 4 months`. Let's say that this simulation will have a runtime of 1 hour.
- once the simulation finishes, gsy-e then starts a Canary Network, at the current now() which is the now() at the start of the simulation in step #2 + 1 hour. Therefore, the simulation and the Canary Network will leave a gap of 1 hour in the results

In order to compensate for that gap, we need to run the simulation in step #2 for longer than the expected duration at the start of it. I decided to set it to 4 hours, in order to take into account any simulations that might take a long time to calculate their results.

hannesdiedrich: Ok, thank you for your explanation. Makes sense.
I was actually confused by the subtraction of SCM_CN_DAYS_OF_DELAY. So if the user starts a CN with start time of 4 months in the past, additional 3 days are subtracted from the start time and the end time. Why is this again? AFAIU it is not a problem, because the start of the actual CN (not the pre-calculation sim) is also 3 days earlier. Right?

spyrostz: Exactly. This variable controls the constant delay of the CN compared to real time. Therefore the Canary Network will always start at the current time - SCM_CN_DAYS_OF_DELAY, and  the simulation that runs before it has to run until that time too.

hannesdiedrich: Can you please remind me of the reason for this please? Is this in order to have results showing up on the UI when the CN is running?

For this PR I would propose an inline comment, that explains the `add(hours=4)` for the future me to not ask again ;-) Thank a lot!

spyrostz: I think that the comment I added will clarify this. The SCM_CN_DAYS_OF_DELAY is there in order to account for delays in sending of measurements from the assets. At the moment, Datadis can send the updated measurements even with days of delay, therefore with such a delay we can be certain that we will have received the correct and updated measurement data from all assets.

hannesdiedrich (approved):
LGTM

PR #1690, opened 2023-09-17
GSYE-637: Added gsy_energy_bill_revenue, gsy_energy_bill_excl_fees an…

spyrostz:
…d gsy_total_benefit calculation.

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/area/scm_manager.py
+++ b/src/gsy_e/models/area/scm_manager.py
@@ -315,12 +315,12 @@ class AreaEnergyBills:  # pylint: disable=too-many-instance-attributes
 
     @property
     def gsy_energy_bill_revenue(self):
-        """Energy bill of the home excluding revenue."""
+        """Total revenue from grid and community."""
         return self.earned_from_grid + self.earned_from_community
 
     @property
     def gsy_energy_bill_excl_fees(self):
-        """Energy bill of the home excluding revenue."""
+        """Energy bill of the home excluding fees."""
         return (self.gsy_energy_bill - self.grid_fees - self.tax_surcharges
                 - self.fixed_fee - self.marketplace_fee - self.assistance_fee)
 
--- a/src/gsy_e/models/area/scm_manager.py
+++ b/src/gsy_e/models/area/scm_manager.py
@@ -228,7 +228,10 @@ class AreaEnergyBills:  # pylint: disable=too-many-instance-attributes
             "home_balance": self.home_balance,
             "gsy_energy_bill_excl_revenue": self.gsy_energy_bill_excl_revenue,
             "gsy_energy_bill_excl_revenue_without_fees":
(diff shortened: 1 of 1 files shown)


src/gsy_e/models/area/scm_manager.py
```diff
@@ -310,6 +313,22 @@ def gsy_energy_bill_excl_revenue_without_fees(self):
         return (self.gsy_energy_bill_excl_revenue - self.grid_fees - self.tax_surcharges
                 - self.fixed_fee - self.marketplace_fee - self.assistance_fee)
 
+    @property
+    def gsy_energy_bill_revenue(self):
+        """Energy bill of the home excluding revenue."""
```
hannesdiedrich, line 318: The docstring does not comply to the implementation. 
How about:
```suggestion
        """Total revenue from gird and community."""
```

src/gsy_e/models/area/scm_manager.py
```diff
@@ -310,6 +313,22 @@ def gsy_energy_bill_excl_revenue_without_fees(self):
...
+        return self.earned_from_grid + self.earned_from_community
+
+    @property
+    def gsy_energy_bill_excl_fees(self):
+        """Energy bill of the home excluding revenue."""
```
hannesdiedrich, line 323: Here too, I think the following should apply, right?
```suggestion
    def gsy_energy_bill_excl_fees(self):
        """Energy bill of the home excluding fees."""
```

src/gsy_e/models/area/scm_manager.py
```diff
@@ -310,6 +313,22 @@ def gsy_energy_bill_excl_revenue_without_fees(self):
...
+        return (self.gsy_energy_bill - self.grid_fees - self.tax_surcharges
```
hannesdiedrich, line 324: Can you please remind me of the sign convention of `gsy_energy_bill` ? What if this is negative? Does this mean the user has a credit balance? If yes, the fees should be added here, right?

hannesdiedrich (commented):
Some questions and comments, LGTM once addressed

src/gsy_e/models/area/scm_manager.py
spyrostz: Good catch, thanks !

src/gsy_e/models/area/scm_manager.py
spyrostz: Indeed, thanks!

src/gsy_e/models/area/scm_manager.py
spyrostz: The energy bill includes the fees right from the start (essentially its default value is not 0 but `monthly_fee + assistance_fee + fixed_fee`) therefore the fees should be subtracted from it.

hannesdiedrich (approved):
LGTM

PR #1694, opened 2023-10-09
GSYE-567: Fix issues in area_serializer and export setup file; add se…

hannesdiedrich:
…rialization method in heat pump strategy

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/area_serializer.py
+++ b/src/gsy_e/gsy_e_core/area_serializer.py
@@ -40,6 +40,7 @@ from gsy_e.models.leaves import (
     Leaf, scm_leaf_mapping, CoefficientLeaf, forward_leaf_mapping,
     forecast_scm_leaf_mapping) # NOQA
 from gsy_e.models.leaves import *  # NOQA  # pylint: disable=wildcard-import
+from gsy_e.models.strategy.trading_strategy_base import TradingStrategyBase
 
 logger = getLogger(__name__)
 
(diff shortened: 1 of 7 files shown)


src/gsy_e/gsy_e_core/area_serializer.py
```diff
@@ -55,8 +56,10 @@ def default(self, o):
             return self._encode_leaf(o)
         if isinstance(o, BaseStrategy):
             return self._encode_subobject(o)
+        if isinstance(o, TradingStrategyBase):
```
spyrostz, line 59: This is because the Heatpump strategies do not inherit from BaseStrategy right ? At any case, I agree with the addition!

src/gsy_e/models/strategy/storage.py
```diff
@@ -60,7 +60,7 @@ def serialize(self):
             "battery_capacity_kWh": self.state.capacity,
             "max_abs_battery_power_kW": self.state.max_abs_battery_power_kW,
             "cap_price_strategy": self.cap_price_strategy,
-            "initial_energy_origin": self.state.initial_energy_origin,
+            "initial_energy_origin": self.state.initial_energy_origin.value,
```
spyrostz, line 63: Thank you!

spyrostz (approved):
LGTM

src/gsy_e/gsy_e_core/area_serializer.py
hannesdiedrich: Exactly.

PR #1697, opened 2023-10-16
Added images to the main blockchain page and to the blockchain system…

spyrostz:
… components overview page.

## Proposed changes

-


wiki/docs/blockchain-system-components-overview.md
```diff
@@ -8,6 +8,13 @@ The [Substrate](https://substrate.io/){target=_blank}-based GSy Node is the back
 
 For more, please see the section on [GSy Node Installation, Operation and Development](blockchain-installation.md).
 
+
+
+![alt_text](img/blockchain-system-components-overview-1.png)
+
+***Figure 1***. *GSY DEX On-Chain Operations via GSY Node (Note: Order Book is where bids and offers are stored, to be then matched based on the applicable trading mechanism for price verification and executed based on energy delivery verification).*
```
hannesdiedrich, line 15: Hey, I took the liberty in my last PR to homogenize the images and used the following html snippet (under: "Add figure caption" here: https://github.com/gridsingularity/gsy-e/tree/master/wiki#images
This will connect the image with its caption
Could you also please follow this?

hannesdiedrich: Also, I also prepended the section number in the figure number:
For the BC section this would result in:
`Figure 6.1`

spyrostz: Darn, it makes the documentation files a bit ugly, too bad that mkdocs does not natively support this. Will fix, thanks

hannesdiedrich: I also hate that this is not supported be default, but also open to alternatives. 
I remember quite some digging though.

spyrostz: > Also, I also prepended the section number in the figure number:
For the BC section this would result in:
Figure 6.1

BTW I do not like this at all. I will follow this for consistency's purpose however this means that if we insert one new section after current section #1, we wo;; have to change the figure numbers of all the sections apart from 1. Regardless, let me do it and if we stumble upon it we will see how to deal with it.

hannesdiedrich: I also thought about this, but thought that this will be kind of easy with a search & replace. Of course, this search&replace will have to be performed n times and if the number of sections is increased a lot, this is not so quick any more. 
If we do not follow this "sectionisation", we will have to change all figure names if we want to insert a figure in section 1 by hand or we will have a Figure 82 below Figure 1 (not that I care that much about that, just wanted to outline my reasoning).

hannesdiedrich (approved):
LGTM

PR #1703, opened 2023-10-18
Feature/gsye 651

hannesdiedrich:
Before I looked into the HP, I added a mixin for the market maker rate and only realised afterwards, that in the heat pump we deal with the price settings differently. 
Only a little tweak had to be done in the HP in order for the FE to just set the final buying rate in the heat pump which leads to using the market maker rate. 

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-651


src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -134,18 +134,16 @@ def deserialize_args(constructor_args: Dict) -> Dict:
...
-                        initial_rate=constructor_args.get(
-                            "initial_buying_rate",
-                            ConstSettings.HeatPumpSettings.BUYING_RATE_RANGE.initial),
-                        final_rate=constructor_args.get(
-                            "final_buying_rate",
-                            ConstSettings.HeatPumpSettings.BUYING_RATE_RANGE.final
-                        ))
+                        update_interval=(duration(
+                            minutes=constructor_args.get("update_interval"))
+                                    if constructor_args.get("update_interval") else None),
```
spyrostz, line 139: This clause is only in order to avoid `update_interval=0`, right ? Does not hurt to revert to None however I am not certain whether this is preferable to let the validator handle and reject this value. Leave it as is at any case.

src/gsy_e/models/strategy/load_hours.py
```diff
@@ -259,11 +254,8 @@ def area_reconfigure_event(self, *args, **kwargs):
 
     def event_activate_price(self):
         """Update the strategy prices upon the activation and validate them afterwards."""
-        # If use_market_maker_rate is true, overwrite final_buying_rate to market maker rate
         if self.use_market_maker_rate:
```
spyrostz, line 257: Why not including the `use_market_maker_rate` check in the mixin too ? That way it might be easier in the future if we revert from inheritance to composition (convert the mixin to a normal class that will be constructed in the LoadHoursStrategy constructor)

spyrostz (approved):
Some comments, LGTM in general

src/gsy_e/models/strategy/load_hours.py
hannesdiedrich: Makes sense, will do.

src/gsy_e/models/strategy/infinite_bus.py
```diff
@@ -44,15 +44,15 @@ def __init__(self, energy_sell_rate=None, energy_rate_profile=None, energy_buy_r
         # buy
         if all(arg is None for arg in [
                buying_rate_profile, buying_rate_profile_uuid, energy_buy_rate]):
-            energy_buy_rate = ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE
+            energy_buy_rate = GlobalConfig.FEED_IN_TARIFF
```
spyrostz, line 47: I am hesitant to have yet another round for this, but I think that since this parameter is mutable, it is better to follow the same approach as for the MARKET_MAKER_RATE. Essentially to create a new constant named DEFAULT_FEED_IN_TARIFF, and in the GlobalConfig class, to assign it `FEED_IN_TARIFF = ConstSettings.DEFAULT_FEED_IN_TARIFF`. That way the default value will remain unchanged, even if the GlobalConfig.FEED_IN_TARIFF changes.

src/gsy_e/models/strategy/infinite_bus.py
```diff
@@ -44,15 +44,15 @@ def __init__(self, energy_sell_rate=None, energy_rate_profile=None, energy_buy_r
...
         self._buy_energy_profile = EnergyProfile(
             buying_rate_profile, buying_rate_profile_uuid, energy_buy_rate,
             profile_type=InputProfileTypes.IDENTITY)
 
         # sell
         if all(arg is None for arg in [
                energy_rate_profile, energy_rate_profile_uuid, energy_sell_rate]):
-            energy_sell_rate = ConstSettings.GeneralSettings.DEFAULT_MARKET_MAKER_RATE
+            energy_sell_rate = GlobalConfig.MARKET_MAKER_RATE
```
spyrostz, line 55: Why not keep using the default here? The MARKET_MAKER_RATE is mutable and changes throughout the simulation.

tests/strategies/test_heat_pump.py
```diff
@@ -37,6 +39,7 @@ def fixture_heatpump_strategy(request) -> Tuple["TradingStrategyBase", "Area"]:
                 gsye_root_path, "resources", "hp_external_temp_C.csv"),
             **strategy_params)
     else:
+        print(strategy_params)
```
spyrostz, line 42: I think that this can be removed.

spyrostz (commented):

src/gsy_e/models/strategy/heat_pump.py
spyrostz: Sorry missed this, agreed of course and thanks!

spyrostz (approved):
LGTM

PR #1706, opened 2023-10-20
GSYE-640: Remove 'area_stats' key from market area stats, also remove…

hannesdiedrich:
… market area stats from asset states

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-640

### DIFF
--- a/src/gsy_e/models/area/area.py
+++ b/src/gsy_e/models/area/area.py
@@ -91,17 +91,15 @@ class Area(AreaBase):
     def get_state(self):
         """Get the current state of the area."""
         state = super().get_state()
-
-        state.update(**{
-            "current_tick": self.current_tick,
-            "area_stats": self.stats.get_state()
-        })
+        state.update({"current_tick": self.current_tick})
+        if not self.strategy:
+            state.update(**self.stats.get_state())
         return state
 
     def restore_state(self, saved_state):
         """Restore a previously-saved state."""
         self.current_tick = saved_state["current_tick"]
-        self.stats.restore_state(saved_state["area_stats"])
+        self.stats.restore_state(saved_state)
         super().restore_state(saved_state)
 
     def area_reconfigure_event(self, **kwargs):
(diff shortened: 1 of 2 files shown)


src/gsy_e/models/area/area.py
```diff
@@ -91,17 +91,15 @@ def __init__(self, name: str = None, children: List["Area"] = None,
     def get_state(self):
         """Get the current state of the area."""
         state = super().get_state()
-
-        state.update(**{
-            "current_tick": self.current_tick,
-            "area_stats": self.stats.get_state()
-        })
+        state.update({"current_tick": self.current_tick})
+        if not self.strategy:
```
spyrostz, line 95: I assume that the strategy areas do not need the AreaStats and that's why this clause was added, right?

src/gsy_e/models/area/area.py
```diff
@@ -91,17 +91,15 @@ def __init__(self, name: str = None, children: List["Area"] = None,
     def get_state(self):
         """Get the current state of the area."""
         state = super().get_state()
-
-        state.update(**{
-            "current_tick": self.current_tick,
-            "area_stats": self.stats.get_state()
-        })
+        state.update({"current_tick": self.current_tick})
```
spyrostz, line 94: Nothing to be done now, but I think that it is not needed to repeat the `current_tick` for every area, and it should be stored only once. However it is part of a larger refactoring and there is a good chance once we implement it we will design this in a way that avoids repetition of state. TLDR nothing to do now

spyrostz (approved):
LGTM

src/gsy_e/models/area/area.py
hannesdiedrich: This is my proposal actually. Currenlty, area_stats contain the followeing members:
```
exported_energy
imported_energy
rate_stats_market
```
which are market area specific. 
Do you foresee that the assets might also export statistics that should be part of the state in the future? 
If yes, let me revert this.

src/gsy_e/models/area/area.py
hannesdiedrich: I was also thinking about this. Indeed, lets tackle this in the future.

src/gsy_e/models/area/area.py
spyrostz: This is only part of the state, therefore I do not think that it will be necessary for the future. Leave it as is and we will revert the clause if needed.

spyrostz (approved):
LGTM

PR #1717, opened 2023-12-08
GSYE-666: Also track unmatched demand of the heat pump when max_power…

hannesdiedrich:
…_rating is lower than demand

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/strategy/energy_parameters/heat_pump.py
+++ b/src/gsy_e/models/strategy/energy_parameters/heat_pump.py
@@ -190,7 +190,12 @@ class HeatPumpEnergyParameters(HeatPumpEnergyParametersBase):
             self.state.get_temp_decrease_K(time_slot)) / self.state.get_cop(time_slot)
 
         assert max_energy_consumption > -FLOATING_POINT_TOLERANCE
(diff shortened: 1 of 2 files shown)


spyrostz (approved):
LGTM, mainly because it is mostly an unrealistic scenario for undersized heatpumps, which is normally not the case. However, I am not certain if this is the right way to go, because if the heat demand of the home is greater than the power rating of the heatpump, one can speculate that the unmatched demand should not be limited at all by the power rating. Having said that, and since it is a matter of debate, feel free to merge this and if it becomes an issue we can change it later down the road.

hannesdiedrich:
> LGTM, mainly because it is mostly an unrealistic scenario for undersized heatpumps, which is normally not the case. However, I am not certain if this is the right way to go, because if the heat demand of the home is greater than the power rating of the heatpump, one can speculate that the unmatched demand should not be limited at all by the power rating. Having said that, and since it is a matter of debate, feel free to merge this and if it becomes an issue we can change it later down the road.

I was also thinking about the fact that we mix unmatched demands due to different reasons. 
Still, I was confused to not see unmatched demand when the demand was higher than the max power rating, so I still stand by this PR. 
I agree though that we might have to discuss this again in the future. Thanks!

PR #1718, opened 2023-12-11
Feature/gsye 664

hannesdiedrich:

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
+++ b/src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
@@ -31,6 +31,7 @@ from gsy_e.models.strategy.scm.load import SCMLoadHoursStrategy, SCMLoadProfileS
 from gsy_e.models.strategy.scm.pv import SCMPVUserProfile
 from gsy_e.models.strategy.storage import StorageStrategy
 from gsy_e.models.strategy.heat_pump import HeatPumpStrategy
+from gsy_e.models.strategy.scm.storage import SCMStorageStrategy
 from gsy_e.models.strategy.virtual_heatpump import VirtualHeatpumpStrategy
 
 if TYPE_CHECKING:
@@ -389,6 +390,7 @@ class CoefficientDataExporter(BaseDataExporter):
 
 
 class CoefficientLeafDataExporter(BaseDataExporter):
+    # pylint: disable=protected-access
     """LeafDataExporter for SCM simulations."""
 
     def __init__(self, area: "CoefficientArea"):
@@ -396,6 +398,8 @@ class CoefficientLeafDataExporter(BaseDataExporter):
 
     @property
     def labels(self) -> List:
+        if isinstance(self._area.strategy, SCMStorageStrategy):
+            return ["slot", "traded energy [kWh]"]
         if isinstance(self._area.strategy, (SCMLoadHoursStrategy, SCMLoadProfileStrategy)):
             return ["slot", "desired energy [kWh]", "deficit [kWh]"]
         if isinstance(self._area.strategy, SCMPVUserProfile):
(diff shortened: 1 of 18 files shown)


src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
```diff
@@ -397,8 +396,6 @@ def __init__(self, area: "CoefficientArea"):
 
     @property
     def labels(self) -> List:
-        if isinstance(self._area.strategy, SCMStorageStrategy):
-            return ["slot", "charge [kWh]", "offered [kWh]", "charge [%]"]
```
spyrostz, line 401: I understand the reason why these are removed, but why not converting to the new SCMStorageStrategy attributes? Or did you decide that since we will not run SCM in CLI mode, it is better to leave as is?

src/gsy_e/models/strategy/state/storage_state.py
```diff
@@ -453,3 +453,19 @@ def get_results_dict(self, current_time_slot: DateTime) -> dict:
         return {
             "soc_history_%": self.charge_history.get(current_time_slot, 0)
         }
+
+
+class ScmStorageState(StateInterface):
```
spyrostz, line 458: Out of curiosity, is the empty class needed in this case ? Is it just for having a state object, thus allowing the EndpointBuffer class to access all states the same way?

spyrostz (approved):
1 main comment and 1 clarification comment, LGTM from now though

src/gsy_e/models/strategy/state/storage_state.py
hannesdiedrich: Exactly, I am not proud of this. but it is needed unfortunately.

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
hannesdiedrich: > but why not converting to the new SCMStorageStrategy attributes

Which attributes, there are none to be plotted.

src/gsy_e/models/strategy/state/storage_state.py
spyrostz: In this case though, I think that it can either be better named by this being called "DummyState" , since it does not do anything. Or possibly using default implementation on these abstractmethods.

hannesdiedrich: > Or possibly using default implementation on these abstractmethods.

I tried using `StateInterface` directly, but I remember It was not allowed as it was an ABC. I try again, if not, I will rename indeed.

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
spyrostz: There is always one attribute that can be shown: traded energy. This is also the only thing reported for the infinite bus, and I think it is important to have just for the purpose of completeness

src/gsy_e/models/strategy/state/storage_state.py
spyrostz: Might be because the abstractmethods do not have a default implementation, regardless though I think it is better to create a dummy class that essentially disables the interface. Thanks!

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
hannesdiedrich: Hmm, this is the SCM. All energy is traded, right? So this will be only a duplicate of the input profile of the SCMStorage. Do you think this is valuable to have?

spyrostz: The thing is that there is no other place where we can see it as part of the results. Maybe we can see it as part of the setup file / input profile, however I think it will be more convenient and more harmonized with the other results if we had it here as well.

hannesdiedrich: done [e0d6333](https://github.com/gridsingularity/gsy-e/pull/1718/commits/e0d633312df3c7226bc7d869d65f5ee80eedc5d1)

PR #1719, opened 2023-12-11
Feature/gsye 649

spyrostz:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-649
**GSY_FRAMEWORK_BRANCH**=feature/GSYE-649

### DIFF
--- a/src/gsy_e/models/strategy/energy_parameters/virtual_heat_pump.py
+++ b/src/gsy_e/models/strategy/energy_parameters/virtual_heat_pump.py
@@ -19,6 +19,108 @@ WATER_SPECIFIC_HEAT_CAPACITY = 4182  # [J/kg°C]
 GROUND_WATER_TEMPERATURE_C = 12
 
 
+class HeatpumpStorageEnergySolver:
+    # pylint: disable=too-many-instance-attributes,too-many-arguments
+    """Solver that calculates heatpump storage temperature and energy."""
+    def __init__(
+            self, tank_volume_l: float, current_storage_temp: float, dh_supply_temp: float,
+            dh_return_temp: float, dh_flow_m3_per_hour: float,
+            target_storage_temp_C: Optional[float] = None, energy_kWh: Optional[float] = None):
+        # Optional inputs
+        self.energy_kWh = energy_kWh
+        self.target_storage_temp_C = target_storage_temp_C
+        # Inputs
+        self.current_storage_temp = current_storage_temp
+        self.dh_supply_temp = dh_supply_temp
+        self.dh_return_temp = dh_return_temp
+        self.dh_flow_kg_per_sec = dh_flow_m3_per_hour * 1000 / 3600
+        if self.dh_flow_kg_per_sec < FLOATING_POINT_TOLERANCE:
+            self.dh_flow_kg_per_sec = 0
+        self._tank_volume_l = tank_volume_l
+        # Outputs
+        self.temp_differential_per_sec = None
+        self.q_out_J = None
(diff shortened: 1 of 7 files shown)


src/gsy_e/models/strategy/energy_parameters/virtual_heat_pump.py
```diff
@@ -197,9 +197,7 @@ def _calc_temp_decrease_K(self, time_slot: DateTime) -> float:
     def _calc_temp_increase_K(self, time_slot: DateTime, traded_energy_kWh: float) -> float:
         storage_temp = self._energy_to_storage_temp(traded_energy_kWh, time_slot)
         current_storage_temp = self.state.get_storage_temp_C(time_slot)
-        if storage_temp <= current_storage_temp:
-            return 0
-        return storage_temp - current_storage_temp
+        return storage_temp - current_storage_temp + self.state.get_temp_decrease_K(time_slot)
```
hannesdiedrich, line 216: Took me a while to understand. I think I got it now.
So in the ideal scenario, the traded energy == demanded energy in which case 
`storage_temp - current_storage_temp == self.state.get_temp_decrease_K(time_slot)` 
applies, right?

hannesdiedrich: I thought this is all done in the `update_storage_temp` and therefore this is a duplication here. Or am I missing something here?

hannesdiedrich: If this ( `+ self.state.get_temp_decrease_K(time_slot)`) has to reside in here, I would vote to rename this method to `calc_storage_temp_change`

hannesdiedrich (commented):
Left some comments to the same issue. Apart from that, LGTM

src/gsy_e/models/strategy/energy_parameters/virtual_heat_pump.py
spyrostz: This confused me too in the base heatpump implementation as well. In general the algorithm is like this:
- calculate temp decrease due to heat demand (e.g. -2 degrees C)
- calculate temp increase due to energy trades, excluding heat demand (e.g. +3 degrees C)
- update the next temperature of the storage. Assuming current temp is 25 degrees, the next temperature will be 25 + 3 - 2  = 26 degrees C

In the virtual heatpump though, we already have the next temperature of the storage already, because the heat demand is part of the linear system already. Therefore, the formula `current_temp + increase - decrease = next_temp` will change to `increase = next_temp - current_temp + decrease`, because the unknown factor is the `increase`. Then the `update_storage_temp` (which is reused by both strategies) can recalculate correctly the next_temp, which is of course already calculated by this model, but is not already calculated by the base heatpump strategy, and both strategy classes need to be compatible. 

 For a similar reason I cannot easily rename this `calc_storage_temp_change`. This is a base class method that has to be overridden in order to disable the base implementation and enable the new one.

hannesdiedrich: Thank you for your verbose comment, but:
Sorry, I still not get the following :
The `update_storage_temp` calculates the new temp by:
```
        new_temp = (self.get_storage_temp_C(...)
                    - self.get_temp_decrease_K(...)
                    + self.get_temp_increase_K(...))
```
If the `_calc_temp_increase_K` in the virtual heat pump also adds the decrease, then AFAIS, the decrease is double counted.

PR #1720, opened 2023-12-12
Feature/gsye 665

hannesdiedrich:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-665


src/gsy_e/models/leaves.py
```diff
@@ -232,19 +238,25 @@ class ForecastSCMPV(CoefficientLeaf):
     strategy_type = ForecastSCMPVStrategy
 
 
+class ForecastSCMHeatPump(CoefficientLeaf):
+    strategy_type = ForecastScmHeatPumpStrategy
```
spyrostz, line 242: I think that we discussed it some time ago too, but I do not think that we will need Forecast strategies for the SCM. Even now, we only read from profiles from the database even for Canary Networks, therefore we do not take advantage of the Redis API that is invoked with the Forecast strategies. I would suggest to remove this strategy completely, and if you agree I will add another story to remove the rest of the ForecastScm strategies.

src/gsy_e/models/strategy/scm/heat_pump.py
```diff
@@ -0,0 +1,44 @@
...
+
+class ScmHeatPumpEnergyParameters(DefinedLoadEnergyParameters):
+    """
+    Heat Pump Energy Parameters for the ScmHeatPumpStrategy
+    Only needed in order to only allow energy profiles
+    """
+
+    def __init__(self, consumption_kWh_profile=None, consumption_kWh_profile_uuid: str = None):
+        # pylint: disable= super-init-not-called
+        """super() should not be called"""
```
spyrostz, line 18: Why not? I think that it could be invoked directly via:
```
super().__init__(daily_load_profile= consumption_kWh_profile, daily_load_profile_uuid= consumption_kWh_profile_uuid)
```
If it is just for avoiding this line:
```
        super().__init__(avg_power_W=0, hrs_per_day=24, hrs_of_day=list(range(0, 24)))
```
as far as I recall this is just fo disabling the load hours strategy. Therefore should not hurt to have in the heatpump too.

spyrostz (commented):
Some comments, looks good in general though!

src/gsy_e/models/leaves.py
hannesdiedrich: Agreed. Will do. Thanks for adding the ticket!

src/gsy_e/models/strategy/scm/heat_pump.py
hannesdiedrich: Yes, I decided against it because of the line you are mentioning and and more importantly this one:
```python:
        self.energy_profile = EnergyProfile(daily_load_profile, daily_load_profile_uuid)
```
This will load the profile as power profile, however, we have an energy profile here.
Do you have a better solution for this?

spyrostz: Yes, profile type can be overridden by this in the constructor of the ScmHeatPumpEnergyParameters:
self.energy_profile.profile_type = InputProfileTypes.ENERGY_KWH. Keep in mind that constructors do not read the profiles, they only setup the EnergyProfile class, therefore you can change the profile type in the HeatPump constructor and it will apply to new profiles.

hannesdiedrich: > Keep in mind that constructors do not read the profiles, they only setup the EnergyProfile class

Ah, I was confused by the `should_read_profile_from_db` if statement in the constructor, but you are right of course.

hannesdiedrich: I was wrong, I wanted to make sure, that the `ScmHeatPumpEnergyParameters` area assigned to ` ScmHeatPumpStrategy.energy_profile` and not the `DefinedLoadEnergyParameters`. In there I will call `super().__init__` though in order to align with our discussion here.

spyrostz (approved):
LGTM

PR #1733, opened 2024-01-18
Bug iaa

spyrostz:

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=bug-iaa

### DIFF
--- a/src/gsy_e/constants.py
+++ b/src/gsy_e/constants.py
@@ -23,7 +23,10 @@ import os
 from gsy_framework.constants_limits import DATE_TIME_FORMAT, DATE_TIME_UI_FORMAT, TIME_ZONE  # NOQA
 from gsy_framework.constants_limits import TIME_FORMAT, DATE_FORMAT, GlobalConfig  # NOQA
 
-FLOATING_POINT_TOLERANCE = 0.00001
+# In order to cover conversion and reverse-conversion to 5 decimal points, the tolerance has to be
+# 0.00002. That way off-by-one consecutive rounding errors would not be treated as errors, e.g.
+# when recalculating the original energy rate in trade chains.
+FLOATING_POINT_TOLERANCE = 0.00002
 ROUND_TOLERANCE = 5
 
 # Percentual standard deviation relative to the forecast energy, used to compute the (simulated)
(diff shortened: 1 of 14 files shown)


hannesdiedrich:
One general remark WRT the branch name: We call it MarketAgent nowadays ;-)

src/gsy_e/constants.py
```diff
@@ -23,7 +23,10 @@
 from gsy_framework.constants_limits import DATE_TIME_FORMAT, DATE_TIME_UI_FORMAT, TIME_ZONE  # NOQA
 from gsy_framework.constants_limits import TIME_FORMAT, DATE_FORMAT, GlobalConfig  # NOQA
 
-FLOATING_POINT_TOLERANCE = 0.00001
+# In order to cover conversion and reverse-conversion to 5 decimal points, the tolerance has to be
+# 0.00002. That way off-by-one consecutive rounding errors would not be treated as errors, e.g.
+# when recalculating the original energy rate in trade chains.
```
hannesdiedrich, line 28: ❤️

spyrostz:
> One general remark WRT the branch name: We call it MarketAgent nowadays ;-)

Of course you are right but old habits die hard!

src/gsy_e/models/market/grid_fees/base_model.py
```diff
@@ -66,7 +66,7 @@ def update_forwarded_bid_trade_original_info(self, trade_original_info, market_b
         if not trade_original_info:
             return None
         trade_offer_info = TradeBidOfferInfo(
-            original_bid_rate=market_bid.original_price / market_bid.energy,
+            original_bid_rate=market_bid.original_energy_rate,
```
hannesdiedrich, line 69: nice! this will save also some calculation effort, right?

hannesdiedrich:
Ah, you added the totral traded energy part here. I thought I put it into my heat pump branch. Thematically it shold not be here, right?

src/gsy_e/models/market/grid_fees/base_model.py
spyrostz: Unfortunately not, because this is a property. We will not have to remind ourselves to do the rounding though, because it is handled by the property.

spyrostz:
> Ah, you added the totral traded energy part here. I thought I put it into my heat pump branch. Thematically it shold not be here, right?

Does not belong thematically, but I was under the impression that you would not add it in your PR. Regardless, I will merge both changes, should not introduce many conflicts.

src/gsy_e/models/market/grid_fees/base_model.py
hannesdiedrich: I saw this now in the framework PR. Thanks

PR #1737, opened 2024-01-24
Bug/gsye 674 proposal2

hannesdiedrich:

## Proposed changes

-

### DIFF
--- a/src/gsy_e/constants.py
+++ b/src/gsy_e/constants.py
@@ -23,7 +23,10 @@ import os
 from gsy_framework.constants_limits import DATE_TIME_FORMAT, DATE_TIME_UI_FORMAT, TIME_ZONE  # NOQA
 from gsy_framework.constants_limits import TIME_FORMAT, DATE_FORMAT, GlobalConfig  # NOQA
 
-FLOATING_POINT_TOLERANCE = 0.00001
+# In order to cover conversion and reverse-conversion to 5 decimal points, the tolerance has to be
+# 0.00002. That way off-by-one consecutive rounding errors would not be treated as errors, e.g.
+# when recalculating the original energy rate in trade chains.
+FLOATING_POINT_TOLERANCE = 0.00002
 ROUND_TOLERANCE = 5
 
 # Percentual standard deviation relative to the forecast energy, used to compute the (simulated)
--- a/src/gsy_e/gsy_e_core/area_serializer.py
+++ b/src/gsy_e/gsy_e_core/area_serializer.py
@@ -23,7 +23,7 @@ from gsy_framework.utils import convert_pendulum_to_str_in_dict, key_in_dict_and
 from gsy_framework.constants_limits import ConstSettings, SpotMarketTypeEnum, GlobalConfig
 from pendulum import Duration
 
-from gsy_e.models.area import CoefficientArea, Area # NOQA
+from gsy_e.models.area import CoefficientArea, Area, Market, Asset # NOQA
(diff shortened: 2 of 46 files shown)


hannesdiedrich:
@spyros The following integration test is failing and I think it has to be corrected:
`Simulation publishes intermediate and final results`
This is the step that fails is: 
```
@then('intermediate results are transmitted on every slot and final results once')
def interm_res_report(context):
    # Add an extra result for the start of the simulation.
    # Add another extra result for: initializing -> running
    assert context.interm_results_count == 14
```
I understand that `Add an extra result for the start of the simulation.` is triggered [here](https://github.com/gridsingularity/gsy-e/blob/a89cf46124942a45ede542cdbf63327263b43410/src/gsy_e/gsy_e_core/simulation/simulation.py#L162) .
And that `Add another extra result for: initializing -> running` is triggered [here](https://github.com/gridsingularity/gsy-e/blob/a89cf46124942a45ede542cdbf63327263b43410/src/gsy_e/gsy_e_core/simulation/simulation.py#L116).
However, in the integration test, the simulation is already inititated in a step before and the latter is not counting into the number of transmitted results, because the `simulation._results.kafka_connection.is_enabled()` is not mocked yet.
TLDR; The expected count should be 13 (1 for the start of the simulation and 12 for the 12 simulated market slots).
And this should be also in the current state where we send the results an additional time at the end of the simulation ([here](https://github.com/gridsingularity/gsy-e/blob/a89cf46124942a45ede542cdbf63327263b43410/src/gsy_e/gsy_e_core/simulation/simulation.py#L256)).

spyrostz (approved):
LGTM. I am not 100% convinced or the current_market -> spot_market rename, however the former name was also not clear therefore I do not have a strong opinion on it.

spyrostz:
> @spyros The following integration test is failing and I think it has to be corrected: `Simulation publishes intermediate and final results` This is the step that fails is:
> 
> ```
> @then('intermediate results are transmitted on every slot and final results once')
> def interm_res_report(context):
>     # Add an extra result for the start of the simulation.
>     # Add another extra result for: initializing -> running
>     assert context.interm_results_count == 14
> ```
> 
> I understand that `Add an extra result for the start of the simulation.` is triggered [here](https://github.com/gridsingularity/gsy-e/blob/a89cf46124942a45ede542cdbf63327263b43410/src/gsy_e/gsy_e_core/simulation/simulation.py#L162) . And that `Add another extra result for: initializing -> running` is triggered [here](https://github.com/gridsingularity/gsy-e/blob/a89cf46124942a45ede542cdbf63327263b43410/src/gsy_e/gsy_e_core/simulation/simulation.py#L116). However, in the integration test, the simulation is already inititated in a step before and the latter is not counting into the number of transmitted results, because the `simulation._results.kafka_connection.is_enabled()` is not mocked yet. TLDR; The expected count should be 13 (1 for the start of the simulation and 12 for the 12 simulated market slots). And this should be also in the current state where we send the results an additional time at the end of the simulation ([here](https://github.com/gridsingularity/gsy-e/blob/a89cf46124942a45ede542cdbf63327263b43410/src/gsy_e/gsy_e_core/simulation/simulation.py#L256)).

I think that this was clarified and fixed with commit #a079584

hannesdiedrich:
> LGTM. I am not 100% convinced or the current_market -> spot_market rename, however the former name was also not clear therefore I do not have a strong opinion on it.

Yeah, I know. I kept the old the naming towards the results and only renamed it internally.

PR #1748, opened 2024-03-19
GSYE-689: Add profile parameters for measuerment profiles that are st…

hannesdiedrich:
…ored in the profile DB to straategies

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-689


src/gsy_e/models/strategy/energy_parameters/load.py
```diff
@@ -216,6 +222,14 @@ def update_energy_requirement(self, time_slot):
         self.state.set_desired_energy(load_energy_kwh * 1000, time_slot, overwrite=False)
         self.state.update_total_demanded_energy(time_slot)
 
+    def set_energy_measurement_kWh(self, time_slot: DateTime) -> None:
+        if gsy_e.constants.CONNECT_TO_PROFILES_DB:
+            measurement_from_profile_kWh = self.measurement_profile.profile.get(time_slot)
+            if measurement_from_profile_kWh is not None:
+                self.state.set_energy_measurement_kWh(measurement_from_profile_kWh, time_slot)
```
spyrostz, line 229: I have concerns about this actually. I thought that the energy parameters would be the place to decide how to deal with missing datapoints, and I would prefer that the lack of measurements is set to a reasonable default. If we omit updating the state here, the simulation results service would have to deal with the gap/missing entry in the measurement profiles. 
Better to discuss this offline as well, but I see 2 options here:
- leave it as is, and have the simulation results service / library deal with the missing datapoints
- use 0 as a default, in order to not have gaps 
- a combination of 1 and 2, using None as a default value, and then have the simulation results service handle the None values by omitting them from profiles

src/gsy_e/models/strategy/profile.py
```diff
@@ -54,3 +91,12 @@ def read_or_rotate_profiles(self, reconfigure=False):
...
             profile=profile,
             profile_uuid=self.input_profile_uuid)
+
+
+def profile_factory(input_profile=None, input_profile_uuid=None,
+                    input_energy_rate=None,
+                    profile_type: InputProfileTypes = None) -> EnergyProfileBase:
+    """Return correct profile handling class for input parameters."""
+    return (EmptyProfile(profile_type)
+            if input_profile == input_profile_uuid == input_energy_rate is None
```
spyrostz, line 101: I think that this is missing the CLI case, where the input profile is non-None, but the profile_uuid is None. Might be mistaken though, so if you could check whether gsy-e CLI works as expected with input profiles it would be great. Thanks!

spyrostz (commented):
2 comments, LGTM once resolved. I am tempted to suggest to add some more test coverage, but leave it be for now for 2 reasons:
- we do not use the measurement profiles yet
- even if we use them, there is a chance that we will e2e-test them, which will make the unit tests redundant.

src/gsy_e/models/strategy/energy_parameters/load.py
hannesdiedrich: Thanks for this! I would vote for your third option, the combination of both.

src/gsy_e/models/strategy/profile.py
hannesdiedrich: I will test, however, if the `input_profile` is not None, then it is good to select the `EnergyProfile` class, it deals with reading the in-setup-profiles too. 
Or did I get your concern wrong?

src/gsy_e/gsy_e_core/simulation/time_manager.py
```diff
@@ -192,6 +192,18 @@ def handle_slowdown_and_realtime_scm(
...
         self.slot_time_counter = int(time())
 
+    @staticmethod
+    def get_start_time(config: "SimulationConfig") -> DateTime:
+        """Return the start tim of the simulation."""
+        if gsy_e.constants.RUN_IN_REALTIME:
+            today = pendulum.today(tz=TIME_ZONE)
+            seconds_since_midnight = time() - today.int_timestamp
+            slot_no = int(seconds_since_midnight // config.slot_length.seconds) + 1
+            start_time = config.start_date + duration(seconds=slot_no*config.slot_length.seconds)
```
spyrostz, line 202: I am fairly certain that we discussed this, however why not adding support for the SCM days of delay for Canary Networks? There is a chance that it is already included in the `config.start_date` however I will see it later on the PR if so.

hannesdiedrich: The problem is, that here we are in the init of the simulation, where the "slot resumed" is not calculated yet. Here we like to buffer the profiles and only want to start at 12:00 on the day the CN was started then and not on the config.start_date, that could be e.g. 3 month in the past.

spyrostz: Got it, I just checked the callers of this method and indeed this is not called elsewhere so all good. Only thing I suggest though is to name this `get_start_time_on_init` instead, in order to avoid confusion when trying to use this method in the future. What do you think?

spyrostz (approved):
Minor comments, LGTM already

src/gsy_e/gsy_e_core/simulation/time_manager.py
hannesdiedrich: Agreed, let me change this.

PR #1756, opened 2024-04-30
Feature/gsye 722

hannesdiedrich:

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-722

### DIFF
--- a/src/gsy_e/models/area/scm_manager.py
+++ b/src/gsy_e/models/area/scm_manager.py
@@ -312,7 +312,7 @@ class AreaEnergyBills:  # pylint: disable=too-many-instance-attributes
         output_dict = asdict(self)
         del output_dict["energy_rates"]
         output_dict.update({
-            **self.energy_rates.area_fees.prices_as_dict(),
+            "fees": self.energy_rates.area_fees.prices_as_dict(),
             "savings": self.savings,
             "savings_percent": self.savings_percent,
             "energy_benchmark": self.energy_benchmark,
--- a/src/gsy_e/models/area/scm_manager.py
+++ b/src/gsy_e/models/area/scm_manager.py
@@ -8,6 +8,7 @@ from uuid import uuid4
 from gsy_framework.constants_limits import (
     ConstSettings, GlobalConfig, is_no_community_self_consumption)
 from gsy_framework.data_classes import Trade, TraderDetails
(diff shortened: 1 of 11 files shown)


src/gsy_e/models/area/scm_manager.py
```diff
@@ -448,97 +526,94 @@ def calculate_community_after_meter_data(self):
...
             if self._intracommunity_base_rate_eur is None
             else self._intracommunity_base_rate_eur)
-        market_maker_rate_decreased_fees = (
-                intracommunity_base_rate_eur
-                + grid_fees * (1.0 - self._grid_fees_reduction)
-                + taxes_surcharges)
-        market_maker_rate_normal_fees = market_maker_rate + grid_fees + taxes_surcharges
 
+        area_fees = AreaFees(
+            grid_fee=home_data.fee_properties.get(
```
spyrostz, line 539: I am tempted to suggest to change the fee_properties from a Dict to a separate class, in order to be able to provide helper methods that generate these dicts, however feel free to omit it, this will work for now.

spyrostz (approved):
LGTM

src/gsy_e/models/area/scm_manager.py
hannesdiedrich: Let me do this actually, we can also omit the validation on the Coefficient area then (or simplify it).

hannesdiedrich: Done: [88dd3ae](https://github.com/gridsingularity/gsy-e/pull/1756/commits/88dd3ae8a0bfed15b1ebaef3f96343f6158add2e)

I also took the liberty to move all dataclasses into its own module for SCM in order to reduce the length of the scm_manager module and to not have to scroll so much when working inside of the module, even though this will break the git blame chain of most of the scm dataclasses: https://github.com/gridsingularity/gsy-e/pull/1756/commits/e290d75c2ad318d3afd067f9549214b697acf600
What do you think?

spyrostz (approved):
LGTM

src/gsy_e/models/area/scm_manager.py
spyrostz: You did well and I was thinking about that too in the past, the module was already large, thanks!

PR #1757, opened 2024-05-07
GSYE-724: Configure hours of delay that was previously a constant

BigTava:

## Proposed changes

-


src/gsy_e/gsy_e_core/rq_job_handler.py
```diff
@@ -160,6 +160,8 @@ def _configure_constants_constsettings(
     ConstSettings.SCMSettings.MARKET_ALGORITHM = CoefficientAlgorithm(
         settings.get("scm_coefficient_algorithm", 1)).value
 
+    ConstSettings.SCMSettings.HOURS_OF_DELAY = settings.get("scm_cn_hours_of_delay", 72)
```
hannesdiedrich, line 163: Why is this needed here?

hannesdiedrich (commented):
One question, once addressed, LGTM

src/gsy_e/gsy_e_core/rq_job_handler.py
spyrostz: Same as Hannes, I also believe that this is no longer required. Probably only useful if we want to somehow access this variable globally, however I do not foresee the need for it.

spyrostz (commented):
Same comment as Hannes, LGTM once resolved.

src/gsy_e/gsy_e_core/rq_job_handler.py
BigTava: Exactly. Thanks

spyrostz (approved):
LGTM

PR #1765, opened 2024-05-17
Bump requirements

spyrostz:

## Proposed changes

-

### DIFF
--- a/fabfile.py
+++ b/fabfile.py
@@ -90,6 +90,9 @@ def compile_requirements(_ctx, upgrade="", package=None):
     _pre_check()
     upgrade = upgrade.lower() in {"true", "upgrade", "1", "yes", "up"}
     print("Updating requirements")
+    cnx.local(f"rm -rf {REQ_DIR / 'base.txt'}")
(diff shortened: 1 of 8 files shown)


fabfile.py
```diff
@@ -90,6 +90,9 @@ def compile_requirements(_ctx, upgrade="", package=None):
     _pre_check()
     upgrade = upgrade.lower() in {"true", "upgrade", "1", "yes", "up"}
     print("Updating requirements")
+    cnx.local(f"rm -rf {REQ_DIR / 'base.txt'}")
```
hannesdiedrich, line 93: Can you please explain?

hannesdiedrich (approved):
One question, LGTM already

fabfile.py
spyrostz: This is what we talked about in our morning meeting. If we run pip-compile without having deleted the .txt files, the existing content of the .txt files will be taken into account when updating the requirements, thus not allowing updates in major versions. If we delete the files before running pip-compile though, the compiler will not be able to see the generated files and will update the versions as much as possible.

PR #1774, opened 2024-07-24
GSYE-736: Add imported_exported_energy to json exports

hannesdiedrich:

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-736

### DIFF
--- a/src/gsy_e/gsy_e_core/export.py
+++ b/src/gsy_e/gsy_e_core/export.py
@@ -15,6 +15,7 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+
 import csv
 import json
 import logging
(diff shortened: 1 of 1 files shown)


src/gsy_e/gsy_e_core/export.py
```diff
@@ -69,7 +81,8 @@
     "random_seed": "random_seed",
     "simulation_state": "simulation_state",
     "status": "status",
-    "trade_profile": "trade_profile"
+    "trade_profile": "trade_profile",
+    "imported_exported_energy": "imported_exported_energy",
```
hannesdiedrich, line 85: This is the actual change (rest by black)

spyrostz (approved):
LGTM. What I am missing from the series of PRs is a plot that shows the aggregated imported / exported energy from each home when running gsy-e locally. Nevertheless, it can be added on the project simulations and not necessarily here.

PR #1775, opened 2024-07-25
Feature/gsye 745

spyrostz:

## Proposed changes

-


src/gsy_e/models/strategy/virtual_heatpump.py
```diff
@@ -72,24 +55,39 @@ def __init__(
...
-            tank_parameters=tank_parameters,
-            water_supply_temp_C_profile=water_supply_temp_C_profile,
-            water_supply_temp_C_profile_uuid=water_supply_temp_C_profile_uuid,
-            water_return_temp_C_profile=water_return_temp_C_profile,
-            water_return_temp_C_profile_uuid=water_return_temp_C_profile_uuid,
-            dh_water_flow_m3_profile=dh_water_flow_m3_profile,
-            dh_water_flow_m3_profile_uuid=dh_water_flow_m3_profile_uuid,
-        )
+        # Repeat the validation for each of the tanks in order to conform to the validate method.
+        # TODO: Multiple tank validator will need to be created.
```
hannesdiedrich, line 59: IMO it is fine to loop here and keep as is.

src/gsy_e/models/strategy/energy_parameters/virtual_heatpump_solver.py
```diff
@@ -0,0 +1,214 @@
...
+
+    def __str__(self):
+        return (
+            f"Tank Volume: {self.tank_volume_L} Current Temp: {self.current_storage_temp_C}"
+            f"Target Temp: {self.target_storage_temp_C} "
+            f"Temp Differential: {self.temp_differential_per_sec}"
+        )
+
+
+class HeatpumpStorageEnergySolver:
```
hannesdiedrich, line 56: Minor, but I would like to show that this is only for the virtual HP:
```suggestion
class VirtualHeatpumpStorageEnergySolver:
```

src/gsy_e/models/strategy/energy_parameters/heat_pump_tank.py
```diff
@@ -0,0 +1,311 @@
...
+            # unmatched heat demand event in order to be calculated and tracked.
+            self._state.set_temp_decrease_K(time_slot, 0.0)
+            return True
+        assert temp_decrease_C <= 0.0
+        self._state.set_temp_decrease_K(time_slot, abs(temp_decrease_C))
+        return False
+
+    def get_max_energy_consumption(self, cop: float, time_slot: DateTime):
+        """Calculate max energy consumption that a heatpump with provided COP can consume."""
+        max_desired_temp_diff = (
```
hannesdiedrich, line 96: Here I find "desired" misleading. AFAIU, this method returns the maximal energy consumption regarding the temperature limits. If it is desired to read the max temp is not clear at this point right?
How about: `max_temp_diff` or `max_possible_temp_diff`?

src/gsy_e/models/strategy/energy_parameters/heat_pump_tank.py
```diff
@@ -0,0 +1,311 @@
...
+    """Manage the operation of heating and extracting temperature from multiple tanks."""
+
+    def __init__(self, tank_parameters: List[TankParameters]):
+        self._tanks_energy_parameters = [
+            TankEnergyParameters(tank, GlobalConfig.slot_length) for tank in tank_parameters
+        ]
+
+    def increase_tanks_temp_from_heat_energy(self, heat_energy: float, time_slot: DateTime):
+        """Increase the temperature of the water tanks with the provided heat energy."""
+        # Split heat energy equally across tanks
```
hannesdiedrich, line 189: Nothing to do now, but we might implement a mechanism to distribute the produced heat differently for different tanks, e.g. with coefficients for each tank (the sum over all coefficients is 1).
But only if this is a requirement of course.

src/gsy_e/models/strategy/energy_parameters/heat_pump_tank.py
```diff
@@ -0,0 +1,311 @@
...
+        Returns true if there is unmatched heat demand in any of the tanks.
+        """
+        heat_energy_per_tank = heat_energy / len(self._tanks_energy_parameters)
+        unmatched_heat_demand = [
+            tank.decrease_tank_temp_vhp(heat_energy_per_tank, time_slot)
+            for tank in self._tanks_energy_parameters
+        ]
+        # TODO: Implemented the current behavior, which is wrong because setting the current tank
+        # temp to target tank temp ignores the energy that is already consumed from the heatpump
+        # and the corresponding temp increase.
```
hannesdiedrich, line 214: Isn't "energy that is already consumed from the heatpump" dealt with [later on](https://github.com/gridsingularity/gsy-e/pull/1775/files#diff-2f2237c0fd6b6d25a18ddfe4800b4eec7b658c686d250fcbd9404cab7764beb4R176) ?
IMO in this method here we should not deal with the increase of the storage temp because of the name of the method.

hannesdiedrich (commented):
I raise my head to you for dealing with this.
I had a close look, also into the equations in the solver. However, I may have not seen all the corners. Tests indeed would be great and I am happy to add them later. 
I have left some comments and questions. And I have a general one:
How do you plan to incorporate the multiple tank functionality into the "traditional" HP strategy ? As we discussed, we might only use that one for the research simulation.

src/gsy_e/models/strategy/virtual_heatpump.py
spyrostz: Thanks! Will remove the TODO in this case since you also agree

src/gsy_e/models/strategy/energy_parameters/virtual_heatpump_solver.py
spyrostz: Agreed and fixing, thanks

src/gsy_e/models/strategy/energy_parameters/heat_pump_tank.py
spyrostz: I was also thinking about that, and I also thought about equally simplistic alternatives, e.g. distribute heat proportionally to the tank size. However, heat share across tanks depends largely on the heat exchanger and plumbing / connections between the heatpump and each water tank. Since we omitted simulating even the heat exchanger though, I decided to skip it altogether and share the heat proportionally across tanks. 
If it becomes a requirement though you are right, this definitely qualifies a new mechanism in a separate class that decides on the distribution of heat amongst the water tanks.

src/gsy_e/models/strategy/energy_parameters/heat_pump_tank.py
spyrostz: Agreed, this is a remnant from refactoring. Fixing

spyrostz:
> I raise my hat to you for dealing with this. I had a close look, also into the equations in the solver. However, I may have not seen all the corners. Tests indeed would be great and I am happy to add them later. I have left some comments and questions. And I have a general one: How do you plan to incorporate the multiple tank functionality into the "traditional" HP strategy ? As we discussed, we might only use that one for the research simulation.

Thanks for the review of this long PR! WRT the comment for the traditional HP strategy, I agree and I already adapted it to work with multiple water tanks.

src/gsy_e/models/strategy/energy_parameters/heat_pump_tank.py
spyrostz: I was wrong when I left this comment, I checked again and indeed the consumed energy is dealt with in the calculation of the unmatched heat demand. Also:
> we should not deal with the increase of the storage temp
this method does not handle that, it only saves the decrease of the temp, and reports if the temp decrease brings the water tank at a state lower than the min temp. So I think that we agree on all, deleting the comment and thanks for pointing it out!

hannesdiedrich (approved):
Thanks! LGTM

PR #1795, opened 2024-09-18
GSYE-767: Enable passing profiles to OrderUpdaterParameters

hannesdiedrich:

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/strategy/order_updater.py
+++ b/src/gsy_e/models/strategy/order_updater.py
@@ -1,6 +1,9 @@
 from dataclasses import dataclass
-from typing import List
+from typing import Optional, Union, List
 
+from gsy_framework.constants_limits import ConstSettings
+from gsy_framework.exceptions import GSyException
+from gsy_framework.utils import convert_pendulum_to_str_in_dict
 from pendulum import duration, DateTime, Duration
 
 from gsy_e.models.market import MarketSlotParams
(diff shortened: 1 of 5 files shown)


src/gsy_e/models/strategy/order_updater.py
```diff
@@ -12,36 +15,89 @@ class OrderUpdaterParameters:
...
         }
 
+    def get_update_interval(self):
+        """Return update_interval. If not set, return default value."""
+        if self.update_interval is None:
+            return duration(minutes=ConstSettings.GeneralSettings.DEFAULT_UPDATE_INTERVAL)
+        return self.update_interval
+
+    @staticmethod
+    def _get_default_value_initial_rate(_time_slot: DateTime):
```
BigTava, line 46: Hmmm, it feels we can simplify here no and just return 0 instead of calling this function? Or the idea is to have a default value for every time_slot?

BigTava (approved):
LG, but left 1 comment

src/gsy_e/models/strategy/order_updater.py
hannesdiedrich: Please have a look at the `HeatPumpOrderUpdaterParameters` that inherits from the `OrderUpdaterParameters`. This is to inject a different default value to every implementation of `OrderUpdaterParameters`. Here we do not need a specific default value. Does it make sense?

BigTava: I see we are overriding the function. Looks fine then

PR #1796, opened 2024-09-19
GSYE-568: Add create_hierarchy_stats to SimulationEndpointBuffer

hannesdiedrich:

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -15,12 +15,17 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+
 import logging
 from collections import defaultdict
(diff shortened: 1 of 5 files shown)


src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -170,9 +188,34 @@ def _create_results_validator(self):
...
     @staticmethod
-    def _create_endpoint_buffer(should_export_plots):
+    def _create_results_handler(should_export_plots):
         return ResultsHandler(should_export_plots)
 
+    def create_hierarchy_stats(self, area: "AreaBase"):
+        """Calculate hierarchy related statistics."""
+        hierarchy_self_consumption_list = {}
+        self._calc_hierarchy_stats(area, 0, hierarchy_self_consumption_list)
+        for level, consumption_list in hierarchy_self_consumption_list.items():
```
spyrostz, line 198: The KPI value is not the intended one here. We actually want to replicate this:
image

Therefore I think that this can be done a bit more elaborate here. Let's discuss it on Slack to settle. I will also add the screenshot in the ticket in order to have it available.

PR #1798, opened 2024-09-23
Feature/gsye 766

hannesdiedrich:

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-766

### DIFF
--- a/src/gsy_e/gsy_e_core/non_p2p_handler.py
+++ b/src/gsy_e/gsy_e_core/non_p2p_handler.py
@@ -1,4 +1,5 @@
 from gsy_framework.constants_limits import GlobalConfig
+from gsy_framework.exceptions import GSyException
 
 
 class NonP2PHandler:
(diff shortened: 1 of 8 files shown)


BigTava (dismissed):
Forgot to submit my review here. LGTM once print removed.

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -237,6 +238,8 @@ def _structure_results_from_area_object(target_area: "AreaBase") -> Dict:
         area_dict = {}
         area_dict["name"] = target_area.name
         area_dict["uuid"] = target_area.uuid
+        if target_area.is_home_area and gsy_e.constants.RUN_IN_NON_P2P_MODE:
+            area_dict["non_p2p"] = True
```
spyrostz, line 242: This is the only think that raises my eyebrow in this list of PRs actually. Not a problem since it is "only" in the results, however Not certain, however it might be better if we added a global constant as a part of the results. The disadvantage of this is that the detection of "home areas" will need to happen in gsy-framework, thus being slightly more complex than here. Thus, please ignore this suggestion and if we find the need for additional logic in the results service in the future, we can rethink about this approach.

src/gsy_e/gsy_e_core/non_p2p_handler.py
```diff
@@ -1,19 +1,35 @@
...
 
 
 class NonP2PHandler:
     """Handles non-p2p case"""
 
     def __init__(self, scenario: dict):
         self.non_p2p_scenario = scenario
+        self._energy_sell_rate = 0.0
+        self._energy_buy_rate = 0.0
+        self._get_energy_rates_from_infinite_bus(scenario)
```
spyrostz, line 12: For now this is good. However, keep in mind that in the future we might need to set different energy rates per home. We will deal with it when we need it though, nothing to do for now.

spyrostz (dismissed):
LGTM

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
hannesdiedrich: Yeah, I was not 100% happy with this implementation either. But you suggestion also does not convince me actually. I will take your offer to ignore it. Let's resume this discussion once we stumble over issues with this, ok?

src/gsy_e/gsy_e_core/non_p2p_handler.py
hannesdiedrich: Yes, having different rates per home will lead to more changes anyway. Let's deal with this then. Thanks for pointing out though.

PR #1799, opened 2024-09-26
GSYE-645: Add loss calculation to storage state

hannesdiedrich:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-645

### DIFF
--- a/src/gsy_e/models/strategy/__init__.py
+++ b/src/gsy_e/models/strategy/__init__.py
@@ -516,8 +516,7 @@ class BaseStrategy(EventMixin, AreaBehaviorBase, ABC):
     def _create_settlement_market_strategy(cls):
         return SettlementMarketStrategyInterface()
 
-    @classmethod
-    def _create_future_market_strategy(cls):
+    def _create_future_market_strategy(self):
         return FutureMarketStrategyInterface()
 
     def energy_traded(self, market_id: str, time_slot: DateTime = None) -> float:
--- a/src/gsy_e/models/strategy/__init__.py
+++ b/src/gsy_e/models/strategy/__init__.py
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
 """
 Copyright 2018 Grid Singularity
 This file is part of Grid Singularity Exchange.
@@ -15,6 +16,7 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with this program.  If not, see <http://www.gnu.org/licenses/>.
 """
+
 import json
 import logging
 import sys
@@ -60,6 +62,7 @@ INF_ENERGY = int(sys.maxsize)
 @dataclass
 class AcceptOfferParameters:
     """Parameters for the accept_offer MarketStrategyConnectionAdapter methods"""
+
     market: Union["OneSidedMarket", str]
     offer: Offer
     buyer: TraderDetails
@@ -68,22 +71,27 @@ class AcceptOfferParameters:
 
     def to_dict(self) -> dict:
         """Convert dataclass to dict in order to be able to send these arguments via Redis."""
-        return {"offer_or_id": self.offer.to_json_string(),
-                "buyer": self.buyer.serializable_dict(),
-                "energy": self.energy,
-                "trade_bid_info": self.trade_bid_info.serializable_dict()}
+        return {
+            "offer_or_id": self.offer.to_json_string(),
+            "buyer": self.buyer.serializable_dict(),
+            "energy": self.energy,
+            "trade_bid_info": self.trade_bid_info.serializable_dict(),
+        }
 
     def accept_offer_using_market_object(self) -> Trade:
         """Calls accept offer on the market object that is contained in the dataclass,
-         using the arguments from the other dataclass members"""
+        using the arguments from the other dataclass members"""
         return self.market.accept_offer(
-            offer_or_id=self.offer, buyer=self.buyer,
+            offer_or_id=self.offer,
+            buyer=self.buyer,
             energy=self.energy,
-            trade_bid_info=self.trade_bid_info)
+            trade_bid_info=self.trade_bid_info,
+        )
 
 
 class _TradeLookerUpper:
     """Find trades that concern the strategy from the selected market"""
+
     def __init__(self, owner_name: str):
         self.owner_name = owner_name
 
@@ -94,8 +102,9 @@ class _TradeLookerUpper:
                 yield trade
 
 
-def market_strategy_connection_adapter_factory() -> Union["MarketStrategyConnectionAdapter",
-                                                          "MarketStrategyConnectionRedisAdapter"]:
+def market_strategy_connection_adapter_factory() -> (
+    Union["MarketStrategyConnectionAdapter", "MarketStrategyConnectionRedisAdapter"]
+):
     """
     Return an object of either MarketStrategyConnectionRedisAdapter or
     MarketStrategyConnectionAdapter, depending on whether the flag EVENT_DISPATCHING_VIA_REDIS is
@@ -114,6 +123,7 @@ class MarketStrategyConnectionRedisAdapter:
     the class methods. Useful for markets that are run in a remote service, different from
     the strategy.
     """
+
     def __init__(self):
         self.redis = BlockingCommunicator()
         self._trade_buffer: Optional[Trade] = None
@@ -132,8 +142,7 @@ class MarketStrategyConnectionRedisAdapter:
 
         """
         market_id = market.id if not isinstance(market, str) else market
-        self._send_events_to_market("OFFER", market_id, offer_args,
-                                    self._redis_offer_response)
+        self._send_events_to_market("OFFER", market_id, offer_args, self._redis_offer_response)
         offer = self._offer_buffer
         assert offer is not None
         self._offer_buffer = None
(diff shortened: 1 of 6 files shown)


src/gsy_e/models/strategy/__init__.py
```diff
@@ -1,3 +1,4 @@
+# pylint: disable=too-many-lines
```
hannesdiedrich, line 1: This module has only be changed by Black because I fixed one method interface. You can ignore reviewing this. Thanks

BigTava (dismissed):
Well, I am bit skeptic that charging and discharging losses are fixed values and not percentage over the charging and discharging power. But the logic seems good ! Left minor comments but looks good

src/gsy_e/models/strategy/state/storage_state.py
```diff
@@ -371,14 +438,19 @@ def register_energy_from_one_sided_market_accept_offer(
...
+        time_slot: DateTime,
+        energy_origin: ESSEnergyOrigin = ESSEnergyOrigin.UNKNOWN,
+    ):
         """Register energy from a bid trade event."""
         assert energy >= -FLOATING_POINT_TOLERANCE
         self.pledged_buy_kWh[time_slot] += energy
         self.offered_buy_kWh[time_slot] -= energy
         self._track_energy_bought_type(energy, energy_origin)
         self._clamp_energy_to_buy_kWh([time_slot])
+        # if self.charging_loss_per_kWh > 0:
```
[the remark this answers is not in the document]
hannesdiedrich: Of course, let me remove

hannesdiedrich:
> Well, I am bit skeptic that charging and discharging losses are fixed values and not percentage over the charging and discharging power.

They are indeed implemented as percentages (have a look here: https://github.com/gridsingularity/gsy-e/pull/1799/files#diff-4581ce8152beebe7aa76014b4a741adac802ef0fcd8ac7a578d7ec4aabf39dfbR362-R363) 
But you are right, the names should reflect that. Will fix.

src/gsy_e/models/strategy/state/storage_state.py
```diff
@@ -43,16 +49,28 @@ class ESSEnergyOrigin(Enum):
 EnergyOrigin = namedtuple("EnergyOrigin", ("origin", "value"))
 
 
+@dataclass
+class StorageLosses:
+    """Container for all loss related Storage settings."""
+
+    charging_loss_percent: float = 0.0
+    discharging_loss_percent: float = 0.0
+    self_discharge_per_day_kWh: float = 0.0
```
spyrostz, line 58: All good, however the unit is striking as a bit odd here. Did you find it in any spec sheet of a battery? Reason I am asking is because I have mostly seen this unit being the percentage of the total capacity of the battery, e.g. https://www.batterypowertips.com/why-self-discharge-is-important-in-batteries/ or more specifically here https://www.batterypowertips.com/wp-content/uploads/2023/02/Why-is-self-discharge-important-in-batteries-Figure-1.jpeg .

src/gsy_e/models/strategy/state/storage_state.py
```diff
@@ -43,16 +49,28 @@ class ESSEnergyOrigin(Enum):
...
-                 min_allowed_soc=StorageSettings.MIN_ALLOWED_SOC):
+    def __init__(
+        self,
+        initial_soc=StorageSettings.MIN_ALLOWED_SOC,
+        initial_energy_origin=ESSEnergyOrigin.EXTERNAL,
+        capacity=StorageSettings.CAPACITY,
+        max_abs_battery_power_kW=StorageSettings.MAX_ABS_POWER,
+        min_allowed_soc=StorageSettings.MIN_ALLOWED_SOC,
+        losses: Optional[StorageLosses] = StorageLosses(),
+    ):
```
spyrostz, line 73: Minor, but since it is `Optional` I would prefer `None` as default value:
```suggestion
        losses: Optional[StorageLosses] = None,
    ):
        if not losses:
            losses = StorageLosses()
```

src/gsy_e/models/strategy/state/storage_state.py
```diff
@@ -304,6 +357,17 @@ def market_cycle(self, past_time_slot, current_time_slot: DateTime,
...
                 self.time_series_ess_share[past_time_slot][energy_type.origin] += energy_type.value
 
+    def _apply_losses_at_market_cycle(self, sold_energy_kWh: float, bought_energy_kWh: float):
+
+        charging_loss_kWh = bought_energy_kWh * self.losses.charging_loss_percent
+        discharging_loss_kWh = sold_energy_kWh * self.losses.discharging_loss_percent
+        self_discharging_kWh = (
+            self.losses.self_discharge_per_day_kWh * GlobalConfig.slot_length.total_days()
+        )
+        total_loss_kWh = charging_loss_kWh + discharging_loss_kWh + self_discharging_kWh
```
spyrostz, line 367: Admittedly an overkill, but do you think that it makes sense to also keep the total_loss_kWh in the results for debugging purposes? On second thought, better to omit it and add it if necessary in the future.

src/gsy_e/models/strategy/storage.py
```diff
@@ -67,75 +69,103 @@ def serialize(self):
...
+            max_abs_battery_power_kW=max_abs_battery_power_kW,
+            min_allowed_soc=min_allowed_soc,
+            losses=losses,
+        )
         self.cap_price_strategy = cap_price_strategy
         self.balancing_energy_ratio = BalancingRatio(*balancing_energy_ratio)
 
+    @staticmethod
+    def deserialize_args(constructor_args: Dict) -> Dict:
+        """Deserialize the constructor arguments for the HeatPump strategy."""
```
spyrostz, line 160: Minor
```suggestion
        """Deserialize the constructor arguments for the Storage strategy."""
```

src/gsy_e/models/strategy/storage.py
```diff
@@ -67,75 +69,103 @@ def serialize(self):
...
+        initial_selling_rate: Union[float, dict] = StorageSettings.SELLING_RATE_RANGE.initial,
+        final_selling_rate: Union[float, dict] = StorageSettings.SELLING_RATE_RANGE.final,
+        initial_buying_rate: Union[float, dict] = StorageSettings.BUYING_RATE_RANGE.initial,
+        final_buying_rate: Union[float, dict] = StorageSettings.BUYING_RATE_RANGE.final,
+        fit_to_limit=True,
+        energy_rate_increase_per_update=None,
         energy_rate_decrease_per_update=None,
         update_interval=None,
         initial_energy_origin: Enum = ESSEnergyOrigin.EXTERNAL,
+        losses: Optional[StorageLosses] = StorageLosses(),
```
spyrostz, line 87: Can be `None` here too AFAIS.

src/gsy_e/models/strategy/storage.py
```diff
@@ -67,75 +69,103 @@ def serialize(self):
...
+            capacity=battery_capacity_kWh,
+            max_abs_battery_power_kW=max_abs_battery_power_kW,
+            min_allowed_soc=min_allowed_soc,
+            losses=losses,
+        )
         self.cap_price_strategy = cap_price_strategy
         self.balancing_energy_ratio = BalancingRatio(*balancing_energy_ratio)
 
+    @staticmethod
+    def deserialize_args(constructor_args: Dict) -> Dict:
```
spyrostz, line 159: I think that this method only needs to be implemented if we want to involve gsy-web. Since we do not this could be completely omitted. Having said that, no problem keeping it in case we need it in the future.

spyrostz (commented):
Some comments, LGTM in general, thanks!

src/gsy_e/models/strategy/state/storage_state.py
hannesdiedrich: Oh yes, it is also a percentage value. I also found the links that you shared. Let me rename. Also, I need to fix the calculation then. I don't know why I came up with the current implementation.

src/gsy_e/models/strategy/state/storage_state.py
hannesdiedrich: I think it is not an overkill and indeed interesting to know if you run storages with losses. Let me add it to the storage CSV file.

src/gsy_e/models/strategy/storage.py
```diff
@@ -67,75 +69,103 @@ def serialize(self):
...
-        StorageSettings.BUYING_RATE_RANGE.initial,
-        final_buying_rate: Union[float, dict] =
-        StorageSettings.BUYING_RATE_RANGE.final,
-        fit_to_limit=True, energy_rate_increase_per_update=None,
+        initial_selling_rate: Union[float, dict] = StorageSettings.SELLING_RATE_RANGE.initial,
+        final_selling_rate: Union[float, dict] = StorageSettings.SELLING_RATE_RANGE.final,
+        initial_buying_rate: Union[float, dict] = StorageSettings.BUYING_RATE_RANGE.initial,
+        final_buying_rate: Union[float, dict] = StorageSettings.BUYING_RATE_RANGE.final,
+        fit_to_limit=True,
+        energy_rate_increase_per_update=None,
```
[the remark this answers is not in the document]
hannesdiedrich: Definitely! IMO we should do it in the frame of https://gridsingularity.atlassian.net/browse/GSYE-769.

src/gsy_e/models/strategy/storage.py
hannesdiedrich: OK, let me remove it here, only to explicitly add it once/if we add the losses to the FE.

src/gsy_e/models/strategy/state/storage_state.py
hannesdiedrich: I did add it here: https://github.com/gridsingularity/gsy-e/pull/1799/commits/3c11349d9e08f9fef7ea69b04f20cf990fe3a7d8
However, there is a corner case for that last market slot. We only calculate the losses for the past market slot in the market_cycle event method. But after the last market slot this is not triggered, hence the losses are not calculated. I gave it some time to fix, but could not find a solution for this. As you named this an overkill, I left it like this now. Let us deal with the last market slot once we need it. OK?

spyrostz: Agreed and thanks!

spyrostz (approved):
LGTM

PR #1800, opened 2024-10-02
GSYE-568: Change hierarchy self consumption into percentage values

hannesdiedrich:

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -201,8 +201,10 @@ class SimulationEndpointBuffer:
         for level, consumption_list in hierarchy_self_consumption_list.items():
             if len(consumption_list):
                 hierarchy_self_consumption[level] = sum(consumption_list) / len(consumption_list)
-
-        if len(hierarchy_self_consumption) > 0 and next(iter(hierarchy_self_consumption)) == 0:
+        if (
+            len(hierarchy_self_consumption) > 0
+            and hierarchy_self_consumption[next(iter(hierarchy_self_consumption))] == 0
+        ):
             # early return in case there is no self consumpton at all (in order to not devide by 0)
             return
 
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -195,21 +195,18 @@ class SimulationEndpointBuffer:
 
     def create_hierarchy_stats(self, area: "AreaBase"):
         """Calculate hierarchy related statistics."""
-        hierarchy_self_consumption_list = {}
-        hierarchy_self_consumption = {}
-        self._calc_hierarchy_self_consumption(area, 0, hierarchy_self_consumption_list)
-        for level, consumption_list in hierarchy_self_consumption_list.items():
+        hierarchy_self_consumption_lists = {}
+        hierarchy_self_consumption = []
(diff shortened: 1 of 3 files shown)


src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -195,14 +196,35 @@ def _create_results_handler(should_export_plots):
...
             if len(consumption_list):
-                self.hierarchy_self_consumption[level] = sum(consumption_list) / len(
-                    consumption_list
-                )
+                hierarchy_self_consumption[level] = sum(consumption_list) / len(consumption_list)
+        if (
+            len(hierarchy_self_consumption) > 0
+            and hierarchy_self_consumption[next(iter(hierarchy_self_consumption))] == 0
+        ):
+            # early return in case there is no self consumpton at all (in order to not devide by 0)
```
spyrostz, line 208:
```suggestion
            # early return in case there is no self consumption at all (in order to not devide by 0)
```
Nit

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -195,14 +196,35 @@ def _create_results_handler(should_export_plots):
...
+                ) * 100
+            else:
+                self.hierarchy_self_consumption_percent[level] = (
+                    (self_consumption - hierarchy_self_consumption[level + 1])
+                    / hierarchy_self_consumption[0]
+                ) * 100
+        if self.hierarchy_self_consumption_percent:
+            assert isclose(
+                sum(self.hierarchy_self_consumption_percent.values()), 100
+            ), "Self consumption percentages do not sum up to 100%"
```
spyrostz, line 225: This is a great assert, it will always confirm that this feature is working as expected, thanks!

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -195,14 +196,35 @@ def _create_results_handler(should_export_plots):
...
+        if (
+            len(hierarchy_self_consumption) > 0
+            and hierarchy_self_consumption[next(iter(hierarchy_self_consumption))] == 0
+        ):
+            # early return in case there is no self consumpton at all (in order to not devide by 0)
+            return
 
-    def _calc_hierarchy_stats(self, area: "AreaBase", level: int, results: Dict):
+        for level, self_consumption in hierarchy_self_consumption.items():
+            if level + 1 not in hierarchy_self_consumption:
```
spyrostz, line 212: Not that big of an issue, since it probably makes the solution simpler, however I think that `hierarchy_self_consumption` could be a list and not a dict, since the index / position in the list could be used as the level number. Regardless if you believe it is more explicit that way leave it as is.

spyrostz (approved):
2 minor comments, LGTM

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
hannesdiedrich: https://github.com/gridsingularity/gsy-e/pull/1800/commits/87ff961819ddf20482f0f1f6fd929f622c390d18

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
hannesdiedrich: I do not mind either way, so I changed it to be a list: https://github.com/gridsingularity/gsy-e/pull/1800/commits/87ff961819ddf20482f0f1f6fd929f622c390d18

BigTava (approved):
LGTM

spyrostz (approved):
LGTM

PR #1818, opened 2024-11-29
Feature/gsye 810

hannesdiedrich:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-810
**GSY_FRAMEWORK_BRANCH**=feature/GSYE-810


tests/strategies/energy_parameters/test_tank_energy_parameters.py
```diff
@@ -28,59 +28,59 @@ def setup_method(self):
         self._datetime = datetime(2023, 1, 1, 0, 0)
 
     def test_increase_tanks_temp_from_heat_energy(self):
-        self._tanks.increase_tanks_temp_from_heat_energy(1, self._datetime)
+        self._tanks.increase_tanks_temp_from_heat_energy(5000, self._datetime)
```
hannesdiedrich, line 31: @spyrostz You need 4186 J to heat up one liter of water by 1 degree. Hence, in our cases where we talk about hundrets of liters we should change to use kJ as unit for the heat demand profile in order to not have too large numbers, for the input as well as for the output. What do you think?
If you agree, I will add a ticket for this.

src/gsy_e/constants.py
```diff
@@ -22,12 +22,13 @@
 import os
 
 from gsy_framework.constants_limits import DATE_TIME_FORMAT, DATE_TIME_UI_FORMAT, TIME_ZONE  # NOQA
-from gsy_framework.constants_limits import TIME_FORMAT, DATE_FORMAT, GlobalConfig  # NOQA
+from gsy_framework.constants_limits import (  # NOQA
+    TIME_FORMAT,
+    DATE_FORMAT,
```
spyrostz, line 27: Are these 2 and the NOQA required?

src/gsy_e/models/strategy/energy_parameters/heatpump/cop_models/cop_models.py
```diff
@@ -0,0 +1,141 @@
...
+        # Power consumption (P) calculation
+        return self._model["Pref"] * CAPFT * HEIRFT * HEIRFPLR
+
+    def calc_cop(self, source_temp_C: float, tank_temp_C: float, heat_demand_kW: float):
+        assert heat_demand_kW is not None, "heat demand should be provided"
+        if heat_demand_kW == 0:
+            return 0
+        electrical_power_kW = self._calc_power(source_temp_C, tank_temp_C, heat_demand_kW)
+        if electrical_power_kW <= 0:
+            log.debug(
```
spyrostz, line 90: I would write an error here, and also I would prepend the configuration / simulation id in case this error will be logged in deployed environment.

src/gsy_e/models/strategy/energy_parameters/heatpump/cop_models/cop_models.py
```diff
@@ -0,0 +1,141 @@
...
+                self.model_type.name,
+                round(source_temp_C, 2),
+                round(tank_temp_C, 2),
+                round(heat_demand_kW, 2),
+                round(electrical_power_kW, 2),
+            )
+            return 0
+        cop = heat_demand_kW / electrical_power_kW
+        if cop > 6:
+            log.debug(
```
spyrostz, line 103: Same here, an error log and a reference to the configuration uuid will be nice to have

src/gsy_e/constants.py
hannesdiedrich: I had to check myself again, too.
yes, we have a lot of imports in other modules that source these ones:
```python
from gsy_e.constants import TIME_ZONE
```
I am not sure whether this is good practice. Should I remove them and directly import from gsy-framework?

src/gsy_e/models/strategy/energy_parameters/heatpump/cop_models/cop_models.py
hannesdiedrich: This might clutter or logs, but OK, let's do it and see how often this happens.

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -294,20 +306,17 @@ def _calc_cop(self, time_slot: DateTime) -> float:
...
+        # 1 J = 1 W s
+        heat_demand_kW = (
+            self._heat_demand_Q_J.get_value(time_slot) / self._slot_length.total_seconds() / 1000
+            if self._heat_demand_Q_J
+            else None
+        )
+        return self._cop_model.calc_cop(
+            source_temp_C=self._source_temp_C.get_value(time_slot),
+            tank_temp_C=self._state.tanks.get_average_tank_temperature(time_slot),
+            heat_demand_kW=heat_demand_kW,
```
spyrostz, line 318: Nothing to do here, I took a long time reviewing whether this will break the existing strategies, but as far as I saw the old strategies should operate like before so all good!

spyrostz (commented):
Some comments, LGTM already

src/gsy_e/constants.py
spyrostz: I think that the best practice is to remove them from here actually and directly import them. I checked too and I think that the only reason why we had them like these were in order to use them as `gsy_e.constants.TIME_FORMAT`, however unless I do not see something, this can be replaced by a direct import. Therefore I would reecommend to take the time and remove them, and thanks!

hannesdiedrich: ok, will do.

PR #1824, opened 2025-01-09
Feature/gsye 824

hannesdiedrich:

## Proposed changes

-


src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_models.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class PCMModelBase:
+
+    def __init__(self, slot_length: duration, mass_flow_rate_kg_s: float):
+        self._mass_flow_rate = mass_flow_rate_kg_s
+        self._slot_length = slot_length
+
+    @property
+    def _soc_lut(self):
```
BigTava, line 11: Please provide a """ """ for the property meaning. What is "lut" ?

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
```diff
@@ -185,6 +189,14 @@ def _specific_labels(self):
                 "heat demand J",
             ]
         # pylint: disable=unidiomatic-typecheck
+        if type(self.area.strategy) == PCMHeatPump:
```
BigTava, line 192: Could it be a Tank class and not a HeatPump class? Thinking about HSB LL. We have a MultipleTankHeatPumpStrategy. Will we replace it one PCMHeatPump, three PCMHeatPump or a MultipleTankHeatPumpStrategy that takes three PCMTank in tank_parameters?

src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_models.py
```diff
@@ -0,0 +1,41 @@
...
+
+    @property
+    def _soc_lut_key(self) -> str:
+        return ""
+
+    def get_soc(self, current_storage_temps: list):
+        return 50
+
+
+class PCMChargeModel(PCMModelBase):
```
BigTava, line 22: Please provide brief explanation """ """

src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -340,3 +346,52 @@ def __init__(
             preferred_buying_rate,
             cop_model_type=cop_model_type,
         )
+
+
+class PCMHeatPump(HeatPumpStrategy):
+    """Heat pump with PCM tank"""
```
BigTava, line 352: It could be confusing to think PCM as a type of heat pump in our context.

src/gsy_e/setup/strategy_tests/heat_pump_pcm.py
```diff
@@ -0,0 +1,83 @@
...
+def get_setup(config):
+    area = Area(
+        "Grid",
+        [
+            Area(
+                "House 1",
+                [
+                    Area(
+                        "PCM",
+                        strategy=PCMHeatPump(
```
BigTava, line 42: It looks like the usual HeatPumpStrategy

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
hannesdiedrich: There will only be one PCM tank on HSB LL. We decided to only implement a single PCM tank and not allow multiple, because there will not be a use in the near and far future.

src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_models.py
hannesdiedrich: LUT is short for Look-Up-Table. Let me add a doc string, indeed. Thanks

src/gsy_e/models/strategy/heat_pump.py
hannesdiedrich: You have a point, but we need to distinguish between heat pumps that have water and PCM tanks, because they need different energy-parameter classes. 
I do not have a better naming in order to omit this confusion. Do you maybe?

src/gsy_e/setup/strategy_tests/heat_pump_pcm.py
hannesdiedrich: yes, the required input parameters are identical. Only the energy-paramer and tank state class is different.
Once we agree on the implementation of these, we can maybe indeed switch between PCM and water tankes via a boolean variable that is part of the tank parameters and omit having a PCMHeatpPump strategy class. For now I wanted to make it work first.

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
BigTava: Ok. So one PCM tank will replace the 3 water tanks. In that case, PCMHeatPump looks fine. But I think we should simplify the three classes (MultipleTankHeatPumpStrategy, HeatPumpStrategy, PCMHeatPump) and only have a HeatPumpStrategy that takes an array of tanks as arguments, and optional parameters in the HeatPump in case, which are validated in case a PCM tank is included

src/gsy_e/models/strategy/heat_pump.py
BigTava: Cound't we just have a MultipleTankHeatPumpStrategy with tank_parameters: an array of tanks, and optional parameters in the HeatPump, which are validated in case a PCM tank is included

src/gsy_e/setup/strategy_tests/heat_pump_pcm.py
BigTava: And very well. Is good to have something running above all. I just peaked into the code for the report purposes.

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
```diff
@@ -260,6 +272,18 @@ def _specific_row(self, slot, market):
                 round(self.area.strategy.state.heatpump.get_cop(slot), ROUND_TOLERANCE),
                 round(self.area.strategy.state.heatpump.get_heat_demand(slot), ROUND_TOLERANCE),
             ]
+        if type(self.area.strategy) == PCMHeatPump:
+            print()
```
spyrostz, line 276: Can be removed

spyrostz (dismissed):
LGTM, I agree with Tiago's comments that probably a separate TankParameters class would be more preferable as a constructor argument and would decrease the number of classes, however I would recommend to do it at a later stage due to:
1. this will break the integration with gsy-web and the existing setup files 
2. it will require careful decoupling between the tank and heatpump strategy classes in order to be sanely implemented

src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_tank.py
```diff
@@ -0,0 +1,504 @@
...
+class PCMTankParameters:
+    """Nameplate parameters of a water tank."""
+
+    min_temp_C: float = ConstSettings.HeatPumpSettings.MIN_TEMP_C
+    max_temp_C: float = ConstSettings.HeatPumpSettings.MAX_TEMP_C
+    initial_temp_C: float = ConstSettings.HeatPumpSettings.INIT_TEMP_C
+    max_capacity_kWh: float = 6.0  # todo: put default value somewhere
+
+
+class PCMTankState(StateInterface):
```
spyrostz, line 53: This breaks a bit the file hierarchy, e.g. this class should be under `state` package. Having said that we can leave it here and refactor at a later stage.

src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_models.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def get_temp_after_charging(self, current_storage_temps: list, charging_temp: float) -> list:
+        return current_storage_temps
+
+
+class PCMDischargeModel(PCMModelBase):
+
+    @property
+    def _soc_lut(self):
+        return {}
```
spyrostz, line 36: Why do you need to override these properties? Since the PCMModelBase already implements this property method, it will be anyways inherited to the child classes.

src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_tank.py
```diff
@@ -0,0 +1,504 @@
...
+    def get_soc(self, time_slot: DateTime) -> float:
+        """Return SOC level."""
+        return self._soc[time_slot]
+
+    def get_htf_temp_C(self, time_slot: DateTime) -> Optional[float]:
+        """Return mean temperature of the heat transfer fluid"""
+        storage_temps = self.get_storage_temp_C(time_slot)
+        if storage_temps is None:
+            return None
+        return mean(storage_temps[0:-2:2])
```
spyrostz, line 135: For readability, why not distinguishing the HTF and the PCM temps in different lists ?

src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_tank.py
```diff
@@ -0,0 +1,504 @@
...
+        """Get storage_temp_C for specific time_slot."""
+        return self._storage_temp_C.get(time_slot, None)
+
+    def set_soc_after_charging(self, time_slot: DateTime):
+        """Calculate and set SOC level after charging."""
+        self._soc[time_slot] = self._pcm_charge_model.get_soc(self._storage_temp_C[time_slot])
+
+    def set_soc_after_discharging(self, time_slot: DateTime):
+        """Calculate and set SOC level after discharging."""
+        self._soc[time_slot] = self._pcm_discharge_model.get_soc(self._storage_temp_C[time_slot])
```
spyrostz, line 124: These are never been called by other parts of the code, and I think that they are needed for SOC levels to be reported.

src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_models.py
hannesdiedrich: Ah yes, this is only a placeholder for the code that is closed source until now. We are overwriting those for the charging and discharging. In order to not confuse the code here, I will remove the overwriting.

src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_tank.py
hannesdiedrich: I am not very sure about this any more. Wenn calling the (dis-)charge model we will need to combine these lists again into ne (in order to not having to change in the greybox model). And then we need to return two lists again from the models side in order to dave them here in the state separately. Do you think it is worth it?

src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_tank.py
```diff
@@ -0,0 +1,504 @@
...
+
+    def update_consumed_energy(self, energy: float, time_slot: DateTime) -> None:
+        """Add provided energy value to the value in consumed_energy for the provided time_slot."""
+        self._consumed_energy[time_slot] = self._consumed_energy.get(time_slot, 0.0) + energy
+
+    def get_consumed_energy(self, time_slot: DateTime):
+        """Return consumed_energy for the provided time_slot"""
+        return self._consumed_energy.get(time_slot, 0.0)
+
+    def init_storage_temps(self, time_slot: DateTime):
```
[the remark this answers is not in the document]
hannesdiedrich: Done here [b34d973](https://github.com/gridsingularity/gsy-e/pull/1824/commits/b34d973a8a87aa65e78b8811382752f0781c4342)

src/gsy_e/models/strategy/heat_pump.py
spyrostz: This is already tracked and should be done in the context of this ticket https://gridsingularity.atlassian.net/browse/GSYE-864

src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_models.py
spyrostz: Got it and thanks for the clarification!

spyrostz (approved):
LGTM

PR #1825, opened 2025-01-12
GSYE-839: Corrected savings calculation, taking into account by inclu…

spyrostz:
…ding the savings from the revenue in addition to the savings from buying energy from community. Corrected the default grid fee reduction value to be 1, so that intracommunity trading does not include any grid fees by default.

## Proposed changes

-

### DIFF
--- a/src/gsy_e/models/area/scm_dataclasses.py
+++ b/src/gsy_e/models/area/scm_dataclasses.py
@@ -385,8 +385,6 @@ class AreaEnergyBills:  # pylint: disable=too-many-instance-attributes
             {
                 "fees": self.energy_rates.area_fees.prices_as_dict(),
                 "savings": self.savings,
-                "savings_from_sell_to_community": self.savings_from_sell_to_community,
-                "savings_from_buy_from_community": self.savings_from_buy_from_community,
                 "savings_percent": self.savings_percent,
                 "energy_benchmark": self.energy_benchmark,
                 "home_balance_kWh": self.home_balance_kWh,
--- a/src/gsy_e/models/area/scm_dataclasses.py
+++ b/src/gsy_e/models/area/scm_dataclasses.py
@@ -278,7 +278,7 @@ class AreaFees:
 
     grid_import_fee_const: float = 0.0
     grid_export_fee_const: float = 0.0
-    grid_fees_reduction: float = 0.0
+    grid_fees_reduction: float = 1.0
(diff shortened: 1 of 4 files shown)


src/gsy_e/models/area/scm_dataclasses.py
```diff
@@ -278,7 +278,7 @@ class AreaFees:
 
     grid_import_fee_const: float = 0.0
     grid_export_fee_const: float = 0.0
-    grid_fees_reduction: float = 0.0
+    grid_fees_reduction: float = 1.0
```
hannesdiedrich, line 281: 👍

src/gsy_e/models/area/scm_dataclasses.py
```diff
@@ -539,14 +557,17 @@ class AreaEnergyBillsWithoutSurplusTrade(AreaEnergyBills):
     """Dedicated AreaEnergyBills class, specific for the non-suplus trade SCM."""
```
hannesdiedrich, line 557:
```suggestion
    """Dedicated AreaEnergyBills class, specific for the non-surplus trade SCM."""
```

src/gsy_e/models/area/scm_manager.py
```diff
@@ -299,7 +300,15 @@ def community_bills(self) -> Dict:
...
 
-        return community_bills.to_dict()
+            savings_from_buy_from_community += data.savings_from_buy_from_community
+            savings_from_sell_to_community += data.savings_from_sell_to_community
+            savings += data.savings
+
+        comm_bills = community_bills.to_dict()
+        comm_bills["savings"] = savings
+        comm_bills["savings_from_buy_from_community"] = savings_from_buy_from_community
+        comm_bills["savings_from_sell_to_community"] = savings_from_sell_to_community
```
hannesdiedrich, line 310: So you decided to export all of them, even though the "savings" already include the other two for the "normal" AreaEnergyBills ?
I guess this is in order not to break the existing workflow, right? However, I do not see the last two begin used on the gsy-web. Can you please point me to them?

hannesdiedrich (approved):
LGTM, only left some minor comments

src/gsy_e/models/area/scm_manager.py
spyrostz: I was actually thinking to save these in the DB for debugging purposes, but later on I figured out that it is probably an overkill to maintain them just for debugging. Will remove these both from here and from the gsy-framework, thanks for catching this!

PR #1826, opened 2025-01-14
GSYE-835: Adapt TankEnergyParameters.get_min_energy_consumption so th…

hannesdiedrich:
…e storage temperature can drop to its mininum

## Proposed changes

-


spyrostz (approved):
LGTM, however this does not handle the unmatched heatpump demand if I recall this feature correctly. Frankly I was never too much of a fan of this feature and I do not mind skipping its implementation, just asking to have the full picture here.

hannesdiedrich:
> LGTM, however this does not handle the unmatched heatpump demand if I recall this feature correctly. Frankly I was never too much of a fan of this feature and I do not mind skipping its implementation, just asking to have the full picture here.

AFAIS, the implementation of the unmatched demand is not influenced by this change.
https://github.com/gridsingularity/gsy-e/blob/b9f5e655218e4af5c5eb635e94695a0891942f72/src/gsy_e/models/strategy/energy_parameters/heatpump/tank.py#L89
It depends on the `get_temp_increase_K` which in turn depends on the traded energy. 

In this PR we only change the value of the bid energy, so before trading.

spyrostz:
> > LGTM, however this does not handle the unmatched heatpump demand if I recall this feature correctly. Frankly I was never too much of a fan of this feature and I do not mind skipping its implementation, just asking to have the full picture here.
> 
> AFAIS, the implementation of the unmatched demand is not influenced by this change.
> 
> https://github.com/gridsingularity/gsy-e/blob/b9f5e655218e4af5c5eb635e94695a0891942f72/src/gsy_e/models/strategy/energy_parameters/heatpump/tank.py#L89
> 
> 
> It depends on the `get_temp_increase_K` which in turn depends on the traded energy.
> In this PR we only change the value of the bid energy, so before trading.

Got it and thanks for the clarification!

PR #1827, opened 2025-01-15
GSYE-842: Provide filepath to ProfilesHandler.rotate_profile in order…

hannesdiedrich:
… to be used when rotating profile provided by file

## Proposed changes

-

### DIFF
--- a/src/gsy_e/gsy_e_core/user_profile_handler.py
+++ b/src/gsy_e/gsy_e_core/user_profile_handler.py
@@ -480,7 +480,11 @@ class ProfilesHandler:
         )
 
     def rotate_profile(
-        self, profile_type: InputProfileTypes, profile, profile_uuid: str = None
+        self,
+        profile_type: InputProfileTypes,
+        profile,
+        profile_uuid: str = None,
+        input_profile_path: str = None,
     ) -> Dict[DateTime, float]:
         """Reads a new chunk of profile if the buffer does not contain the current time stamp
         Profile chunks are either generated from single values, input daily profiles or profiles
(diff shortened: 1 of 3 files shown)


BigTava (approved):
LGTM

src/gsy_e/gsy_e_core/user_profile_handler.py
```diff
@@ -480,7 +480,11 @@ def _read_new_datapoints_from_buffer_or_rotate_profile(
         )
 
     def rotate_profile(
-        self, profile_type: InputProfileTypes, profile, profile_uuid: str = None
+        self,
+        profile_type: InputProfileTypes,
+        profile,
+        profile_uuid: str = None,
+        input_profile_path: str = None,
```
spyrostz, line 487: I agree but why the extra parameter here? If I recall correctly, this is exactly the reason why we allow `str` as the type of the `profile` parameter, in order to be able to support file paths. Did you try setting the profile path to the `profile` parameter and it did not work out?

hannesdiedrich: Please have a look into the [StrategyProfile](https://github.com/gridsingularity/gsy-e/pull/1827/files#diff-be775c2f1ca4ae57dbc0db44e06d4a7793116288106c1e06e6ade6f4793f8d52L128). If the profiles was already read once, the profile will be the profile for he first week. This will be used by the read_arbitrary_profile which will re-use it, but even worse: it will convert it again into kWh. The latter resulted in very low energy demands that are close to or lower than our Floating_point_tollerance.

I am open to better solutions here.

spyrostz: I think that you can use the `input_profile` in the `StrategyProfile` class. Let me leave a possible suggestion that you can also double check if it works out.

src/gsy_e/models/strategy/strategy_profile.py
```diff
@@ -128,7 +128,10 @@ def read_or_rotate_profiles(self, reconfigure=False):
             profile = self.profile
```
spyrostz, line 129: It needs to be checked whether it violates the previous if clauses, however the solution can be similar to this, so that the profile variable is overridden if we need to rotate a profile that originates from a CSV file: 

```suggestion
if isinstance(self.input_profile, str):
    profile = self.input_profile
```

hannesdiedrich: ` if isinstance(self.input_profile, str):` will always we True for all the market slot in the simulation which will result in calling the read_arbitrary_profile inside the ProfilesHandler on every market slot, which is not what we want.

spyrostz (approved):
LGTM

PR #1839, opened 2025-02-03
GSYE-846: export carbon_emissions when imported_exported_energy is no…

BigTava:
…n empty dict

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-846

### DIFF
--- a/src/gsy_e/gsy_e_core/export.py
+++ b/src/gsy_e/gsy_e_core/export.py
@@ -35,6 +35,7 @@ from gsy_framework.data_classes import (
 )
 from gsy_framework.enums import AvailableMarketTypes, BidOfferMatchAlgoEnum
 from gsy_framework.utils import mkdir_from_str
+from gsy_framework.sim_results.carbon_emissions.results import CarbonEmissionsHandler
 from pendulum import DateTime
 
 import gsy_e.constants
@@ -93,12 +94,18 @@ class ExportAndPlot:
 
     # pylint: disable=too-many-arguments
     def __init__(
-        self, root_area: Area, path: str, subdir: str, endpoint_buffer: "SimulationEndpointBuffer"
+        self,
+        root_area: Area,
+        path: str,
+        subdir: str,
+        endpoint_buffer: "SimulationEndpointBuffer",
+        country_code: str,
     ):
         self.area = root_area
         self.endpoint_buffer = endpoint_buffer
         self.file_stats_endpoint = file_export_endpoints_factory()
         self.raw_data_subdir = None
+        self.country_code = country_code
         try:
             if path is not None:
                 path = os.path.abspath(path)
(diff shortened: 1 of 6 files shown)


src/gsy_e/gsy_e_core/export.py
```diff
@@ -124,12 +131,31 @@ def _export_json_data(self) -> None:
...
                 json.dump(value, outfile, indent=2)
 
+        if self.country_code:
+            carbon_emissions_handler = CarbonEmissionsHandler()
+
+            imported_exported_energy = json_report.get("imported_exported_energy")
+            carbon_emissions = {}
+            if imported_exported_energy != {}:
+                carbon_emissions = (
+                    carbon_emissions_handler.calculate_from_gsy_imported_exported_energy(
```
hannesdiedrich, line 150: From here, I do not see an option to provide a carbon ratio csv file, which we discussed should be possible for a cli user. 
Providing the path to the carbon ration file should probably be a cli option and saved in the simulation config. Maybe there is a better way. Let's discuss. 
In any case, there should be a validator that checks if the carbon ration file contains values for all time slots in the simulation .

hannesdiedrich (changes requested):
Please enable the cli user to provide a path to a carbon ration CSV file. Thanks a lot!

src/gsy_e/gsy_e_core/export.py
BigTava: Yes, I should have explained why I decided to keep the country code. What we are solving with the CarbonEmissionsHandler is essentially the calculation of the carbon ratio. It is very unlikely that a cli user will spend time to generate a file with the carbon ratio for every hour compatible with our interface, because that is essentially the problem we are solving for him. In addition, the calculate_from_carbon_ratio_and_imported_energy offers a simple interface to calculate the results in a post-processing step, which is what we use in research-scripts. So, with this approach we offer a simple and accessible solution to calcule the carbon footprint from non-granular and out-of-date sources. With the file path option we could offer a flexible and precise solution to calculate the carbon footprint, but most likely nobody will use it, particularly when the real sauce for retrieving precise hourly carbon ratios from Entosoy-e  is now in our private repo.

BigTava: Let me know if you agree with the explanation and if it needs further explanation. Very grateful for your attention in reviewing and discussing

spyrostz: IMO this option should be provided. I am fairly certain that we will use this feature in future project studies / simulations, in which case we will need to provide detailed CO2 emission files with 15-min / hourly resolution, in order to have substantial and detailed results. Therefore we cannot rely in monthly / yearly aggregated data, but we should give the capability to the CLI user (us, essentially) to provide CO2 emission data profiles.

spyrostz: In order to avoid doing it in the postprocessing for future projects, I would recommend to just add the carbon_ratio_file_path. In order to not complicate the structure of the solution, as a starting point I would recommend to save the timeseries data from the file in the `GlobalObjects` singleton class, by creating a new class that is instantiated there and manages only the carbon footprint file data. This class can then be imported in the CSV results files generator, and use the Carbon Emissions in order to calculate the footprint and savings.

spyrostz (approved):
LGTM

PR #1840, opened 2025-02-07
GSYE-775: Use price settings of loads and pvs for the FIT and MMR in …

hannesdiedrich:
…the homes when running in non-P2P mode

## Proposed changes

-


src/gsy_e/gsy_e_core/non_p2p_handler.py
```diff
@@ -43,15 +47,31 @@ def _get_energy_rates_from_infinite_bus(self, scenario: Area):
...
+        mmr = None
+        fit = None
+        for area in home_area.children:
+            if isinstance(area.strategy, (LoadHoursStrategy, DefinedLoadStrategy)):
+                mmr = area.strategy.bid_update.final_rate_input
+            if isinstance(area.strategy, (PVStrategy, PVUserProfileStrategy)):
+                fit = area.strategy.offer_update.final_rate_input
+
+        energy_sell_rate = self._energy_sell_rate if mmr is None else mmr
+        energy_buy_rate = self._energy_buy_rate if fit is None else fit
```
hannesdiedrich, line 60: The `mmr` and `fit` will never be None, because of the default values of the strategies. 
So, this if statement does not have an effect and the user can not globally change the MMR and FIT via the one market maker. They have to individually change the rates in the strategies.
I wonder if this is a good approach or wether we should add some setting/flag that toggles using the global market maker or not. 
What do you think @spyrostz ?

spyrostz: Hmm , this I need to check, because if I recall correctly these strategy parameters are overridden if a global market maker rate is set. Editing the comment in a bit

hannesdiedrich: We toogle using the global setting with the `use_market_maker_rate` in the strategies. But this only applies to the market maker rate. 
My point here was, that the if statement here does not have an effect, because either the strategy setting will be used, or if not set, the default value is set. If `use_market_maker_rate` is True, the global market maker will be used. 
So, I could remove these if statements and it would work the same. 
The only missing part: The user will not be able to globally set the FIT. This is OK IMO and could be handled when needed. What do you think?

spyrostz: Thanks for the clarification here, and I agree, the non-P2P should explicitly demand by the user to set up custom rates per home. If you can remove the ifs it would be great, thanks!

spyrostz (approved):
LGTM once the comment thread is fixed

src/gsy_e/gsy_e_core/non_p2p_handler.py
hannesdiedrich: @spyrostz I was about to change this, but I was seeing one corner case: Let's say that there is no Load or PV in the house, e.g. only a heat pump. Then, the global MMR and FIT should be used, right?
So, I propose to leave it as is now. Do you agree?

spyrostz: Fair point, I think that the custom rates apply also to the heat pumps and storages, despite that better to play it safe and leave it as is as you suggested.

PR #1847, opened 2025-03-18
GSYE-871: Use ProfileUuidScenarioKeyNames in profile handling

hannesdiedrich:

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-871


src/gsy_e/gsy_e_core/user_profile_handler.py
```diff
@@ -45,20 +49,20 @@ class ProfileDBConnectionException(Exception):
 
 
 PROFILE_UUID_NAMES = [
```
spyrostz, line 51: This can be directly imported from gsy-framework and completely removed from here.

spyrostz (approved):
LGTM

PR #1851, opened 2025-04-09
GSYE-864: Moved parts of the heat pump calculation from the energy pa…

spyrostz:
…rameters to the combined state class.

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-864
**GSY_FRAMEWORK_BRANCH**=feature/GSYE-864

### DIFF
--- a/src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
+++ b/src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
@@ -123,14 +123,14 @@ class CombinedHeatpumpTanksState:
 
     def get_results_dict(self, current_time_slot: DateTime) -> dict:
         """Results dict for all heatpump and tanks results."""
+        if current_time_slot is None:
+            return {}
         tanks_state = self._charger.get_tanks_results_dict(current_time_slot)
-        if isinstance(tanks_state, list):
-            return {
-                "tanks": tanks_state,
-                **self._hp_state.get_results_dict(current_time_slot),
-                "storage_temp_C": mean([tank["storage_temp_C"] for tank in tanks_state]),
-            }
-        return {**self._hp_state.get_results_dict(current_time_slot), **tanks_state}
+        return {
+            "tanks": tanks_state,
+            **self._hp_state.get_results_dict(current_time_slot),
+            "storage_temp_C": mean([tank["storage_temp_C"] for tank in tanks_state]),
+        }
 
     def get_state(self) -> Dict:
         """Return the current state of the device."""
--- a/src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
+++ b/src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
@@ -20,7 +20,7 @@ from gsy_e.models.strategy.energy_parameters.heatpump.cop_models import (
 )
 from gsy_e.models.strategy.energy_parameters.heatpump.tank import (
     TankParameters,
-    AllTanksEnergyParameters,
+    AllTanksState,
 )
 from gsy_e.models.strategy.state import HeatPumpState
 from gsy_e.models.strategy.strategy_profile import profile_factory
@@ -78,7 +78,7 @@ class CombinedHeatpumpTanksState:
     def __init__(
         self,
         hp_state: HeatPumpState,
-        tanks_state: AllTanksEnergyParameters,
+        tanks_state: AllTanksState,
         cop_model: BaseCOPModel,
         slot_length: pendulum.Duration,
         max_energy_consumption_kWh: float,
@@ -124,7 +124,7 @@ class CombinedHeatpumpTanksState:
         return self._hp_state
 
     @property
-    def tanks(self) -> AllTanksEnergyParameters:
+    def tanks(self) -> AllTanksState:
         """Exposes the tanks state."""
         return self._tanks_state
 
@@ -231,9 +231,6 @@ class CombinedHeatpumpTanksState:
             return 0
         return convert_kJ_to_kWh(Q_energy_kJ / cop)
 
-    def _last_time_slot(self, current_market_slot: DateTime) -> DateTime:
-        return current_market_slot - GlobalConfig.slot_length
-
 
 class HeatPumpEnergyParametersBase(ABC):
     """
@@ -257,7 +254,7 @@ class HeatPumpEnergyParametersBase(ABC):
             maximum_power_rating_kW * self._slot_length.total_hours()
         )
         state = HeatPumpState(self._slot_length)
-        tanks = AllTanksEnergyParameters(tank_parameters)
+        tanks = AllTanksState(tank_parameters)
         self._state = CombinedHeatpumpTanksState(
             state, tanks, cop_model, self._slot_length, self._max_energy_consumption_kWh
         )
--- a/src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
+++ b/src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
@@ -34,11 +34,14 @@ Description of physical units and parameters:
 """
 
 
+HEAT_EXCHANGER_EFFICIENCY = 1.0
+
+
 class HeatPumpEnergyParametersException(Exception):
     """Exception raised in the HeatPumpEnergyParameters"""
 
 
-class HeatCharger:
+class HeatChargerDischarger:
     """
     Modelling of the heat exchangers between:
      - the condenser side of the heat pump and the heat storage
@@ -47,28 +50,57 @@ class HeatCharger:
 
     def __init__(self, tanks):
         self.tanks = tanks
+        self._efficiency = HEAT_EXCHANGER_EFFICIENCY
 
     def get_condenser_temperature_C(self, time_slot: DateTime):
         """Get the temperature on the condenser side of the heat pump."""
-        return self.tanks.get_average_tank_temperature(time_slot)
+        return self.tanks.get_average_tank_temperature(time_slot) * self._efficiency
 
     def charge(self, heat_energy_kJ: float, time_slot: DateTime):
         """
         Increase temperature of the heat storage by the provided heat energy from the heat pump.
         """
-        self.tanks.increase_tanks_temp_from_heat_energy(heat_energy_kJ, time_slot)
+        self.tanks.increase_tanks_temp_from_heat_energy(
+            heat_energy_kJ * self._efficiency, time_slot
+        )
 
     def discharge(self, heat_energy_kJ: float, time_slot: DateTime):
         """Decrease temperature from the heat storage by the provided heat demand energy."""
-        self.tanks.decrease_tanks_temp_from_heat_energy(heat_energy_kJ, time_slot)
+        self.tanks.decrease_tanks_temp_from_heat_energy(
+            heat_energy_kJ * self._efficiency, time_slot
+        )
 
     def get_max_heat_energy_charge_kJ(self, time_slot: DateTime):
         """Get the maximum heat energy that the heat storage can accomodate."""
-        return self.tanks.get_max_heat_energy_consumption_kJ(time_slot)
+        return self.tanks.get_max_heat_energy_consumption_kJ(time_slot) / self._efficiency
 
     def get_min_heat_energy_charge_kJ(self, time_slot: DateTime):
         """Get the minimum heat energy that the heat storage can accomodate."""
-        return self.tanks.get_min_heat_energy_consumption_kJ(time_slot)
+        return self.tanks.get_min_heat_energy_consumption_kJ(time_slot) / self._efficiency
+
+    def get_tanks_results_dict(self, current_time_slot: DateTime) -> dict:
+        """Results dict for tanks results."""
+        return self.tanks.get_results_dict(current_time_slot)
+
+    def get_state(self) -> Dict:
+        """Return the current state of the charger / tanks."""
+        return self.tanks.get_state()
+
+    def restore_state(self, state_dict: Dict):
+        """Update the state of the charger using the provided dictionary."""
+        self.tanks.restore_state(state_dict)
+
+    def delete_past_state_values(self, current_time_slot: DateTime):
+        """Delete the state of the charger before the given time slot."""
+        self.tanks.delete_past_state_values(current_time_slot)
+
+    def update_tanks_temperature(self, time_slot: DateTime):
+        """
+        Update temperature of the tanks at the end of the market slot. Sets updated value to the
+        storage temperature in order to be used from the next market slot as a base for the
+        temperature calculations.
+        """
+        self.tanks.update_tanks_temperature(time_slot)
 
 
 class CombinedHeatpumpTanksState:
@@ -84,15 +116,14 @@ class CombinedHeatpumpTanksState:
         max_energy_consumption_kWh: float,
     ):
         self._hp_state = hp_state
-        self._tanks_state = tanks_state
-        self._charger = HeatCharger(self._tanks_state)
+        self._charger = HeatChargerDischarger(tanks_state)
         self._cop_model = cop_model
         self._slot_length = slot_length
         self._max_energy_consumption_kWh = max_energy_consumption_kWh
 
     def get_results_dict(self, current_time_slot: DateTime) -> dict:
         """Results dict for all heatpump and tanks results."""
-        tanks_state = self._tanks_state.get_results_dict(current_time_slot)
+        tanks_state = self._charger.get_tanks_results_dict(current_time_slot)
         if isinstance(tanks_state, list):
             return {
                 "tanks": tanks_state,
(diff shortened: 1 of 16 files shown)


src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -76,6 +121,108 @@ def tanks(self) -> AllTanksEnergyParameters:
...
+        cop = self._cop_model.calc_cop(
+            source_temp_C=source_temp_C,
+            condenser_temp_C=self._charger.get_condenser_temperature_C(time_slot),
+            heat_demand_kW=max_heat_demand_kW,
+        )
+
+        if cop == 0:
+            return 0
+
+        max_energy_consumption_kWh = max_heat_demand_kW / cop
```
hannesdiedrich, line 141: The COP does not convert between kW and kWh. It only converts between heat and electricity energy (or power). 
This is a bug AFAIU.

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -76,6 +121,108 @@ def tanks(self) -> AllTanksEnergyParameters:
...
+        cop = self._cop_model.calc_cop(source_temp_C, condenser_temp, energy_kWh)
+        return convert_kWh_to_kJ(cop * energy_kWh)
+
+    def _calc_energy_kWh_from_Q_kJ(self, time_slot: DateTime, Q_energy_kJ: float) -> float:
+        cop = self._hp_state.get_cop(time_slot)
+        if cop == 0:
+            return 0
+        return convert_kJ_to_kWh(Q_energy_kJ / self._hp_state.get_cop(time_slot))
+
+    def _last_time_slot(self, current_market_slot: DateTime) -> DateTime:
```
hannesdiedrich, line 223: I do not see this being used anywhere.

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -31,12 +32,56 @@ class HeatPumpEnergyParametersException(Exception):
...
+    """
+    Modelling of the heat exchangers between:
+     - the condenser side of the heat pump and the heat storage
+     - the heat storage and the heat demand
+    """
+
+    def __init__(self, tanks):
+        self.tanks = tanks
+
+    def get_condenser_temperature_C(self, time_slot: DateTime):
```
hannesdiedrich, line 45: I am a bit confused here. IMO the condensor temperature should not be an output of the charger, but an input for the charging process (if at all).
I also remember that we discussed this, but I do not remember what the outcome was.
Getting the average of the tank temperatures should be done in a method in the tank state, not here in the charger. Or am I missing a corner case here?

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -252,73 +394,56 @@ def _rotate_profiles(self, current_time_slot: Optional[DateTime] = None):
...
+
+    def _update_last_time_slot_data(self, time_slot: DateTime):
+        last_time_slot = self.last_time_slot(time_slot)
+        if last_time_slot not in self._source_temp_C.profile:
+            return
+        self.combined_state.increase_tank_temp_from_bought_energy(
+            last_time_slot, self._bought_energy_kWh, self._source_temp_C.profile[time_slot]
+        )
+        self._bought_energy_kWh = 0.0
+        self._state.tanks.update_tanks_temperature(last_time_slot)
```
hannesdiedrich, line 414: AFAIS, this line is doing the same as line 410. Please correct me.

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -252,73 +394,56 @@ def _rotate_profiles(self, current_time_slot: Optional[DateTime] = None):
...
+
+        self.combined_state.update_cop(
```
hannesdiedrich, line 416: I understand that you are updating the cop for the last market slot here. Why is that? Only for reporting? I am asking this, because the COP of the last_market_slot is not being used after this point, correct?

src/gsy_e/models/strategy/energy_parameters/heatpump/tank.py
hannesdiedrich, line 1: I propose to merge TankEnergyParameters and HeatPumpTankState into one class and to name it HeatPumpTankState.
Furthermore, I would rename AllTanksEnergyParameters into AllTanksState.

tests/strategies/energy_parameters/test_heat_pump.py
```diff
@@ -140,16 +140,18 @@ def test_event_traded_energy_updates_temp_increase(energy_params):
         energy_params.event_market_cycle(CURRENT_MARKET_SLOT)
         traded_energy = 0.1
         energy_params.event_traded_energy(CURRENT_MARKET_SLOT, traded_energy)
+        # energy_params._update_last_time_slot_data(CURRENT_MARKET_SLOT)
```
hannesdiedrich, line 143: This can be removed now.
IMO the test method name should also be renamed, for example: `test_event_market_cycle_updates_temp_increase_if_energy_was_traded`

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
spyrostz: Indeed and this should be removed by now.

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
spyrostz: This is a remnant from an old implementation, removing.

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
spyrostz: To my understanding, the condenser is directly connected via a heat exchanger to the water tanks (or any other heat storage e.g. the PCM storages). We have assumed that the heat exchanger has an efficiency of 1 at the moment. Therefore the condenser temperature is directly correlated and equal to the average tank temperature, and is defined at any point in time from the temperature of the tanks. The charging operation is essentially determining on the rate of change of the condenser / tank temperature, and not the current value. 

> IMO the condensor temperature should not be an output of the charger, but an input for the charging process (if at all)

I do not agree with this, because the condenser temperature is essentially the result of the charging / heating operation, and not the "setpoint". I am aware that some heat pump models use this as a setpoint, however it makes sense only from the standpoint of the heat pump vendor, because they need a way to control the output temperature of the heat pump. Our goal is the opposite, to control the heat pump using power setpoints.

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -252,73 +394,56 @@ def _rotate_profiles(self, current_time_slot: Optional[DateTime] = None):
...
+        )
+
+    def _update_last_time_slot_data(self, time_slot: DateTime):
+        last_time_slot = self.last_time_slot(time_slot)
+        if last_time_slot not in self._source_temp_C.profile:
+            return
+        self.combined_state.increase_tank_temp_from_bought_energy(
+            last_time_slot, self._bought_energy_kWh, self._source_temp_C.profile[time_slot]
+        )
+        self._bought_energy_kWh = 0.0
```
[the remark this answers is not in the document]
spyrostz: I think that we can also retrieve this via the traded_energy_kWh, however indeed let's discuss it further!

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
spyrostz: This might have changed by the last commit, however this is not the same. `increase_tank_temp_from_bought_energy` only updates the `temp_increase_K` of the tanks, while the latter updates the `_storage_temp_C` by subtracting `increase_temp_K - decrease_temp_K`.  If you would prefer that, I can remove one of the 2 functions and merge the 2 operations in one. Let me know what you think!

tests/strategies/energy_parameters/test_heat_pump.py
spyrostz: Fixed both and thanks!

src/gsy_e/models/strategy/energy_parameters/heatpump/tank.py
spyrostz: Agreed and fixed, thanks!

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
hannesdiedrich: > Therefore the condenser temperature is directly correlated and equal to the average tank temperature, and is defined at any point in time from the temperature of the tanks.

Agreed!

> the condenser temperature is essentially the result of the charging / heating operation, and not the "setpoint".

This is where I do not agree. If you exchange "condensor" with "tank" I would agree. And this is essentially what I propose: to rename `get_condenser_temperature_C` to `get_tank_temperature_C` or similar.

As far as I see it, the condensor temperature is indeed the setpoint for the tank, at least in the PCM tank implementation that we have now. Please have a look [here](https://github.com/gridsingularity/gsy-e/blob/master/src/gsy_e/models/strategy/energy_parameters/heatpump/pcm_tank/pcm_tank.py#L157) .
The condensor temperature is the input of the tank model which then adapts the tank (or its partions/parts) to the this input.
Our goal is to harmonize those in order to use water and PCM tank interchangeably, or did I get it wrong?

In any case, and in my opinion, getting the average tank temperature should not be part of the charger class but of the tank class.

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -408,29 +428,29 @@ def _update_last_time_slot_data(self, time_slot: DateTime):
         if last_time_slot not in self._source_temp_C.profile:
             return
         self.combined_state.increase_tank_temp_from_bought_energy(
-            last_time_slot, self._bought_energy_kWh, self._source_temp_C.profile[time_slot]
+            last_time_slot, self._bought_energy_kWh, self._source_temp_C.profile[last_time_slot]
         )
         self._bought_energy_kWh = 0.0
-        self._state.tanks.update_tanks_temperature(last_time_slot)
+        self._state.tanks.update_tanks_temperature(time_slot)
```
hannesdiedrich, line 431: I do not agree on this. Why are you updating the tank temperatures before the trading?

tests/strategies/energy_parameters/test_tank_energy_parameters.py
hannesdiedrich, line 1: NIT, please also rename `TestAllTanksEnergyParameters` to `Test AllTanksState`

hannesdiedrich (commented):
Reviewed again, a couple of comments remaining. Thanks

src/gsy_e/models/strategy/energy_parameters/heatpump/tank.py
```diff
@@ -1,219 +1,106 @@
...
     def get_unmatched_demand_kWh(self, time_slot: DateTime):
         """Get unmatched demand of all water tanks."""
-        return sum(
-            tank.get_unmatched_demand_kWh(time_slot) for tank in self._tanks_energy_parameters
-        )
+        return sum(tank.get_unmatched_demand_kWh(time_slot) for tank in self._tanks_states)
 
     def serialize(self) -> Union[Dict, List]:
         """Serializable dict with the parameters of all water tanks."""
         # TODO: Convert the AVRO schemas to be able to serialize multiple tanks
```
hannesdiedrich, line 72: You did that in this PR: https://github.com/gridsingularity/gsy-framework/pull/595
Right?
If so, you can remove this comment. :-)

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
spyrostz: The reason why I insist with naming this "Condenser" is because the charger class was designed in order to act as an intermediary between the heat pump and the tank. First of all, this method is only used in the COP model calculation. The COP model does not assume any "tank" or other heat storage, it only cares about how much the temperature on the condenser side of the heat pump will increase if the heat pump consumes a fixed amount of energy. This makes sense, because the COP model only cares about the heat pump and does not attempt to model a heat storage.

> The condensor temperature is the input of the tank model which then adapts the tank (or its partions/parts) to the this input. 

Not exactly. As we discussed above, the Charger class also incorporates the heat exchanger between the condenser and the tank. You are right that the PCM does not have a heat exchanger (that's why we currently use a default efficiency of 1) however the normal water tanks usually have a heat exchanger. IMO, this heat exchanger should not be part of the heat pump, but part of the Charger class, because it controls the heat transfer from the condenser side of the heat pump to the heat storage. 

> In any case, and in my opinion, getting the average tank temperature should not be part of the charger class but of the tank class.

The average tank temperature is already part of the tank class and not of the charger class (as you can see, get_average_tank_temperature is only called from the Charger class, and is part of the AllTanksState class). IMO getting the condenser temperature though should be part of the Charger class and not part of the heat pump class.

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -182,9 +181,17 @@ def decrease_tank_temp_from_heat_demand(self, heat_demand_kJ: float, time_slot:
...
+        self,
+        source_temp_C: float,
+        time_slot: DateTime,
+        last_time_slot: DateTime,
+    ):
         """Update the COP of the heat pump in its state class."""
-        cop = self._calc_cop(heat_demand_J, source_temp_C, time_slot)
+        heat_demand_J = self._hp_state.get_heat_demand(last_time_slot)
+        cop = self._calc_cop(heat_demand_J, source_temp_C, last_time_slot)
+        # Set the calculated COP on both the last and the current time slot to use in calculations
```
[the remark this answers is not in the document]
spyrostz: Indeed this is a remnant of an old implementation IIRC. Let me remove this and thanks for catching it

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
hannesdiedrich: I feel like we had this conversation a lot of times already :-) 
Let me rephrase what I understand from your explanation above in my words in order to check if I understood:
1. the charger acts as a connection between condensor and tank, a heat exchanger
2. We assume an efficiency of 1 which is why the temperature of the condensor side is equal to the tank side
3. Only as long the efficiency is 1, we return the average tank temperature here. Once there are losses, we need to be more sophisticated  

I need to point to one thing again though. Given the above, my following statement ist still correct
> The condensor temperature is the input of the tank model which then adapts the tank (or its partions/parts) to the this input.

And I have to insist here. The condensor side of the heat exchanger is the source of heat and the tank side is the sink. This is where our disagreement is originating from AFAIS. 

> IMO getting the condenser temperature though should be part of the Charger class and not part of the heat pump class.

I only agree on this because we say that there is a heat exchanger between HP and tank to act as "middle man". In order to enforce this even more, we should remove the `self. _tanks_state ` from the  `CombinedHeatpumpTanksState` class and only let the charger have access to the tanks. 
What do you think?

spyrostz: I think that we clarified part of the comment yesterday, however just in case some parts remain unclear:

> The condensor temperature is the input of the tank model which then adapts the tank (or its partions/parts) to the this input.

I think that we agreed on this, and despite the fact that we do not use it as a setpoint we discussed that this should stay as is. 

> I only agree on this because we say that there is a heat exchanger between HP and tank to act as "middle man". In order to enforce this even more, we should remove the self. _tanks_state  from the CombinedHeatpumpTanksState class and only let the charger have access to the tanks.

This should be done now, _tanks_state should be removed from the CombinedheatPumpTanksState.

tests/strategies/energy_parameters/test_tank_energy_parameters.py
spyrostz: Done, thanks!

src/gsy_e/models/strategy/energy_parameters/heatpump/tank.py
spyrostz: This is still pending and no good way to resolve this comes to mind :( This needs changes in the simulation results service in order to be able to save the tanks results as a list. Let me give it a quick try just in case it is not breaking too many things.

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -462,6 +498,9 @@ def _populate_state(self, time_slot: DateTime):
         )
 
     def _calculate_and_set_unmatched_demand(self, time_slot: DateTime) -> None:
-        unmatched_energy_demand = self._state.tanks.get_unmatched_demand_kWh(time_slot)
+        # pylint: disable=protected-access
+        unmatched_energy_demand = self.combined_state._charger.tanks.get_unmatched_demand_kWh(
+            time_slot
+        )
```
hannesdiedrich, line 504: I think it would be better if we remove all heat pump related variables from the tank and charger class and propose to add a method in the charger that acquires the net temperature difference and calculates the unmatched demand here. What do you think?
Nothing to do now, I can also add it in my PR

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -48,13 +48,13 @@ class HeatChargerDischarger:
...
     """
 
-    def __init__(self, tanks):
+    def __init__(self, tanks: AllTanksState):
         self.tanks = tanks
         self._efficiency = HEAT_EXCHANGER_EFFICIENCY
 
     def get_condenser_temperature_C(self, time_slot: DateTime):
         """Get the temperature on the condenser side of the heat pump."""
-        return self.tanks.get_average_tank_temperature(time_slot) * self._efficiency
```
hannesdiedrich: Great catch, thanks!

hannesdiedrich (approved):
LGTM

PR #1853, opened 2025-04-24
Feature/gsye 862

hannesdiedrich:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-862


src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
```diff
@@ -263,28 +248,12 @@ def _specific_row(self, slot, market):
         if type(self.area.strategy) == HeatPumpStrategy:
             return [
                 round(
-                    self.area.strategy.state.tanks.get_unmatched_demand_kWh(slot),
-                    ROUND_TOLERANCE,
-                ),
-                round(
-                    self.area.strategy.state.tanks.get_average_tank_temperature(slot),
+                    self.area.strategy.state.charger.tanks.get_average_tank_temperature(slot),
```
spyrostz, line 251: I understand that making the `charger` member of the CombinedHeatPumpState class public was mandatory due to this, and for now it is fine. However, at some point it would be great if we used the output of `get_results_dict` here, instead of accessing the class directly. That way the charger member could stay private, and the data required here would still be accessible via the dict. IMO this applies to all strategy classes and not just for this one, but again probably it is better to be dealt in a separate story.

src/gsy_e/models/strategy/state/heatpump_all_tanks_state.py
```diff
@@ -1,22 +1,32 @@
...
     def __init__(self, tank_parameters: List[TankParameters]):
         self._tanks_states = [
-            HeatPumpTankState(tank, GlobalConfig.slot_length) for tank in tank_parameters
+            (
+                HeatPumpTankState(tank_parameters=tank, slot_length=GlobalConfig.slot_length)
+                if tank.type == HeatpumpTankTypes.WATER
+                else PCMTankState(tank_parameters=tank)
+            )
+            for tank in tank_parameters
         ]
```
spyrostz, line 30: Minor, but I think I prefer this in order to get warned if the enum selection is not supported:
```
if tank.type == HeatpumpTankTypes.WATER:
    state = HeatPumpTankState(tank_parameters=tank, slot_length=GlobalConfig.slot_length)
elif tank.type == HeatpumpTankTypes.PCM:
    state = PCMTankState(tank_parameters=tank)
else:
   assert False, f"Unsupported heat pump tank type {tank.type}"

```

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
hannesdiedrich: Ah yes, I agree. Lets refactor this in a small separate story. 
I quickly created one: https://gridsingularity.atlassian.net/browse/GSYE-882

spyrostz (commented):
Some small comments, will LGTM as soon as solved or discussed. 

Another comment, only thing I do not like that much about the new file paths is the state package, which now has a lot of heatpump_* state files. One possible option would be to group all heat pump tank states in a separate package inside the state package. That way the state package will not have that many files, and it will be possible to remove the heatpump_ prefix from most of the file names. Leaving the comment here because it affects multiple files.

hannesdiedrich:
> Another comment, only thing I do not like that much about the new file paths is the state package, which now has a lot of heatpump_* state files. One possible option would be to group all heat pump tank states in a separate package inside the state package. That way the state package will not have that many files, and it will be possible to remove the heatpump_ prefix from most of the file names. Leaving the comment here because it affects multiple files.

Agreed, I was thinking the same. Will do it.

src/gsy_e/models/strategy/state/heatpump_pcm_tank_state.py
```diff
@@ -0,0 +1,280 @@
...
+        temp_cond_C = self._get_condensor_temp_from_heat_demand_kWh(heat_energy_kWh, time_slot)
+        self._increase_storage_temp_from_condensor_temp(temp_cond_C, time_slot)
+
+    def decrease_tank_temp_from_heat_energy(self, heat_energy_kWh: float, time_slot: DateTime):
+        """Decrease the temperature of the water tank with the provided heat energy."""
+        temp_cond_C = self._get_condensor_temp_from_heat_demand_kWh(-heat_energy_kWh, time_slot)
+        self._decrease_storage_temp_from_condensor_temp(temp_cond_C, time_slot)
+        self._set_heat_demand_kJ(convert_kWh_to_kJ(heat_energy_kWh), time_slot)
+
+    def _increase_storage_temp_from_condensor_temp(
```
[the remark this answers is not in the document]
hannesdiedrich: Upsi, I think you left this comment some months/years back already but I did not keep it in mind. Will correct of course.

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
spyrostz: Thanks!

spyrostz (approved):
LGTM

PR #1854, opened 2025-04-25
Feature/gsye 881

hannesdiedrich:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-881
**GSY_FRAMEWORK_BRANCH**=feature/GSYE-881


src/gsy_e/models/strategy/energy_parameters/heatpump/virtual_heatpump_tank.py
```diff
@@ -103,9 +103,7 @@ class VirtualHeatpumpAllTanksState(AllTanksState):
 
     # pylint: disable=super-init-not-called
     def __init__(self, tank_parameters: List[TankParameters]):
-        self._tanks_states = [
-            VirtualHeatpumpTankState(tank, GlobalConfig.slot_length) for tank in tank_parameters
-        ]
+        self.tanks_states = [VirtualHeatpumpTankState(tank) for tank in tank_parameters]
```
spyrostz, line 106: Same comment as in the other PR, this probably can be avoided by the followup ticket where we will use the `get_results_dict` output to access the data in dict format instead of directly accessing the classes.

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
```diff
@@ -244,64 +229,86 @@ def _specific_row(self, slot, market):
...
+            for tank in self.area.strategy.state.charger.tanks.tanks_states:
+                if isinstance(tank, WaterTankState):
+                    rows.append(round(tank.current_tank_temperature(slot), ROUND_TOLERANCE))
+                    rows.append(round(tank.get_soc(slot), ROUND_TOLERANCE))
+        if file_key == "pcm_tanks":
+            for tank in self.area.strategy.state.charger.tanks.tanks_states:
+                if isinstance(tank, PCMTankState):
+                    rows.append(round(tank.current_tank_temperature(slot), ROUND_TOLERANCE))
+                    rows.append(round(tank.get_soc(slot), ROUND_TOLERANCE))
+        return rows
```
spyrostz, line 291: Do you plan to change this in the future and report different information for PCM tanks and Water tanks? Because if not, then there is no need for having 2 separate files for the water and pcm tanks, one file should suffice for both.

spyrostz (commented):
Some comments, will LGTM once resolved or discussed

src/gsy_e/models/strategy/state/base_states.py
```diff
@@ -402,6 +403,11 @@ def get_energy_production_forecast_kWh(self, time_slot: DateTime, default_value:
 class TankStateBase(StateInterface):
     """Base class for heat tank states"""
 
+    def __init__(self, tank_parameters: TankParameters):
+        self.name = tank_parameters.name
+        self._min_storage_temp_C = tank_parameters.min_temp_C
+        self._max_storage_temp_C = tank_parameters.max_temp_C
```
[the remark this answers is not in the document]
hannesdiedrich: Why not. Let's do it!

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
hannesdiedrich: I wanted to separate the two very different storages. Also I wanted to make sure that the number of columns does not grow that much. Imagine, with the current implementation, if you have 4 water tanks, you have 9 columns in the CSV file which gets unhandy.
Main reason is the one you mentioned though. We might export different things for the PCM tank than the water tank, even in this study already.

spyrostz (dismissed):
LGTM

PR #1855, opened 2025-04-28
Feature/gsye 880

hannesdiedrich:

## Proposed changes

-


src/gsy_e/models/strategy/state/heatpump_tank_states/all_tanks_state.py
```diff
@@ -31,19 +34,36 @@ class AllTanksState:
     def __init__(self, tank_parameters: List[TankParameters]):
         self.tanks_states = [heatpump_state_factory(tank) for tank in tank_parameters]
 
+    def _get_scaling_factors_for_charging_energy(self, time_slot):
+        _current_energy_demands_tanks = [
+            tank.get_heat_consumption_kJ(time_slot) for tank in self.tanks_states
+        ]
+        total_energy = sum(_current_energy_demands_tanks)
+        if total_energy == 0:
+            return [0 for _ in range(len(self.tanks_states))]
```
spyrostz, line 43:
```suggestion
            return [0] * len(self.tanks_states)
```

src/gsy_e/models/strategy/state/heatpump_tank_states/all_tanks_state.py
```diff
@@ -31,19 +34,36 @@ class AllTanksState:
...
+        return [energy / total_energy for energy in _current_energy_demands_tanks]
+
+    def _get_scaling_factors_for_discharging(self, time_slot):
+        available_energies = [
+            tank.get_available_energy_kJ(time_slot) for tank in self.tanks_states
+        ]
+        total_available_energy = sum(available_energies)
+        if total_available_energy == 0:
+            return [0 for _ in range(len(self.tanks_states))]
```
spyrostz, line 52: Same as above

src/gsy_e/models/strategy/state/heatpump_tank_states/all_tanks_state.py
```diff
@@ -31,19 +34,36 @@ class AllTanksState:
     def __init__(self, tank_parameters: List[TankParameters]):
         self.tanks_states = [heatpump_state_factory(tank) for tank in tank_parameters]
 
+    def _get_scaling_factors_for_charging_energy(self, time_slot):
+        _current_energy_demands_tanks = [
+            tank.get_heat_consumption_kJ(time_slot) for tank in self.tanks_states
```
spyrostz, line 39: I think that this formula will lead to unbalanced tank temperatures / tank SOCs. I do not recall our discussion anymore, however did we also evaluate using the available heat capacity of the water tank - in other words, the max heat energy that can be stored in the water tank? That way, charging would be distributed proportionally to all water tanks. 

I vaguely recall that we discussed about this though, and we though that the more the consumption of a tank is, the more it should be charged, however I am not sure about that. Feel free to ignore the comment if we came to such a conclusion

spyrostz (commented):
Some comments, LGTM once fixed / discussed, thanks!

PR #1861, opened 2025-05-15
Feature/gsye 882

hannesdiedrich:

## Proposed changes

-

**INTEGRATION_TESTS_BRANCH**=feature/GSYE-882


src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -366,8 +367,10 @@ def event_activate(self):
 
     def event_market_cycle(self, current_time_slot):
         """To be called at the start of the market slot."""
-        self._rotate_profiles(current_time_slot)
+        # Order is very important here, because _populate_state relies on last market slot,
```
hannesdiedrich, line 370: This was the root cause of the error that I was mentioning earlier. @spyrostz

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
```diff
@@ -306,9 +302,9 @@ def _water_tank_labels(self):
     @property
     def _pcm_tank_labels(self):
         labels = ["slot"]
-        for number, tank in enumerate(self.area.strategy.state.charger.tanks.tanks_states):
-            if isinstance(tank, PCMTankState):
-                tank_name_str = tank._params.name if tank._params.name else f"tank {number + 1}"
+        for number, tank in enumerate(self.area.strategy.state.get_results_dict()["tanks"]):
+            if tank["type"] == "PCM":
```
spyrostz, line 306: I would prefer an enum here if we had more than 2 types, still ok though since these are just 2.

spyrostz (dismissed):
LGTM

PR #1869, opened 2025-06-19
GSYE-889: Call TemplateStrategyUpdaterBase._read_or_rotate_rate_profi…

hannesdiedrich:
…les on every market slot and call the rotation of the profiles with the correct parameters

## Proposed changes

-

**SCM_ENGINE_BRANCH**=master


spyrostz (approved):
LGTM

PR #1890, opened 2025-08-19
Feature/gsye 906

hannesdiedrich:

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-906
**SCM_ENGINE_BRANCH**=master

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
+++ b/src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
@@ -240,7 +240,7 @@ class HeatPumpDataExporter(BaseDataExporter):
                 "heat demand [kJ]",
                 "average SOC",
                 "average tank temp [C]",
-                "average condenser temperature [C]",
+                "storage inlet temperature [C]",
                 "net heat consumed [kJ]",
             ],
             "water_tanks": self._water_tank_labels,
--- a/src/gsy_e/gsy_e_core/sim_results/results_plots.py
+++ b/src/gsy_e/gsy_e_core/sim_results/results_plots.py
@@ -746,7 +746,7 @@ class PlotHPPhysicalStats:
                 "Average SOC [%]",
                 "Coefficient of Performance",
                 "Average Tank Temperature [°C]",
-                "Condenser Temperature [°C]",
+                "Storage Inlet Temperature [°C]",
                 "Net Charging Heat energy [kJ]",
             ),
         )
@@ -780,8 +780,8 @@ class PlotHPPhysicalStats:
         fig.add_trace(
             go.Scatter(
                 x=data["slot"],
-                y=self._convert_to_float_list(data["average condenser temperature [C]"]),
-                name="Condenser Temperature [°C]",
+                y=self._convert_to_float_list(data["storage inlet temperature [C]"]),
+                name="Storage Inlet Temperature [°C]",
             ),
             row=4,
             col=1,


src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -134,7 +141,11 @@ def get_results_dict(self, current_time_slot: Optional[DateTime] = None) -> dict
             "average_soc": self._charger.get_average_soc(current_time_slot),
             **self._hp_state.get_results_dict(current_time_slot),
             "storage_temp_C": mean([tank["storage_temp_C"] for tank in tanks_results]),
-            "condenser_temp_C": self._charger.get_condenser_temperature_C(current_time_slot),
+            "average_tank_temp_C": self._charger.get_average_tank_temp_C(current_time_slot),
```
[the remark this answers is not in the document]
hannesdiedrich: I does not seem like this, so let me remove the "storage_temp_C"

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -134,7 +141,11 @@ def get_results_dict(self, current_time_slot: Optional[DateTime] = None) -> dict
...
+            "average_condenser_temp_C": self._charger.get_average_condenser_temp_C(
```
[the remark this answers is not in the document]
hannesdiedrich: As discussed offline, will rename to "average_inlet_temperature"

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -231,7 +244,11 @@ def update_cop(
     ):
         """Update the COP of the heat pump in its state class."""
         heat_demand_kJ = self._hp_state.get_heat_demand_kJ(last_time_slot)
-        cop = self._calc_cop(heat_demand_kJ, source_temp_C, last_time_slot)
+        if heat_demand_kJ < FLOATING_POINT_TOLERANCE:
+            cop = self._hp_state.get_cop(last_time_slot)
+        else:
+            cop = self._calc_cop(heat_demand_kJ, source_temp_C, last_time_slot)
```
spyrostz, line 250: I am second-guessing the initial implementation unfortunately, but shouldn't the COP depend on the produced heat energy and not just the heat demand of the home ?

hannesdiedrich: `_calc_cop` calls either the `UniversalCOPModel` or the `IndividualCOPModel` (the latter was provided by Tekniker and uses the heat demand as input. The default model `UniversalCOPModel` ignores this parameter. 

My change here is only to populate the COP value in the hp_state even if the `heat_demand_kJ` is 0. This is true for some of the market slots unfortunately. 
I hope it makes sense.

src/gsy_e/models/strategy/heat_pump.py
```diff
@@ -212,15 +214,14 @@ def event_tick(self):
         self._update_open_orders()
 
     def event_bid_traded(self, *, market_id: str, bid_trade: Trade) -> None:
+        if bid_trade.buyer.origin_uuid != self.owner.uuid:
+            return
```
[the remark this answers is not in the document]
hannesdiedrich: No, it did not :-) 
I moved it temporarily because I had some prints here that were also shown for non-HP assets. 
I left it as is in order to save the effort for the commands in between, which is none really, but might play a role in large simulations with lots of heat pumps :-)

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
hannesdiedrich: I agree, that we pass the traded energy to this method that makes it independent on the heat_demand=0 case.
Still, there are cases when there is no traded energy (e.g. when there is no cheep energy available and the demand is also 0).
In this case we need to set the COP value from the last market slot in order to have a continues profile for COP. 
Does it make sense?

src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py
```diff
@@ -236,25 +234,29 @@ def update_tanks_temperature(
...
     ):
         """Update the COP of the heat pump in its state class."""
-        heat_demand_kJ = self._hp_state.get_heat_demand_kJ(last_time_slot)
-        if heat_demand_kJ < FLOATING_POINT_TOLERANCE:
+        if bought_energy_kWh < FLOATING_POINT_TOLERANCE:
             cop = self._hp_state.get_cop(last_time_slot)
         else:
-            cop = self._calc_cop(heat_demand_kJ, source_temp_C, last_time_slot)
+            cop = self._calc_cop(
+                produced_heat_kJ=convert_kWh_to_kJ(bought_energy_kWh),
```
spyrostz, line 249: Thank you! This indeed was a bug in the reporting of the cop which was important to solve.

spyrostz (approved):
LGTM

PR #1896, opened 2025-09-08
Feature/gsye 912

spyrostz:

## Proposed changes

-

**GSY_FRAMEWORK_BRANCH**=feature/GSYE-912
**SCM_ENGINE_BRANCH**=master

### DIFF
--- a/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
+++ b/src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
@@ -67,15 +67,15 @@ class SimulationResultValidator:
 
     def validate_simulation_raw_data(self, data: Dict):
         """Validate flattened_area_core_stats_dict."""
-        self.simulation_raw_data_validator.validate(data=data, raise_exception=True)
+        # self.simulation_raw_data_validator.validate(data=data, raise_exception=True)
 
     def validate_configuration_tree(self, data: Dict):
         """Validate configuration_tree dict."""
-        self.simulation_configuration_tree_validator.validate(data=data, raise_exception=True)
+        # self.simulation_configuration_tree_validator.validate(data=data, raise_exception=True)
 
     def validate_simulation_state(self, data: Dict):
         """Validate simulation state."""
-        self.simulation_state_validator.validate(data, raise_exception=True)
+        # self.simulation_state_validator.validate(data, raise_exception=True)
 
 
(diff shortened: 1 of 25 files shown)


src/gsy_e/models/strategy/__init__.py
```diff
@@ -48,12 +48,11 @@
...
-    from gsy_framework.data_classes import TradeBidOfferInfo
-
-    from gsy_e.models.market.one_sided import OneSidedMarket
-    from gsy_e.models.market.two_sided import TwoSidedMarket
+    from gsy_framework.data_classes import TradeBidOfferInfo  # noqa: F401
+    from gsy_e.models.market.one_sided import OneSidedMarket  # noqa: F401
+    from gsy_e.models.market.two_sided import TwoSidedMarket  # noqa: F401
 
-INF_ENERGY = int(sys.maxsize)
+INF_ENERGY = int(10000000)
```
hannesdiedrich, line 55: I was already surprised when I saw the offer from the Infinite bus that you shared. Can you please explain why you went with 10000000. Is this to better locate those offers in the logs?

hannesdiedrich (commented):
Left to questions. 
Once the temporal logging is removed, I will LGTM

src/gsy_e/models/strategy/__init__.py
spyrostz: I removed it in order to avoid possible integer overflow errors. In theory, if we do this operation for example `1.5 * sys.maxsize` , the integer would overflow and the result of this operation would be `inf`. To exclude this possibility from being the cause of floating-point precision errors, I decided to use a much smaller integer (but still huge in this context, this is actually 10 GWh)

src/gsy_e/gsy_e_core/sim_results/endpoint_buffer.py
```diff
@@ -67,15 +67,15 @@ def __init__(self, is_scm: bool):
 
     def validate_simulation_raw_data(self, data: Dict):
         """Validate flattened_area_core_stats_dict."""
-        self.simulation_raw_data_validator.validate(data=data, raise_exception=True)
+        # self.simulation_raw_data_validator.validate(data=data, raise_exception=True)
```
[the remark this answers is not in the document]
spyrostz: Yes and thanks for the remark! Removing

hannesdiedrich (approved):
LGTM

PR #1897, opened 2025-09-09
Feature/gsye 906 soc management

spyrostz:

## Proposed changes

-

**SCM_ENGINE_BRANCH**=master


src/gsy_e/models/strategy/heat_pump_soc_management.py
```diff
@@ -0,0 +1,137 @@
...
+        return True
+
+    def _get_energy_from_target_state(
+        self, target_state: HeatPumpChargingState, time_slot: DateTime
+    ) -> float:
+        if target_state == HeatPumpChargingState.CHARGE:
+            return self._energy_params.get_max_energy_demand_kWh(time_slot)
+        if target_state == HeatPumpChargingState.DISCHARGE:
+            return 0
+        return self._energy_params.get_min_energy_demand_kWh(time_slot)
```
hannesdiedrich, line 114: This does not work unfortunately. If the heat demand is higher than the current charge, the storage will be "over-drained". We need to return the current stored energy energy ("soc_energy") here. The ``
already deals with this actually both in the [PCM tank state](https://github.com/gridsingularity/gsy-e/blob/a66e67fd92a64bbfcdf6f34f1f9966a1c26c6fac/src/gsy_e/models/strategy/state/heatpump_tank_states/pcm_tank_state.py#L256) and in the [water tank state](https://github.com/gridsingularity/gsy-e/blob/a66e67fd92a64bbfcdf6f34f1f9966a1c26c6fac/src/gsy_e/models/strategy/state/heatpump_tank_states/water_tank_state.py#L138).
So my proposal is the following:
```suggestion
        if target_state == HeatPumpChargingState.CHARGE:
            return self._energy_params.get_max_energy_demand_kWh(time_slot)
        return self._energy_params.get_min_energy_demand_kWh(time_slot)
```

hannesdiedrich: We somehow need to divide the responsibilities between the states and this SOC management.

hannesdiedrich: Unfortunately, the SOC Management is only applied before trading. 
The actual charging and discharging happens in the energy parameters of the heat pump [here](https://github.com/gridsingularity/gsy-e/blob/a66e67fd92a64bbfcdf6f34f1f9966a1c26c6fac/src/gsy_e/models/strategy/energy_parameters/heatpump/heat_pump.py#L215)

If there is no energy traded, the tank indeed could over-discharge if the heat demand is higher that the SOC of the tank.

hannesdiedrich: We discussed offline, that this approach is correct, because the heat demand of one market slot should not exceed the `MIN_SOC_TOLERANCE`. The latter should be configured accordingly.

spyrostz: Adapted, thanks!

PR #1918, opened 2026-04-20
Feature/research 306

spyrostz:

## Proposed changes

-

**SCM_ENGINE_BRANCH**=master


.pylintrc
```diff
@@ -41,10 +41,6 @@ load-plugins=
 # Pickle collected data for later comparisons.
 persistent=yes
 
-# When enabled, pylint would attempt to guess common misconfiguration and emit
-# user-friendly hints instead of false-positive error messages.
-suggestion-mode=yes
```
hannesdiedrich, line 46: Can you please explain the impact what this change has?

src/gsy_e/external/proxy/connection.py
```diff
@@ -0,0 +1,365 @@
...
+    @abstractmethod
+    def subscribe(
+        self,
+        on_market_slot: Callable[[MarketSlotInfo], None] = None,
+        on_trade: Callable[[EnergyTrade], None] = None,
+    ) -> None:
+        """Define callbacks for market slot and trade events."""
+
+
+class EWClientGatewayConnection(Connection):
```
hannesdiedrich: IMO this calls deserves its one module. What do you think?

src/gsy_e/external/proxy/connection.py
```diff
@@ -0,0 +1,365 @@
...
+        """
+        try:
+            with ws_connect(
+                self._ws_url,
+                subprotocols=[_CG_WS_SUBPROTOCOL],
+                additional_headers=self._auth_headers(),
+            ) as ws:
+                while not self._stop_event.is_set():
+                    try:
+                        raw = ws.recv(timeout=1.0)
```
hannesdiedrich: Do you think the timeout deserves it's constant variable at the top of the module?

src/gsy_e/external/proxy/connection.py
```diff
@@ -0,0 +1,365 @@
...
+                additional_headers=self._auth_headers(),
+            ) as ws:
+                ws.send(json.dumps(envelope))
+                raw = ws.recv()
+            return json.loads(raw).get("clientGatewayMessageId")
+        except Exception:
+            logger.exception("Failed to send order via WebSocket (%s)", self._ws_url)
+            return None
+
+    def subscribe(
```
hannesdiedrich: NIT: let's keep the public methods at the top of the class and move the private ones to the bottom. Makes reading easier.

src/gsy_e/external/proxy/connection.py
```diff
@@ -0,0 +1,365 @@
...
+        """Send a request to delete an offer to the Client Gateway."""
+        payload = {
+            "int:orderId": order_id,
+            "int:orderType": "Offer",
+            "int:timeSlot": int(time_slot.timestamp()),
+        }
+        self._send_order(payload)
+
+
+class StubConnection(Connection):
```
hannesdiedrich: Why do we need this class? I was thinking it is for testing, but I do not see this.

src/gsy_e/external/proxy/ewcg_strategy.py
```diff
@@ -0,0 +1,279 @@
...
+        """
+        Open new market slots and post initial orders. Expired slots are cleaned up first; then
+        active slots are fetched from the connector, and any slot not yet tracked is initialised.
+        """
+        self._clean_expired_slots()
+        self._open_slot(market_slot_info)
+        if self._new_market_callback:
+            self._new_market_callback(market_slot_info)
+
+    def on_tick(self) -> None:
```
hannesdiedrich, line 123: AFAIS in the connector class, this is not implemented, or at least I do not see a related event in the subscribe method. Is this only to have it ready in case we need it or am I missing something?

src/gsy_e/external/proxy/ewcg_strategy.py
```diff
@@ -0,0 +1,279 @@
...
+            state.remaining_offer_energy_kWh = max(
+                0.0, state.remaining_offer_energy_kWh - trade.energy_kWh
+            )
+
+        if self._trade_callback:
+            self._trade_callback(trade)
+
+    def _open_slot(self, market_slot_info: MarketSlotInfo) -> None:
+        open_market_ids = [state.slot_info.market_id for state in self._slot_states]
+        if market_slot_info.market_id in open_market_ids:
```
hannesdiedrich, line 174: This is only a safetynet, this should never be true, right?

src/gsy_e/external/proxy/ewcg_strategy.py
```diff
@@ -0,0 +1,279 @@
...
+        self._bid_inputs = bid_inputs
+
+    def update_offer_inputs(self, offer_inputs: List[OrderInput]) -> None:
+        """Replace the offer input list before the next market cycle."""
+        self._offer_inputs = offer_inputs
+
+    @property
+    def now(self) -> DateTime:
+        """Return the current time."""
+        return now("UTC")
```
hannesdiedrich: NIT: "UTC" could be a global constant (we have it in the framework)

src/gsy_e/external/proxy/ewcg_strategy.py
```diff
@@ -0,0 +1,279 @@
...
+            state.bid_updaters.append(
+                SlotPriceUpdater(
+                    initial_rate=Decimal(str(bid_input.min_price)),
+                    final_rate=Decimal(str(bid_input.max_price)),
+                    opening_time=market_slot_info.opening_time,
+                    closing_time=market_slot_info.closing_time,
+                    update_interval=self._update_interval,
+                )
+            )
+            state.open_bid_ids.append(None)
```
hannesdiedrich: I do not understand. What is this for? (the same for the offers)

src/gsy_e/external/proxy/ewcg_strategy.py
```diff
@@ -0,0 +1,279 @@
...
+        for bid_input in bids_for_slot:
+            state.bid_updaters.append(
+                SlotPriceUpdater(
+                    initial_rate=Decimal(str(bid_input.min_price)),
+                    final_rate=Decimal(str(bid_input.max_price)),
+                    opening_time=market_slot_info.opening_time,
+                    closing_time=market_slot_info.closing_time,
+                    update_interval=self._update_interval,
+                )
+            )
```
hannesdiedrich, line 205: I have a general question: Why would I provide two bids per slot, if the price updater will be the same for both? 
If we only allow one bid and one offer per slot, the code will be simpler AFAIS. 
Or am I missing something?

src/gsy_e/external/proxy/connection.py
```diff
@@ -0,0 +1,365 @@
...
+        return order_id
+
+    def delete_bid(self, time_slot: DateTime, order_id: str) -> None:
+        """Send a request to delete a bid to the Client Gateway."""
+        payload = {
+            "int:orderId": order_id,
+            "int:orderType": "Bid",
+            "int:timeSlot": int(time_slot.timestamp()),
+        }
+        self._send_order(payload)
```
hannesdiedrich: I am curious how this behaves when there is an issue on the server site (I am referring to the standard time out fo 10 seconds)

src/gsy_e/external/proxy/price_updater.py
```diff
@@ -0,0 +1,66 @@
+from __future__ import annotations
```
hannesdiedrich: This is not used AFAIS

src/gsy_e/external/rest_strategy.py
```diff
@@ -0,0 +1,262 @@
...
+    def replace_pending_offers(self, inputs: List[OrderInputRequest]) -> dict:
+        """Replace the pending offer inputs from a REST request body."""
+        parsed = [inp.to_external_order_input() for inp in inputs]
+        with self._pending_lock:
+            self._pending_offer_inputs = parsed
+        return {"status": "ok", "count": len(parsed)}
+
+    def snapshot_pending_bids(self) -> List[OrderInputRequest]:
+        """Return the pending bid inputs as serialisable request objects."""
+        with self._pending_lock:
```
hannesdiedrich, line 250: Why do we need this lock here for a lookup?

hannesdiedrich (commented):
I have left some comments and questions. I have a general one before I can continue
 I started with the EWCG strategy only to realise that this is not the external strategy (my bad). 
I understand the gsy "hosts" the rest API for any external commands that are performed via the rest interface. And the Energy Web Client Gateway acts as a proxy (as your folder structure says correctly). However, why don't we re-use the OrderUpdater also in the proxy strategy and re-implement a new one?

src/gsy_e/external/proxy/connection.py
spyrostz: Agreed and moved to its own module

src/gsy_e/external/proxy/connection.py
spyrostz: Agreed and fixing

src/gsy_e/external/proxy/connection.py
spyrostz: Moved private methods to the bottom

src/gsy_e/external/proxy/connection.py
spyrostz: It is a remnant from a former implementation. Removed this class completely and, because the connection module has no purpose anymore, I renamed it to dataclasses.py in order to accomodate the dataclasses for the client gateway connection.

src/gsy_e/external/proxy/ewcg_strategy.py
spyrostz: Yes, this is a placeholder that has not been defined yet. The tick events should either be triggered from FOS (that constructs this class therefore it has access to it) or from EWDS (from a new message that has not been defined yet).

src/gsy_e/external/proxy/ewcg_strategy.py
spyrostz: Indeed, but IMO necessary because MarketSlotInfo is received from EWCG, therefore it is good to sanity-check the data that we receive from it.

src/gsy_e/external/proxy/ewcg_strategy.py
spyrostz: The reason why I do not import gsy-framework is because I want this package to be independent from the rest of our code. That way we will be able to easily take it out and host it in a dedicated public repo - which is most probably how we will distribute this package in the future. Adding gsy-framework does not complicate the distribution too much, still better to avoid it completely.

src/gsy_e/external/proxy/ewcg_strategy.py
spyrostz: open_bid_ids and bid_updaters are both index-mapped. Each bid_updater corresponds to one open bid in the open_bid_ids list; therefore for each bid updater that is added we need to add one element in the open_bid_ids. 
Let me think and come up with a better design here, because index-mapping is not ideal and quite error prone.

src/gsy_e/external/proxy/ewcg_strategy.py
spyrostz: The price updater will not be the same, IIRC there is one bid updater per bid. That way, one can submit two bids with different min / max prices which update independently.

src/gsy_e/external/proxy/connection.py
spyrostz: Indeed, in this case the error will be muted and the bid will not be deleted. Of course there will be a log that the bid was not deleted, but it is not explicitly handled.

src/gsy_e/external/proxy/price_updater.py
spyrostz: This is actually used explicitly for typing. It automatically converts all types in type hints to strings, therefore we do not have to care to wrap types in double-quotes when defining custom types. Many use it by default in every new module that is created, but for us since we do not have it in other parts of our codebase, we can remove it. I'll try to remove it and use it in other modules where typing is more complex.

src/gsy_e/external/rest_strategy.py
spyrostz: This is due to the thread of the gunicorn server and the main thread, and to avoid concurrent write (for the gunicorn thread) and read (from the main thread) on these variables.

.pylintrc
spyrostz: I do not have the exact error at hand, but during development this caused lots of errors that were not really errors, but suggestions like "Maybe you meant to call this method"? After a brief search online, the suggestion was to omit it from pylint

src/gsy_e/external/proxy/connection.py
hannesdiedrich: Perfect, I was also thinking about this.

src/gsy_e/external/proxy/ewcg_connection.py
```diff
@@ -0,0 +1,253 @@
...
+            "fqcn": _CG_FQCN,
+            "topicName": _CG_ORDER_TOPIC_NAME,
+            "topicVersion": _CG_TOPIC_VERSION,
+            "topicOwner": _CG_TOPIC_OWNER,
+            "transactionId": str(uuid.uuid4()),
+            "payload": json.dumps(inner_payload),
+            "anonymousRecipient": [],
+        }
+        try:
+            with ws_connect(
```
hannesdiedrich: Why do we use websocket and not a simple HTTP request for sending?
For `_listener` I get it, but not here.

src/gsy_e/external/proxy/ewcg_strategy.py
hannesdiedrich: I understand. If this is the only time we set the time-zone that this is fine. If not, we could have a constant on package level.

src/gsy_e/external/proxy/ewcg_strategy.py
hannesdiedrich: Ah yes, `_bid_inputs` is a list of `OrderInput` that could easily contain n elements for the same market slot. Thanks!

src/gsy_e/external/proxy/ewcg_strategy.py
hannesdiedrich: But you are appending `None` here. I still do not get it, sorry.

src/gsy_e/external/proxy/connection.py
hannesdiedrich: I was more focussing on the time-out of 10 seconds. If this comment for some reason takes 10 seconds, everything else will be delayed by 10 seconds, right?

.pylintrc
hannesdiedrich: OK, makes sense

hannesdiedrich (commented):
Thanks a lot for the explanations on- and offline. 
Left a few follow up questions. Will LGTM once answered.

src/gsy_e/external/proxy/ewcg_connection.py
spyrostz: Mostly to reuse the same protocol when sending and receiving data. However I think that you are correct and it is much more reliable to use HTTP for sending data. Fixing this

src/gsy_e/external/proxy/ewcg_strategy.py
spyrostz: This is because the bid_id and offer_id are not known yet. They are populated with the response of the HTTP request, so they have to be None here and the id value will be assigned after the bid / offer have been posted. Since this is obviously confusing, I think it makes more sense to add the id of the bid and the offer in the price updaters. Let me give it a try.
Edit: I restructured the code to avoid this

src/gsy_e/external/proxy/ewcg_strategy.py
spyrostz: I added a constant to avoid the hardcoded values.

src/gsy_e/external/proxy/ewcg_connection.py
```diff
@@ -46,7 +49,9 @@ class EWClientGatewayConnection:
     def __init__(self, actor_id: str, actor_type: str) -> None:
         self._actor_id = actor_id
         self._actor_type = actor_type
-        self._ws_url = _CG_GATEWAY_URL.rstrip("/") + _CG_WS_PATH
+        base_url = _CG_GATEWAY_URL.rstrip("/")
+        self._ws_url = base_url + _CG_WS_PATH
+        self._http_url = base_url + _CG_HTTP_MESSAGES_PATH
```
hannesdiedrich: NIT: We could use urllib.parse.urljoin here

src/gsy_e/external/proxy/ewcg_strategy.py
hannesdiedrich: Now I get it. I like the new implementation much more. Thanks a lot!

hannesdiedrich (dismissed):
LGTM

PR #1919, opened 2026-05-13
Feature/gsye 891

hannesdiedrich:

## Proposed changes

-

**SCM_ENGINE_BRANCH**=master


src/gsy_e/models/strategy/heatpump_with_sortes_tank.py
```diff
@@ -0,0 +1,560 @@
...
+        )
+
+    def _calc_energy_to_buy_minimum(self, time_slot: DateTime) -> float:
+        available_stored_heat_kWh = self._calc_available_stored_heat_kWh(time_slot)
+        energy_to_be_bought_for_heat = self._get_total_electricity_demand_for_time_slot_kWh(
+            time_slot
+        ) - self._calc_heat_capacity_into_electricity_kWh(available_stored_heat_kWh)
+
+        if energy_to_be_bought_for_heat < FLOATING_POINT_TOLERANCE:
+            # corner case when the demand is lower than the discharging energy
```
hannesdiedrich, line 475: Lets talk about this

src/gsy_e/constants.py
```diff
@@ -107,3 +107,17 @@ class HeatPumpSettingsDefaultParameters:
...
+class SorTesConfiguration:
+    """Collection of SorTes heat tank configuration parameters."""
+
+    MINUTES_BEFORE_SWITCH_ALLOWED = 2 * 60
+    MIN_SOC_TOLERANCE = 10
+    MAX_SOC_TOLERANCE = 90
+    CAPACITY_KWH = 25
+    COP_HEAT_SOURCE = 1
+    COP_CONDENSER = 1
+    COP_EVAPORATOR = 1  # to be updated
```
hannesdiedrich: We need to ask for this again

src/gsy_e/models/strategy/heatpump_with_sortes_tank.py
```diff
@@ -0,0 +1,560 @@
...
+        """Return the current charging state of the tank."""
+        return self._current_state
+
+    def _get_tank_soc(self, time_slot: DateTime):
+        return self._energy_params.get_soc(time_slot)
+
+    def _is_energy_affordable(self, time_slot: DateTime, _buy_rate: float) -> bool:
+        return (
+            self._average_trade_rate.get_value(time_slot)
+            < GlobalConfig.market_maker_rate[time_slot]
```
spyrostz: As we discussed also privately, this can be optimized to look at the future values, similar to this (pseydocode):


```suggestion
            all(self._average_trade_rate.get_value(t) < GlobalConfig.market_maker_rate[t] for t in [
                time_slot.add(interval * i)
                for i in range(0, time_horizon_mins / slot_length_mins)
                if time_slot.add(interval * i) <= interval + time_horizon_mins
            ])
```

src/gsy_e/constants.py
spyrostz: Indeed this needs to be clarified

src/gsy_e/models/strategy/heatpump_with_sortes_tank.py
hannesdiedrich: Let's not discuss until further notice. I am investigating an issue here.

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
```diff
@@ -407,6 +416,47 @@ def _row(self, slot, market):
...
+
+    @property
+    def labels(self) -> List:
+        return [
+            "slot",
+            "energy traded [kWh]",
+            "COP",
+            "heat demand [kJ]",
+            "SOC",
+            "total_charged_energy_kWh",
```
spyrostz: No need for now since we do not have KPIs for this, but it would make sense to also separate the electricity consumption (traded energy) in 2 components: the electricity consumption of the condenser / evaporator, and the consumption required to produce the heat. Nothing to do of course, just for the sake of the argument.

src/gsy_e/models/strategy/heat_pump_soc_management.py
```diff
@@ -83,13 +83,19 @@ def calculate(self, time_slot: DateTime, _buy_rate: float = 0.0):
         if not self._is_time_for_state_change(time_slot):
             # If state change is not possible, but at the same time the soc is below the min or
             # above the max SOC value of the tank, then maintain the SOC.
-            if self._charger.get_average_soc(time_slot) >= self.MAX_SOC_TOLERANCE:
+            if (
+                self._get_tank_soc(time_slot) >= self.MAX_SOC_TOLERANCE
+                and self._current_state == HeatPumpChargingState.CHARGE
```
spyrostz: Ok, now I see in detail what you meant in our discussion. Essentially if the SOC has overshoot to a value greater than the MAX_SOC_TOLERANCE, and the state changed to DISCHARGE for some reason, then the state would not switch to MAINTAIN_SOC. For SorTES where we do not care about interrupting the charging / discharging state I think this is ok, however for FractLES I think that the former implementation is valid. The storage should not switch to MAINTAIN_SOC in this case, it should just discharge, unless I misunderstood something.

src/gsy_e/models/strategy/heatpump_with_sortes_tank.py
spyrostz: On first glance this looks reasonable

src/gsy_e/models/strategy/heatpump_with_sortes_tank.py
```diff
@@ -0,0 +1,588 @@
...
+        discharge_energy_kWh = self._get_performance_energy_discharge_kWh(
+            self.last_time_slot(time_slot)
+        )
+        evaporator_energy_kWh = self._calc_evaporator_electricity_kWh(discharge_energy_kWh)
+        assert (net_traded_energy_kWh - evaporator_energy_kWh) < FLOATING_POINT_TOLERANCE
+
+        self._update_soc(time_slot, -discharge_energy_kWh)
+        self._state.update_total_charged_energy_kWh(-discharge_energy_kWh)
+
+    def _update_soc(self, time_slot: DateTime, heat_energy_kWh: float):
```
spyrostz, line 542: Nit, but if the heat_energy_kWh can be negative, please leave a comment to indicate so. Thanks!

spyrostz (dismissed):
Some comments, LGTM already

src/gsy_e/models/strategy/heat_pump_soc_management.py
hannesdiedrich: I spend some time to remember why I added these changes here. Unfortunately, I am not sure, but I think we saw a case, when the status was changed to DISCHARGE before, but the SOC was still higher than the MAX value. This lead this section to change the status again to MAINTAIN_SOC and the storage was never discharged. 
How about I move this implementation to SorTesTankMinimiseSwitchStrategy and leave the original implementation here for the fractLES?

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
hannesdiedrich: Good idea. Let me even do it now, because I foresee that we have to investigate the results a bit more (already saw some unusual numbers in the KELVIN results).

src/gsy_e/models/strategy/heatpump_with_sortes_tank.py
hannesdiedrich: Ah sorry, I did not remove this comment. Indeed, the current implementation is OK IMO.

src/gsy_e/models/strategy/heat_pump_soc_management.py
hannesdiedrich: Did that here [5e53114](https://github.com/gridsingularity/gsy-e/pull/1919/commits/5e53114dfff315f39106e8b66ab62e9419392091)

src/gsy_e/gsy_e_core/sim_results/file_export_endpoints.py
hannesdiedrich: Done [e940291](https://github.com/gridsingularity/gsy-e/pull/1919/commits/e94029101808d920cc15e176c1a517af09caba9d)

src/gsy_e/models/strategy/heatpump_with_sortes_tank.py
hannesdiedrich: [e940291](https://github.com/gridsingularity/gsy-e/pull/1919/commits/e94029101808d920cc15e176c1a517af09caba9d)

spyrostz (dismissed):
LGTM

PR #1922, opened 2026-06-03
Small adaptions to SorTES heat pump strategy

hannesdiedrich:

## Proposed changes

-

**SCM_ENGINE_BRANCH**=master


spyrostz (approved):
LGTM, nothing to do and I know that we discussed already about it and different parameters are used for different calculations, still the 3 temperature input values look very weird to me. Can you explain these in the public documentation, for example that:
- ambient temp: used to select the correct value from the performance maps 
- source temp: used for the SOC calculation
- target temp: used to define the desired target temperature of the building

hannesdiedrich:
> LGTM, nothing to do and I know that we discussed already about it and different parameters are used for different calculations, still the 3 temperature input values look very weird to me. Can you explain these in the public documentation, for example that:
> 
> * ambient temp: used to select the correct value from the performance maps
> * source temp: used for the SOC calculation
> * target temp: used to define the desired target temperature of the building

I explained the ambient_temp_C_profile with 
```
 temperature profile of the air temperature at the location of the SorTES tank that is used for selecting the correct performance power of the SorTES tank
```
For the target temperature I wrote:
```
 Temperature profile of the targeted condenser temperature in °C
```
And for the source temperature I referred to the "normal" heat pump wiki. 
OK?

